PR #41, opened 2023-05-16
add openai proxy site support.

zhuyunxiang:
for someone who cannot access openai.com, add openai proxy site support.


crazywoola (commented):
LGTM

PR #96, opened 2023-05-18
Add Japanese Documents

eltociear:
I created Japanese translated documents.


takatost (approved):
thx for contributing.

PR #102, opened 2023-05-19
Feat: support image multi platform

takatost:
supports:
- linux/amd64
- linux/arm64


crazywoola (approved):

takatost:
I think this multi-platform image building is take too much resource and time consuming, which is unacceptable to users. I think we need to come up with another solution. 🤔

takatost:
Done in another PR

PR #127, opened 2023-05-21
fix: bootstrap env

yuhao1118:
Assuming I haven't missed any information, the current tutorial for starting a project seems to be missing some details. To launch a web project, developers need to manually set a series of NEXT-related environment variables in order for it to start normally. Secondly, the APP_URL for the API project is actually only used for front-end display, so it should be the front-end project's URL (3000). The real APP_URL (5001) should be specified in the web project, as this is the base URL when sending requests.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -3,7 +3,7 @@
   "version": "0.2.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "source env.sh && next dev",
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
--- a/web/package.json
+++ b/web/package.json
@@ -3,7 +3,7 @@
   "version": "0.2.0",
   "private": true,
   "scripts": {
-    "dev": "source env.sh && next dev",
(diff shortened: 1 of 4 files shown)


web/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.2.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "source env.sh && next dev",
```
iamjoel: Windows doesn't support command:  `source env.sh`. In stead, You can use `.env` file: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#env

iamjoel (changes requested):
Windows doesn't support command: source env.sh. In stead, You can use .env file: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#env

web/package.json
yuhao1118: fixed

iamjoel (approved):

PR #136, opened 2023-05-22
fix: more than 6th options would be hide

iamjoel:


web/app/components/share/chat/config-scence/index.tsx
```diff
@@ -5,7 +5,7 @@ import Welcome from '../welcome'
 
 const ConfigSence: FC<IWelcomeProps> = (props) => {
   return (
-    <div className='mb-5 antialiased font-sans overflow-hidden shrink-0'>
```
takatost, line 8: ok

takatost (approved):

PR #175, opened 2023-05-23
fix: #148 message truncated

kense-lab:


iamjoel:
Can you provide an example of the processing that needs to be done? For instance, what is the data structure of the API's response?

yuhao1118:
I am facing a similar issue. When I make a request to the /v1/chat-message API with the streaming option enabled, the response stream seems to be truncated accidentally, and this causes the failure in the JSON parsing process. This solution works for me.
image

takatost:
https://github.com/openai/openai-python/issues/399
I've noticed quite a few cases of stream returns being interrupted, it's probably related to the OpenAI issue, but it's hard to reproduce. If you guys find anything, can you send it to me so I can take a look?

PR #217, opened 2023-05-26
fix: robot emoji

yuhao1118:
- My macOS cannot display the default "robot" logo. I found that the emoji used is different (even though it looks the same visually). So I added the following improvements. It's best to test it on other operating systems before merging the code.
-  For "text generation" applications, use the logo information in siteInfo instead of hardcoding the logo.

# Before
image

image

# After
image

image


crazywoola (approved):
LGTM

PR #282, opened 2023-06-02
Add tooltips to display error details when token validation error

weifpeng:
image


crazywoola (commented):
We need i18n support for these kind of messages.

weifpeng:
> We need i18n support for these kind of messages.

It's just show error msg in response data to help location reason

PR #371, opened 2023-06-15
fix: notion-page-selector

zxhlyh:


pengjeck:
I found that after importing Notion as a data source, some subpages were not successfully imported. Will this commit fix this issue?

PR #422, opened 2023-06-20
fix: filter empty value in xlsx to improve vector similarity hit

### LINKED ISSUE #388: The hit-test result is not right
I use the original question in my document to test. But Dify can not hit the corresponding text chunk. Here is the original text chunk in dataset:
image

Here is the hitting-test result:
image

Dify version: Cloud 

## Steps To Reproduce

1. The original dataset can be downloaded here: [Test Dataset](https://docs.google.com/spreadsheets/d/1Zy9Fag-DmadBeRU-Kur7KHRX4D3-gzSN/edit#gid=567952901)
2. Upload to dataset and test with keyword: "酒店找谁"

## The current behavior
The hit result is not what we expected.

## The expected behavior
Dify should hit the text chunk contains the keyword “酒店找谁”

saifeiLee:
Exclude empty value during xlsx parsing, to make hit-test result more accurate.
By the way, we should exclude any redundant text in original dataset, so that the indexing and query result can be more accurate.

Fix: https://github.com/langgenius/dify/issues/388

### DIFF
--- a/api/core/index/readers/xlsx_parser.py
+++ b/api/core/index/readers/xlsx_parser.py
@@ -28,6 +28,6 @@ class XLSXParser(BaseParser):
                         keys = list(map(str, row))
                     else:
                         row_dict = dict(zip(keys, row))
-                        row_dict = {k: v for k, v in row_dict.items() if v is not None}
+                        row_dict = {k: v for k, v in row_dict.items() if v}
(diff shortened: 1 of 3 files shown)


takatost (approved):

takatost:
Btw, How much of an impact does None have on dataset retrieval? Is there any comparison that we can refer to?

JohnJyong (approved):

PR #438, opened 2023-06-25
bugfix: conversation parameters

iamsk:
image
The document and SDK are not match, the 'last_id' is right, tested on api.dify.ai.


takatost:
thanks

takatost (approved):

PR #451, opened 2023-06-26
fix: Optimized the private deployment check version

qin8948050:
The local network may be disconnected from https://updates.dify.ai, causing a failure to jump to the dashboard after login.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -23,6 +23,9 @@ services:
       # console or api domain.
       # example: http://udify.app
       APP_URL: ''
+      # Check the URL for version updates.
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yaml
```diff
@@ -23,6 +23,9 @@ services:
       # console or api domain.
       # example: http://udify.app
       APP_URL: ''
+      # Check the URL for version updates.
+      # example: https://updates.dify.ai
+      CHECK_UPDATE_URL: ''
```
takatost, line 28: Sorry about that, you should have the url filled in by default, not deleted by default.

takatost (changes requested):

PR #621, opened 2023-07-20
fix(web): fix Embedded copy status when toggle options

JunIce:
fix Embedded Modal when toggle options but copy status was not reset


crazywoola (approved):
LGTM

JzoNgKVO (approved):

PR #622, opened 2023-07-20
fix(web): using Tooltip unique selector key

JunIce:
if page has multi Tooltip Instance with same selector key, when click the Tooltip, will reference the error instance and will update error

image

### DIFF
--- a/web/app/components/develop/secret-key/input-copy.tsx
+++ b/web/app/components/develop/secret-key/input-copy.tsx
@@ -1,7 +1,8 @@
 'use client'
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import copy from 'copy-to-clipboard'
 import { t } from 'i18next'
+import { randomString } from '../../app-sidebar/basic'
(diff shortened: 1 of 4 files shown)


web/app/components/develop/secret-key/input-copy.tsx
```diff
@@ -1,7 +1,8 @@
 'use client'
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import copy from 'copy-to-clipboard'
 import { t } from 'i18next'
+import { randomString } from '../../app-sidebar/basic'
```
crazywoola: We prefer using `'@/app/xxxx'` instead of relative path.

crazywoola (changes requested):
The rest part is fine to me

web/app/components/develop/secret-key/input-copy.tsx
JunIce: copy!

crazywoola (approved):

PR #634, opened 2023-07-22
设置管理员账户页面也支持密码的显示与隐藏

Selenium39:
设置管理员账户密码时由于无法显示密码，导致密码设置错误也不知道。
修改为与登录页面一致，支持密码的显示与隐藏

Screenshot 2023-07-22 at 23 22 42
Screenshot 2023-07-22 at 23 23 03

### DIFF
--- a/api/controllers/console/datasets/datasets.py
+++ b/api/controllers/console/datasets/datasets.py
@@ -238,8 +238,10 @@ class DatasetIndexingEstimateApi(Resource):
         elif args['info_list']['data_source_type'] == 'notion_import':
 
             indexing_runner = IndexingRunner()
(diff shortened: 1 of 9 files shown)


takatost (changes requested):
MySQL import is something we need to consider separately, and this feature should not be included in the Cloud version. Can we split it into two PRs? We will evaluate the MySQL import feature separately.

Selenium39:
> MySQL import is something we need to consider separately, and this feature should not be included in the Cloud version. Can we split it into two PRs? We will evaluate the MySQL import feature separately.

Of course, no problem. This was my mistake, and I only intended to submit a pull request related to viewing passwords.

PR #647, opened 2023-07-26
feat: chat in explore support agent 

iamjoel:


takatost:
@CodiumAI-Agent /describe

PR #707, opened 2023-08-01
Fix ImportError: cannot import name 'default_tfidf' from 'jieba.analyse'

daedalus2022:
ImportError: cannot import name 'default_tfidf' from 'jieba.analyse'


takatost:
the issue with jieba is that it automatically upgraded to the latest version. But when we found out that the new version had a Breaking change, we fixed the version to 0.42.1. However, running "pip install -r requirements.txt" doesn't downgrade it. You need to execute "pip install -r requirements.txt --upgrade --force-reinstall" to force the downgrade.

PR #722, opened 2023-08-02
fix: generate_more_like_this function issue

MatriQ:
The parameter `chain_output` has been removed from `completion.py/get_main_llm_prompt`, but it is still incorrectly being used here. The `generate_more_like_this` functionality will work properly when the parameter is removed.


takatost (approved):

takatost:
thanks!

PR #768, opened 2023-08-07
fix: members page z-index bug

MatriQ:
Fixing the operation column and popup menu incorrectly covers the title bar.

Bug screenshot:

QA screenshot:


web/app/components/header/account-setting/members-page/operation/index.tsx
```diff
@@ -3,12 +3,12 @@ import { useTranslation } from 'react-i18next'
 import { Fragment } from 'react'
```
MatriQ, line 3: Changes of this whole file are in order to fix linting issues

crazywoola (approved):

PR #784, opened 2023-08-09
fix: frontend permission check

MatriQ:
There are many places that do not hide UI elements that should be hidden when users do not have permission to access them.

* Hide `Create New App` for normal members, Hide `Delete` button for apps:
> image
* Disable buttons for normal members:
> image
> image
* Hide `Proptps Eng.` for normal members:
> image
* Disable `Invite` for workspace members:
> image
* Disable `Add workspace` for data source:
> image
* Hide `Dataset`:
> image

### DIFF
--- a/web/app/(commonLayout)/datasets/Datasets.tsx
+++ b/web/app/(commonLayout)/datasets/Datasets.tsx
@@ -7,7 +7,7 @@ import NewDatasetCard from './NewDatasetCard'
 import DatasetCard from './DatasetCard'
 import type { DataSetListResponse } from '@/models/datasets'
 import { fetchDatasets } from '@/service/datasets'
-import { useSelector } from '@/context/app-context'
+import { useAppContext, useSelector } from '@/context/app-context'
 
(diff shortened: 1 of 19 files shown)


web/app/(commonLayout)/datasets/Datasets.tsx
```diff
@@ -44,7 +45,7 @@ const Datasets = () => {
       {data?.map(({ data: datasets }) => datasets.map(dataset => (
         <DatasetCard key={dataset.id} dataset={dataset} onDelete={mutate} />),
       ))}
-      <NewDatasetCard ref={anchorRef} />
+      { ['owner', 'admin'].includes(currentWorkspace.role) && <NewDatasetCard ref={anchorRef} /> }
```
crazywoola: Can you make it as a common function in `AppContext`, and use a enum to describe `['owner', 'admin'] normal`,  same as the rest.

crazywoola (changes requested):
The rest LGTM

crazywoola (approved):

PR #785, opened 2023-08-09
fix: properly respond when no tenant linked to current account

MatriQ:
Fix 500 response when no tenant is linked to the current account.

### DIFF
--- a/api/controllers/console/explore/installed_app.py
+++ b/api/controllers/console/explore/installed_app.py
@@ -42,6 +42,9 @@ class InstalledAppsListApi(Resource):
     @account_initialization_required
     @marshal_with(installed_app_list_fields)
     def get(self):
+        if not current_user.current_tenant:
+            return {'installed_apps': []}
+
         current_tenant_id = current_user.current_tenant_id
         installed_apps = db.session.query(InstalledApp).filter(
             InstalledApp.tenant_id == current_tenant_id
--- a/api/controllers/console/explore/recommended_app.py
+++ b/api/controllers/console/explore/recommended_app.py
@@ -45,6 +45,9 @@ class RecommendedAppListApi(Resource):
     def get(self):
         language_prefix = current_user.interface_language if current_user.interface_language else 'en-US'
 
+        if not current_user.current_tenant:
+            return {'recommended_apps': [], 'categories': []}
+
         recommended_apps = db.session.query(RecommendedApp).filter(
             RecommendedApp.is_listed == True,
             RecommendedApp.language == language_prefix
--- a/api/controllers/console/workspace/members.py
+++ b/api/controllers/console/workspace/members.py
@@ -36,7 +36,11 @@ class MemberListApi(Resource):
     @account_initialization_required
     @marshal_with(account_list_fields)
     def get(self):
-        members = TenantService.get_tenant_members(current_user.current_tenant)
+        members = [
+            TenantService.get_tenant_members(current_user.current_tenant)
+            if current_user.current_tenant
+            else []
+        ]
         return {'result': 'success', 'accounts': members}, 200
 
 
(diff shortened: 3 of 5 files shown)


takatost (changes requested):
This idea seems off. There shouldn't be a situation where `current_tenant` is `None`. That would be considered dirty data.
btw, when do you usually encounter this issue?

MatriQ:
> This idea seems off. There shouldn't be a situation where `current_tenant` is `None`. That would be considered dirty data.
> btw, when do you usually encounter this issue?

You can reproduce the corresponding issue with the following steps:
1.  Invite a new user: xxx@xxx.com.
2.  Signup with xxx@xxx.com.
3.  Remove xxx@xxx.com from your workspace.
4.  Then you can reproduce it with login xxx@xxx.com.

I also considered creating a tenant and setting it as the default workspace if a user just removed it from all tenants, but it shouldn't have happened in the SELF_HOSTED mode, right?

However, I think these changes can make the code to be more strong.

takatost:
Maybe the problem lies in the logic of retrieving the team/tenant after logging in: api.app.load_user.
According to the normal logic:

If a user logs in and the team corresponding to the workspace_id retrieved from the session does not exist, they should switch to their own team.
If a user does not belong to any team and is in a detached state, they should register a team of their own.
What do you think? 😊

MatriQ:
> > This idea seems off. There shouldn't be a situation where `current_tenant` is `None`. That would be considered dirty data.
> > btw, when do you usually encounter this issue?
> 
> You can reproduce the corresponding issue with the following steps:
> 
> 1. Invite a new user: [xxx@xxx.com](mailto:xxx@xxx.com).
> 2. Signup with [xxx@xxx.com](mailto:xxx@xxx.com).
> 3. Remove [xxx@xxx.com](mailto:xxx@xxx.com) from your workspace.
> 4. Then you can reproduce it with login [xxx@xxx.com](mailto:xxx@xxx.com).
> 
> I also considered creating a tenant and setting it as the default workspace if a user just removed it from all tenants, but it shouldn't have happened in the SELF_HOSTED mode, right?
> 
> However, I think these changes can make the code to be more strong.

It's a good idea. but I still think this PR is necessary, else the issue still can be triggered when the user xxx@xxx.com never sign out.

BTW, Can it have more tenants in `SELF_HOSTED` mode?

takatost:
Indeed, your code can avoid some issues, but these interfaces are not all. There are still many interfaces that need to query the current user's workspace. It may increase in the future.

What I mean is, if we fix this logic vulnerability at `@login_manager.user_loader`, then these interfaces will not have a situation where `current_user.current_tenant` is empty when retrieving it.

As I mentioned above, if the user's workspace does not exist, switch to an existing workspace. If there is no workspace at all, create a workspace with the owner being the current user.

MatriQ:
> Indeed, your code can avoid some issues, but these interfaces are not all. There are still many interfaces that need to query the current user's workspace. It may increase in the future.
> 
> What I mean is, if we fix this logic vulnerability at `@login_manager.user_loader`, then these interfaces will not have a situation where `current_user.current_tenant` is empty when retrieving it.
> 
> As I mentioned above, if the user's workspace does not exist, switch to an existing workspace. If there is no workspace at all, create a workspace with the owner being the current user.

It's a simpler and more useful approach. I can implement it here, and if it works I'll remove these redundant changes.

MatriQ:
The `workspace_id ` will be incorrect if the user does not re-signin in this case.

takatost:
Or log out before returning None.
```
flask.session.pop('workspace_id', None)
flask_login.logout_user()
```

MatriQ:
I created a new [PR]( https://github.com/langgenius/dify/pull/793) instead of this.

PR #798, opened 2023-08-10
add streaming event end flag field finished

conghaoyuan:


takatost (dismissed):
thx!

takatost (changes requested):
Please submit it as two PRs, thank you!

PR #812, opened 2023-08-12
add streaming event end flag field finished

conghaoyuan:
{"event": "end", "finished": true}
{"event": "message", "task_id": "f9e6c33d-3cbd-45b8-94ad-e311f47eb1ac", "id": "2b829d7f-86fa-49af-8831-66017e1997a2", "answer": "", "created_at": 1691482603, "conversation_id": "28c9d3b3-06ee-41a3-8808-0ae3b95484c3", "finished": false}


takatost:
We have already provided "event: end" to indicate that the message has ended. Why do we need to add "finished" again? I feel like these two indicators are redundant.

PR #816, opened 2023-08-12
fix: showing dataset error tip

MatriQ:
Adding dataset document error tip

image


SuperevanZ:
@MatriQ UI
error message ui

PR #837, opened 2023-08-14
fix: change config string variable limit

MatriQ:
* Change the config string variable limit to 1024.
 * Fix linting issue.

### DIFF
--- a/web/app/components/app/configuration/config-var/config-string/index.tsx
+++ b/web/app/components/app/configuration/config-var/config-string/index.tsx
@@ -7,7 +7,7 @@ export type IConfigStringProps = {
   onChange: (value: number | undefined) => void
 }
 
-const MAX_LENGTH = 1024
+const MAX_LENGTH = 256
 
 const ConfigString: FC<IConfigStringProps> = ({
   value,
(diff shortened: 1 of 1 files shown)


web/app/components/app/configuration/config-var/config-string/index.tsx
```diff
@@ -1,32 +1,27 @@
...
+import React from 'react'
 
-export interface IConfigStringProps {
+export type IConfigStringProps = {
   value: number | undefined
   onChange: (value: number | undefined) => void
 }
 
-const MAX_LENGTH = 64
+const MAX_LENGTH = 1024
```
crazywoola: LGTM Please set to 256 instead

crazywoola (approved):

web/app/components/app/configuration/config-var/config-string/index.tsx
crazywoola: ;)

crazywoola (changes requested):

web/app/components/app/configuration/config-var/config-string/index.tsx
crazywoola:
```suggestion
const MAX_LENGTH = 256
```

crazywoola (dismissed):

crazywoola (approved):

PR #862, opened 2023-08-16
Fix/price calc

lingfengchencn:
use self.price_config to load model price configs.

TODO:

load price config from model settings

### DIFF
--- a/api/core/model_providers/models/embedding/azure_openai_embedding.py
+++ b/api/core/model_providers/models/embedding/azure_openai_embedding.py
@@ -48,12 +48,25 @@ class AzureOpenAIEmbedding(BaseEmbedding):
         # calculate the number of tokens in the encoded text
         return len(tokenized_text)
 
-    def get_token_price(self, tokens: int):
-        tokens_per_1k = (decimal.Decimal(tokens) / 1000).quantize(decimal.Decimal('0.001'),
-                                                                  rounding=decimal.ROUND_HALF_UP)
+    @property
+    def model_unit_prices_config(self):
+        """
+        get model unit prices config.
 
-        total_price = tokens_per_1k * decimal.Decimal('0.0001')
-        return total_price.quantize(decimal.Decimal('0.0000001'), rounding=decimal.ROUND_HALF_UP)
+        :return: object format {
+            "model_name": {
+                'completion': decimal.Decimal('0'),
+                'unit': decimal.Decimal('0'),
+            }
+        }
+        """
+        config = {
+            'text-embedding-ada-002':{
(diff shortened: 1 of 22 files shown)


api/core/model_providers/models/llm/azure_openai_model.py
```diff
@@ -97,44 +98,48 @@ def get_num_tokens(self, messages: List[PromptMessage]) -> int:
...
-        tokens_per_1k = (decimal.Decimal(tokens) / 1000).quantize(decimal.Decimal('0.001'),
-                                                                  rounding=decimal.ROUND_HALF_UP)
-
-        total_price = tokens_per_1k * unit_price
-        return total_price.quantize(decimal.Decimal('0.0000001'), rounding=decimal.ROUND_HALF_UP)
-
-    def get_currency(self):
-        return 'USD'
+    @property
+    def model_unit_prices_config(self):
```
takatost: In Azure, the base_model_name is the name of the base model, while name is the deployment name in Azure, which can be customized by the user. Has this been considered here?

api/core/model_providers/models/embedding/azure_openai_embedding.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, model_provider: BaseModelProvider, name: str):
             max_retries=1,
             **self.credentials
         )
+        # TODO load price_config from configs(db)
```
takatost: I personally think that storing in the db is not a good distribution solution, as it involves the issue of needing to perform migration to update every time there is a change. But you've reminded me, placing it in the api/core/model_providers/rules in a JSON configuration file might make updating more convenient.

takatost (changes requested):
Thank you very much for your optimization commit. This part was indeed written in a hurry. I have written some notes after the review; please take a look to see if there are any questions. If there are no issues, we will arrange comprehensive testing to confirm that everything is correct.

api/core/model_providers/models/llm/azure_openai_model.py
lingfengchencn: em...so , we need to add base model as the config key?

api/core/model_providers/models/embedding/azure_openai_embedding.py
lingfengchencn: Because there may be discounts available through partnerships, the actual price and the listed price may not necessarily be the same. so, I think it is better for config by user self.

takatost (approved):
Thank you so much for your contribution! There might be a few minor issues, but I'll fix them after merging.

PR #877, opened 2023-08-16
Add support for Llama2, Palm, Cohere, Anthropic, Replicate Models - using litellm

ishaan-jaff:
This PR adds support for models from all the above mentioned providers using https://github.com/BerriAI/litellm/

`ChatLiteLLM()` is integrated into langchain and allows you to call all models using the OpenAI I/O interface 
https://python.langchain.com/docs/integrations/chat/litellm

Here's an example of how to use ChatLiteLLM()
```python
ChatLiteLLM(model="gpt-3.5-turbo")
ChatLiteLLM(model="claude-2", temperature=0.3)
ChatLiteLLM(model="command-nightly")
ChatLiteLLM(model="replicate/llama-2-70b-chat:2c1608e18606fad2812020dc541930f2d0495ce32eee50074220b87300bc16e1")

```

### DIFF
--- a/api/core/third_party/langchain/llms/chat_open_ai.py
+++ b/api/core/third_party/langchain/llms/chat_open_ai.py
@@ -2,11 +2,11 @@ import os
 
 from typing import Dict, Any, Optional, Union, Tuple
 
-from langchain.chat_models import ChatOpenAI
+from langchain.chat_models import ChatLiteLLM
 from pydantic import root_validator
 
 
-class EnhanceChatOpenAI(ChatOpenAI):
+class EnhanceChatOpenAI(ChatLiteLLM):
     request_timeout: Optional[Union[float, Tuple[float, float]]] = (5.0, 300.0)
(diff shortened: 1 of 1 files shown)


ishaan-jaff:
@takatost @crazywoola can you please take a look at this PR😊

Happy to add more tests/ docs if this initial commit looks good

takatost:
I'm glad you can provide us with more model support, but it seems that this only modifies the OpenAI ChatLLM. However, our support is not limited to just architecture and interface. We need to define providers/rules/llm models in api/core/model_providers and adapt the frontend web to add support for suppliers in order to be supported in Dify. It's great to see a library that can support adding multiple models.
Currently, we officially support OpenAI, Anthropic, and Replicate Models. Others can be supported as new suppliers. Contributions are welcome, or you can wait for us to adapt them when we have time later on. 😊

ishaan-jaff:
@takatost thanks for the feedback -  Is adding new LLMs not a priority for dify currently ?

PR #1021, opened 2023-08-25
feat: add LocalAI local embedding model support

takatost:


nitrag:
Thanks for this! Weird though...I can't see it in the UI settings under show more model providers. I'm running `0.3.19` in docker-compose.

nitrag:
@takatost ? I still don't see in `0.3.22` docker version? Is there an ENV variable needed for this?

image

takatost:
> @takatost ? I still don't see in `0.3.22` docker version? Is there an ENV variable needed for this?
> 
> image

Sorry, the front-end here missed displaying LocalAI. It has been fixed in #1169 and will be included in the next version.

PR #1041, opened 2023-08-28
fix openpyxl dimensions error

### LINKED ISSUE #1039: Load WPS xlsx error
Dify version: Cloud and  Self Host

## Steps To Reproduce

When using Excel files generated by MS Office, openpyxl can successfully load the data. 

However, when using Excel files generated by WPS, openpyxl can only read the data from the first cell, which is A1:A1.

I found the reason from this document (https://openpyxl.readthedocs.io/en/stable/optimized.html#worksheet-dimensions):

> Read-only mode relies on applications and libraries that created the file providing correct information about the worksheets, specifically the used part of it, known as the dimensions. Some applications set this incorrectly. You can check the apparent dimensions of a worksheet using ws.calculate_dimension(). If this returns a range that you know is incorrect, say A1:A1 then simply resetting the max_row and max_column attributes should allow you to work with the file:

```
ws.reset_dimensions()
```

If you need some test data from WPS, I can send it to you privately.

wangbooth:
fix #1039

### DIFF
--- a/api/core/data_loader/loader/excel.py
+++ b/api/core/data_loader/loader/excel.py
@@ -30,6 +30,8 @@ class ExcelLoader(BaseLoader):
         wb = load_workbook(filename=self._file_path, read_only=True)
         # loop over all sheets
         for sheet in wb:
+            if 'A1:A1' == sheet.calculate_dimension():
+                sheet.reset_dimensions()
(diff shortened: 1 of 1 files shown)


crazywoola (commented):
LGTM

JohnJyong (approved):

PR #1052, opened 2023-08-30
fix: handle AttributeError for datasets and index

yezhwi:
Handle AttributeError when accessing properties of Python None objects


crazywoola (approved):
LGTM

PR #1069, opened 2023-08-31
chore(api): api image multistage build

JunIce:
Api image size reduce to 1.8 GB

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -3,6 +3,9 @@ FROM python:3.10-slim AS base
 
 LABEL maintainer="takatost@gmail.com"
 
+RUN apt-get update \ 
+    && apt-get install -y --no-install-recommends gcc g++ python3-dev libc-dev libffi-dev
+
 COPY requirements.txt /requirements.txt
 
 RUN pip install --prefix=/pkg -r requirements.txt
@@ -24,7 +27,7 @@ EXPOSE 5001
 WORKDIR /app/api
 
 RUN apt-get update \ 
-    && apt-get install -y --no-install-recommends bash curl wget vim gcc g++ python3-dev libc-dev libffi-dev nodejs \
+    && apt-get install -y --no-install-recommends bash curl wget vim nodejs \
     && apt-get autoremove \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


api/Dockerfile
```diff
@@ -1,7 +1,15 @@
-FROM python:3.10-slim
+# packages install stage
+FROM python:3.10-slim AS base
 
 LABEL maintainer="takatost@gmail.com"
 
+COPY requirements.txt /requirements.txt
+
+RUN pip install --prefix=/pkg -r requirements.txt
```
takatost, line 11: Before pip install, you need to install the following packages. These packages are required for the pip install process on the arm architecture. 
```
RUN apt-get update \
    && apt-get install -y --no-install-recommends gcc gcc g++ python3-dev libc-dev libffi-dev \
    && apt-get autoremove \
    && rm -rf /var/lib/apt/lists/*
```

api/Dockerfile
```diff
@@ -15,13 +23,12 @@ EXPOSE 5001
...
 WORKDIR /app/api
 
-RUN apt-get update && \
-    apt-get install -y bash curl wget vim gcc g++ python3-dev libc-dev libffi-dev nodejs
-
-COPY requirements.txt /app/api/requirements.txt
-
-RUN pip install -r requirements.txt
+RUN apt-get update \ 
+    && apt-get install -y --no-install-recommends bash curl wget vim gcc g++ python3-dev libc-dev libffi-dev nodejs \
```
takatost: This side can reduce the installation of the aforementioned dependencies, and it can be simplified to `bash curl wget vim nodejs`

takatost (changes requested):

api/Dockerfile
JunIce: Have been modified.

JunIce:
Why we need `vim` pkg, which scenario we use it?

takatost:
> Why we need `vim` pkg, which scenario we use it?

You can remove it, it was originally for debugging purposes.

takatost (approved):
thx😊

JunIce:
> You can remove it, it was originally for debugging purposes.

No need. I just want to know the details. Thank for your explaination.

PR #1100, opened 2023-09-04
fix: shortening invite url

MatriQ:
Shortening invite url from `/activate?workspace_id=****&email=****&token=***` to  `/activate?token=***`, 

The old pattern of invite url is still supported to activate.

### DIFF
--- a/api/services/account_service.py
+++ b/api/services/account_service.py
@@ -1,5 +1,6 @@
 # -*- coding:utf-8 -*-
 import base64
+import json
 import logging
 import secrets
 import uuid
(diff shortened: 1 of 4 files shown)


api/services/account_service.py
```diff
@@ -401,7 +406,7 @@ def invite_new_member(cls, tenant: Tenant, email: str, role: str = 'normal',
         # send email
         send_invite_member_mail_task.delay(
             to=email,
-            token=cls.generate_invite_token(tenant, account),
```
MatriQ, line 404: Is it a bug? I'm not sure why it uses different tokens between the front end and the invite email.

crazywoola (commented):
LGTM

api/services/account_service.py
crazywoola: Black history

takatost: yep😅

crazywoola (dismissed):

MatriQ:
Based on the discussion with @takatost , added the email to the returned invite URL:
604662ef94dbd4f884a93783279b871

takatost (approved):
thx!

PR #1107, opened 2023-09-06
fix(web): fix AppCard Menu popover open bug

JunIce:


web/app/components/app/overview/settings/index.tsx
```diff
@@ -47,7 +47,7 @@ const SettingsModal: FC<ISettingsModalProps> = ({
 }) => {
   const [isShowMore, setIsShowMore] = useState(false)
   const { icon, icon_background } = appInfo
-  const { title, description, copyright, privacy_policy, default_language } = appInfo.site
+  const { title, description = undefined, copyright, privacy_policy, default_language } = appInfo.site
```
crazywoola: If this is undefined, this should be undefined. There is no need to give it a value.

crazywoola (changes requested):

crazywoola (approved):

PR #1113, opened 2023-09-06
feat: add health check and depend condition in docker compose

deltaLRD:
Add health check to avoid api error at the begin of the api container


crazywoola (dismissed):

takatost (changes requested):
Can we also add this `condition` parameter to the `api` service? 😊

takatost (approved):
thx!

Issue #1088, opened 2023-09-01
API return NULL answer


cmastro:
The issue was closed, but unfortunately it's not fixed: https://github.com/langgenius/dify/issues/981 .. and I cannot re-open the issue.

Thanks!

PR #1133, opened 2023-09-09
fix: answer returned null when response_mode was blocking

takatost:
## How to reproduce

When selecting multiple datasets, subscribing to the messages of agent_thought, there is no content in the answer, which ultimately results in a Null response.

## Solution

Only subscribe to messages with the event "message"


takatost:
fix #1088

PR #1142, opened 2023-09-10
fix: remove the deprecated depends_on.condition format

takatost:
remove the deprecated depends_on.condition format in docker compose v3


takatost:
fix #1132

PR #1326, opened 2023-10-12
Add Message Suggested Api

charli117:
下一步问题建议获取API及相关文档更新，并同步删除get_message、get_conversation、delete_conversation用户筛选参数

### DIFF
--- a/api/controllers/console/explore/conversation.py
+++ b/api/controllers/console/explore/conversation.py
@@ -52,7 +52,7 @@ class ConversationApi(InstalledAppResource):
 
         conversation_id = str(c_id)
         try:
-            ConversationService.delete(app_model, conversation_id)
+            ConversationService.delete(app_model, conversation_id, current_user)
         except ConversationNotExistsError:
             raise NotFound("Conversation Not Exists.")
         WebConversationService.unpin(app_model, conversation_id, current_user)
--- a/api/controllers/console/explore/conversation.py
+++ b/api/controllers/console/explore/conversation.py
@@ -52,7 +52,7 @@ class ConversationApi(InstalledAppResource):
 
         conversation_id = str(c_id)
         try:
-            ConversationService.delete(app_model, conversation_id, current_user)
+            ConversationService.delete(app_model, conversation_id)
         except ConversationNotExistsError:
             raise NotFound("Conversation Not Exists.")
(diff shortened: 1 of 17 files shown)


crazywoola (commented):
LGTM

api/controllers/web/conversation.py
```diff
@@ -49,7 +49,7 @@ def delete(self, app_model, end_user, c_id):
 
         conversation_id = str(c_id)
         try:
-            ConversationService.delete(app_model, conversation_id, end_user)
+            ConversationService.delete(app_model, conversation_id)
```
takatost: Web API can't remove the user's filtering criteria because web API is publicly accessible. If a user somehow obtains the id of other conversations, they can delete someone else's conversation, which is an abuse of authority. 😕

api/controllers/web/message.py
```diff
@@ -178,7 +178,6 @@ def get(self, app_model, end_user, message_id):
         try:
             questions = MessageService.get_suggested_questions_after_answer(
                 app_model=app_model,
-                user=end_user,
```
takatost: Same as above, this will lead to abuse of power.

api/controllers/console/explore/conversation.py
```diff
@@ -52,7 +52,7 @@ def delete(self, installed_app, c_id):
 
         conversation_id = str(c_id)
         try:
-            ConversationService.delete(app_model, conversation_id, current_user)
```
takatost: Actually, I don't understand why we need to remove the `user` parameter. If the current user obtains the conversation id through the `conversation list`, then when deleting a conversation or retrieving suggested questions, it should be matched with the current user without any issues. Can you explain the reason behind this decision? 🤔

takatost (changes requested):

api/controllers/console/explore/conversation.py
charli117: 不好意思只考虑唯一性问题忽略了越权，已经恢复并重新调整逻辑

api/controllers/web/message.py
charli117: 已调整

takatost (approved):
thx!

crazywoola (approved):

PR #1454, opened 2023-11-06
fix: Use correct typehint for return values

lethe3000:
suppress style check warnings and get rid of potential errors.


takatost (approved):
thx!

PR #1460, opened 2023-11-06
feat: support chatglm_turbo model #1443

takatost:


lchustc:
hello, why chatglm2 couldn't remember chat history?
image
image

PR #1510, opened 2023-11-11
feat: [backend] vision support

takatost:
support OpenAI GPT-4V image upload


barnett-yuxiang (commented):
Is it still necessary to avoid submitting such large commit?

PR #1551, opened 2023-11-16
feat: fetch stream compatibility enhance

rickythink:
The purpose of this PR is to fix an issue that fetch stream feature would not work because of compatibility.

## Steps to reproduce

- Use the QQ iOS app
- browser any dify demo app
- try to send any word and we won't get any reply

## The implementation approach is:

Detect browser version, and do not use fetch streams for older browsers
- For older browsers, use traditional request mode for chatting
- For newer browsers, continue using fetch streams

## Reference

[https://developer.chrome.com/articles/fetch-streaming-requests/](https://developer.chrome.com/articles/fetch-streaming-requests/)

### DIFF
--- a/web/global.d.ts
+++ b/web/global.d.ts
@@ -1,2 +1,5 @@
 declare module 'lamejs';
 declare module 'react-18-input-autosize';
+declare module 'fetch-readablestream' {
+  export default function fetchReadableStream(url: string, options?: RequestInit): Promise<Response>
+}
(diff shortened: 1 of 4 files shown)


crazywoola (approved):
LGTM

zxhlyh (approved):

takatost:
How did you deploy dify? Did you use Docker Compose or deploy from source code? 
I guess it might be a source code deployment without updating the Python dependencies:

```
cd api
pip install -r requirements.txt
```

Issue #1583, opened 2023-11-21
SQLAlchemy Transaction error


waltcow:
### Dify version

0.3.21

### Cloud or Self Hosted

Self Hosted

### Steps to reproduce

> psycopg2.InterfaceError: cursor already closed

```python
        with db.engine.begin() as conn:
            rs = conn.execute(db.text(sql_query), arg_dict)

        response_data = []

        for i in rs:
            response_data.append({
                'date': str(i.date),
                'interactions': float(i.interactions.quantize(Decimal('0.01')))
            })

        return jsonify({
            'data': response_data
        })

```
(https://github.com/langgenius/dify/blob/cc35d0645a862c7eea28ff34b044ba8ebf29ff6e/api/controllers/console/app/statistic.py#L259C1-L272C11)

<img width="2421" alt="image" src="https://github.com/langgenius/dify/assets/3941174/4851b484-5896-4a9d-ac38-845fa377d25f">

### ✔️ Expected Behavior

no error

### ❌ Actual Behavior

sql error

waltcow:
Okay

waltcow:
> Can you please make a PR if possible? I would be extremely grateful!

https://github.com/langgenius/dify/pull/1586

PR #1586, opened 2023-11-21
fix sql transaction error in statistic API

waltcow:
https://github.com/langgenius/dify/issues/1583

### DIFF
--- a/api/controllers/console/app/statistic.py
+++ b/api/controllers/console/app/statistic.py
@@ -62,11 +62,10 @@ class DailyConversationStatistic(Resource):
 
         sql_query += ' GROUP BY date order by date'
 
+        response_data = []
+
         with db.engine.begin() as conn:
             rs = conn.execute(db.text(sql_query), arg_dict)
-
-            response_data = []
-
             for i in rs:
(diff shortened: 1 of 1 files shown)


api/controllers/console/app/statistic.py
```diff
@@ -65,13 +65,13 @@ def get(self, app_id):
         with db.engine.begin() as conn:
             rs = conn.execute(db.text(sql_query), arg_dict)
 
-        response_data = []
+            response_data = []
```
takatost: Defining `response_data` inside the `with` statement might cause issues when trying to access it in the final `return` statement. Put it above `with` would be better.

takatost: The changes below are the same, you know. 
Whatever, thank you very much for your contribution! 😊

takatost (changes requested):

api/controllers/console/app/statistic.py
waltcow: Fixed @takatost

takatost (approved):
thx!

PR #1608, opened 2023-11-23
fix ascii codec error, by using utf8

yunyu-Mr:
postgresql url config `client_encoding=utf8`


crazywoola (changes requested):
Can you use environment variables instead of hard code here. Just make sure the behavior here is the same

takatost:
Can you provide the original scene to reproduce the ASCII codec error?

barnett-yuxiang (commented):
hardcode？ 😭

crazywoola (approved):

PR #1678, opened 2023-12-03
fix wrong syntax of type definitions

Yeuoly:
fix: fix wrong syntax of return type definitions of AgentExecutor._init_agent


takatost (approved):
Thank you so much!

PR #1747, opened 2023-12-13
fix: add 'sentence-similarity' in allowed hugging face model tasks

guchenhe:
many popular embedding models fall under 'sentence-similarity' tasks on hugging face such as [this one](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2). this fix makes them accessible.

### DIFF
--- a/api/core/model_providers/providers/huggingface_hub_provider.py
+++ b/api/core/model_providers/providers/huggingface_hub_provider.py
@@ -118,9 +118,9 @@ class HuggingfaceHubProvider(BaseModelProvider):
                 if 'inference' in model_info.cardData and not model_info.cardData['inference']:
                     raise ValueError(f'Inference API has been turned off for this model {model_name}.')
 
-                VALID_TASKS = ("text2text-generation", "text-generation", "feature-extraction")
+                VALID_TASKS = ("text2text-generation", "text-generation", "feature-extraction", "sentence-similarity")
                 if model_info.pipeline_tag not in VALID_TASKS:
-                    raise ValueError(f"Model {model_name} is not a valid task, "
+                    raise ValueError(f"Task {model_info.pipeline_tag} of model {model_name} is not valid, "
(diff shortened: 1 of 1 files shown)


crazywoola (approved):

takatost (changes requested):
Sentence Similarity Task is not an Embedding task, ya know. 😄
Although the model used here is indeed an Embedding model, its task is to calculate similarity scores by inputting query and docs. We utilize the ability of the Embedding model to convert embeddings into vectors, and the retrieval and similarity comparison are performed in a vector database.

The principle is that the vector database stores hundreds or thousands of segments, and the query vector after embedding is used to retrieve the top K segments that best match from this large batch of segments. It is a retrieval index technique.
However, Sentence Similarity Task is not suitable for retrieval tasks with such a large number of segments. The core of the vector database lies in the indexing performance and optimization under large data volume, and the similarity comparison algorithm in text is a simple dot product operation (cosine similarity algorithm). Sentence Similarity Task is more suitable for NLP tasks such as classification, intent recognition, and other simple tasks.

jafrank88:
How about permitting models tagged as 'bert'?

GarfieldDai:
You can use huggingface [Inference Endpoints](https://huggingface.co/docs/inference-endpoints/index) to deploy target embedding model. Pay attention, you should select task type as "Feature Extraction" when you do deployment.
image
image

jafrank88:
This might be obvious to others, but why would I want to pay to run an embedding model for my local docs when the models are small, the documents are private, and it does not take that much processing power to embed if you have a GPU.

GarfieldDai:
Only [Hosted Inference API](https://huggingface.co/docs/api-inference/quicktour) is enabled that can be used in Dify. Not all the embedding model is available.
image
The model you show is unavailable at Dify for two reason. One task type is not "feature extraction" and the other is the model owner turn off the Inference API.
image

PR #1754, opened 2023-12-14
change to SpacyTextSplitter

kztao:


JohnJyong:
`zh_core_web_sm`  this Chinese pipeline is not suitable , we should support multi-language

kztao:
Okay, I'll close it

PR #1853, opened 2023-12-29
Refactor BaseVectorIndex delete method

waltcow:
vendor index base class should not contain any error type from particular providers


crazywoola (approved):

barnett-yuxiang (commented):
LGTM

PR #1912, opened 2024-01-04
ci: Properly cache pip packages

bowenliang123:
- the existed step for pip caching seems often miss hitting caching 
- resolve the problem of pip packages caching by configuring `actions/setup-python`, with os-aware caching support

With this PR, pip packages cache is properly reused in `Setup Python` step:
image

### DIFF
--- a/.github/workflows/api-model-runtime-tests.yml
+++ b/.github/workflows/api-model-runtime-tests.yml
@@ -31,28 +31,19 @@ jobs:
       HUGGINGFACE_EMBEDDINGS_ENDPOINT_URL: c
       MOCK_SWITCH: true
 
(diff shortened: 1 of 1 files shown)


Yeuoly (commented):
Thanks! but pytest requires single line command, otherwise there may cause a FileNotFound error

bowenliang123:
> Thanks! but pytest requires single line command, otherwise there may cause a FileNotFound error

Sure, let me revert that change. Btw, it's still a single-line command indeed.

Yeuoly (approved):
Thanks for your contributions!

takatost (approved):

PR #1948, opened 2024-01-05
fix: reranking_provider_name to str type

ethuwlwfu3288:


JohnJyong:
the dict is below format :

```
                'reranking_model': {
                    'reranking_provider_name': '',
                    'reranking_model_name': ''
                }
```
So there is no need to make such modifications.

crazywoola (changes requested):
```
'reranking_model': {
    'reranking_provider_name': '',
    'reranking_model_name': ''
}
```

PR #1956, opened 2024-01-05
fix: cover missed source paths for eslint

bowenliang123:
- NextJs only lints  files in the pages/, app/, components/, lib/, and src/ directories by default , referring to doc (https://nextjs.org/docs/app/building-your-application/configuring/eslint#linting-custom-directories-and-files)
- This PR covers the missed source paths for eslint, including /i18n , /context, etc.


bowenliang123:
cc @crazywoola @zxhlyh

crazywoola (dismissed):

iamjoel (approved):

PR #2012, opened 2024-01-12
fixup /stop api

killpanda:
1. ignore user_model from decorator `validate_app_token` since it's always None
2. get `end_user_id` from json body


takatost (approved):
lgtm

PR #2047, opened 2024-01-16
fix qdrant tag in docker-compose.yaml

takatost:
use v1.7.3 instead.


AndyMik90:
I'm getting "Error response from daemon: manifest for langgenius/qdrant:1.7.3 not found: manifest unknown: manifest unknown"

PR #2072, opened 2024-01-17
fix: bypass admin users to use dataset api with API key

bowenliang123:
- Before: only owner users successfully use datasetapi, for the admin users, it failes with `Unauthorized("Tenant is not exist.")`
- After: both owner and admin users passes

### DIFF
--- a/api/controllers/service_api/wraps.py
+++ b/api/controllers/service_api/wraps.py
@@ -2,8 +2,6 @@
 from datetime import datetime
 from functools import wraps
 
-from sqlalchemy import or_
-
 from extensions.ext_database import db
 from flask import current_app, request
 from flask_login import user_logged_in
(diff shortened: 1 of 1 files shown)


JohnJyong:
The logic here is only used to obtain owner information, not to verify permissions. @bowenliang123

bowenliang123:
> The logic here is only used to obtain owner information, not to verify permissions. @bowenliang123

Yes, but without this change, the request by admin users will fall to the failure "Unauthorized("Tenant is not exist.")".

crazywoola (approved):

takatost:
Just a reminder: 
Currently, the knowledge base API key is not tied to individual account identity. So, when you make calls to the knowledge base API, you'll be using the workspace owner's identity.

bowenliang123:
OK. How about changing the word from `allow` to `bypass` in the title?

takatost (approved):

bowenliang123:
Updated the title and description, if you check.

PR #2084, opened 2024-01-18
fix: yarn install command in web Dockerfile

bowenliang123:
- `--only=prod` is removed as `--only` is not a valid option for `yarn` or `yarn install command`
- for the `packages` stage in dockerfile, all the npm packages should be installed, as it used for building production files later in `builder` stage. and the npm packages are not shipped with `production` stage
- the `entrypoint.sh` has been set to executable since #1966

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -1,16 +1,16 @@
 # base image
 FROM node:20.11.0-alpine AS base
+LABEL maintainer="takatost@gmail.com"
 
 # install packages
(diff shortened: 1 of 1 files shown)


crazywoola (changes requested):
This file is conflicted with previous pull request regarding the node version.

bowenliang123:
will fix it soon. thanks.

crazywoola (approved):

PR #2091, opened 2024-01-19
support setting timezone in docker images

bowenliang123:
- install `tzdata` module in web docker image for timezone support
- in web docker image, set the default `TZ` system env to `UTC` and be linked to `/etc/localtime`, referring Alpine docs: https://wiki.alpinelinux.org/wiki/Setting_the_timezone
- in web docker image, set the default `TZ` system env to `UTC`

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -6,15 +6,15 @@ LABEL maintainer="takatost@gmail.com"
 RUN apt-get update \
     && apt-get install -y --no-install-recommends gcc g++ python3-dev libc-dev libffi-dev
 
-# set timezone
-ENV TZ UTC
-
 COPY requirements.txt /requirements.txt
 
 RUN pip install --prefix=/pkg -r requirements.txt
 
 # build stage
 FROM python:3.10-slim AS builder
+
+# set timezone
+ENV TZ UTC
 
 ENV FLASK_APP app.py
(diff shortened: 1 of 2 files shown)


api/Dockerfile
```diff
@@ -6,6 +6,9 @@ LABEL maintainer="takatost@gmail.com"
 RUN apt-get update \
     && apt-get install -y --no-install-recommends gcc g++ python3-dev libc-dev libffi-dev
 
+# set timezone
+ENV TZ UTC
```
takatost: I think TZ ENV should be added in the builder stage.

web/Dockerfile
```diff
@@ -1,6 +1,12 @@
 # base image
 FROM node:20.11.0-alpine AS base
 
+RUN apk add --no-cache tzdata
+
+# set timezone
+ENV TZ UTC
```
takatost: I think it should be included in the production stage.

takatost (changes requested):

api/Dockerfile
bowenliang123: Reasonable suggestion.  Moved.

web/Dockerfile
bowenliang123: Resolved.

takatost (approved):
lgtm

PR #2099, opened 2024-01-19
fix: skip linking /etc/localtime file first in api docker image

bowenliang123:
#2091 followup


takatost (dismissed):

bowenliang123:
As discussed offline, skip creating the /etc/localtime, and the Debian system respects the TZ env. Remain the TV env with explicit default value as the hint.

takatost (approved):

PR #2139, opened 2024-01-23
feat: support assistant frontend

iamjoel:


zxhlyh (approved):

zachariahtimothy:
Why was the Chat component moved from Explore? Is there a way to achieve this without custom code? We use it frequently but I cannot see how to duplicate that functionality in Studio because there is no user input for context.

PR #2171, opened 2024-01-24
Support JSONL output

JzoNgKVO:


crazywoola (dismissed):

web/app/components/app/annotation/header-opts/index.tsx
```diff
@@ -26,21 +30,36 @@ type Props = {
...
 }) => {
   const { t } = useTranslation()
   const { locale } = useContext(I18n)
   const language = getModelRuntimeSupported(locale)
   const { CSVDownloader, Type } = useCSVDownloader()
   const [list, setList] = useState<AnnotationItemBasic[]>([])
+
+  const listTransformer = (list: AnnotationItemBasic[]) => list.map(
+    (item: AnnotationItemBasic) => {
+      const dataString = `{"messages": [{"role": "system", "content": ""}, {"role": "user", "content": "${item.question.replace(/\n/g, '\\n')}"}, {"role": "assistant", "content": "${item.answer.replace(/\n/g, '\\n')}"}]}`
```
takatost: JSON encoding transforms data into a JSON format string. Besides replacing newline characters with `\n`, it also involves the following operations:

1. Escapes double quotes (`"`) as `\"`.
2. Escapes backslashes (`\`) as `\\`.
3. Transforms control characters, like tab (`\t`), into their escape sequences.
4. Converts non-printable characters into Unicode escape sequences (e.g., `\uXXXX`).

takatost (changes requested):

web/app/components/app/annotation/header-opts/index.tsx
JzoNgKVO: changed with 'JSON.stringify'.

takatost (approved):

PR #2195, opened 2024-01-25
improve api docker file and lock Debian version in base image tag

bowenliang123:
- reuse the base stage with base image
- `3.10-slim-bookworm` is equivalent tag with `3.10-slim`, with explicitly set the Debian version in the base image tag, showing the `bookworm` for reference (eg. preparing apt-get source mirror URL)
- skip reinstalling the `bash` package as the Debian slim image is shipping with it
- skip resetting `/entrypoint.sh` to executable as the source script `api/docker/entrypoint.sh` is executable

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,5 +1,5 @@
 # packages install stage
-FROM python:3.10-slim AS base
+FROM python:3.10-slim-bookworm AS base
 
 LABEL maintainer="takatost@gmail.com"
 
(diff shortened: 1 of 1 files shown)


api/Dockerfile
```diff
@@ -30,15 +32,14 @@ ENV TZ UTC
...
+    && apt-get install -y --no-install-recommends curl wget vim nodejs ffmpeg \
     && apt-get autoremove \
     && rm -rf /var/lib/apt/lists/*
 
-COPY --from=base /pkg /usr/local
+COPY --from=packages /pkg /usr/local
 COPY . /app/api/
 
 COPY docker/entrypoint.sh /entrypoint.sh
-RUN chmod +x /entrypoint.sh
```
takatost: This could be necessary, if building on a Windows host, you might run into issues with execution permissions being lost.

takatost (changes requested):

api/Dockerfile
bowenliang123: OK, reverted this change.

takatost (approved):

Issue #2210, opened 2024-01-25
As long as an administrator is assigned to the workspace, an error will be reported when calling the API.


IAMZn1018:
### Self Checks

- [X] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [X] I confirm that I am using English to file this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).

### Dify version

0.5.0 and 0.4.9

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

When there is an administrator in my workspace and I call the **knowledge** API
```
import requests

headers = {
    "Authorization": "Bearer dataset-vmFnLm3LkmAdwuLzXerAJUbq",
    "Content-Type": "application/json"
}

requests.get("http://10.254.22.222/v1/datasets?page=1&limit=20", headers=headers).json()
```  

I get the following error:
<img width="635" alt="微信图片_20240125210555" src="https://github.com/langgenius/dify/assets/24623176/82b0661e-525c-414f-a3ae-4b4bfe51309e">

After reading the source code, I found that there is a problem with the following code
![image](https://github.com/langgenius/dify/assets/24623176/0b6712e3-cf31-4427-8e19-1e43bb23862e)
**one_or_none()** means that an error will be reported when there are multiple query results.  
  
I checked the postgres database and found that the **tenant_id** of owner and admin are the same.
![image](https://github.com/langgenius/dify/assets/24623176/14539367-97fc-4b2b-9080-21183c91a47d)
After looking at the tables account, tenant_account_joins, api_tokens, and tenants, I thought the best solution is to change the code to
![image](https://github.com/langgenius/dify/assets/24623176/93581bc0-05ce-4512-9c63-5296ac8b2838)
It is completely unknown from the token which account is logging in, so there is no need to judge the account below in the code.You only need to determine whether the token was created by the owner.  

I'm not sure if the code in the cloud version is the same as on github, so my suggestion is just the code on github

### ✔️ Expected Behavior

get the success response

bowenliang123:
I've different suggestions for this problem.
Although removing ’admin‘ helps to avoid the error, it does prevent admin users to use the API.
How about changing `.one_or_none` to `.first` to bypass the multiple result rows situation?

@takatost

crazywoola:
> I'm not sure if the code in the cloud version is the same as on github, so my suggestion is just the code on github

[...]

PR #2211, opened 2024-01-25
fix:determine multiple result exceptions caused by admin

IAMZn1018:
When the table tenant_account_joins contains a user whose role is admin, calling the knowledge API will cause an error.


JohnJyong (approved):

bowenliang123:
Hey, this PR is fixing the https://github.com/langgenius/dify/issues/2210, and comes with reverting the https://github.com/langgenius/dify/pull/2072. They should be at least linked.

PR #2212, opened 2024-01-25
fix:determine multiple node exceptions caused by admin

IAMZn1018:
When the table tenant_account_joins contains a user whose role is admin, calling the knowledge base API will cause an error.I think this method is the smallest change and the most reasonable method at present.


takatost (changes requested):
Seems like a lot of unrelated frontend code got included 🤔

PR #2219, opened 2024-01-26
fix: multiple rows were found correctly

### LINKED ISSUE #2210: As long as an administrator is assigned to the workspace, an error will be reported when calling the API.

crazywoola:


crazywoola:
@IAMZn1018 @bowenliang123

bowenliang123:
LGTM. As originally in #2212, `first` should be used instead of `one_or_none`.

takatost (approved):

PR #2238, opened 2024-01-26
use pm2 to guard and monitor the web service in docker file

bowenliang123:
- pm2 is a production process manager for Node. js applications
- pm2 also provides handy access to check the status of the process


bowenliang123:
Already tested in my local environment.

takatost:
@iamjoel @zxhlyh Need to check if the environment variables are working?

iamjoel (approved):

iamjoel:
Pm2 start would take the environment variables checked.

Issue #2267, opened 2024-01-29
openai_api_compatible stream mode delimiter error


bowenliang123:
### Self Checks

- [X] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [X] I confirm that I am using English to file this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).

### Dify version

0.5.2

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. set the stream delimiter to '\n\r' for custom `qwen` 千问 72b model with openai_api_compatible provider
2. llm chat failed with (red block in the screenshot)
`Error: can only concatenate str (not "NoneType") to str`

### ❌ Actual Behavior

During steamlingly displaying the reply:
<img width="813" alt="image" src="https://github.com/langgenius/dify/assets/1935105/4986bfc2-0a61-45e9-bf9d-5fd00e5a08ed">

When the reply ends with error:
<img width="780" alt="image" src="https://github.com/langgenius/dify/assets/1935105/047cf6c1-3958-4e61-8de0-1309e64afb58">

Code:

![image](https://github.com/langgenius/dify/assets/1935105/75afb2cf-b148-4fc8-baee-f8b84589b7f5)

bowenliang123:
The orginal feature and bug introduced by https://github.com/langgenius/dify/pull/2190.
I don't know the reason that the guard conditions in purple block does not work as expectedly and I suggested a temporary workaround that works for me in blue block, for reference.

cc @geosmart @takatost

PR #2270, opened 2024-01-29
fix: remove finish_reason condition logic when deltaContent is empty

geosmart:
fix openai_api_compatible stream mode delimiter error #2267

### DIFF
--- a/api/core/model_runtime/model_providers/openai_api_compatible/llm/llm.py
+++ b/api/core/model_runtime/model_providers/openai_api_compatible/llm/llm.py
@@ -372,15 +372,9 @@ class OAIAPICompatLargeLanguageModel(_CommonOAI_API_Compat, LargeLanguageModel):
 
                 if 'delta' in choice:
                     delta = choice['delta']
-                    if delta.get('content') is None or delta.get('content') == '':
-                        if finish_reason is not None:
-                            yield create_final_llm_result_chunk(
-                                index=chunk_index,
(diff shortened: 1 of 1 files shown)


api/core/model_runtime/model_providers/openai_api_compatible/llm/llm.py
```diff
@@ -372,15 +372,9 @@ def create_final_llm_result_chunk(index: int, message: AssistantPromptMessage, f
...
-                        if finish_reason is not None:
-                            yield create_final_llm_result_chunk(
-                                index=chunk_index,
-                                message=AssistantPromptMessage(content=choice.get('text', '')),
-                                finish_reason=finish_reason
-                            )
-                        else:
-                            continue
+                    delta_content = delta.get('content', '')
+                    if delta_content == '':
```
bowenliang123:
```suggestion
                    if delta_content is None or delta_content == '':
```
Add checking None to conditions.
As in https://github.com/langgenius/dify/issues/2267, `delta.get('content', '')` still turns out to be a `None`, which only checks whether `content` key exists in `delta` dict but does not guarantee it's not a None.

geosmart: thanks,done

bowenliang123: Thanks. LGTM.

guchenhe (approved):

PR #2306, opened 2024-01-31
bump version to 0.5.3

takatost:
version to 0.5.3 and reformat python imports

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -30,7 +30,7 @@ from flask import Flask, Response, request
 from flask_cors import CORS
 from libs.passport import PassportService
 # DO NOT REMOVE BELOW
-from models import account, dataset, model, source, task, tool, web, tools
+from models import account, dataset, model, source, task, tool, tools, web
 from services.account_service import AccountService
 
 # DO NOT REMOVE ABOVE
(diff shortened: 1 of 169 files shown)


crazywoola (dismissed):

bowenliang123:
Would you like to consider applying `dev/reformat` to python import styling before the release ofevery version? 
This would be great for rebasing our self-hosted repos. Thanks.
@takatost

takatost:
> Would you like to consider applying `dev/reformat` to python import styling before the release ofevery version? This would be great for rebasing our self-hosted repos. Thanks. @takatost

What action should I take?

takatost:
> Would you like to consider applying `dev/reformat` to python import styling before the release ofevery version? This would be great for rebasing our self-hosted repos. Thanks. @takatost

BTW, this operation is quite a breaking change for many of us developers and contributors. When merging the reformatted version, we'll need to resolve a lot of conflicts. Is there a more elegant solution?

bowenliang123:
We use the style check test and the reformat tools together, which leads to a good practice in Apache Kyuubi.
The key point is that both approaches should be brought to stable outputs. And the lining tools should provide a one-key reformatting solution. The `dev/reformat` does so.

bowenliang123:
> What action should I take?

It's an optional suggestion anyway. We could apply the `dev/reformat` script and commit the changes in a new PR before cutting the branch for bumping to 0.5.3.

PR #2313, opened 2024-01-31
remove unused session store codes

takatost:


bowenliang123:
This PR introduced a bug that prevents users from switching workspaces. Consider reverting this change.

takatost:
> This PR introduced a bug that prevents users from switching workspaces. Consider reverting this change.

Yep, I noticed that the JWT doesn't store the workspace id, still using session instead. I'll roll back for now and then update it to store in the JWT later.

PR #2318, opened 2024-01-31
fix: typo

Yeuoly:


takatost (dismissed):

api/core/app_runner/assistant_app_runner.py
```diff
@@ -199,7 +199,7 @@ def run(self, application_generate_entity: ApplicationGenerateEntity,
         llm_model = cast(LargeLanguageModel, model_instance.model_type_instance)
         model_schema = llm_model.get_model_schema(model_instance.model, model_instance.credentials)
 
-        if set([ModelFeature.MULTI_TOOL_CALL, ModelFeature.TOOL_CALL]).intersection(model_schema.features):
+        if set([ModelFeature.MULTI_TOOL_CALL, ModelFeature.TOOL_CALL]).intersection(model_schema.features or []):
```
bowenliang123: 1. this is not a typo, which would be better to get fixed in a separate PR, if you like. Small PR is more traceable.
2. it's suggested to apply a replacement in the default value for the field `features` of the class `ProviderModel` from `None` to an empty list, instead of patching a magic value `[]` here and in other usage of reading this field.

Yeuoly: 1. I found this issue when testing the branch and fixed it together in this branch.
2. In order to ensure minimal impact to modules owned by other developers, we prefer using `or []` instead. btw, the reason why it's None is the parent class of model schema is pydantic.BaseModel, it prefer to set an Optional fields to None when constructing an Object if it's an empty array, and there are also a method called `_get_customizable_model_schema` in model_runtime which would return an object has None features. There are many similar problems with pydnatic, and many places will become None because of the yaml configuration

Thanks for the suggestions.

bowenliang123: OK. Thanks. I got your point in No. 2 reasoning.
For the point no.1 , I have also encountered locally and patched in my repo.

Still strongly recommended to use small PRs for different purposes. It's always a better way for readability and agile development.

The minimum acceptance to the changes should be mentioned and covered in the tile or the description if you Dify community agrees it's a bottom line.

bowenliang123: LGTM overall. Thanks for explaination.

Yeuoly: revert.

Thanks

crazywoola (approved):

api/core/app_runner/assistant_app_runner.py
Yeuoly: solved #2330

PR #1983, opened 2024-01-10
improve: introduce isort for linting Python imports

bowenliang123:
- introduce `isort` (https://pycqa.github.io/isort/index.html) for linting the Python imports
- provide `dev/reformat` script for one-key fixing Python import styles with `isort`
- the isort package is treated as dev dependency and not added to `requirements.txt` as runtime dependency
- ~~apply `isort` checks via superlinter in style workflow~~
- put the commit id of applying isort fixes into `.git-blame-ignore-revs`, so that git blame and github ignores this changes

PR #2335, opened 2024-02-01
automate isort styling by adding to pre-commit hooks

bowenliang123:
- as #1983 introduced isort and one-key linting scripts for linting Python imports, it helps to conform a steady output in sorted imports within a unified style.
- this PR add the isort script into Husky for git's pre-commit hooks, which is to be executed automatically before each commit
- This will greatly help the developer out of the dirty work when rebasing onto the new release and branches without worrying the conflict in Python imports.

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -48,4 +48,5 @@ jobs:
           VALIDATE_BASH_EXEC: true
           VALIDATE_GITHUB_ACTIONS: true
           VALIDATE_DOCKERFILE_HADOLINT: true
+          VALIDATE_PYTHON_ISORT: true
           VALIDATE_YAML: true
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -25,6 +25,16 @@ jobs:
           cache: yarn
           cache-dependency-path: ./web/package.json
 
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.10
+
+      - name: Python style check
+        run: |
+          pip install isort
+          isort --check-only --settings ./.github/linters/.isort.cfg --gitignore ./
+
       - name: Web dependencies
         run: |
           cd ./web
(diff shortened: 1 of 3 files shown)


bowenliang123:
Hi @takatost , this PR has been rebased onto latest commit, and it's ready to go.

takatost:
What happens if the developer only modifies the frontend code without having the Python environment or without installing `isort`? 
In that case, the pre-commit hook won't be able to run, I think.

takatost:
`
#!/usr/bin/env sh

# get the list of modified files
files=$(git diff --cached --name-only)

api_modified=false
web_modified=false

for file in $files
do
    if [[ $file == "api/"* && $file == *.py ]]; then
        # set api_modified flag to true
        api_modified=true
    elif [[ $file == "web/"* ]]; then
        # set web_modified flag to true
        web_modified=true
    fi
done

# check if api or web directory is modified
if $api_modified; then
    echo "do something for api"
fi

if $web_modified; then
    echo "do something for web"
fi
`

This script can determine what subsequent actions need to be taken by retrieving file modification records. For example, if the file is located in the "api" directory and has a .py extension, it will execute the "reformat" command.

crazywoola (approved):

wlbksy:
why not bring in `black` as well?

bowenliang123:
Several popular Python linter tools were considered, including black, isort, ruff and etc.
As dify community requested minimum impact to existed code for commit history traceability, `isort` is introduced to sort and refomrated the Python's imports solely.

In the future, I would prefer introduce the `ruff` for general Python linter for compatible for Flake8,black and isort. And `ruff` is extremely fast which is implemented in Rust.

Ffdjki (approved):

bowenliang123:
Closing this PR.
`ruff` seems to to be a much better option for lining Python imports.
As we want to apply Python imports linting in pre-commit stage, we'd better choose some tools like `ruff` does.
It's extremely fast.

PR #2341, opened 2024-02-01
Feat/move tenant id into db

crazywoola:
This aims to remove the session usage in our code base, but to set this in jwt token is not that good, because we do not change that token very often.


crazywoola:
@bowenliang123 @charli117  Please take a look at this if you have some spare time.

api/controllers/console/auth/login.py
```diff
@@ -30,11 +30,6 @@ def post(self):
         except services.errors.account.AccountLoginError:
             return {'code': 'unauthorized', 'message': 'Invalid email or password'}, 401
 
-        try:
```
crazywoola, line 33: Since we already setup the account in `account_service` while creating the `tenant`, there is no need to do this anymore.

api/services/account_service.py
```diff
@@ -277,18 +256,19 @@ def get_current_tenant_by_account(account: Account):
     @staticmethod
     def switch_tenant(account: Account, tenant_id: int = None) -> None:
         """Switch the current workspace for the account"""
-        if not tenant_id:
-            tenant_account_join = TenantAccountJoin.query.filter_by(account_id=account.id).first()
-        else:
-            tenant_account_join = TenantAccountJoin.query.filter_by(account_id=account.id, tenant_id=tenant_id).first()
+
+        TenantAccountJoin.query.filter_by(account_id=account.id).update({'current': False})
```
crazywoola: Update all records of current account to false.

api/models/account.py
```diff
@@ -1,6 +1,5 @@
 import enum
 import json
```
charli117, line 2: json is no longer in use and can be deleted, the rest is fine

charli117 (approved):

api/services/account_service.py
```diff
@@ -277,18 +256,19 @@ def get_current_tenant_by_account(account: Account):
...
+        tenant_account_join = TenantAccountJoin.query.filter_by(account_id=account.id, tenant_id=tenant_id).first()
 
         # Check if the tenant exists and the account is a member of the tenant
         if not tenant_account_join:
             raise AccountNotLinkTenantError("Tenant not found or account is not a member of the tenant.")
```
bowenliang123: How about moving the looking-up to in front of updating all the current accounts to false?

api/migrations/versions/16830a790f0f_.py
```diff
@@ -0,0 +1,32 @@
...
+# revision identifiers, used by Alembic.
+revision = '16830a790f0f'
+down_revision = '380c6aa5a70d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('tenant_account_joins', schema=None) as batch_op:
```
takatost, line 21: What about moving `current_tenant_id` to `accounts`? 
Setting a flag(current) on `tenant_account_joins` record may result in multiple data being `true` at the same time, lacking atomicity. However, setting `accounts.current_tenant_id` can ensure that an account is only bound to one current tenant.

takatost (changes requested):

takatost:
If we switch suddenly, will the data on the page still be in the context of the other workspace while the backend context has already switched to the workspace where the owner is?

takatost (dismissed):

api/services/account_service.py
```diff
@@ -277,18 +257,24 @@ def get_current_tenant_by_account(account: Account):
...
+            try:
+                tenant_account_join = TenantAccountJoin.query.filter_by(account_id=account.id, tenant_id=tenant_id).first()
+                TenantAccountJoin.query.filter_by(account_id=account.id).update({'current': False})
 
-        # Set the current tenant for the account
-        account.current_tenant_id = tenant_account_join.tenant_id
-        session['workspace_id'] = account.current_tenant.id
+                # Check if the tenant exists and the account is a member of the tenant
+                if not tenant_account_join:
+                    raise AccountNotLinkTenantError("Tenant not found or account is not a member of the tenant.")
```
bowenliang123: Better to move to in front of starting the db transaction.
If no tenant account was found, why not just return the error and pervert create and rollback the transaction?

takatost (approved):

bowenliang123 (approved):
LGTM.

PR #2349, opened 2024-02-02
fix [baichuan] Error: argument of type 'NoneType' is not iterable

baiyansong:
When the Baichuan model API is called frequency_penalty parameters must be between 1 and 2, otherwise an error will be reported, and abnormal data will be returned, resulting in the error: [baichuan] Error: argument of type 'NoneType' is not iterable.


crazywoola:
Hello thank you for your pr, could you add my wechat: crazyphage

crazywoola:
Please merge code into main branch instead.

PR #2371, opened 2024-02-04
add support mysql and use pymysql . all test local

golangpkg:
增加了个 type 类型，可以区分 mysql 和 postgres 了。

其他参数都可以支持，就是看 ID 是否可以替换成：

default=lambda: uuid.uuid4()  这样就可以兼容 mysql 和 PG 了。本地测试一些创建功能是没有问题的。

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -2,10 +2,12 @@
 import os
 
 import dotenv
+import logging
 
 dotenv.load_dotenv()
 
 DEFAULTS = {
+    'DB_TYPE': 'postgres',
     'DB_USERNAME': 'postgres',
     'DB_PASSWORD': '',
     'DB_HOST': 'localhost',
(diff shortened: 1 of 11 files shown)


crazywoola:
Please add my wechat: crazyphage

api/config.py
```diff
@@ -140,7 +142,17 @@ def __init__(self):
...
 
-        self.SQLALCHEMY_DATABASE_URI = f"postgresql://{db_credentials['DB_USERNAME']}:{db_credentials['DB_PASSWORD']}@{db_credentials['DB_HOST']}:{db_credentials['DB_PORT']}/{db_credentials['DB_DATABASE']}{db_extras}"
+        # db_type = postgresql or mysql
+        # https://docs.sqlalchemy.org/en/20/core/pooling.html
+        db_type = get_env('DB_TYPE')
+        self.SQLALCHEMY_DATABASE_TYPE = db_type
+        if "mysql" == db_type:
+            db_type = "mysql+pymysql"
+
+        self.SQLALCHEMY_DATABASE_URI = f"{db_type}://{db_credentials['DB_USERNAME']}:{db_credentials['DB_PASSWORD']}@{db_credentials['DB_HOST']}:{db_credentials['DB_PORT']}/{db_credentials['DB_DATABASE']}{db_extras}"
```
takatost, line 152: Maybe we can use `db_driver` instead of `db_type`?

takatost (changes requested):
Any requirements for the MySQL version? 
We're using PostgreSQL's `uuid4` and `jsonb` features, so MySQL might need a higher version to support them.

api/config.py
golangpkg: OK。

golangpkg: OK。

takatost:
Based on our communication, here's the summary:
1. We hope that other developers don't need to worry about the differences in the underlying database when declaring `Model` class and writing `migrations`. Every developers use the same set of conventions.
2. If there's a need for pure SQL execute, se compatible syntax as much as possible.
3. If there are some PgSQL-specific syntax in codes, it may be necessary to modify the codes to support different databases.

quicksandznzn:
+1

kaktos:
+1

takatost:
Sorry, but due to the lack of response for a long time and MySQL not being in our roadmap, this PR will be closed.

simplast:
+1

Bhawna07:
Can you help me? I want to use a PostgreSQL database as an external knowledge base through endpoints for chatbot. Is that possible with dify cloud?

PR #2390, opened 2024-02-05
add built-in maths tool for local expression evaluation on NumExpr

bowenliang123:
Introduce a built-in tool for maths expression evaluation, running locally on `numexpr` (already required by Numpy as well).

- NumExpr is a fast numerical array expression evaluator for Python, NumPy, Pandas
- NumExpr's `evaluate` method is used for math expression evaluation, doc: https://numexpr.readthedocs.io/en/latest/user_guide.html#usage-notes
  - supported operators: https://numexpr.readthedocs.io/en/latest/user_guide.html#supported-operators
  - supported functions: https://numexpr.readthedocs.io/en/latest/user_guide.html#supported-functions 

The incorrect result on GLM3 without maths tool:
企业微信截图_1a8f1273-441c-4855-9298-46a46ce7214b

The correct result on GLM3 with maths tool:
企业微信截图_2d6197c8-9c93-4a54-b1ba-a06b22c657f8

image


takatost:
We need to check if this solution has any security risks. @Yeuoly

crazywoola (dismissed):

bowenliang123:
> We need to check if this solution has any security risks. @Yeuoly

Thanks. 
To avoid possible injection risk, NumExpr's `evaluate` is reused for expression evaluation, rather than Python's raw `eval()`. It parses expressions in maths and rejects incorrect violations in syntax.NumExpr is also used by Numpy.

Yeuoly:
> > We need to check if this solution has any security risks. @Yeuoly
> 
> Thanks. To avoid possible injection risk, NumExpr's `evaluate` is reused for expression evaluation, rather than Python's raw `eval()`. It parses expressions in maths and rejects incorrect violations in syntax.NumExpr is also used by Numpy.

Yep, use NumExpr is a better solution rather than native Python, the expression parser is written in C and compiled into native codes, will not cause code injection security  risk.

crazywoola (dismissed):

takatost (approved):

PR #2408, opened 2024-02-06
fix recreating users' default tenant relations when loading user

bowenliang123:
- fix bugs in `load_user` introduced by #2341
- 1. skip auto creating default tenant in this method. for the reasons that 
    - 1.1 it causes leak as it does not acquire locks to guarantee only one default tenant will be created
    - 1.2 it's not a right timing to create user's default tenant. A better way is to do it in register service.
- 2. reuse existed available tenant for the user

Caused by 1.2.
image

### DIFF
--- a/api/services/account_service.py
+++ b/api/services/account_service.py
@@ -56,18 +56,19 @@ class AccountService:
 
         if account.status in [AccountStatus.BANNED.value, AccountStatus.CLOSED.value]:
             raise Forbidden('Account is banned or closed.')
-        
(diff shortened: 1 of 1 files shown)


bowenliang123:
cc @crazywoola @takatost

takatost:
We need to create a personal tenant as the owner when invited to another tenant. 
We need to make this logic before we can merge together.

takatost (approved):
LGTM

PR #2419, opened 2024-02-08
chore: apply ruff's pyupgrade linter rules to modernize Python code with targeted version

bowenliang123:
- to modernize Python code with the targeted version
- `pyupgrade` rules in Ruff: https://docs.astral.sh/ruff/rules/#pyupgrade-up
  - to prefer native builtin collections over from typing, e.g. `list` instead of `List`
  - to prefer collections from `collections.abc` (abstract base class container from Python3.3) over from `collections`
  - skip UTF-8 encoding in the headline, as it's already the default encoding in Python
  - prefer `yield from` in generator
- ignoring the following rules
  - UP007: Use X \| Y for type annotations
    - to prevent enforcing `Unin[type1, type2]` into `type1, type2` or breaking `Optional[]`
  - UP032: Use f-string instead of format call
- Ruff respects the required Python version from config file, which is `>=3.10`.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -1,4 +1,3 @@
-# -*- coding:utf-8 -*-
 import os
 
 from werkzeug.exceptions import Unauthorized
--- a/api/config.py
+++ b/api/config.py
@@ -1,4 +1,3 @@
-# -*- coding:utf-8 -*-
 import os
 
 import dotenv
--- a/api/controllers/console/app/app.py
+++ b/api/controllers/console/app/app.py
@@ -1,4 +1,3 @@
-# -*- coding:utf-8 -*-
 import json
 import logging
 from datetime import datetime
--- a/api/controllers/console/app/audio.py
+++ b/api/controllers/console/app/audio.py
@@ -1,4 +1,3 @@
-# -*- coding:utf-8 -*-
 import logging
 
 from flask import request
--- a/api/controllers/console/app/completion.py
+++ b/api/controllers/console/app/completion.py
@@ -1,7 +1,7 @@
-# -*- coding:utf-8 -*-
 import json
 import logging
-from typing import Generator, Union
+from collections.abc import Generator
+from typing import Union
 
 import flask_login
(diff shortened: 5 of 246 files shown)


takatost:
Is it possible to add all the necessary constraints at once, like "E": PEP 8? 
Making such a large number of file changes each time puts a lot of pressure on our branch merging/rebasing during developing.

bowenliang123:
Hi, @takatost. Thanks for reviewing this PR.
And it's possibly the last batche of necessary rules we required.

There're some clearification in the objectives and aproaches for introducing the linter rules.

The objectives with priority as follows:
0. linting over formatting
1. avoiding misuses or bugs in code is the priority, which brings concret improvements in security and readibility
2. no goal to enforcing the specs or suggestions of pure linting style, eg. PEP 8 suggests double quote for strings
3. provide a executable and progressive auto-fixable tools and plan to deliver the improvements in each concerned aspects, eg. the Ruff for the tool, and the batches of rules for different clear purposes

Looking back what we have done and we are doing for these objectives in 4 major steps:
0. prepare proper linter tool, the Ruff linter (not the formatter of Ruff)
1. sort and eliminate the unused imports, which improves the quality of imports and avoids duplicate imports
2. introduce the pyflake rules (also partly conforms the spec PEP 8) for linting the possible risks in Python code
3. fix the mixed usage of legacy and modern usage in Python syntax and packages, aligning with the same standard of Python versions, e.g. preferring the builtin `list` over the `typing.List` and mixing usage in them 


This PR is the forseeable last step here as we listed.


As for the developing or rebasing braches, it does provide a executable way to conform the introdueced rules. Either of the following way would help.
- use or copy the `pyproject.yaml` for rules, and replay the linting locally before merging
- merging the latest commit of mater branch, and merging all the changes in the developing branches, run `dev/reformat` with the merged code with introdueced rules. It garuntees a stable output for fixable out

takatost (approved):
LGTM

PR #2420, opened 2024-02-08
chore: apply ruff's pyflakes linter rules

bowenliang123:
- `pyflake` rules in ruff: https://docs.astral.sh/ruff/rules/#pyflakes-f
  - fix 1 missing placeholder `format` for format string in `libs/helper`
 ```
error = ('Invalid {arg}: {val}. {arg} must be conform to the format {format}'
               .format(arg=self.argument, val=value, format=self.format))
```
  - flattern unnecessary string formatting into static stirngs
- ignoring the following rules
  - to allow star imports
    - F403: undefined-local-with-import-star
    - F405: undefined-local-with-import-star-usage
  - to allow undefined name, F821, for the class method returning types of the class itself, e.g.
    ```
    class PromptType:
     @classmethod
        def value_of(cls, value: str) -> 'PromptType':
     ```
  - to allow unused variable, F841

### DIFF
--- a/api/controllers/console/app/app.py
+++ b/api/controllers/console/app/app.py
@@ -133,8 +133,8 @@ class AppListApi(Resource):
 
                 if not model_instance:
                     raise ProviderNotInitializeError(
-                        f"No Default System Reasoning Model available. Please configure "
-                        f"in the Settings -> Model Provider.")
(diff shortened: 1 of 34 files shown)


takatost (approved):
LGTM

PR #2422, opened 2024-02-08
fix: possible unsent function call in the last chunk of streaming response in OpenAI provider

bowenliang123:
- fix possible unsent function call in the last chunk of streaming response in OpenAI provider
- the `delta_assistant_message_function_call_storage` caches the function calls when function_call is found and then it continues to process the next chuck. However, if the last message itself contains a full function calling, it's skipped and never sent
- the corner case comes with using OpenAI provider to call the LLMs under Open-compatible proxies, like we have GLM3 and Qwen

### DIFF
--- a/api/core/model_runtime/model_providers/openai/llm/llm.py
+++ b/api/core/model_runtime/model_providers/openai/llm/llm.py
@@ -497,8 +497,9 @@ class OpenAILargeLanguageModel(_CommonOpenAI, LargeLanguageModel):
                 continue
 
             delta = chunk.choices[0]
+            has_finish_reason = delta.finish_reason is not None
 
-            if delta.finish_reason is None and (delta.delta.content is None or delta.delta.content == '') and \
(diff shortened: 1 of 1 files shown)


bowenliang123:
cc @Yeuoly @takatost

Yeuoly (approved):

Yeuoly:
BTW, openai_api_compatible will support function call soon, you can then use it as your proxies of GLM3 and Qwen.

Thanks! best wishes to your new year.

bowenliang123:
Thanks. We probably continue to use the openai provider to connect the LLMs via the openai compatible proxies.
1. for the earlier support of new features, Dify often consider OpenAI services first.
2. a set of predefined models of OpenAI helps us to provide all the selected services to out internal users, instead of forcing them to add the LLMs one by one in every workspace.

I raised this PR against OpenAI provider, for the reason that it does wrongly skipped the last message of functional call. Especially in the future, OpenAI models may be able to return the arguments with the function calling at one message.

And Happy new Year!

PR #2423, opened 2024-02-08
enhancement: speed up app lauching by lazy importing transformers

bowenliang123:
- importing the `transformers` package cost much amount of the app launching time, while it's not a required pre-load package for API serivce
- last importing `transformers` package as where it's used to speed up app launching

- before:
```
[app.py][100] App started in 7.410 seconds
```

- after:
```
[app.py][101] App started in 2.511 seconds
```

### DIFF
--- a/api/core/model_runtime/model_providers/__base/tokenizers/gpt2_tokenzier.py
+++ b/api/core/model_runtime/model_providers/__base/tokenizers/gpt2_tokenzier.py
@@ -2,8 +2,6 @@ from os.path import abspath, dirname, join
 from threading import Lock
 from typing import Any
 
-from transformers import GPT2Tokenizer as TransformerGPT2Tokenizer
-
 _tokenizer = None
 _lock = Lock()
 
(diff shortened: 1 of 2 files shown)


bowenliang123:
cc @crazywoola @takatost

takatost:
Will it cause a high delay for the first request using the `transformer`?

bowenliang123:
The loading time should be postponed to when it is used for the first time.
And I set the break points to the import statement when debugging locally; no one was touched or stopped from rebooting the app.
So the main purpose of this PR is to minimize the downtime of the API service.

takatost:
> The loading time should be postponed to when it is used for the first time. And I set the break points to the import statement when debugging locally; no one was touched or stopped from rebooting the app. So the main purpose of this PR is to minimize the downtime of the API service.

I don't think so. The long loading time can be solved by implementing instance rotation strategy, and it's not a good user experience to make users to bear with this long delay when first request. If there are multiple instances rotating, then there will definitely be several user who have to bear this delay before they can smoothly make requests.

takatost:
Thank you for your contribution. Due to the fact that the tokenizer in the transformer is a necessary method for computing tokens in most models, moving the import process to runtime would result in a delay for the first request involving token computation, which would lead to a poor user experience. 
We will close this PR. Sorry for the inconvenience.

bowenliang123:
Okay, thanks.

PR #2431, opened 2024-02-11
FIX: List from typing is not imported in account service, cause error…

Yash-1511:
Error while starting Celery Services or API. 
Screenshot 2024-02-10 195501


takatost:
Thank you so much!
However, due to the current lint rules(#2419), we are required to use lowercase "list" instead of "List". As a result, "List" has been deprecated. I have already submitted a new PR #2434 to fix this issue. 
I will close this PR. Thanks again for your submission. 🙏

PR #2435, opened 2024-02-11
fix: account has no owner workspace by member inviting

takatost:
- If the current account is not associated with any existing team, it will log out directly. 
- If a disconnected account logs in next time, it will automatically create a team with owner permissions. 
- If an invitation is sent to a non-existent email user, it will automatically create the user and a team with owner permissions.


takatost:
related with #2408

Issue #2352, opened 2024-02-02
Historical Configuration Management and Evaluation


wlbksy:
### Self Checks

- [X] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [X] I confirm that I am using English to file this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).

### 1. Is this request related to a challenge you're experiencing?

In the current dify code, each time the publish button is clicked on the app configuration page, the current app configuration is saved in the datatable `app_model_configs`, which can be considered as a commit of the configuration.

However, it does not meet the following needs:
1. Sometimes we want to revert to an old configuration commit, but currently dify lacks tools for this.
2. Sometimes we want to evaluate two historical configuration commits, and need to be able to call an API for each of the configurations of two different commits.

### 2. Describe the feature you'd like to see

1. Have a webpage for managing historical commits
    - Display all historical commits/tags/release
    - Tag or release a historical commit, fill in a description
    - Show diff between two commits like `git diff`

2. backend API provide a way to call to make AI task for a specific commit/tag/release configuration, defaulting to the latest commit configuration.

### 3. How will this feature improve your workflow or experience?

It would make Historical Configuration useful both for version control and evaluation

## Proposal

1. Add tag or release fields to the `app_model_configs` datatable, and add a description field.
2. Add an API interface for adding tag/release/description to a configuration commit.
3. Add a webpage search interface, which returns all related configurations of provided configuration commit/tag/release ID, along with the datasets/documents/upload_files/tools/tool_api_providers used by this commit (obtained by joining through id across datatables)
4. Add a webpage to display a historical configuration
5. Add a tag/release management page
5. Add a diff page between two configurations.

### 5. Can you help us with this feature?

- [ ] I am interested in contributing to this feature.

VincePotato:
Good suggestion! We are considering adding version control to our app builder(studio). Thanks for your useful feedback.

wlbksy:
> Good suggestion! We are considering adding version control to our app builder(studio). Thanks for your useful feedback.

What's your opinion about the idea of a call for a specific version, so that we users could evaluate and compare two versions, helping us make decision about which version is good to go?

wlbksy:
bot being fooling around likes it's blind on the ref from the other PR

wlbksy:
bad bot

wlbksy:
bad bot

PR #2465, opened 2024-02-18
feat: use a specific version of model config

wlbksy:
Make the backend API use an `app_model` specified by `app_config_model_id`, so that a specific version of the model config can be called without worrying about the config being changed afterward.

### DIFF
--- a/api/controllers/console/app/completion.py
+++ b/api/controllers/console/app/completion.py
@@ -49,6 +49,7 @@ class CompletionMessageApi(Resource):
         parser.add_argument('model_config', type=dict, required=True, location='json')
         parser.add_argument('response_mode', type=str, choices=['blocking', 'streaming'], location='json')
         parser.add_argument('retriever_from', type=str, required=False, default='dev', location='json')
+        parser.add_argument('app_model_config_id', type=str, required=False, location='json')
         args = parser.parse_args()
 
         streaming = args['response_mode'] != 'blocking'
@@ -124,6 +125,7 @@ class ChatMessageApi(Resource):
         parser.add_argument('conversation_id', type=uuid_value, location='json')
         parser.add_argument('response_mode', type=str, choices=['blocking', 'streaming'], location='json')
         parser.add_argument('retriever_from', type=str, required=False, default='dev', location='json')
+        parser.add_argument('app_model_config_id', type=str, required=False, location='json')
         args = parser.parse_args()
 
         streaming = args['response_mode'] != 'blocking'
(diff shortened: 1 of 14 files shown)


crazywoola:
Please add my wechat: crazyphage I will invite you to our contributors' group

takatost:
Thank you so much for your contribution! 😊
I have questions:

What is the purpose of this change and I'm not sure if it's appropriate to add this logic to all APIs.

wlbksy:
> What is the purpose of this change

This is a pr towarding https://github.com/langgenius/dify/issues/2352. A studio app may have been modified several times, however sometimes we need a API call for a fixed version of configuration such that no further modification would interfere with this version.

User case: After I pushed a app, I would like provide a way for external service to call the API of this dify app, and I want to make sure that the configuration does not change even if I made some change to the app. As long as the version is given, the external service is expecting a steady response.


> I'm not sure if it's appropriate to add this logic to all APIs.

I made it default to latest configuration, so all the API would see no diff for now. However, discussions and suggestions on this are welcomed.

takatost:
> > What is the purpose of this change
> 
> This is a pr towarding #2352. A studio app may have been modified several times, however sometimes we need a API call for a fixed version of configuration such that no further modification would interfere with this version.
> 
> User case: After I pushed a app, I would like provide a way for external service to call the API of this dify app, and I want to make sure that the configuration does not change even if I made some change to the app. As long as the version is given, the external service is expecting a steady response.
> 
> > I'm not sure if it's appropriate to add this logic to all APIs.
> 
> I made it default to latest configuration, so all the API would see no diff for now. However, discussions and suggestions on this are welcomed.

BTW, next month we will unify the design of the app version. This logic will be incompatible with the current design(by app_model_config_id). My suggestion is to keep this logic locally first, not merge it into main. When our version management goes online, we can replace it with our unified solution.

wlbksy:
> > > What is the purpose of this change
> > 
> > 
> > This is a pr towarding #2352. A studio app may have been modified several times, however sometimes we need a API call for a fixed version of configuration such that no further modification would interfere with this version.
> > User case: After I pushed a app, I would like provide a way for external service to call the API of this dify app, and I want to make sure that the configuration does not change even if I made some change to the app. As long as the version is given, the external service is expecting a steady response.
> > > I'm not sure if it's appropriate to add this logic to all APIs.
> > 
> > 
> > I made it default to latest configuration, so all the API would see no diff for now. However, discussions and suggestions on this are welcomed.
> 
> BTW, next month we will unify the design of the app version. This logic will be incompatible with the current design(by app_model_config_id). My suggestion is to keep this logic locally first, not merge it into main. When our version management goes online, we can replace it with our unified solution.

Good to here that. Any roadmap or updates that I could get more info on this design. I'm quite keen on this feature.

VincePotato:
Thanks for your feedback! In April, we will introduce support for app version control, which will include saving app drafts and restoring published versions. We hope this solves your problem.

takatost:
This feature will be replaced by the forthcoming release publish management, so we're closing it here. Thanks a lot!

PR #2765, opened 2024-03-08
SMTP authentication is optional

lmagnien:
# Description

Not all SMTP servers require authentication. It should be possible to configure SMTP servers to use either authenticated or unauthenticated SMTP

Fixes #2764

## Type of Change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [X] I rebuild the docker image and configure an smtp server without username/password, then add a user from the gui and verify that the user receive the email.
- [X] also check the log : Task tasks.mail_invite_member_task.send_invite_member_mail_task[ae4e5264-3990-4370-bd3b-5f9c1e3cbd38] succeeded in 0.2560313250000945s: None

# Suggested Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods
- [ ] `optional` I have made corresponding changes to the documentation 
- [ ] `optional` I have added tests that prove my fix is effective or that my feature works
- [ ] `optional` New and existing unit tests pass locally with my changes

PR #2937, opened 2024-03-22
Authentication is only applied when both the username and password have values.

LeoQuote:
# Description

`no auth` in smtp is implemented in https://github.com/langgenius/dify/pull/2765 , but there's a bug, [SMTP.sendmail](https://docs.python.org/3/library/smtplib.html#smtplib.SMTP.sendmail) requires the first sender address as the first argument.

When username is set to none, the auth is skipped but `sendmail` would fail due to the type error

When username is set to empty string, the auth is skipped , mail was sent, but this mail envelope does not include sender info, which could lead to rejection by server.

So the correct configuration for `no-auth` smtp would be 

```
SMTP_USERNAME=no-reply@example.com
SMTP_PASSWORD="" # or just dont set
```

In the current logic, if a username is provided, authentication will be performed, which is not as expected. Therefore, the logic of authentication at this point needs to be modified from authenticating when a username is present to authenticating only when both username and password are provided.

## Type of Change

# How Has This Been Tested?

- [x] import SMPTClient and try to send a mail

```
>>> from smtp import SMTPClient
>>> c = SMTPClient("mail.example.com", 25, "no-reply@example.com", "", "no-reply@example.com")
>>> c.send({"subject": "test", "to": "my-box@example.com", "html": "<p>hello</p>"})
```

And I can receive the mail.

# Suggested Checklist:

- [x] `optional` New and existing unit tests pass locally with my changes

### DIFF
--- a/api/libs/smtp.py
+++ b/api/libs/smtp.py
@@ -16,7 +16,7 @@ class SMTPClient:
         smtp = smtplib.SMTP(self.server, self.port)
         if self._use_tls:
             smtp.starttls()
-        if (self.username):
+        if self.username and self.password:
             smtp.login(self.username, self.password)
         msg = MIMEMultipart()
         msg['Subject'] = mail['subject']


crazywoola:
Hello can you add my wechat crazyphage I will invite you to the contributors' group

crazywoola (approved):

crazywoola (approved):

PR #2957, opened 2024-03-25
fix: progress bar issue

yikayiyo:
# Description

On the `/datasets` page, clicking on a delete icon triggers the unexpected appearance of the progress bar. The latest version of the `next-nprogress-bar` package provides a solution to this issue. However, it might be worth considering redesigning the `DatasetCard` component, as encapsulating numerous elements within a single `Link` component is not considered a best practice.

## Type of Change

- [x] Dependency upgrade


web/yarn.lock
crazywoola, line 1: There is no actual changes of package.json, so this should not be committed into the vc system.

crazywoola (changes requested):

crazywoola:
Hello please add my wechat crazyphage I will invite you to the contributors' group.

web/yarn.lock
yikayiyo: I ran `yarn upgrade next-nprogress-bar` and it didn't change the package.json file. Should I update the package.json file?

crazywoola: Yes, in this case you should change the deps as well.

yikayiyo: I've updated the deps.

crazywoola (approved):

Issue #2679, opened 2024-03-04
Add OpenRouter as a model provider


richards199999:
### Self Checks

- [X] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [X] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [X] Pleas do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing?

[OpenRouter](https://openrouter.ai/) support [a large amount of models](https://openrouter.ai/docs#models), including many open source and several close source models. Even though we already have together.ai as a model provider, there still has a lot more models on OpenRouter.

![OpenRouter](https://github.com/langgenius/dify/assets/142148415/a335eebb-cca1-4956-b35f-bb4d41c20d4b)

### 2. Describe the feature you'd like to see

I hope dify.ai Team can consider adding OpenRouter as a new model provider.

### 3. How will this feature improve your workflow or experience?

It can let users have more options and more models to use.

### 5. Can you help us with this feature?

- [X] I am interested in contributing to this feature.

Korayem:
@richards199999 wooot!

PR #3042, opened 2024-03-30
feat (new llm): add support for openrouter

Korayem:
Code mainly copied from togetherai llm provider

Covers feature request #2679


Yeuoly:
It would be better to consider changing the icon, as the current ones do not align well with our overall style, and the square icon seems to be completely transparent.
image
image

Korayem:
> @Yeuoly  It would be better to consider changing the icon, as the current ones do not align well with our overall style, and the square icon seems to be completely transparent. 

Check my last commit, I revised both
image

api/core/model_runtime/model_providers/openrouter/openrouter.yaml
```diff
@@ -0,0 +1,75 @@
...
+  en_US: openrouter_square.svg
+icon_large:
+  en_US: openrouter.svg
+background: "#F1EFED"
+help:
+  title:
+    en_US: Get your API key from openrouter.ai
+    zh_Hans: 从 openrouter.ai 获取 API Key
+  url:
+    en_US: https://api.openrouter.ai/
```
Yeuoly: This URL does not work for me, is it https://openrouter.ai/keys?

Yeuoly (changes requested):

api/core/model_runtime/model_providers/openrouter/openrouter.yaml
Korayem: oh you're right! sorry for that... fixed it in the new commit

Yeuoly: Emmm, it looks like it's incorrect also
image

Korayem: Sorry I added `/key` but didn't remove the subdomain part of the url

Please check now

Yeuoly (approved):
LGTM

PR #3122, opened 2024-04-06
Main

iimVishu:
Changes made:

Corrected a typo in a comment: "slow" instead of "slow".
Changed output: 'standalone' to output: { standalone: true } as an object is expected.
Removed an unnecessary comma after the last item in the redirects array.


guchenhe:
This doesn't seem like a productive PR; You've also committed the yarn.lock.

iimVishu:
I apologize, sir. It was my first attempt, so I made a mistake. Could you
please provide guidance so that I can avoid making mistakes in the future.

PR #3490, opened 2024-04-15
Removed the mandatory check in the code that the username and password of the milvus configuration are not empty

sqj8899:
# Description

This change allows the username and password for Milvus to be optional, removing the mandatory validation that previously caused errors when these credentials were left blank. This adjustment aims to improve usability and flexibility in scenarios where Milvus does not require authentication, ensuring the application's smooth operation without unnecessary hindrances.

Fixes # (issue)

## Type of Change

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

You may initiate the code to verify the connection with a default empty Milvus instance.


bowenliang123:
-1 for removing mandatory check in required username and password of Milvus.
1. fast failing for knowing missing username and password
2. Milvus provides username and password by default in deployment

PR #3570, opened 2024-04-17
fix(openai_api_compatible): fixing the error when converting chunk to json

aniaan:
# Description
fixing the error when converting chunk to json

Fixes # (issue)

## Type of Change

- [ ] Dependency upgrade

# How Has This Been Tested?

- [ ] TODO

# Suggested Checklist:

### DIFF
--- a/api/core/model_runtime/model_providers/openai_api_compatible/llm/llm.py
+++ b/api/core/model_runtime/model_providers/openai_api_compatible/llm/llm.py
@@ -154,7 +154,7 @@ class OAIAPICompatLargeLanguageModel(_CommonOAI_API_Compat, LargeLanguageModel):
                 json_result['object'] = 'chat.completion'
             elif (completion_type is LLMMode.COMPLETION and json_result['object'] == ''):
                 json_result['object'] = 'text_completion'
-                
+
             if (completion_type is LLMMode.CHAT
                     and ('object' not in json_result or json_result['object'] != 'chat.completion')):
                 raise CredentialsValidateFailedError(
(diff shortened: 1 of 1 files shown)


bowenliang123:
-1 for this.

As far as I know, the spaces at the front or the end of the string would not fail the json parsing.
Any further explanation or evidence for the issue you're fixing?

aniaan:
chunk is not always a space. In some cases, it may be `\r`. When `chunk='\r'`, an error will occur when executing the corresponding code.

This is the minimal simple reproducing case.
```python
import json

chunk = '\r'
decoded_chunk = chunk.strip().lstrip('data: ').lstrip()
chunk_json = json.loads(decoded_chunk)
```
provide more explanation as to why sometimes the chunk will be `\r`. The reason is actually because the default delimiter for `dify` is `\n\n`, but I am using the library `LLaMA-Factory`, which uses the delimiter `\r\n`. When I first used `dify`, I configured it according to its default value `\n\n`, but there was no corresponding response from `dify` (actually, the program code has already thrown an error, but there was also a lack of logs), which made debugging difficult for me. So, I had to try changing some delimiters and added some code logs. When I tried using '\n' as the delimiter, although there were still errors, they could be resolved by adding one line of code `chunk=chunk.strip()`. That's why this pr exists.

There is no problem with using OpenAI's SDK by default to call `LLaMA-Factory`, so initially it made me suspect that the issue was with `dify`.

Regarding the use of `\r\n` in `LLaMA-Factory`, I discovered this today when looking at the underlying dependency library code. If we had known from the beginning that it uses `\r\n`, maybe we wouldn't have encountered this issue.


As for this specific problem, personally, I think this line of code may still be meaningful as it adds robustness to the program. However, ultimately it depends on your opinion.

ps:

Some key error locations in the code may still benefit from adding some logs, for example.

https://github.com/langgenius/dify/blob/80e390b906b848eae41e185c97acb83fd22cc4fa/api/core/model_runtime/model_providers/openai_api_compatible/llm/llm.py#L437-L443

Yeuoly (approved):
LGTM

bowenliang123:
Thanks for the explanation. SGTM.

PR #3612, opened 2024-04-18
feat: add file log

liuzhenghua:
# Description

add file log

Fixes # (issue)

## Type of Change

- [ X] New feature (non-breaking change which adds functionality)
- [ ] Dependency upgrade

# How Has This Been Tested?

find the log file and format.

- [ ] TODO

# Suggested Checklist:

- [X ] I have performed a self-review of my own code

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -1,4 +1,5 @@
 import os
+import sys
 from logging.handlers import RotatingFileHandler
 
 if not os.environ.get("DEBUG") or os.environ.get("DEBUG").lower() != 'true':
@@ -87,13 +88,12 @@ def create_app(test_config=None) -> Flask:
 
     app.secret_key = app.config['SECRET_KEY']
 
+    log_handlers = None
     log_file = app.config.get('LOG_FILE')
-    log_dir = os.path.dirname(log_file)
-    os.makedirs(log_dir, exist_ok=True)
-    logging.basicConfig(
-        level=app.config.get('LOG_LEVEL'),
-        format=app.config.get('LOG_FORMAT'),
-        handlers=[
+    if log_file:
+        log_dir = os.path.dirname(log_file)
+        os.makedirs(log_dir, exist_ok=True)
+        log_handlers = [
             RotatingFileHandler(
                 filename=log_file,
                 maxBytes=1024 * 1024 * 1024,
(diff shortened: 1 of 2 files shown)


api/app.py
```diff
@@ -86,7 +87,21 @@ def create_app(test_config=None) -> Flask:
 
     app.secret_key = app.config['SECRET_KEY']
 
-    logging.basicConfig(level=app.config.get('LOG_LEVEL', 'INFO'))
+    log_file = app.config.get('LOG_FILE')
```
takatost, line 92: The code here should check if LOG_FILE is empty, then use default stdout / stderr output.

api/config.py
```diff
@@ -38,6 +38,8 @@
     'QDRANT_CLIENT_TIMEOUT': 20,
     'CELERY_BACKEND': 'database',
     'LOG_LEVEL': 'INFO',
+    'LOG_FILE': 'logs/app.log',
```
takatost: Default value should be empty

takatost (changes requested):
Please expand the configuration strategy while keeping the original logic unchanged.

api/config.py
liuzhenghua: LOG_FILE defaults empty string now

api/app.py
liuzhenghua: If LOG_FILE is empty, logs will be output only to stdout in the new commit. Please help review it.

liuzhenghua: I believe that the code app.config.get('LOG_LEVEL', 'INFO') can be misleading to developers, because the 'LOG_LEVEL' key is defined in config.py (even if it is None or an empty string), and the default value afterwards will only be returned when you haven't defined this key (which can be misleading).
So I changed this section to app.config.get('LOG_LEVEL'), and the default value should be modified in the DEFAULTS variable in config.py to take effect.

takatost: Customizing the configuration is fine, but we recommend modifying the configuration values in `.env` rather than changing the contents in config.py `DEFAULTS`.

liuzhenghua: The new version has set the LOG_FILE key in the DEFAULTS variable of the config.py file to an empty string. Could you please take a look when you have time to see if anything else needs to be changed?

takatost (approved):
LGTM

PR #3697, opened 2024-04-23
Add new tool: GitHub Advanced

richards199999:
# Description

Although we have a GitHub tool, however, the previous one does not have some critical feature like search for User, list files and many more, so I add this new advanced one which don't need GitHub token access.

## Type of Change

# How Has This Been Tested?

Haven't tested since I don't have much time (heavy schoolwork).🥲 But hopefully no huge bugs.

# TODO Checklist:

### DIFF
--- a/web/app/components/app/configuration/config-prompt/simple-prompt-input.tsx
+++ b/web/app/components/app/configuration/config-prompt/simple-prompt-input.tsx
@@ -148,7 +148,7 @@ const Prompt: FC<ISimplePromptInput> = ({
             )}
           </div>
           <div className='flex items-center'>
-            {!isAgent && (
+            {!isAgent && !readonly && (
               <AutomaticBtn onClick={showAutomaticTrue} />
             )}
           </div>


bowenliang123:
Some suggestions here for this PR:
1. prevent introducing a completely new tool but extend the existing Github way, if possible. And you could 'upgrade' the existed functions in your wary.
2. Sending raw RESTful HTTP requests may not a good idea to connect a SaaS platform like Github, as it may change the API schema and migrate to new version. `pygithub` SDK a popular wrapper for Github API is a much better approach to achieve the goals and handle the future impacts, see https://github.com/PyGithub/PyGithub.

Yeuoly:
1. I suggest that using GitHub SDK instead, native HTTP requests are not stable enough.
2. There is already a tool provider named GitHub, AdvancedGithub will confuse users, you can try to contact the author of that and add your name to author fields as co-author.

thx!

richards199999:
@bowenliang123 @Yeuoly Thank you for your advice.
You mentioned:
> 'upgrade' the existed functions in your wary.
> already a tool provider named GitHub
The reason I didn't replace or modify the original provider is that I think I should not do this on my own🤣 (of cuz i should not). And yes I would contact the author.

You also mentioned:
> using GitHub SDK instead
The reason I didn't use it is that I'm not familiar with the framework and I'm kinda busy these days, so might consider do it in Summer? idk

richards199999:
@charli117 hi? i think you are the author of the original tool. do you mind replacing your version with mine and i would add your name as coauthor🤗

charli117:
> @charli117 hi? i think you are the author of the original GitHub tool. do you mind replacing your version with mine and i would add your name as coauthor🤗

No problem at all. You adjust

richards199999:
Thanks for your understanding🤗🤗🤗🫡🫡 @charli117

bowenliang123:
@crazywoola May I ask for inviting @richards199999 to join the contributor Wechat group for further and easier discussion? Since this PR has reached a reasonable maturity level.

PR #3785, opened 2024-04-24
 Enhance PDF parsing capabilities

ic-xu:
# Description

Using the enhanced open parsing library to parse PDF documents can maintain the table style of PDF documents, and can also extract the image content in the PDF while ensuring that the user's reading order does not change

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

To test the PDF parsing function individually, use the following code
```
from core.rag.extractor.pdf.openparse import DocumentParser, processing
from core.rag.extractor.pdf.openparse.schemas import ImageElement

if __name__ == '__main__':
    """Lazy load given path as pages."""
    # blob = Blob.from_path(self._file_path)
    # yield from self.parse(blob)
    file_path = "pdf file path here"
    documents = []
    parser = DocumentParser(
        processing_pipeline=processing.BasicIngestionPipeline(),
        table_args={
            "parsing_algorithm": "pymupdf",
            "table_output_format": "markdown"
        }
    )
    parsed_basic_doc = parser.parse(file_path)
    documentContent = ''
    for _index, node in enumerate(parsed_basic_doc.nodes):
        metadata = {"source": file_path, "page": _index}
        for element in node.elements:
            if isinstance(element, ImageElement):
                # pdf images a
                pass
            else:
                print(element.text)
```

there is ImageElement\ TableElement\TextElement
If you parse the image, the focus is ImageElement， which has the following attributes that can be used
    block: dict
    text: str
    image: bytes
    ext: str
    bbox: Bbox

----

To test the overall effect, please follow the complete PDF upload and parsing process.

# Suggested Checklist:

- [ ] normal PDF
- [ ] PDF document containing tables
- [ ] If possible, you can also detect PDF documents with images


api/requirements.txt
```diff
@@ -80,3 +81,10 @@ lxml==5.1.0
 xlrd~=2.0.1
 pydantic~=1.10.0
 pgvecto-rs==0.1.4
+xlrd~=2.0.1
+pdfminer==20191125
+pdfminer.six==20201018
+pypdf==4.2.0
+PyMuPDF==1.24.0
```
takatost, line 86: Seems like some dependencies requiring commercial licenses have been introduced, please check it out. 🧐

takatost (changes requested):

web/yarn.lock
crazywoola, line 1: Hello please leave this file untouched if there is no frontend changes.

crazywoola (changes requested):
And there are some changes related to requirements.txt, please resolve the conflicts. We will take a look at this asap.

PR #3898, opened 2024-04-26
feat: update notion extractor

badbye:
# Description

see this post: https://github.com/langgenius/dify/discussions/3883

This PR combine all the blocks of a notion page into a single document. Headings are converted to the markdown style, so that user can use customized splitter to split it into chunks.
For example, `\n## ` can be used to split by h2 (It may not work if there is a code block and comment in the page).

Anyway, from my experience, too many chunks and less content in each chunk result in poorly performance. This PR could make it better.

## Type of Change
- [X] Improvement, including but not limited to code refactoring, performance optimization, and UI/UX improvement

# How Has This Been Tested?

A test script was added.

# Suggested Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods
- [X] `optional` I have added tests that prove my fix is effective or that my feature works
- [X] `optional` New and existing unit tests pass locally with my changes

### DIFF
--- a/api/core/rag/extractor/notion_extractor.py
+++ b/api/core/rag/extractor/notion_extractor.py
@@ -19,8 +19,12 @@ SEARCH_URL = "https://api.notion.com/v1/search"
 
 RETRIEVE_PAGE_URL_TMPL = "https://api.notion.com/v1/pages/{page_id}"
 RETRIEVE_DATABASE_URL_TMPL = "https://api.notion.com/v1/databases/{database_id}"
-HEADING_TYPE = ['heading_1', 'heading_2', 'heading_3']
-
+# if user want split by headings, use the corresponding splitter
+HEADING_SPLITTER = {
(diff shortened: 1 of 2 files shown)


JohnJyong:
We will add more different splitter rules for user to choose in our roadmap , becauser no one is better than the other but just when one fits more in certain type of questions.

badbye:
> We will add more different splitter rules for user to choose in our roadmap , becauser no one is better than the other but just when one fits more in certain type of questions.

Totally agree.  This PR is actually not about the splitter. It is just trying to combine the blocks into a single document, then you can use a customized splitter to split the page.
You mean you will not consider it before more split rules are added?

crazywoola (approved):

PR #3902, opened 2024-04-26
feat: add start commands to devcontainer

perzeuss:
# Description

I work with the dify repo every week and every time I forget the whole command to start the api or the worker. I added these commands because they are intuitive and easy to remember.

This change just introduces shortcut commands to run services, which are made available via alias in the devcontainer:
- start-api: starts the flask app in debug mode
- start-worker: starts the worker 
- start-web: starts the nextjs web app in dev mode
- start-containers: runs docker compose for the middleware compose file to start containers like dbs and sandbox

## Type of Change

# How Has This Been Tested?

- [ ] Open the dify workspace in a devcontainer
- [ ] Run the VS Code script "Rebuild Container"
- [ ] Watch the logs, after the container is created, it should run an npm install for web (as it did before)
- [ ] After the container has been created, it should run pip install for api (as it did before)
- [ ] run start-containers (just type it in the terminal and press enter)
  - [ ] this should run docker compose to run required containers for local development
- [ ] run start-api
  - [ ] this should start the dify api
- [ ] run start-worker
  - [ ] this should start the the dify worker
- [ ] run start-web
  - [ ] this should start the dify web
- [ ] open `~/.bashrc` you should see 4 commands defined at the end of the file
- [ ] close vs code and reopen it
- [ ] the devcontainer should run pip install but should not create the aliases again
- [ ] open `~/.bashrc` again
- [ ] it should still contain only the 4 commands once

# Suggested Checklist:

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -32,8 +32,8 @@
 			]
 		}
 	},
-	"postStartCommand": "chmod +x ./.devcontainer/post_start_command.sh; ./.devcontainer/post_start_command.sh",
-	"postCreateCommand": "chmod +x ./.devcontainer/post_create_command.sh; ./.devcontainer/post_create_command.sh"
+	"postStartCommand": "./.devcontainer/post_start_command.sh",
+	"postCreateCommand": "./.devcontainer/post_create_command.sh"
 
 	// Features to add to the dev container. More info: https://containers.dev/features.
 	// "features": {},
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -32,8 +32,8 @@
 			]
 		}
 	},
-	"postStartCommand": "cd api && pip install -r requirements.txt",
-	"postCreateCommand": "cd web && npm install"
+	"postStartCommand": "chmod +x ./.devcontainer/post_start_command.sh; ./.devcontainer/post_start_command.sh",
+	"postCreateCommand": "chmod +x ./.devcontainer/post_create_command.sh; ./.devcontainer/post_create_command.sh"
 
 	// Features to add to the dev container. More info: https://containers.dev/features.
 	// "features": {},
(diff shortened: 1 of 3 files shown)


crazywoola (dismissed):

perzeuss:
I did not add docs because I don't want to confuse people with these commands, but maybe you have an idea where to add a hint that these commands exist?

crazywoola:
I think you need to fix the ci builds. I think it's ok to left the docs blank for this pr.

.devcontainer/devcontainer.json
```diff
@@ -32,8 +32,8 @@
 			]
 		}
 	},
-	"postStartCommand": "cd api && pip install -r requirements.txt",
-	"postCreateCommand": "cd web && npm install"
+	"postStartCommand": "chmod +x ./.devcontainer/post_start_command.sh; ./.devcontainer/post_start_command.sh",
+	"postCreateCommand": "chmod +x ./.devcontainer/post_create_command.sh; ./.devcontainer/post_create_command.sh"
```
bowenliang123: For what reason that `chmod +x` should be run every time?

bowenliang123: Is that do it once in this commit and set it executable？

perzeuss: > Is that do it once in this commit and set it executable？
Yes, I had problems running the scripts in the dev container. I'll test again to see if I can leave it out.

.devcontainer/post_start_command.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+cd api && pip install -r requirements.txt
```
takatost, line 3: Do you think we should add "flask db upgrade" at the end? I'm not using devcontainer, so I'm not very familiar with this practice.

perzeuss: Is there a way to make a dry-run and just inform the user that the db is out of sync?
I think we should not overwrite the db just because the developer opens the devcontainer. By the way, I would do this in another PR anyway since I didn't add it, just moved it.

crazywoola (approved):

PR #3965, opened 2024-04-29
🐞 fix(dataset): delete get_dataset

baxiang:
Duplicate query select dataset


bowenliang123:
Seems not a bug but ensuring existence of the dataset.

PR #4188, opened 2024-05-08
chore: use prettier to format code

TinsFox:
# Description

Add prettier to automatically format the code when saving it so that it follows a consistent style. Currently works with eslint

## Type of Change

# How Has This Been Tested?

Feel free to modify the code and save it. But please make sure that format on save of vs code is turned on.

`  "editor.formatOnSave": true,`

# Suggested Checklist:

### DIFF
--- a/web/.eslintrc.json
+++ b/web/.eslintrc.json
@@ -1,13 +1,7 @@
 {
-  "extends": [
-    "next",
-    "@antfu"
-  ],
+  "extends": ["next", "@antfu", "prettier"],
   "rules": {
-    "@typescript-eslint/consistent-type-definitions": [
-      "error",
-      "type"
-    ],
+    "@typescript-eslint/consistent-type-definitions": ["error", "type"],
     "@typescript-eslint/no-var-requires": "off",
     "no-console": "off",
     "indent": "off",
--- a/web/.prettierignore
+++ b/web/.prettierignore
@@ -1,0 +1,5 @@
+node_modules
+.next
+/build
+.env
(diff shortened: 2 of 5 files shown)


TinsFox:
I'm very sorry, I just saw that the project uses antfu's eslint configuration, and antfu does not use prettier. Maybe my PR should be closed.

But there are some issues that I think still need to be discussed

1. The configuration of `@antfu/eslint-config` already supports eslint v9. The version in the project is still 8.x. Do we need to upgrade simultaneously?
3. There are React related rules in `@antfu/eslint-config`, but our project is Next.js and there are still some differences. It may be necessary to rewrite some rules after upgrading to cover them.
4. There are no tailwindcss related rules in `@antfu/eslint-config`. We need to introduce rules to restrict tailwindcss, such as repeated class use, class order, etc. Now there are situations where the same class is used
image

My suggestion is to use eslint + prettier. Using the configuration of [shadcn-ui](https://github.com/shadcn-ui/ui) can meet our basic needs.
https://github.com/shadcn-ui/ui/blob/main/.eslintrc.json

nite-knite:
Prettier's opinionated nature often leads to undesirable results and cannot be optimized through configuration. Therefore, we prefer to use eslint for code formatting.
Thank you for your suggestions on upgrading `@antfu/eslint-config`. Our team will choose an appropriate time to upgrade the plugin and relevant infrastructures, as upgrading lint rules may cause unwanted code conflicts.
Given that we will not adopt prettier, close this for now.

PR #4269, opened 2024-05-10
fix: suggested questions language

YidaHu:
# Description

This PR introduces the ability to specify the language for generating suggested questions based on the app settings. Previously, the language was not specified. Now, the language can be set according to the app's settings, providing a more personalized and accurate generation of suggested questions for different language users.

## Type of Change

# How Has This Been Tested?

- [ ] TODO

# Suggested Checklist:

### DIFF
--- a/api/core/llm_generator/llm_generator.py
+++ b/api/core/llm_generator/llm_generator.py
@@ -51,9 +51,9 @@ class LLMGenerator:
         return name
 
     @classmethod
-    def generate_suggested_questions_after_answer(cls, tenant_id: str, histories: str):
+    def generate_suggested_questions_after_answer(cls, tenant_id: str, histories: str, language="English"):
         output_parser = SuggestedQuestionsAfterAnswerOutputParser()
-        format_instructions = output_parser.get_format_instructions()
+        format_instructions = output_parser.get_format_instructions(language)
 
         prompt_template = PromptTemplateParser(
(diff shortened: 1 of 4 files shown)


leslie2046:
Maybe sometimes the suggested questions language should follow the language from user inputs in multilingal?
Sometimes we cannot set a language to a certain fixed  language i think。

YidaHu:
> Maybe sometimes the suggested questions language should follow the language from user inputs in multilingal? Sometimes we cannot set a language to a certain fixed language i think。

@leslie2046 You're right. Then I think prompt should be modified so that the LLM can give the corresponding language suggestions based on the context language. Not according to the set language.

takatost:
Since the default strategy is not specifically for English, but for all languages, sometimes there may be some issues with language understanding. 
I suggest reverting the default strategy back to its original state here, and if a specific language is passed in, then add the corresponding prompt.

takatost:
> Since the default strategy is not specifically for English, but for all languages, sometimes there may be some issues with language understanding. I suggest reverting the default strategy back to its original state here, and if a specific language is passed in, then add the corresponding prompt.

Any progress?

YidaHu:
> > Since the default strategy is not specifically for English, but for all languages, sometimes there may be some issues with language understanding. I suggest reverting the default strategy back to its original state here, and if a specific language is passed in, then add the corresponding prompt.
> 
> Any progress?

I found that it is the LLM effect problem, and the service does not need to be modified. I close

PR #4451, opened 2024-05-16
fix: the delete method is not supported the write_timeout params in requests library

dwgeneral:
# Description

in v0.6.7, we introduced the timeout params for HTTP node in workflow, but I found the bug that if we call a request with DELETE method, then it will occur the invalid timeout params error, because the `write_timeout` was not supported for DELETE method in requests library, that's why we got this issue.

Fixes # (issue)
The reason why the DELETE method in Python’s requests library supports connect_timeout and read_timeout but not write_timeout is related to the nature of the HTTP DELETE method and how it typically behaves in web applications.

	1.	Connect Timeout: This timeout refers to the time it takes for the client to establish a connection with the server. It’s relevant for all HTTP methods, including DELETE, as establishing a connection is fundamental to any HTTP request.
	2.	Read Timeout: This timeout refers to the time allowed for the server to send a response back to the client after the connection is established. Again, this timeout is relevant for all HTTP methods, including DELETE, as the client expects a response from the server.
	3.	Write Timeout: This timeout refers to the time allowed for the client to send the entire request (including the request body, if any) to the server. It’s more relevant for HTTP methods like POST and PUT, where the client is sending data to the server.

## Type of Change

# How Has This Been Tested?
re-deploy the docker image and run, the issue was gone.

# Suggested Checklist:

### DIFF
--- a/api/core/workflow/nodes/http_request/http_executor.py
+++ b/api/core/workflow/nodes/http_request/http_executor.py
@@ -261,11 +261,16 @@ class HttpExecutor:
         """
             do http request depending on api bundle
         """
+        # the write_timeout parameter is not commonly needed or relevant for the DELETE method in requests
+        timeouts = (self.timeout.connect, self.timeout.read, self.timeout.write)
+        if self.method == 'delete':
+            timeouts = (self.timeout.connect, self.timeout.read)
+
         kwargs = {
             'url': self.server_url,
             'headers': headers,
             'params': self.params,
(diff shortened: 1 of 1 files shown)


Yeuoly:
hi, thanks for your contributions! but this issue has already been solved inside `ssrf_proxy`, we have removed the third timeout param in timeout tuple, eliminating the need for special handling for the delete method like `if xx == delete`, BTW, the comments are not very clear, would you mind add some comments to make it more readable and less confusing?

PR #4487, opened 2024-05-17
improve: generalize transformations and scripts of runner and preloads into TemplateTransformer

bowenliang123:
# Description

- generalize scripts of runner and preload into TemplateTransformer
- properly indent scripts with `dedent` inside methods

## Type of Change

- [ ] Dependency upgrade

# How Has This Been Tested?

- [x] existed tests for code executor

# Suggested Checklist:

### DIFF
--- a/api/core/helper/code_executor/code_executor.py
+++ b/api/core/helper/code_executor/code_executor.py
@@ -12,7 +12,7 @@ from config import get_env
 from core.helper.code_executor.entities import CodeDependency
 from core.helper.code_executor.javascript.javascript_transformer import NodeJsTemplateTransformer
 from core.helper.code_executor.jinja2.jinja2_transformer import Jinja2TemplateTransformer
-from core.helper.code_executor.python3.python3_transformer import PYTHON_STANDARD_PACKAGES, Python3TemplateTransformer
+from core.helper.code_executor.python3.python3_transformer import Python3TemplateTransformer
 from core.helper.code_executor.template_transformer import TemplateTransformer
 
 logger = logging.getLogger(__name__)
@@ -187,7 +187,8 @@ class CodeExecutor:
             response = response.json()
             dependencies = response.get('data', {}).get('dependencies', [])
             return [
-                CodeDependency(**dependency) for dependency in dependencies if dependency.get('name') not in PYTHON_STANDARD_PACKAGES
+                CodeDependency(**dependency) for dependency in dependencies
+                if dependency.get('name') not in Python3TemplateTransformer.get_standard_packages()
             ]
         except Exception as e:
             logger.exception(f'Failed to list dependencies: {e}')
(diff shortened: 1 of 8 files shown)


bowenliang123:
cc @takatost @Yeuoly

bowenliang123:
BTW, during the refactoring, it's found that the approaches for the input objects serialization vary between Python and Javascript. The one for Python is wrapped in Base64 encoding, while the Javascript uses the raw inputs json object. Not the best way for further extensions to other languages.  Is that as designed on purpose?

Yeuoly:
> BTW, during the refactoring, it's found that the approaches for the input objects serialization vary between Python and Javascript. The one for Python is wrapped in Base64 encoding, while the Javascript uses the raw inputs json object. Not the best way for further extensions to other languages. Is that as designed on purpose?

In Python, the boolean value True is capitalized, which is not standard in JSON encoding. As a result, simple JSON encoding might not work as unexpected. However, this issue does not occur in JavaScript.

bowenliang123:
Yes, the differences were noticed. So maybe we could unify the json+base64 as the standard serialization for input objects in the future.

bowenliang123:
And there's another point I may want to have your suggestions. We are still using `replace` with to fill the code and inputs, as the existed way before refactoring. Do you want  `jinja2` to do this for possible injection attacks (but I'm not sure how much MarkupSafe inside jinja2 does help) ?

Yeuoly:
> Yes, the differences were noticed. So maybe we could unify the json+base64 as the standard serialization for input objects in the future.

feel free to do it~

Yeuoly:
> And there's another point I may want to have your suggestions. We are still using `replace` with to fill the code and inputs, as the existed way before refactoring. Do you want `jinja2` to do this for possible injection attacks (but I'm not sure how much MarkupSafe inside jinja2 does help) ?

It's unnecessary I think, the `replace` method doesn't significantly impact performance, maybe format could be a good alternative. As for Jinja2, it's too heavy for this task. It supports many features like loops, which we don't need. Additionally, it introduces some security risks. Even if it's currently safe, security experts can always find new ways to exploit it as I faced before.

bowenliang123:
> > And there's another point I may want to have your suggestions. We are still using `replace` with to fill the code and inputs, as the existed way before refactoring. Do you want `jinja2` to do this for possible injection attacks (but I'm not sure how much MarkupSafe inside jinja2 does help) ?
> 
> 
> 
> It's unnecessary I think, the `replace` method doesn't significantly impact performance, maybe format could be a good alternative. As for Jinja2, it's too heavy for this task. It supports many features like loops, which we don't need. Additionally, it introduces some security risks. Even if it's currently safe, security experts can always find new ways to exploit it as I faced before.

Yes, I agree with you. Let's keep it with "replace".

bowenliang123:
> > Yes, the differences were noticed. So maybe we could unify the json+base64 as the standard serialization for input objects in the future.
> 
> feel free to do it~

Hi, I have get this done. `json + base64` serialization is now a shared mechanism for both Python and Javascript.

loganclark360:
👍

W. Logan Clark

Yeuoly (approved):
LGTM

PR #4518, opened 2024-05-20
fix: workaround db migration error when adding custom_disclaimer column to recommended_apps

bowenliang123:
# Description
- the failing migration script was introduced in #4306 
- error when running `flask db upgrade` for db migration
```
root@5789a1154494:/app/api# flask db upgrade

None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1967, in _exec_single_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 924, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.NotNullViolation: column "custom_disclaimer" of relation "recommended_apps" contains null values


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/flask", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/site-packages/flask/cli.py", line 1105, in main
    cli.main()
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/flask/cli.py", line 386, in decorator
    return ctx.invoke(f, *args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/flask_migrate/cli.py", line 154, in upgrade
[...]

PR #4739, opened 2024-05-28
test: CI test for db migration scripts on changes

bowenliang123:
# Description

- adding DB migration test to CI jobs, to ensure db migration scripts are ready to apply, preventing issues like #4518
- this test only triggers when code changes happen in `api/migration/**` path

## Type of Change

- [ ] Dependency upgrade

# How Has This Been Tested?

- [ ] TODO

# Suggested Checklist:

### DIFF
--- a/.github/workflows/db-migration-test.yml
+++ b/.github/workflows/db-migration-test.yml
@@ -42,4 +42,5 @@ jobs:
       - name: Run DB Migration
         run: |
           cd api
+          cp .env.example .env
           flask db upgrade
--- a/.github/workflows/db-migration-test.yml
+++ b/.github/workflows/db-migration-test.yml
@@ -1,0 +1,45 @@
+name: DB migration test
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'api/migrations/**'
+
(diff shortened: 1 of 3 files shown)


Yeuoly:
How does this test catch issues like #4518? From what I understand, the problem comes from old data. Some records don't have the `custom_disclaimer` column, and during migration, Postgres can't set the `custom_disclaimer` field to null because `nullable` is set to None. But this test only checks the migration process—it doesn't insert any data. It makes sure there won't be two branching paths, but it can't fix issues like #4518.

takatost (approved):
LGTM

takatost:
å

> How does this test catch issues like #4518? From what I understand, the problem comes from old data. Some records don't have the `custom_disclaimer` column, and during migration, Postgres can't set the `custom_disclaimer` field to null because `nullable` is set to None. But this test only checks the migration process—it doesn't insert any data. It makes sure there won't be two branching paths, but it can't fix issues like #4518.

Yeah, adding new fields without data and making them non-nullable will not cause any issues.

bowenliang123:
Agree. Thanks.

PR #4794, opened 2024-05-30
feat: opportunistic tls flag for smtp

chazzhou:
# Description

This pull request introduces an opportunistic TLS flag for the SMTP configuration in Dify and changes the default SMTP port from 587 to 465.

Currently, the existing TLS flag in Dify does not support port 465 (It will hang and then timeout), which is more commonly used nowadays due to its higher security. Port 465 is used for SMTP connections secured with SSL/TLS encryption from the start, establishing a secure connection before any SMTP commands are sent. SSL/TLS encryption is enabled by default on this port, providing a higher level of security compared to port 587.

Port 587, on the other hand, is used for SMTP connections with STARTTLS, where the connection starts as a plain-text SMTP connection and can be upgraded to a secure connection using the STARTTLS command. It allows the client and server to negotiate the use of encryption during the SMTP conversation.

To address the limitation of the current TLS flag and enhance the security of email communication, this pull request introduces an opportunistic TLS flag. This flag provides the option to enable opportunistic TLS encryption for SMTP connections, allowing the flexibility to use either port 587 with STARTTLS or port 465 with SSL/TLS, depending on the email server's requirements and support.

Furthermore, the pull request changes the default SMTP port in the example configs from 587 to 465. This ensures that the SMTP connection is encrypted from the beginning by default, providing a higher level of security out of the box.

The changes include updates to the configuration files, environment variables, and the SMTPClient class to support this new feature and the default port change.

## Type of Change

# How Has This Been Tested?

- [x] Verified that the SMTP configuration updates are reflected in the `api/.env.example` and `docker/docker-compose.yaml` files.
- [x] Tested the SMTPClient class with different combinations of `use_tls` and `opportunistic_tls` flags to ensure the correct behavior.
- [x] Validated that email sending functionality works as expected with the new opportunistic TLS option enabled with port 587.
- [x] Validated that email sending functionality works as expected with the new opportunistic TLS option disabled with port 465.

# Suggested Checklist:

### DIFF
--- a/api/.env.example
+++ b/api/.env.example
@@ -127,10 +127,11 @@ RESEND_API_KEY=
 RESEND_API_URL=https://api.resend.com
 # smtp configuration
 SMTP_SERVER=smtp.gmail.com
-SMTP_PORT=587
+SMTP_PORT=465
 SMTP_USERNAME=123
 SMTP_PASSWORD=abc
-SMTP_USE_TLS=false
+SMTP_USE_TLS=true
+SMTP_OPPORTUNISTIC_TLS=false
 
 # Sentry configuration
 SENTRY_DSN=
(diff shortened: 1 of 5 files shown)


crazywoola (approved):

crazywoola:
Thanks for your contribution.

PR #4804, opened 2024-05-30
fix: remove escape characters from strings in api

LarryPage:
# Description

When setting the API swagger, the interface request is post, and the value corresponding to the submitted parameter key is object type. When converting the job string to object, # remove the escape character from the string, otherwise an error will be reported

Fixes # (issue)

## Type of Change

### DIFF
--- a/api/core/tools/tool/api_tool.py
+++ b/api/core/tools/tool/api_tool.py
@@ -293,7 +293,8 @@ class ApiTool(Tool):
                 elif property['type'] == 'object':
                     if isinstance(value, str):
                         try:
-                            return json.loads(value)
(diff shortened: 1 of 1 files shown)


api/core/tools/tool/api_tool.py
```diff
@@ -293,7 +293,8 @@ def _convert_body_property_type(self, property: dict[str, Any], value: Any) -> A
                 elif property['type'] == 'object':
                     if isinstance(value, str):
                         try:
-                            return json.loads(value)
+                            unescaped_string = value.replace("\\", "")  # 移除字符串中的转义字符
```
crazywoola, line 296: Please sue the English to comment the code.

crazywoola (changes requested):

PR #4809, opened 2024-05-30
fix: remove escape characters from strings in api

LarryPage:
# Description

When setting the API swagger, the interface request is post, and the value corresponding to the submitted parameter key is object type. When converting the job string to object, # remove the escape character from the string, otherwise an error will be reported

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/api/core/tools/tool/api_tool.py
+++ b/api/core/tools/tool/api_tool.py
@@ -293,7 +293,8 @@ class ApiTool(Tool):
                 elif property['type'] == 'object':
                     if isinstance(value, str):
                         try:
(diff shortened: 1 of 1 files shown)


Yeuoly:
As far as I know, The object is generated by model, the format depends on the model itself, something, It could output a valid JSON string without \ but not all the time, this modification could lead an unexpected result if "\" is just a normal character, I suggest use fallback mechanism, just remove escape characters if there is an error raised, otherwise, do nothing.

ghost (commented):
Ok what do I do

PR #4972, opened 2024-06-05
Fixed incorrect recognition of data input in chart

engery:
# Description

The original logic, when processing data such as 12, 23, 34; 230,340,540, the original code will first follow the ";" After splitting, mistakenly taking 12, 23, and 34 as one number for judgment; I made the modifications and will start with ';' After splitting, first perform ',' and then perform int and float judgments to avoid this problem!

## Type of Change

# How Has This Been Tested?

I ran locally and tested different situations, such as when the data is one group or multiple groups.

- [ ] TODO

# Suggested Checklist:

- [ ] I have performed a self-review of my own code

### DIFF
--- a/api/core/tools/provider/builtin/chart/tools/bar.py
+++ b/api/core/tools/provider/builtin/chart/tools/bar.py
@@ -13,37 +13,38 @@ class BarChartTool(BuiltinTool):
         data = tool_parameters.get('data', '')
         if not data:
             return self.create_text_message('Please input data')
-        data = data.split(';')
-
-        # if all data is int, convert to int
-        if all([i.isdigit() for i in data]):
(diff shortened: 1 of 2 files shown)


crazywoola:
Please fix the lint error in this pr.

bowenliang123:
Sorry that I still fail to  understand the key point of the supposed fixes.
Any comparison in parsing outputs for the without or with this fixes?

Yeuoly:
I guess you are trying to enable the drawing of multiple lines on the same chart, it makes sense, but you should update the description of `data` as LLM should be informed  that how to generate the data you want, but also, human_description needs to be optimized, otherwise, developers are confused

PR #4395, opened 2024-05-15
Fix HTTP REQUEST NODE is always waiting but endpoint have responsed

hjlarry:
# Description
Fixes issue (https://github.com/langgenius/dify/issues/4386)

In the PR(https://github.com/langgenius/dify/pull/4360), I use `httpx.Response.headers | requests.Response.headers` instead of `HttpExecutorResponse.headers`, because they are both a version of case-insensitive multi-dict. But cause this issue.

When we run the whole workflow, it has some json encoder/decoder error:
```
Traceback (most recent call last):
  File "C:\Users\hejl\anaconda3\envs\dify\lib\site-packages\werkzeug\wsgi.py", line 256, in __next__
    return self._next()
  File "C:\Users\hejl\anaconda3\envs\dify\lib\site-packages\werkzeug\wrappers\response.py", line 32, in _iter_encoded
    for item in iterable:
  File "C:\Users\hejl\anaconda3\envs\dify\lib\site-packages\flask\helpers.py", line 113, in generator
    yield from gen
  File "C:\Users\hejl\PycharmProjects\dify\api\libs\helper.py", line 156, in generate
    yield from response
  File "C:\Users\hejl\PycharmProjects\dify\api\core\app\apps\base_app_generate_response_converter.py", line 28, in _generate
    for chunk in cls.convert_stream_full_response(response):
  File "C:\Users\hejl\PycharmProjects\dify\api\core\app\apps\workflow\generate_response_converter.py", line 45, in convert_stream_full_response
    for chunk in stream_response:
  File "C:\Users\hejl\PycharmProjects\dify\api\core\app\apps\workflow\generate_task_pipeline.py", line 148, in _to_stream_response
    for stream_response in generator:
  File "C:\Users\hejl\PycharmProjects\dify\api\core\app\apps\workflow\generate_task_pipeline.py", line 188, in _process_stream_response
    workflow_node_execution = self._handle_node_finished(event)
  File "C:\Users\hejl\PycharmProjects\dify\api\core\app\task_pipeline\workflow_cycle_manage.py", line 448, in _handle_node_finished
    workflow_node_execution = self._workflow_node_execution_success(
  File "C:\Users\hejl\PycharmProjects\dify\api\core\app\task_pipeline\workflow_cycle_manage.py", line 223, in _workflow_node_execution_success
    workflow_node_execution.outputs = json.dumps(outputs) if outputs else None
[...]

PR #4975, opened 2024-06-05
Fix: http_request delete method not working

### LINKED ISSUE #4852: HTTP Request with DELETE method errors in any way
### Self Checks

- [X] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).

### Dify version

0.6.9

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Use HTTP Request with DELETE method on workflow.
image
Regardless of its status(this time, the URL will not work because it has not opened), it returns `Node (node-name) run failed: argument of type 'NoneType' is not iterable`.
This might relate to how http(or ssrfproxy) request handles DELETE inside source code, as only DELETE uses `requests` library with some polyfills while all other methods use `httpx`.

### ✔️ Expected Behavior

Should return a response, or corresponding error.

### ❌ Actual Behavior

Request is sent, but causes error regardless of response.

th3n00b13:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #4852 (issue)

## Type of Change

# How Has This Been Tested?

- [ ] TODO

# Suggested Checklist:

### DIFF
--- a/api/core/workflow/nodes/http_request/http_executor.py
+++ b/api/core/workflow/nodes/http_request/http_executor.py
@@ -28,7 +28,7 @@ class HttpExecutorResponse:
         self.headers = {}
         if isinstance(response, httpx.Response | requests.Response):
             for k, v in response.headers.items():
-                self.headers[k.lower()] = v
+                self.headers[k] = v
         self.response = response
 
     @property
@@ -42,7 +42,10 @@ class HttpExecutorResponse:
         return any(v in content_type for v in file_content_types)
 
     def get_content_type(self) -> str:
-        return self.headers.get('content-type')
+        if 'content-type' in self.headers:
+            return self.headers.get('content-type')
+        else:
+            return self.headers.get('Content-Type') or ""
 
     def extract_file(self) -> tuple[str, bytes]:
         """
--- a/api/core/workflow/nodes/http_request/http_executor.py
+++ b/api/core/workflow/nodes/http_request/http_executor.py
@@ -28,7 +28,7 @@ class HttpExecutorResponse:
         self.headers = {}
         if isinstance(response, httpx.Response | requests.Response):
             for k, v in response.headers.items():
-                self.headers[k.lower()] = v
+                self.headers[k] = v
         self.response = response
 
     @property
--- a/api/core/workflow/nodes/http_request/http_executor.py
+++ b/api/core/workflow/nodes/http_request/http_executor.py
@@ -26,9 +26,12 @@ class HttpExecutorResponse:
 
     def __init__(self, response: Union[httpx.Response, requests.Response] = None):
         self.headers = {}
-        if isinstance(response, httpx.Response | requests.Response):
+        if isinstance(response, httpx.Response):
             for k, v in response.headers.items():
                 self.headers[k] = v
+        elif isinstance(response, requests.Response):
+            for k, v in response.headers.items():
+                self.headers[k.lower()] = v
         self.response = response
 
     @property
--- a/api/core/workflow/nodes/http_request/http_executor.py
+++ b/api/core/workflow/nodes/http_request/http_executor.py
@@ -26,10 +26,7 @@ class HttpExecutorResponse:
 
     def __init__(self, response: Union[httpx.Response, requests.Response] = None):
         self.headers = {}
-        if isinstance(response, httpx.Response):
-            for k, v in response.headers.items():
-                self.headers[k] = v
-        elif isinstance(response, requests.Response):
+        if isinstance(response, httpx.Response | requests.Response):
             for k, v in response.headers.items():
                 self.headers[k.lower()] = v
         self.response = response


api/core/workflow/nodes/http_request/http_executor.py
```diff
@@ -26,9 +26,12 @@ class HttpExecutorResponse:
 
     def __init__(self, response: Union[httpx.Response, requests.Response] = None):
         self.headers = {}
-        if isinstance(response, httpx.Response | requests.Response):
+        if isinstance(response, httpx.Response):
```
bowenliang123: Hi, thanks for reporting this. Could you have a try with a tuple-sytle condition as `isinstance(response, (httpx.Response, requests.Response))` ? If it works, it helps to avoid repeating the code.

th3n00b13: Thanks for reviewing, I've deleted elif and relevant codes (reverting ifs) and made header key always lower.

bowenliang123: OK, thx. You could have this conversation resolved now. thx.

bowenliang123:
Hi, could you explain more about why and how the lowered header help to fix the problem?

According to RFC9110 for HTTP 1.1 , the header fields are case-sensitive. 
https://www.rfc-editor.org/rfc/rfc9110.html#name-header-fields

When it comes to HTTP/2, the preset headers are lowered by default.

th3n00b13:
Why error was caused because at the http_executor.py, they always reference "content-type", but while `httpx` will provide Content-Type as `content-type`, `requests` will provide `Content-Type` - there's no `content-type` on headers only for delete method that uses `requests`.
For headers, sorry for the inconvenience, I didn't know it should be case-sensitive.
In this case, if headers should be kept, only replacing `content-type` when it's `requests.Response` would be sufficient. (or just replace requests to httpx, through it could affect other codes)

th3n00b13:
Sorry, I wrongly understood documentation. it actually does *not* lower but it seems to allow any capitalization. (may require experiment how they are handled when it comes to key)

hjlarry:
Hi @th3n00b13,  I debug this problem
the issue is caused by this [line](https://github.com/langgenius/dify/blob/7fa735a43bc88ff3828256c6459b30ef489ab1fa/api/core/workflow/nodes/http_request/http_executor.py#L42) `any(v in content_type for v in file_content_types)`  which `content_type` is `None`

In the `get_content_type()` method, it should response string value, so we just modify `self.headers.get('content-type', '')` is more reasonable.

bowenliang123:
It's suggested to mark this PR as a draft before concrete evidence from further investigation is ready.
1. any idea from original author @Yeuoly , for how and why we are handling response from different HTTP libraries here?
2. why the header field name impacts the DELETE method and impact it only
3. which is the key header name to fix and why
4. the differences (by evidence) in header filed name between httpx's and requets' reponse

hjlarry:
Hi @bowenliang123 
1. I am also confused about why use two request lib. And I found that [ssrf_proxy](https://github.com/langgenius/dify/blob/7fa735a43bc88ff3828256c6459b30ef489ab1fa/api/core/helper/ssrf_proxy.py#L40) do some extra things about DELETE method, maybe httpx can not do that?
2. so just cause DELETE use requests, other's use httpx,  it only impact DELETE
3. Both httpx and requests build their own dict for response header class  which is case-insensitive.  In the `HttpExecutorResponse` class `__init__` we copy each item of the headers to our own dict. Once I try to not do that copy operation, cause this [issue](https://github.com/langgenius/dify/pull/4395)
4. the key is the `get_content_type()` [method](https://github.com/langgenius/dify/blob/7fa735a43bc88ff3828256c6459b30ef489ab1fa/api/core/workflow/nodes/http_request/http_executor.py#L44), it should response an empty string not the `None` as it's type  annotation describes.   The httpx's reponse header is `content-type`, the request's response header is `Content-Type`, so we should check both two keys and if empty response an empty string.

th3n00b13:
It seems it just works fine if I modify `get_content_type` to do as mentioned, or just return an empty string. (always empty string may affect files - otherwise, there's almost no usage of `get_content_type`)
But this time, requests seem to keep original header name when converted with `.items()`, which can cause a different behavior than its httpx. (they didn't for Content-Type, they stay the same as "Content-Type")
image

bowenliang123:
How about have a workaround quick fix this time , skipping the discussion in case-sensitive difference in `requests` and `httpx`, simply modify the `get_content_type` into getting both `Content-Type` and `content-type` from real header values?

bowenliang123:
Btw, please rebase the code to the latest commit of the main branch , which helps to resolve the style issue.

th3n00b13:
I think - I've messed up with branches with rebase, sorry for being somewhat bad. There might be a lot of weird commits, what should I do?

hjlarry:
just makesure the code difference between your current branch and the dify's main branch  is what you want to modify,  no matter how many other commits,rebase things.

th3n00b13:
With this commit(s), it should fix only the error of http-request DELETE method.

bowenliang123 (approved):
LGTM overall.
- it's a proper workaround to handler both reuqests' DELETE and httpx's GET/POST
- the author tested locally and proper screenshots are provided

th3n00b13:
I don't know much about Github, if it's ready, should I mark this pull request as ready for review?

bowenliang123:
sure. and then it's waiting for team  members review and merging.

crazywoola (approved):

PR #5202, opened 2024-06-14
feat: introduce pydantic-settings for config definition and validation

bowenliang123:
# Description

- introduce `pydantic-settings` for configs parsing and validation
  - https://docs.pydantic.dev/latest/concepts/pydantic_settings/
- it benefits the config entry's in the following aspects:
  - entry definition for each config item
  - data-type validation
  - default value
  - proper descriptions
  - grouping
  - deprecation
- no impact on the souce config file `.env`, all the configs is still read via `dotenv` inside `pydantic-settings` 
- no impact on configs reading from Flask's `app.config['XXXX']`

## Type of Change

- [ ] Dependency upgrade

# How Has This Been Tested?

- [ ] TODO

# Suggested Checklist:

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -1,5 +1,7 @@
 import os
 
+from configs.app_configs import DifyConfigs
+
 if not os.environ.get("DEBUG") or os.environ.get("DEBUG").lower() != 'true':
     from gevent import monkey
 
@@ -74,10 +76,19 @@ config_type = os.getenv('EDITION', default='SELF_HOSTED')  # ce edition first
 # Application Factory Function
 # ----------------------------
 
+def create_flask_app_with_configs() -> Flask:
+    """
+    create a raw flask app
+    with configs loaded from .env file
+    """
+    dify_app = DifyApp(__name__)
+    dify_app.config.from_object(Config())
+    dify_app.config.from_mapping(DifyConfigs().model_dump())
+    return dify_app
+
 
 def create_app() -> Flask:
-    app = DifyApp(__name__)
-    app.config.from_object(Config())
+    app = create_flask_app_with_configs()
 
     app.secret_key = app.config['SECRET_KEY']
 
--- a/api/config.py
+++ b/api/config.py
@@ -2,52 +2,27 @@ import os
 
 import dotenv
 
-dotenv.load_dotenv()
-
 DEFAULTS = {
-    'EDITION': 'SELF_HOSTED',
     'DB_USERNAME': 'postgres',
     'DB_PASSWORD': '',
     'DB_HOST': 'localhost',
     'DB_PORT': '5432',
     'DB_DATABASE': 'dify',
     'DB_CHARSET': '',
-    'REDIS_HOST': 'localhost',
-    'REDIS_PORT': '6379',
-    'REDIS_DB': '0',
-    'REDIS_USE_SSL': 'False',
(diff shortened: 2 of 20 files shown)


ghost (approved):

ghost (approved):

ghost (approved):

ghost (approved):

api/app.py
```diff
@@ -74,10 +76,19 @@ class DifyApp(Flask):
...
 # ----------------------------
 
+def create_flask_app_with_configs() -> Flask:
+    """
+    create a raw flask app
+    with configs loaded from .env file
+    """
+    dify_app = DifyApp(__name__)
+    dify_app.config.from_object(Config())
+    dify_app.config.from_mapping(DifyConfigs().dict())
```
laipz8200: It's better to use `model_dump`.

api/tests/unit_tests/settings/test_app_settings.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # configs read from dotenv directly
+    assert config['LOG_LEVEL'] == 'INFO'
+
+    # configs read from pydantic-settings
+    assert config['COMMIT_SHA'] == ''
+    assert config['EDITION'] == 'SELF_HOSTED'
+    assert config['API_COMPRESSION_ENABLED'] is False
+
+    assert config['CONSOLE_API_URL'] == 'https://cloud.dify.ai'
```
laipz8200: This test assumes that developers don't have their own `.env` files, otherwise the test is likely to fail. I think this could be confusing and make local testing difficult. Can we consider other ways to conduct the tests?

laipz8200 (changes requested):
Thanks for your contribution! This PR will help us migrate to a more type-safe system. However, there are still some issues. Would you mind addressing them?

api/app.py
bowenliang123: Previously, I'm trying to use the `.from_object(DifyConfigs().model_dump())`. But unexpected error occurs for iterating the names.

Furthermore, `dict()` outputs a `map[str, Any]` type matching our needs in keys as strings and values as dynamic types with unchanged data type.

laipz8200: Could you share an example where using `model_dump` failed? Since the `dict` method is deprecated, we want to avoid using it.

bowenliang123: OK, using `model_dump` now.

api/tests/unit_tests/settings/test_app_settings.py
bowenliang123: 1. change to use `pytest.fixture` for env file preparation before the tests.
2. move `dotenv.load_dotenv()` from global to inside initialization of the class Config, to prevent causing env variables containmination.

bowenliang123:
@laipz8200 Hi, I have resolved all the problems you listed. Please have a check.

laipz8200 (dismissed):
Awesome! Thanks for your hard work.

laipz8200:
> Please avoid changing unrelated lines in poetry lockfile. Use `dev/sync-poetry` script to refresh the poetry lockfile for good. The `--no-update` option in `poetry lock -C api --no-update` helps in this case.

You're right.

bowenliang123:
I've done it~ Thanks for reviewing.

laipz8200 (approved):

PR #5188, opened 2024-06-14
feat: allow to use IAM Role for Bedrock

tmokmss:
# Description

Fixes #3471 

This PR makes `aws_access_key_id` and `aws_secret_access_key` optional, and let boto to retrieve credentials from its execution environment.

## Type of Change

- [X] Improvement, including but not limited to code refactoring, performance optimization, and UI/UX improvement

# How Has This Been Tested?

- [X] Run web and api locally, and confirmed it successfully uses credentials from running environment when access key is not provided.

# Suggested Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods
- [ ] `optional` I have made corresponding changes to the documentation 
- [ ] `optional` I have added tests that prove my fix is effective or that my feature works
- [ ] `optional` New and existing unit tests pass locally with my changes

PR #5252, opened 2024-06-15
fix: embedding job fails using IAM role

tmokmss:
# Description

Related to #5188

Although we can now uses IAM role credentials when registering Bedrock models, the async embedding job fails due to a missing credential error.

The reason is that there is another code that references aws credentials, which this PR fixes. I confirmed there is no more place that uses these credentials by grep-ing `aws_access_key_id`.

I also edited the label a bit to make it fit with the component width.

## Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [X] run embedding job locally and it sucesses.

# Suggested Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods

### DIFF
--- a/api/core/model_runtime/model_providers/bedrock/bedrock.yaml
+++ b/api/core/model_runtime/model_providers/bedrock/bedrock.yaml
@@ -23,7 +23,7 @@ provider_credential_schema:
     - variable: aws_access_key_id
       required: false
       label:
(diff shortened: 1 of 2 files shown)


takatost (approved):
thx!

PR #5488, opened 2024-06-21
fix: summary of duckduckgo_search

hjlarry:
# Description

The origin summary method of BuildinTool is check how many token in the content, if the token < 60% of the llm's default max token, it won't work.   It seems a support for large context content,  But here I think a summary for the whole search result would be better.

## Type of Change

# How Has This Been Tested?

I test the tool node execute result locally.

- [ ] TODO

# Suggested Checklist:


Yeuoly (approved):
LGTM

PR #5492, opened 2024-06-21
feat:add tts-streaming config and future

### LINKED ISSUE #5251: When requesting /v1/text-to-audio with the streaming parameter set to true, get the TypeError: 'function' object is not iterable
### Self Checks

- [X] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).

### Dify version

0.6.0-fix1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Configure Tongyi Qianwen TTS model
image

2. Request /v1/text-to-audio url, set the parameter 'streaming' to true
curl --location 'http://localhost/v1/text-to-audio' \
--header 'accept: */*' \
--header 'Authorization: Bearer app-xxx' \
--data '{"text":"hello dify","streaming":true}'

### ✔️ Expected Behavior

Returns 200 with audio binary data

### ❌ Actual Behavior

Return code 500
image

ic-xu:
# Description

Add TTS streaming configuration and feature support, add text to speech in the feature configuration options, click on configure to start

Fixes https://github.com/langgenius/dify/issues/5251

## Type of Change

# How Has This Been Tested?

Simply configure the streaming TTS option and test it


crazywoola:
@charli117 Please review this pr as well.

charli117:
@crazywoola ic-xu  is a colleague of our team, and the version submitted now is the version we are currently using. There is also an optimization point that requires adding a blocking mode output logic for models that are not compatible and do not support streaming tts. We have communicated with each other and merged later

api/controllers/web/audio.py
```diff
@@ -104,6 +104,54 @@ def post(self, app_model: App, end_user):
...
+            raise CompletionRequestError(e.description)
+        except ValueError as e:
+            raise e
+        except Exception as e:
+            logging.exception(f"internal server error: {str(e)}")
+            raise InternalServerError()
 
 api.add_resource(AudioApi, '/audio-to-text')
 api.add_resource(TextApi, '/text-to-audio')
+api.add_resource(TextApiWithMessageId, '/text-to-audio/message-id')
```
takatost: It looks like this feature is consistent with the /text-to-audio route, you can add an optional message_id parameter in the /text-to-audio route to achieve the same capability.

api/core/app/apps/advanced_chat/generate_task_pipeline.py
```diff
@@ -515,7 +515,7 @@ def _generate_stream_outputs_when_node_finished(self) -> Optional[Generator]:
         :return:
         """
         if not self._task_state.current_stream_generate_state:
-            return None
+            return
```
takatost, line 587: I think `return None` is consistent with the `Optional` return in the method declaration.

api/core/app/apps/base_app_queue_manager.py
```diff
@@ -51,12 +57,31 @@ def listen(self) -> Generator:
         listen_timeout = current_app.config.get("APP_MAX_EXECUTION_TIME")
         start_time = time.time()
         last_ping_time = 0
-
+        last_message = None
         while True:
             try:
                 message = self._q.get(timeout=1)
                 if message is None:
+                    try:
```
takatost: Can you explain the purpose of this code snippet? It seems like it will cause a sharp increase in the memory usage of Redis.

takatost: I think it's more suitable to subscribe to AppGenerateService.generate() for the streaming return content, rather than putting it in the queue manager.

takatost (changes requested):

api/core/app/apps/base_app_queue_manager.py
ic-xu: image  

Yes, it is indeed possible, but we considered that in the case of multi-container deployment, we need to use a message middleware to cache the messages generated by GPT, and writing to redis is a very fast operation, so we did this

api/controllers/web/audio.py
ic-xu: This can be designed like this. In order not to affect the function of "/text-to-audio", an interface was specially added. I can change it if necessary.

api/core/app/apps/advanced_chat/generate_task_pipeline.py
ic-xu: This is a bug. When the program runs here, it will prompt that None is not a Generator, so I changed it to this

ic-xu: Here I understand that it either does not return or returns an object that can be iterated，When I debug the code, it prompts that None is not iterable.

api/controllers/web/audio.py
takatost: Suggest changing it to an route, with message_id as an optional parameter, it won't affect the previous logic.

api/core/app/apps/base_app_queue_manager.py
takatost: Two Issues:
1. All messages are written to redis here, and the expiration time is very long, which will quickly fill up redis abnormally.
2. `AppQueueManager` only deals with asynchronous communication between the main thread and the sub-thread. It is not appropriate to place this code here. I suggest adding a unified processing method in `core.app.task_pipeline.based_generate_task_pipeline.BasedGenerateTaskPipeline`, and adding this logic in the return of G`enerateTaskPipeline.process()` of the app that needs to be processed.

takatost (changes requested):

api/core/app/apps/base_app_queue_manager.py
ic-xu: It has been processed. Could you please help check it

api/controllers/web/audio.py
ic-xu: It has been processed. Could you please help check it

web/types/app.ts
```diff
@@ -160,6 +160,7 @@ export type ModelConfig = {
     enabled: boolean
     voice?: string
     language?: string
+    autoPlay?: 'enabled' | 'disabled'
```
crazywoola: Please use `ttsAutoPlay.enabled` and `ttsAutoPlay.disabled`here.

web/i18n/zh-Hans/app-debug.ts
```diff
@@ -319,6 +319,9 @@ const translation = {
       language: '语言',
       resolutionTooltip: '文本转语音音色支持语言。',
       voice: '音色',
+      autoPlay: '自动播放',
```
crazywoola, line 322: Please add these keys to `en-US/app-debug.ts` at least.

web/models/debug.ts
```diff
@@ -79,6 +79,7 @@ export type TextToSpeechConfig = {
   enabled: boolean
   voice?: string
   language?: string
+  autoPlay?: string
```
crazywoola:
```suggestion
  autoPlay?: ttsAutoPlay
```

web/app/components/base/features/types.ts
```diff
@@ -14,6 +14,7 @@ export type SuggestedQuestionsAfterAnswer = EnabledOrDisabled
 export type TextToSpeech = EnabledOrDisabled & {
   language?: string
   voice?: string
+  autoPlay?: string
```
crazywoola: enum ttsAutoPlay

crazywoola (commented):
See comments

crazywoola (dismissed):

takatost:
I tested the manual play mode and found it to be unresponsive, with the chrome console reporting the error message as:

```
Uncaught (in promise) SyntaxError: Unexpected token '�', "����bt9�"... is not valid JSON
```

takatost:
Maybe it's because they've all switched to streaming mode now, and the web front-end isn't compatible with manual playback mode.

ic-xu:
Could you provide a test browser? We are using Chrome for testing, which works well and has a fast response speed.

ic-xu:
> Maybe it's because they've all switched to streaming mode now, and the web front-end isn't compatible with manual playback mode.

ic-xu:
> Maybe it's because they've all switched to streaming mode now, and the web front-end isn't compatible with manual playback mode.


Could you provide a test browser? We are using Chrome for testing, which works well and has a fast response speed.

takatost (changes requested):
Ops, now in the `basic chatbot`, clicking the play button does not send a request. 
Additionally, if the TTS function is enabled for the first time in the `Chatflow`, clicking the play button will result in an error. This might be due to using a draft version instead of a published one. The error message is as follows:

```
Traceback (most recent call last):
  File "/Users/takatost/Projects/dify/api/controllers/console/app/audio.py", line 97, in post
    text_to_speech = app_model.workflow.features_dict.get('text_to_speech')
AttributeError: 'NoneType' object has no attribute 'features_dict'
```

crazywoola (approved):

takatost (approved):
thx!

PR #5681, opened 2024-06-27
Fix docker command

WilliamEspegren:
Noobs will struggle on this part a long time if they are using apt, since docker-compose actually is a package which is shown when you try to run `docker-compose -f docker-compose.middleware.yaml -p dify up -d`

image


takatost (approved):
thx!

PR #5787, opened 2024-06-30
feat: correctly delete applications using Celery workers

chazzhou:
# Description

This PR implements a robust application deletion process using Celery workers. The change ensures that app deletion is handled asynchronously, improving system reliability and laying the groundwork for better user experience. It includes the deletion of all related data, including tag bindings.

However, more work is needed to implement polling or WebSocket for actual front-end status updates. This PR focuses on the backend implementation and prepares the system for future frontend enhancements.

## Type of Change

# How Has This Been Tested?
- [x] Manually triggered app deletion and verified in the database that all related data (including tag bindings) are properly deleted
- [x] Tested error scenarios by intentionally causing failures in the deletion process
- [x] Verified that the Celery task retry mechanism works as expected

# Suggested Checklist:


chazzhou:
cc. @crazywoola

laipz8200:
Thank you for your PR! Could you please open an issue first, providing a detailed description of the problem, and include the issue link in this PR? This will help us quickly understand and review your PR.

api/tasks/remove_app_and_related_data_task.py
```diff
@@ -0,0 +1,155 @@
...
+
+def _delete_app_conversations(app_id: str):
+    db.session.query(PinnedConversation).filter(
+        PinnedConversation.conversation_id.in_(
+            db.session.query(Conversation.id).filter(Conversation.app_id == app_id)
+        )
+    ).delete(synchronize_session=False)
+    db.session.query(Conversation).filter(Conversation.app_id == app_id).delete()
+
+def _delete_app_messages(app_id: str):
```
laipz8200, line 129: I noticed several warnings popping up: `SAWarning: Coercing Subquery object into a select() for use in IN(); please pass a select() construct explicitly`.
Please take a look at this.

laipz8200 (changes requested):

api/tasks/remove_app_and_related_data_task.py
chazzhou: Fixed, thank you!

api/tasks/remove_app_and_related_data_task.py
```diff
@@ -0,0 +1,156 @@
...
+
+@shared_task(queue='app_deletion', bind=True, max_retries=3)
+def remove_app_and_related_data_task(self, app_id: str):
+    logging.info(click.style(f'Start deleting app and related data: {app_id}', fg='green'))
+    start_at = time.perf_counter()
+
+    deletion_cache_key = f'app_{app_id}_deletion'
+
+    try:
+        redis_client.set(deletion_cache_key, 'processing', ex=3600)  # Set expiry to 1 hour
```
laipz8200: What is the role of Redis here? It seems like we're only setting values without any place to get them.

laipz8200 (commented):
I still have a few questions. Please check the comments I left in the code.

api/tasks/remove_app_and_related_data_task.py
chazzhou: Redis is used here to accommodate potential status updates on the front-end. For example, a polling mechanism can be added to show the user the status of the delete task.

laipz8200: So, we don't need to use it right now, right? I'm leaning towards removing the Redis-related code because it introduces unnecessary dependencies and increases IO. If we need to retrieve the status in the future, we can always add it back.

chazzhou: I will remove them, but eventually some mechanism is needed for front end notification of these short-running tasks. Currently, if the task fails, the front end would show successfully deleted with the app still there. Not a great user experience. I can imagine the "knowledge" would require a similar task like this, I haven't looked at it yet. I think this is something better designed and implemented by your team as there are tradeoffs in designing the solution for this.

laipz8200 (approved):
Great! Thank you.

PR #5793, opened 2024-07-01
Ensure *.sh are LF-style, so that they can be used directly by Docker for Windows

chjfth:
Before this fix, after running `docker compose up` on Docker Desktop 4.31 for Windows, the nginx-1 container fails to execute `nginx/docker-entrypoint.sh`. It repeatly spout error message:

```
sh: 1: /docker-entrypoint.sh: not found
sh: 1: /docker-entrypoint.sh: not found
sh: 1: /docker-entrypoint.sh: not found
```

image

That's due to docker-entrypoint.sh is having CRLF as line separator.

So, using a `.gitattributes` file at repo root dir solves the problem once and for all.

```
*.sh      text eol=lf
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+# Ensure that .sh scripts use LF as line separator, even if they are checked out
+# to Windows(NTFS) file-system, by a user of Docker for Window. 
+# These .sh scripts will be run from the Container after `docker compose up -d`.
+# If they appear to be CRLF style, Dash from the Container will fail to execute
+# them.
+
+*.sh      text eol=lf


takatost:
Hi @chjfth, we fixed this issue yesterday in #5776. Just update to the latest code on the main branch and give it a shot to see if it resolves your problem!

chjfth:
Hello @takatost , I see your code change in #5776 that uses `sed` to strip off '\r` characters from docker-entrypoint.sh. which can be a workaround.

And I still suggest adding `.gitattributes` to the git repo so that we can avoid introducing cumbersome remedy Unix shell code into `docker-compose.yaml`, for now, and for the future.

crazywoola (approved):

PR #6135, opened 2024-07-10
fix: when http request body is raw or json, if there are line breaks and double quotes, the request will fail

hellof20:
# Description

My workflow is to send the output of the large model to an external API via http request node，when http request body is raw or json, if there are line breaks and double quotes, the request will fail.

image

image

I found that the error was caused by multi-line text or double quotes
image

# Checklist:

> [!IMPORTANT]  

Fixes 

## Type of Change

- [ ] Dependency upgrade

### DIFF
--- a/api/core/workflow/utils/variable_template_parser.py
+++ b/api/core/workflow/utils/variable_template_parser.py
@@ -91,6 +91,7 @@ class VariableTemplateParser:
         def replacer(match):
             key = match.group(1)
             value = inputs.get(key, match.group(0))  # return original matched string if key not found
+            value = value.replace('\n', '\\n').replace('"', '\\"')  # replace the line breaks in the content with /n and the double quotes with /"
             # convert the value to string
             if isinstance(value, list | dict | bool | int | float):
                 value = str(value)


hellof20:
hello @iamjoel , could you help to review this?

iamjoel:
@Yeuoly Please help to review it.

Yeuoly (changes requested):
this trick should be moved to http_request_executor.py instead of variable_template_parser as it will alter many originally normal code logics, such as when it's not in an http node, but in an LLM node

hellof20:
@Yeuoly I tried to modify it in http_request_executor.py, but when the request parameter is raw or json, it is multi-line,
```
{
"msg_type":"text",
"content":{
"text": " {"xxx"}
}
}
```
If it is replaced with '\n', it will also cause the http request parameter format to be incorrect and cause an error.
In addition, I found that there was no problem with the LLM node output, and the problem occurred after being processed by variable_template_parser, so I modified variable_template_parser

Yeuoly:
> @Yeuoly I tried to modify it in http_request_executor.py, but when the request parameter is raw or json, it is multi-line,
> 
> ```
> {
> "msg_type":"text",
> "content":{
> "text": " {"xxx"}
> }
> }
> ```
> 
> If it is replaced with '\n', it will also cause the http request parameter format to be incorrect and cause an error. In addition, I found that there was no problem with the LLM node output, and the problem occurred after being processed by variable_template_parser, so I modified variable_template_parser

Actually, I guess what we need is escaping quotes in plaintext which is a JSON field, I dig into the code and found line breaks are not handled correctly here https://github.com/langgenius/dify/blob/main/api/core/workflow/nodes/http_request/http_executor.py#L347
the original logic is checking if it's already a valid JSON string, if so, escape quotes, otherwise do nothing, maybe you can optimize this.

onevcat:
We are also facing the same problem and seems there is no way to create a correct JSON request for API use. It becomes a blocker for our use of Dify.

Is there any workaround or plan to fix?

PR #6257, opened 2024-07-13
feat: add Mingdao HAP tool, implemented read and maintain HAP application worksheet data.

coldstone:
# Description

This tool can connect to the HAP worksheet via the Mingdao HAP application API, allowing you to add, query, analyze, edit, and delete specific records.

image

## Type of Change

# Testing Instructions

1. Create an application on the Mingdao website ( [https://mingdao.com](https://mingdao.com) ) and obtain the `AppKey` and `Sign` from the "API Documentation" in the application management menu. 

image

image

2. Add the Mingdao HAP node in the flow of Dify Studio, fill in the App Key, Sign and other related parameters, and then you can interact with the HAP application data in Dify. 

image

# Checklist:

> [!IMPORTANT]  


takatost:
Would it be more appropriate to have apikey & sign configured in the tool provider as authorization method?

coldstone:
> Would it be more appropriate to have apikey & sign configured in the tool provider as authorization method?

Not so. Each application of HAP has its own `AppKey` and `Sign` parameters. If provided in the authorization method, it means that only one application in the HAP platform can be interacted with, which is not the actual situation. The fact is that users will have many HAP applications, and in a Dify flow, it is also possible to query the data of multiple HAP applications to combine into the final output content.

Yeuoly:
> > Would it be more appropriate to have apikey & sign configured in the tool provider as authorization method?
> 
> Not so. Each application of HAP has its own `AppKey` and `Sign` parameters. If provided in the authorization method, it means that only one application in the HAP platform can be interacted with, which is not the actual situation. The fact is that users will have many HAP applications, and in a Dify flow, it is also possible to query the data of multiple HAP applications to combine into the final output content.

then you can use `secret-input` as the type for `AppKey`, we also support this type for tool parameters.

coldstone:
> > Not so. Each application of HAP has its own `AppKey` and `Sign` parameters. If provided in the authorization method, it means that only one application in the HAP platform can be interacted with, which is not the actual situation. The fact is that users will have many HAP applications, and in a Dify flow, it is also possible to query the data of multiple HAP applications to combine into the final output content.
> 
> then you can use `secret-input` as the type for `AppKey`, we also support this type for tool parameters.

That's great! But it seems that modifying the type of the tool parameter to `secret-input` didn't work. 

image
image

Yeuoly:
> > > Not so. Each application of HAP has its own `AppKey` and `Sign` parameters. If provided in the authorization method, it means that only one application in the HAP platform can be interacted with, which is not the actual situation. The fact is that users will have many HAP applications, and in a Dify flow, it is also possible to query the data of multiple HAP applications to combine into the final output content.
> > 
> > 
> > then you can use `secret-input` as the type for `AppKey`, we also support this type for tool parameters.
> 
> That's great! But it seems that modifying the type of the tool parameter to `secret-input` didn't work.
> 
> image image

Yes, it works background for Agent only for now

coldstone:
I have modified the type to `secret-input`.

api/core/tools/provider/builtin/mingdao_hap/tools/add_worksheet_record.yaml
```diff
@@ -0,0 +1,78 @@
...
+    label:
+      en_US: App Key
+      zh_Hans: App Key
+    human_description:
+      en_US: The AppKey parameter for the HAP application, typically found in the application's API documentation.
+      zh_Hans: HAP 应用的 AppKey 参数，可以从应用 API 文档中查找到
+    llm_description: the AppKey parameter for the HAP application
+    form: form
+
+  - name: sign
```
takatost: This parameter `sign` should be dynamically calculated as I understand it, does it agree with my understanding here?

takatost (changes requested):

api/core/tools/provider/builtin/mingdao_hap/tools/add_worksheet_record.yaml
coldstone: No, it is static. It will be provided in plaintext as part of the authorization key in each HAP application.

takatost (dismissed):
LGTM

takatost (approved):

PR #6366, opened 2024-07-17
fix: The collect of the vector knowledge base is deleted

xtuhcy:
When updating a document, updating an empty document results in deleting the entire vector knowledge base


crazywoola:
Please open an issue first, also, please keep the PR template and check the required boxes.

PR #7005, opened 2024-08-06
fix: fix tran list issue

hustyichi:
list("abc") = ["a", "b", "c"]

Without specifying a delimiter, the existing implementation converts the text to a list of letters, which is not expected


crazywoola:
Hello thanks for your contribution, but please do not modify the template and fill them accordingly.

PR #7128, opened 2024-08-09
Improvement: join primary key to unique constraint

lpdink:
# Checklist:

> [!IMPORTANT]  

# Description

This submission modifies the table definitions by adding the primary key `id` to all `UniqueConstraint` constraints, ensuring that table creation succeeds in a distributed database. This is because in a distributed database, the distribution key must be a subset of the unique keys in order to perform deduplication. If there are two unique keys, then the distribution key needs to be selected from the intersection of these two unique keys. This PR addresses the issue discussed in https://github.com/langgenius/dify/discussions/6720. However, this requires that all future unique declarations must include the primary key to ensure persistent effectiveness.

I do understand that this may introduce some burden, but I kindly ask the maintainers to consider this suggestion as it will help Dify support large-scale distributed deployments.

## Type of Change

- [ ] Dependency upgrade

# Testing Instructions

- [x] Verified that the updated ORM definitions no longer produce the unique constraint error in a distributed database environment.
- [x] Checked that the primary key and unique constraints work as expected in single-node database


lpdink:
@bowenliang123 Could you please kindly review this PR?

bowenliang123:
cc @takatost

takatost (commented):
However, I’d like to point out that the proposed modification—adding the `id` field to existing unique constraints—could undermine the intended uniqueness of the original combinations, such as `account_id + provider` or any similar pair. Since `id` is always unique, the uniqueness constraint on these combinations would no longer prevent duplicate entries for the same values, as the uniqueness would be enforced only when the `id` differs.

bowenliang123:
> Thank you for your contribution. However, I’d like to point out that the proposed modification—adding the `id` field to existing unique constraints—could undermine the intended uniqueness of the original combinations, such as `account_id + provider` or any similar pair. Since `id` is always unique, the uniqueness constraint on these combinations would no longer prevent duplicate entries for the same values, as the uniqueness would be enforced only when the `id` differs.

+1.

PR #7262, opened 2024-08-14
Fix Site title not updated when app name is updated

RyanxLowz:
Issue: When app_name is updated, the site name is not updated.


crazywoola:
Please do not modify the issue template. Also make sure you have open a pull request first. Thanks :)

Issue #8258, opened 2024-09-11
var-picker not update when edges change in node panel


HiChen404:
### Self Checks

- [X] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [X] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [X] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [X] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [X] Please do not modify this template :) and fill in all the required fields.

### Dify version

0.8.0

### Cloud or Self Hosted

Cloud

### Steps to reproduce

When focusing on a node and changing the connection, the variables in the `var-picker` will not be updated to the latest.

https://github.com/user-attachments/assets/d51d62e5-0c65-4aee-822d-b91d6b268a56

### ✔️ Expected Behavior

 var list updated to the latest when connected edges change.

### ❌ Actual Behavior

Refresh only when switching node panel.

iamjoel:
Fixed: https://github.com/langgenius/dify/pull/8259

PR #8259, opened 2024-09-11
fix(workflow): fix var-selector not update when edges change

HiChen404:
# Checklist:

> [!IMPORTANT]  

# Description

fix var-selector not update when edges change #8258 

## Type of Change

- [ ] Dependency upgrade

# Testing Instructions

- [ ] Test A
- [ ] Test B

### DIFF
--- a/web/app/components/workflow/nodes/_base/components/variable/var-reference-picker.tsx
+++ b/web/app/components/workflow/nodes/_base/components/variable/var-reference-picker.tsx
@@ -8,6 +8,7 @@ import {
 } from '@remixicon/react'
 import produce from 'immer'
 import { useStoreApi } from 'reactflow'
+import useAvailableVarList from '../../hooks/use-available-var-list'
 import VarReferencePopup from './var-reference-popup'
 import { getNodeInfoById, isConversationVar, isENV, isSystemVar } from './utils'
 import ConstantField from './constant-field'
@@ -26,7 +27,6 @@ import {
 } from '@/app/components/base/portal-to-follow-elem'
 import {
   useIsChatMode,
-  useWorkflow,
   useWorkflowVariables,
 } from '@/app/components/workflow/hooks'
 import { VarType as VarKindType } from '@/app/components/workflow/nodes/tool/types'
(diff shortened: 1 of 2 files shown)


web/app/components/workflow/nodes/_base/components/variable/var-reference-picker.tsx
```diff
@@ -102,19 +103,8 @@ const VarReferencePicker: FC<Props> = ({
 
   const [varKindType, setVarKindType] = useState<VarKindType>(defaultVarKindType)
   const isConstant = isSupportConstantValue && varKindType === VarKindType.constant
-  const outputVars = useMemo(() => {
-    if (availableVars)
-      return availableVars
-
-    const vars = getNodeAvailableVars({
```
iamjoel, line 109: It seems that there is a bug in getNodeAvailableVars. Why not fix the problem of getNodeAvailableVars and use useAvailableVarList directly? Because there are other places in the project that use getNodeAvailableVars.

HiChen404: There is no error in `getNodeAvailableVars` itself.
The issue arises from it being wrapped in` useMemo`, causing the variable list not to refresh correctly. A simple solution is to remove useMemo.
 However, since I noticed the same logic in `useAvailableVarList`, I opted to use useAvailableVarList instead.

HiChen404: > It seems that there is a bug in getNodeAvailableVars. Why not fix the problem of getNodeAvailableVars and use useAvailableVarList directly? Because there are other places in the project that use getNodeAvailableVars.

There is no error in `getNodeAvailableVars` itself.
The issue arises from it being wrapped in useMemo, causing the variable list not to refresh correctly. A simple solution is to remove useMemo.
However, since I noticed the same logic in `useAvailableVarList`, I opted to use useAvailableVarList instead.

iamjoel: Great job! I will check it later.

iamjoel (approved):

iamjoel (approved):

Issue #8652, opened 2024-09-23
Add an option to stream for Gemini.


Hisir0909:
### Self Checks

- [X] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [X] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [X] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [X] Please do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing? Tell me about your story.

![image](https://github.com/user-attachments/assets/56d7b445-a5f9-4139-ade6-7ab380cad4be)
It is possible that the gemini interface's review mechanism causes the stream interface to have a duration that is three times longer than the no stream interface. 
Model: gemini-1.5-flash

### 2. Additional context or comments

Someone told me how to add whether to use stream configuration for LLM nodes. 

### 3. Can you help us with this feature?

- [X] I am interested in contributing to this feature.

AAEE86:
Sorry, I didn't notice that you checked the third item. If you submit a pull request, I can close it.

Hisir0909:
![image](https://github.com/user-attachments/assets/3e02f0ca-8663-4040-b83d-c97a5950ffad)
![image](https://github.com/user-attachments/assets/8e1156db-0de1-47eb-ae0f-610b63ac4d70)
In fact, @AAEE86 ‘s pr is incorrect. Simply adding the `stream` flag in the YAML file is ineffective; it's passed to the model's internal parameters, not to the `stream` flag itself.

AAEE86:
I found this problem and created a new issue #8998 
Can you solve this problem?

PR #9098, opened 2024-10-09
fix: The LLM node setting with stream set to False.

### LINKED ISSUE #8998: How to fix Gemini stream mode？
@CXwudi 在config_kwargs中设置 
···
config_kwargs.pop("stream", None) 
····
Although it can solve the error reporting problem, it is still not stream output. The code _handle_generate_stream_response seems to have a problem.

_Originally posted by @AAEE86 in https://github.com/langgenius/dify/issues/8678#issuecomment-2391088192_

Hisir0909:
# Checklist:

> [!IMPORTANT]  

# Description

Fixed the issue where the LLM node returned no results when the configured model did not use stream returns.
Fixes #8998 

## Type of Change

- [ ] Dependency upgrade

# Testing Instructions

- [x] Gemini works correctly with Stream = False.
- [x] Gemini works correctly with stream = True.


CXwudi:
The stream setting for Gemini in the workflow and chatflow works for me now. Since the change is modifying in the core part, I also tested with other providers like OpenAI, Arthorpic, and a few open sourced LLM from OpenRouter to make sure it doesn't produce regression.

The bug still exists in chatbot. I believe it is the same fix for chatbot, but it is totally fine if you just want to fix the bug in workflow and chatflow.
image

Hisir0909:
> The stream setting for Gemini in the workflow and chatflow works for me now. Since the change is modifying in the core part, I also tested with other providers like OpenAI, Arthorpic, and a few open sourced LLM from OpenRouter to make sure it doesn't produce regression.
> 
> The bug still exists in chatbot. I believe it is the same fix for chatbot, but it is totally fine if you just want to fix the bug in workflow and chatflow. image

I've barely used this chatbot, so I've forgotten about it.😰

Hisir0909:
In COT, there's no handling for the return values of non-stream modes.  However, in fc, the stream mode is determined using `stream_tool_call`. I think the `stream` flag shouldn't be used in the chatbot; modifying the YAML description of `stream` is sufficient.

CXwudi:
Sorry for the late reply. I agree that `stream` mode is not likely to be used in chatbot. Given that, user would usually expect the streaming behavior on the UI/UX side for chatbot. Hence, it is totally fine to not bother fixing it.

Hisir0909:
@laipz8200  Why was it closed?  We mean that fixing it under `workflow` and `chatflow` is correct; however, this stream setting should be disabled in `agent` mode, and the error can be ignored.

laipz8200:
> @laipz8200 Why was it closed? We mean that fixing it under `workflow` and `chatflow` is correct; however, this stream setting should be disabled in `agent` mode, and the error can be ignored.

Oops! Sorry for the misunderstanding.

Hisir0909:
> > @laipz8200 Why was it closed? We mean that fixing it under `workflow` and `chatflow` is correct; however, this stream setting should be disabled in `agent` mode, and the error can be ignored.为什么会关闭？我们的意思是，在 `workflow` 和 `chatflow` 模式下进行修复是正确的；但是，在 `agent` 模式下应禁用此流设置，错误可以忽略。
> 
> Oops! Sorry for the misunderstanding.哎呀！对不起，让您误会了。

Thank you！

laipz8200:
Our CI has been updated, could you sync your code with the main branch to pass the CI?

CXwudi:
> 2. all the other models not have a `stream` config,  the dify control whether stream mode itself

If that's so, then I believe this PR is not needed.

> This PR seems to address the issue that the Gemini model's stream mode responds five times slower than when the stream mode is not enabled. The proposed solution is to let the user decide whether to activate the stream mode. However, I don't believe this is the right approach. The core issue is understanding why the Gemini model's stream mode is so slow. Stream mode should enhance UX, but currently, it detracts from it. This fundamental performance issue should be resolved by Google, not us.

Uhhh, I think there is a misunderstanding here. We are not addressing performance issue, nor do we experienced any slowness from Gemini. This PR is mainly addressing https://github.com/langgenius/dify/issues/8998 which is a regression from https://github.com/langgenius/dify/pull/8678 as I mentioned in https://github.com/langgenius/dify/pull/8678#issuecomment-2378313174

hjlarry:
> Uhhh, I think there is a misunderstanding here. We are not addressing performance issues, nor have we experienced any slowness from Gemini. This PR is mainly addressing #8998, which is a regression from #8678, as I mentioned in [#8678 (comment)](https://github.com/langgenius/dify/pull/8678#issuecomment-2378313174).

the performance issue is  https://github.com/langgenius/dify/issues/8652  which seems https://github.com/langgenius/dify/pull/8678  to resolved.

CXwudi:
I see, sorry that I wasn't aware of the root issue.

Hisir0909:
> > Uhhh, I think there is a misunderstanding here. We are not addressing performance issues, nor have we experienced any slowness from Gemini. This PR is mainly addressing #8998, which is a regression from #8678, as I mentioned in [#8678 (comment)](https://github.com/langgenius/dify/pull/8678#issuecomment-2378313174).
> 
> the performance issue is #8652 which seems #8678 to resolved.

I'd like to mention that I originally raised the performance issue #8652, and I'm willing to fix it. However, I didn't have time at that point, and someone else submitted PR #8678. Their approach simply added the stream flag in the YAML file without modifying the code. In DIFY, all LLMs are designed to handle both stream and non-stream processing, but the stream parameter is rarely used. This PR now adapts to the stream flag in the YAML file. If you think this is unnecessary or the approach is problematic,  I will close the PR.

hjlarry:
> I'd like to mention that I originally raised the performance issue #8652, and I'm willing to fix it. However, I didn't have time at that point, and someone else submitted PR #8678. Their approach simply added the stream flag in the YAML file without modifying the code. In DIFY, all LLMs are designed to handle both stream and non-stream processing, but the stream parameter is rarely used. This PR now adapts to the stream flag in the YAML file. If you think this is unnecessary or the approach is problematic, I will close the PR.

Thank you for explaining the whole story from beginning to end.  And the stream flag was already removed from all the YAML file,  so this PR seems not so mush necessary,  what do you think about this? @laipz8200

laipz8200:
Thank you all for providing the context. If the performance issues with Gemini have been resolved, we won't need this PR anymore.

Hisir0909:
> Thank you all for providing the context. If the performance issues with Gemini have been resolved, we won't need this PR anymore.

Clearly, no.  Only version 002 of Gemini-flash is normal because version 002 defaults to having the review interface turned off (according to Google).  Other flash versions, including version 8B, are very slow in stream mode.

laipz8200:
Since the steam option for the model has been removed, I think this PR is unnecessary. Let's wait for Gemini to fix this issue.

PR #9799, opened 2024-10-24
Marking the last piece of data on each page is a duplicate issue, which can be solved by adding the id field to the order by rig and using a unique field

yuanboao:
# Checklist:

> [!IMPORTANT]  

- [ ] I have performed a self-review of my own code

# Description

This problem is in the intelligent assistant Q&A function, in which the annotation option of the last piece of data on each page is duplicated, the problem here is because the value of the create_at field in the database table is the same for all records, and in the case of a large number of duplicate create_at values, the database may have errors in determining the boundaries of pagination. When the database performs a pagination query, it needs to determine which records belong to the current page, and the internal logic of the database in handling pagination is messed up with a large number of duplicate create_at values. Because it can't distinguish records based on create_at field. This can cause the database to incorrectly include the same record multiple times in the result set in some cases, especially near page boundaries. This can result in the last record on each page being duplicated.
Therefore, to solve this problem, you can use a field that can uniquely identify each record after the create_at as the main basis for sorting. When the value of the create_at id is the same, the id field will be pagination according to the value of the id field, so that the pagination can be controlled more accurately.

Fixes 

## Type of Change

- [ ] Dependency upgrade

# Testing Instructions

- [ ] Test A
- [ ] Test B

### DIFF
--- a/api/services/annotation_service.py
+++ b/api/services/annotation_service.py
@@ -132,14 +132,14 @@ class AppAnnotationService:
                         MessageAnnotation.content.ilike("%{}%".format(keyword)),
                     )
                 )
-                .order_by(MessageAnnotation.created_at.desc(),MessageAnnotation.id.desc())
+                .order_by(MessageAnnotation.created_at.desc(), MessageAnnotation.id.desc())
                 .paginate(page=page, per_page=limit, max_per_page=100, error_out=False)
             )
         else:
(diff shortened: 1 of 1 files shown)


crazywoola:
Please fix the lint errors.

crazywoola (approved):

PR #9802, opened 2024-10-24
feat: add health check for dify api service

hwzhuhao:
# Checklist:

> [!IMPORTANT]  

# Description
1. Add health check for dify api service
image

Fixes 

## Type of Change

- [ ] Dependency upgrade

# Testing Instructions

- [ ] Test A
- [ ] Test B

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -254,6 +254,8 @@ services:
     volumes:
       # Mount the storage directory to the container, for storing user files.
       - ./volumes/app/storage:/app/api/storage
+    healthcheck:
+      test: ['CMD', 'curl', '-f', 'http://localhost:5001/health']
     networks:
       - ssrf_proxy_network
       - default


docker/docker-compose.yaml
```diff
@@ -254,6 +254,8 @@ services:
     volumes:
       # Mount the storage directory to the container, for storing user files.
       - ./volumes/app/storage:/app/api/storage
+    healthcheck:
+      test: ['CMD', 'curl', '-f', 'http://localhost:5001/health']
```
crazywoola, line 258: This doesn't seems correct to me, I think this should not be hard coded.

crazywoola (changes requested):
See comments

docker/docker-compose.yaml
hwzhuhao: Could you show me the log after execute `docker inspect docker-api-1` ? I have already tested it locally

bowenliang123: His worry comes with the hardcoded value for binded IP and port , rather than use the system environment variables. Try to use `DIFY_BIND_ADDRESS` and `DIFY_PORT` like the way in `entrypoint.sh` entry script.

hwzhuhao: I think using localhost does not mean hardcoding it. Generally, web services are designed to support access from the local machine. If need to change the hard-coded ip and port, I suggest adding a health check shell script in the Dockerfile. This script would reference the DIFY_BIND_ADDRESS  and DIFY_PORT environment variable, and then the healthcheck in docker-compose would execute this script to assess the service's health. However, I feel that this approach might be overcomplicating a relatively simple issue.

bowenliang123: Alignment is the key design and the prerequisite. There's no point in assuming it must be bonded onto localhost:5001.

PR #10735, opened 2024-11-15
Fix issue with using an JSON object with single quote in a string when testing API tools.

weimzh:
# Summary

Fix issue with using an JSON object with single quote in a string when testing API tools.

Simply replacing the single quote with double quote will cause issue if some string in the JSON object contains single quotes.  This fixes the issue.

# Checklist

### DIFF
--- a/api/core/tools/tool/api_tool.py
+++ b/api/core/tools/tool/api_tool.py
@@ -263,12 +263,14 @@ class ApiTool(Tool):
                 elif property["type"] == "object" or property["type"] == "array":
                     if isinstance(value, str):
                         try:
-                            # an array str like '[1,2]' also can convert to list [1,2] through json.loads
-                            # json not support single quote, but we can support it
-                            value = value.replace("'", '"')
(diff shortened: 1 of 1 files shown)


crazywoola (commented):
[JSON syntax](http://www.json.org/) is not Python syntax. JSON requires double quotes for its strings.

weimzh:
> [JSON syntax](http://www.json.org/) is not Python syntax. JSON requires double quotes for its strings.

What I mean is the string uses double quotes but the string CONTAINS a single quote, eg:

{"key": "value 'test'"}

weimzh:
> [JSON syntax](http://www.json.org/) is not Python syntax. JSON requires double quotes for its strings.

What I mean is the string uses double quotes but the string CONTAINS a single quote, eg:

{"key": "value 'test'"}

the original code in Dify will simply replace the simple quote into double quotes, ie:

{"key": "value 'test'"} -> {"key": "value "test""}, which will definitely cause problems.

PR #10876, opened 2024-11-20
add milvus full text search function

applepieiris:
# Summary

add the full-text search for milvus vector store using the sparse vectors provided by "BAAI/bge-m3"


crazywoola:
Please do not modify this pr template directly.

PR #10883, opened 2024-11-20
fix: refactor all 'or []' and 'or {}' logic to make code more clear

yihong0618:
# Summary

refactor most of the `or []` and `or {}` logic
main is as it
- since we use dict.get(xxx) can use default value
- some value is already [] by default there is no need to `or it` 

> [!Tip]

# Screenshots

<table>
  <tr>
  <td>Before: </td>
  <td>After: </td>
  </tr>
  <tr>
  <td>...</td>
  <td>...</td>
  </tr>
</table>

# Checklist

> [!IMPORTANT]  

### DIFF
--- a/api/core/model_runtime/model_providers/openllm/llm/openllm_generate.py
+++ b/api/core/model_runtime/model_providers/openllm/llm/openllm_generate.py
@@ -106,7 +106,7 @@ class OpenLLMGenerate:
             timeout = 120
 
         data = {
-            "stop": stop or [],
+            "stop": stop,
             "prompt": "\n".join([message.content for message in prompt_messages]),
             "llm_config": default_llm_config,
         }
--- a/api/core/model_runtime/model_providers/openllm/llm/openllm_generate.py
+++ b/api/core/model_runtime/model_providers/openllm/llm/openllm_generate.py
@@ -41,9 +41,9 @@ class OpenLLMGenerate:
         model_name: str,
         stream: bool,
         model_parameters: dict[str, Any],
-        stop: list[str],
-        prompt_messages: list[OpenLLMGenerateMessage],
-        user: str,
+        stop: list[str] | None = None,
+        prompt_messages: list[OpenLLMGenerateMessage] = [],
+        user: str | None = None,
     ) -> Union[Generator[OpenLLMGenerateMessage, None, None], OpenLLMGenerateMessage]:
         if not server_url:
             raise InvalidAuthenticationError("Invalid server URL")
@@ -106,7 +106,7 @@ class OpenLLMGenerate:
             timeout = 120
 
         data = {
-            "stop": stop,
+            "stop": stop or [],
             "prompt": "\n".join([message.content for message in prompt_messages]),
             "llm_config": default_llm_config,
         }
--- a/api/core/model_runtime/model_providers/openllm/llm/openllm_generate.py
+++ b/api/core/model_runtime/model_providers/openllm/llm/openllm_generate.py
@@ -37,12 +37,13 @@ class OpenLLMGenerateMessage:
 class OpenLLMGenerate:
     def generate(
         self,
+        *,
         server_url: str,
         model_name: str,
         stream: bool,
         model_parameters: dict[str, Any],
         stop: list[str] | None = None,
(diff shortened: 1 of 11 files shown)


laipz8200:
This behavior is unexpected because sometimes the dictionary contains entries with `{'key': None}`.

laipz8200:
However, some parts of the changes are worth keeping, and we can retain those. Could you avoid changing the `dict.get('key')` pattern while keeping the other modifications?

yihong0618:
> This behavior is unexpected because sometimes the dictionary contains entries with `{'key': None}`.

yes I did not change that logic, please check~

yihong0618:
copy that, will change the dict `key, None` part I am wrong

yihong0618:
> However, some parts of the changes are worth keeping, and we can retain those. Could you avoid changing the `dict.get('key')` pattern while keeping the other modifications?

fixed thanks

api/core/model_runtime/model_providers/openllm/llm/openllm_generate.py
```diff
@@ -106,7 +106,7 @@ def generate(
             timeout = 120
 
         data = {
-            "stop": stop or [],
+            "stop": stop,
```
laipz8200: There might be a type hint error for the `stop` parameter here. In the caller, the type hint for the `stop` parameter is:

```python
    def _generate(
        self,
        model: str,
        credentials: dict,
        prompt_messages: list[PromptMessage],
        model_parameters: dict,
        tools: list[PromptMessageTool] | None = None,
        stop: list[str] | None = None,
        stream: bool = True,
        user: str | None = None,
    ) -> LLMResult | Generator:
        ...
```

Would you mind fixing the type hint for the `stop` parameter?

api/services/app_service.py
```diff
@@ -158,7 +158,7 @@ def get_app(self, app: App) -> App:
             model_config: AppModelConfig = app.app_model_config
             agent_mode = model_config.agent_mode_dict
             # decrypt agent tool parameters if it's secret-input
-            for tool in agent_mode.get("tools") or []:
+            for tool in agent_mode.get("tools", []):
```
laipz8200: If a dict could originate from a Text or JSON field in the database, now might not be the best time to use the `dict.get(key, default)` pattern.

laipz8200 (changes requested):
Two points might need changed, please check my comments. 😊

api/core/model_runtime/model_providers/openllm/llm/openllm_generate.py
yihong0618: yes will do that, thanks

api/services/app_service.py
yihong0618: cool will change that

api/core/model_runtime/model_providers/openllm/llm/openllm_generate.py
```diff
@@ -41,9 +41,9 @@ def generate(
         model_name: str,
         stream: bool,
         model_parameters: dict[str, Any],
-        stop: list[str],
-        prompt_messages: list[OpenLLMGenerateMessage],
-        user: str,
+        stop: list[str] | None = None,
+        prompt_messages: list[OpenLLMGenerateMessage] = [],
```
laipz8200: Please avoid using reference types as default parameter values, as this can lead to potential memory leak errors.

laipz8200 (changes requested):

api/core/model_runtime/model_providers/openllm/llm/openllm_generate.py
yihong0618: yes, but I think we need to change the order, for default value, python needs them in the last of the function args.
I will use Options type here for `stop` what do you think

laipz8200: I have a simpler proposal: we can enforce users to pass arguments by name by adding a `*` after the `self` parameter. This approach will make the code more readable.

```python
    def generate(
        self,
        *,
        server_url: str,
        ...
```

laipz8200 (approved):
Thank you for your detailed work~

bowenliang123 (approved):
LGTM. Nice shots~

PR #11004, opened 2024-11-23
Add grok-vision-beta to xAI + Update grok-beta Features

taowang1993:
# Summary
This PR adds grok-vision-beta to the xAI provider.
It also updates the grok-beta model with more features.

# Screenshots
image

image

image

### DIFF
--- a/api/core/model_runtime/model_providers/x/llm/grok-vision-beta.yaml
+++ b/api/core/model_runtime/model_providers/x/llm/grok-vision-beta.yaml
@@ -1,15 +1,13 @@
-model: grok-beta
+model: grok-vision-beta
 label:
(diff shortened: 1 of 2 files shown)


taowang1993:
One caveat:

In grok-vision-beta.yaml, the feature section should only include:
  - agent-thought
  - vision

The function/tool calling feature will work without adding
  - tool-call
  - multi-tool-call
  - stream-tool-call

The function/tool calling feature will stop working if you add them to the feature section.
It will return an error saying something like "the model does not support function/tool calling".

However, the function/tool calling works just fine witout specifying these features in the feature section.

crazywoola (approved):

PR #11179, opened 2024-11-28
feat: introduce a new environment variable in order to disable Scarf analytics

realethanhsu:
…le Scarf analytics in the unstructured library.

# Summary

This PR introduces a new environment variable that allows users to disable Scarf analytics in the unstructured library.
By setting `SCARF_NO_ANALYTICS="true"`, users will benefit from the following:
1. Data Protection: Disabling analytics helps safeguard user privacy and data security, as the unstructured library collect user data without explicit consent.
2. Improved Startup Speed: The application can start up faster, particularly in poor network conditions. The unstructured library's logger module sends a request during application startup, which can lead to a poor user experience if the network is unreliable. In fact, I encountered this situation, I could not start the application at all due to this issue. Because my computer cannot access https://packages.unstructured.io .
See [github:unstructured:about analytics](https://github.com/Unstructured-IO/unstructured/blob/main/README.md#chart_with_upwards_trend-analytics) for more info.

> [!Tip]

# Screenshots

<table>
  <tr>
  <td>Before: </td>
  <td>After: </td>
  </tr>
  <tr>
  <td>...</td>
  <td>...</td>
  </tr>
</table>

# Checklist

> [!IMPORTANT]  


crazywoola:
Please add this env to `docker/.env.example` as well.

api/configs/feature/__init__.py
```diff
@@ -585,6 +585,11 @@ class RagEtlConfig(BaseSettings):
         default=None,
     )
 
+    SCARF_NO_ANALYTICS: Optional[str] = Field(
+        description="This is about whether to disable Scarf analytics in Unstructured library.",
+        default="false",
```
crazywoola, line 590: For boolean value, we prefer to use `SCARF_NO_ANALYTICS: bool`. and `default=False`

realethanhsu: Thank you for your suggestion! I agree that we should use `SCARF_NO_ANALYTICS: bool` as you proposed.
However, I would like to point out that if we implement it this way, when `app_factory.py` populates the configurations into the system environment variables, the assignment `os.environ[key] = str(value)` will convert `SCARF_NO_ANALYTICS` to `"True"` in system environment variables. 
image

In fact, the Unstructured library checks the condition using `os.getenv("SCARF_NO_ANALYTICS") != "true"`. This means that `os.getenv("SCARF_NO_ANALYTICS")` should be `"true"` rather than `"True"`.
image

crazywoola (approved):

api/configs/feature/__init__.py
crazywoola: Sounds good to me. :)

PR #11390, opened 2024-12-05
Remove the processing of single quote when testing API tools.

weimzh:
# Summary

Fix issue with using an JSON object with single quote in a string when testing API tools.

Simply replacing the single quote with double quote will cause issue if some string in the JSON object contains single quotes.  This fixes the issue.

For example:

{"key": "value 'test'"}

the original code in Dify will simply replace the simple quotes into double quotes, ie:

{"key": "value 'test'"} -> {"key": "value "test""}, which will definitely cause problems.

# Checklist

### DIFF
--- a/api/core/tools/tool/api_tool.py
+++ b/api/core/tools/tool/api_tool.py
@@ -263,9 +263,6 @@ class ApiTool(Tool):
                 elif property["type"] == "object" or property["type"] == "array":
                     if isinstance(value, str):
                         try:
-                            # an array str like '[1,2]' also can convert to list [1,2] through json.loads
-                            # json not support single quote, but we can support it
-                            value = value.replace("'", '"')
                             return json.loads(value)
                         except ValueError:
                             return value


laipz8200:
@hjlarry Could you please review this issue? I believe we do not require support for single quotes in this instance. Perhaps we could consider removing this feature instead.

hjlarry:
> @hjlarry Could you please review this issue? I believe we do not require support for single quotes in this instance. Perhaps we could consider removing this feature instead.

I think we should remove the processing of single quote,  @weimzh please just remain `json.loads(value)`. :)

weimzh:
> > @hjlarry Could you please review this issue? I believe we do not require support for single quotes in this instance. Perhaps we could consider removing this feature instead.
> 
> I think we should remove the processing of single quote, @weimzh please just remain `json.loads(value)`. :)

Done

api/core/tools/tool/api_tool.py
```diff
@@ -263,9 +263,7 @@ def _convert_body_property_type(self, property: dict[str, Any], value: Any) -> A
                 elif property["type"] == "object" or property["type"] == "array":
                     if isinstance(value, str):
                         try:
-                            # an array str like '[1,2]' also can convert to list [1,2] through json.loads
-                            # json not support single quote, but we can support it
-                            value = value.replace("'", '"')
+                            # an array str like '[1,2]' can also be converted to list [1,2] through json.loads
```
laipz8200: please also remove this comment

weimzh: done

hjlarry (approved):

laipz8200 (approved):

PR #11617, opened 2024-12-12
feat: Add AgentQL integration

colriot:
AgentQL is a query language and a set of supporting developer tools designed to identify web elements and their data using natural language and return them in the shape you define.
Added AgentQL data extraction component. See: https://docs.agentql.com/rest-api/api-reference

This pull request introduces a new component, AgentQL, to both the backend and frontend of the project. The changes include the implementation of the AgentQL component, its integration into the frontend, and the addition of a new icon for AgentQL.

Tested by running in UI:
image

image


shuhaodo:
Does dify support multiline editor for the AgentQL query input?

image

api/core/tools/provider/builtin/agentql/tools/query_data.yaml
```diff
@@ -0,0 +1,78 @@
...
+      en_US: The public URL of the webpage to extract data from.
+    form: form
+  - name: query
+    type: string
+    required: true
+    label:
+      en_US: AgentQL Query
+    human_description:
+      en_US: The AgentQL query to execute. Read more at https://docs.agentql.com/agentql-query.
+    form: form
```
colriot, line 26: I noticed that `form` value requires users to provide value manually, but `llm` value requires another block to pass the value in. Is there a way to allow both? And if not, it seems that `llm` is better for both url and query in this case. Am I on the right track?

On the other side, I don't expect LLM to "generate" a query, but it's more like "I need a way to dynamically provide it, not by LLM exactly"

api/core/tools/provider/builtin/agentql/tools/query_data.yaml
```diff
@@ -0,0 +1,78 @@
...
+  - name: url
+    type: string
+    required: true
+    label:
+      en_US: URL
+    human_description:
+      en_US: The public URL of the webpage to extract data from.
+    form: form
+  - name: query
+    type: string
```
colriot, line 20: Is there a way to configure multiline UI for a string-based param?

api/core/tools/provider/builtin/agentql/tools/query_data.py
```diff
@@ -0,0 +1,63 @@
...
+    ) -> Union[ToolInvokeMessage, list[ToolInvokeMessage]]:
+        """
+        invoke tools
+        """
+        url = tool_parameters["url"]
+        query = tool_parameters["query"]
+        timeout = tool_parameters["timeout"]
+
+        params = {}
+        params["mode"] = tool_parameters.get("mode", "fast")
```
colriot, line 26: Do I still need to provide default values here or can I just rely on default values in yaml config?

colriot (commented):
I have a few additional questions along the lines of code below (forgot to submit them yesterday)

lancemao:
I have a simple AgentQL and output workflow as the following:

Screenshot 2025-05-03 at 10 01 18

In AgentQL node, there is JSON output:

Screenshot 2025-05-03 at 10 00 40

However, in my response node, there is no JSON output from the previous node

Screenshot 2025-05-03 at 10 01 27

Cannot show any response to user. Please help me

PR #12131, opened 2024-12-26
feat: Auto-switch to authorized workspace when accessing URL

aconeshana:
# Summary

user story:
Our product managers usually share prototype URLs to each other. They usually don’t understand the program or dify product logic, so we need to guide them to switch workspaces. I made a small optimization to automatically switch workspaces if there is permission.

# Screenshots

| Before | After |
|--------|-------|
|     | no longer when accessible   |

# Checklist

> [!IMPORTANT]  

### DIFF
--- a/api/controllers/console/app/wraps.py
+++ b/api/controllers/console/app/wraps.py
@@ -27,29 +27,24 @@ def get_app_model(view: Optional[Callable] = None, *, mode: Union[AppMode, list[
             )
 
             if not app_model:
-                # If app not found in current tenant, check other workspaces
-                accessible_tenants = (
-                    db.session.query(Tenant)
+                # If app not found in current tenant, check if user has access to the app in other workspaces
+                app_model = (
+                    db.session.query(App)
+                    .join(Tenant, Tenant.id == App.tenant_id)
                     .join(TenantAccountJoin, TenantAccountJoin.tenant_id == Tenant.id)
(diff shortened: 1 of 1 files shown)


api/controllers/console/app/wraps.py
```diff
@@ -26,6 +26,31 @@ def decorated_view(*args, **kwargs):
...
             )
 
+            if not app_model:
+                # If app not found in current tenant, check other workspaces
+                accessible_tenants = (
+                    db.session.query(Tenant)
+                    .join(TenantAccountJoin, TenantAccountJoin.tenant_id == Tenant.id)
+                    .filter(TenantAccountJoin.account_id == current_user.id)
+                    .all()
+                )
```
bowenliang123: It's not the best way to check the user's accessiblilty of the targeted single tenant by loading all the tenants in to memory in one time. Try to use subquery or join to ensure that and minimize the momory footprints.

aconeshana: ok, new commit enhance this q

bowenliang123: Thx. It looks much better now in looking-up the targeting tenant.

bowenliang123:
Auto-switching workspace may conflict with the existing design of single current workspace and would bring possible impacts in the user experience. It would be a good-to-have feature but not a compulsory one, especially considering switching the workspace in every page of apps of different workspace. Let the Dify team to make th e call. cc @laipz8200

PR #12188, opened 2024-12-28
fix: default value of google storage sa to empty

kazukitakamatsu:
# Summary

Fixed an error when using Default Service Account in Google Storage.
When using Default Service Account, the environment variable GOOGLE_STORAGE_SERVICE_ACCOUNT_JSON_BASE64 must be an empty string. However, even if an empty string was specified in .env, the default string was specified in docker-compose.yaml. Therefore, the default string was changed to the empty string.

## Additional Information.
As an alternative fix, it is possible to change the default to be specified only when undefined, but since the scope of impact is large, this was addressed in this issue.

current style
`SOME_VARIABLE: ${SOME_VARIABLE:-default}`
alternative style
`SOME_VARIABLE: ${SOME_VARIABLE-default}`

see https://docs.docker.com/reference/compose-file/interpolation/

# Screenshots

# Checklist


crazywoola (approved):
LGTM

PR #12262, opened 2024-12-31
fix: dataset_clean_day

huangzhuo1949:
# Summary
fix a CLEAN_DAY bug


crazywoola:
This variable is saas only, you won't need to adjust them.

PR #12807, opened 2025-01-17
Fix the LLM response to not return Unicode text

junyan-zg:
Fix the LLM response to not return Unicode text


yihong0618 (commented):
do we have other case?

junyan-zg:
> do we have other case?

I haven't found it for the time being, but I think that everything that uses this API should be dealt with like this

crazywoola:
Please do not modify the template and link or create an issue.

PR #12994, opened 2025-01-23
Fix docker-compose command

kanoshiou:
This PR reverts #5681.

Since the [README](https://github.com/langgenius/dify/blob/d1ca492563c176ea120d732263b840ad64bfc1b4/README.md#quick-start) already instructs users to ensure that Docker Compose is installed, it is reasonable to assume that Docker Compose is already installed.
https://github.com/langgenius/dify/blob/d1ca492563c176ea120d732263b840ad64bfc1b4/README.md#L62

Moreover, this is a backend document, and I believe it is essential to ensure the accuracy of the documentation. Especially since this is within a markdown code block, every line must run correctly (if there is an error, you should Google where the problem is or raise an issue).

And if they are indeed noobs, the original modification would actually lead them in the wrong direction.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,6 @@
 # Ensure that .sh scripts use LF as line separator, even if they are checked out
 # to Windows(NTFS) file-system, by a user of Docker for Window. 
-# These .sh scripts will be run from the Container after `docker compose up -d`.
+# These .sh scripts will be run from the Container after `docker-compose up -d`.
 # If they appear to be CRLF style, Dash from the Container will fail to execute
 # them.
 
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ The easiest way to start the Dify server is through [docker compose](docker/dock
 cd dify
 cd docker
 cp .env.example .env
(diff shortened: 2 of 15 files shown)


kurokobo:
I don't think this PR is required, since the current documentation is valid. The `docker compose` (not `docker-compose`) command in the documentation is the correct command to use recent Docker Compose.

See the `CLI` section in the official `How Compose works` page: 

> The Docker CLI lets you interact with your Docker Compose applications through the `docker compose` command, and its subcommands.
> https://docs.docker.com/compose/intro/compose-application-model/

kanoshiou:
Thank you @kurokobo , I used to use standalone Docker Compose all the time, now I know there is another way of installing it as a plugin. 


Yes, you are correct. Since standalone installed Docker Compose is not recommended in the installation document, I’ll close this PR.

kanoshiou:
[Differences between `docker-compose` and `docker compose`](https://docs.docker.com/compose/releases/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2).

PR #13571, opened 2025-02-12
Non-Streaming Models Do Not Return Results Properly in _handle_invoke_result

### LINKED ISSUE #13569: Non-Streaming Models Do Not Return Results Properly in _handle_invoke_result
### Self Checks

- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）

### Dify version

0.15.2

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Steps to Reproduce
Use a model that does not support streaming (e.g., OpenAI-compatible models without streaming enabled).
```
Invoke LLMNode with stream=False.
_handle_invoke_result receives an LLMResult instead of a generator.
_handle_invoke_result returns nothing, leading to an empty generator.
The loop in _run(): 

for event in generator:
    # This never executes because generator is empty

```
does not execute, and the model result is lost.

### ✔️ Expected Behavior

When stream=False, _handle_invoke_result should yield a ModelInvokeCompletedEvent with the final response.
The for event in generator: loop should execute and process the result properly.
The model response should be correctly returned.

### ❌ Actual Behavior

_handle_invoke_result does not yield any event when stream=False, causing the generator to be empty.
The for event in generator: loop inside _run() never runs, and no response is returned.
This affects models that do not support streaming, making them unresponsive

vasu-negi:
Fix: Non-Streaming Models Do Not Return Results Properly in _handle_invoke_result
Summary
This PR fixes an issue where models that do not support streaming (stream=False) fail to return results properly. The _handle_invoke_result function was not yielding any events when it received an LLMResult, leading to an empty generator. As a result, the for event in generator: loop in _run() never executed, causing missing responses for non-streaming models.

Issue:
Non-streaming models (stream=False) return an LLMResult, but _handle_invoke_result does not yield anything.
This causes _run() to receive an empty generator, resulting in no response being sent.
Streaming models (stream=True) work fine, as their results are handled iteratively.
Root Cause:
The _handle_invoke_result method contained this condition:

FIx https://github.com/langgenius/dify/issues/13569

if isinstance(invoke_result, LLMResult):
    return  # ❌ This exits without yielding any event
Since _handle_invoke_result did not yield an event for non-streaming models, the generator was empty.
_run() expects to iterate over generator, but since it was empty, the loop never executed.
Fix Implemented
The fix ensures _handle_invoke_result always yields a ModelInvokeCompletedEvent when stream=False, preventing an empty generator.

Updated _handle_invoke_result Implementation:

def _handle_invoke_result(self, invoke_result: LLMResult | Generator) -> Generator[NodeEvent, None, None]:

    # Handle non-streaming models
    if isinstance(invoke_result, LLMResult):
        yield ModelInvokeCompletedEvent(
            text=invoke_result.message.content,
            usage=invoke_result.usage,
            finish_reason=None,  # Adjust as needed
        )
        return  # Ensures at least one event is yielded

   
Impact of the Fix
✅ Non-streaming models now return responses correctly.
✅ Streaming models (stream=True) remain unaffected.
✅ Prevents missing responses by ensuring _handle_invoke_result always yields an event.

Testing & Validation
 Verified that non-streaming models (stream=False) now return valid responses.
 Ensured that streaming models (stream=True) function as expected.
 Confirmed that _handle_invoke_result always yields at least one event, preventing an empty generator.

### DIFF
--- a/api/core/workflow/nodes/llm/node.py
+++ b/api/core/workflow/nodes/llm/node.py
@@ -246,7 +246,7 @@ class LLMNode(BaseNode[LLMNodeData]):
         return self._handle_invoke_result(invoke_result=invoke_result)
 
     def _handle_invoke_result(self, invoke_result: LLMResult | Generator) -> Generator[NodeEvent, None, None]:
-         if isinstance(invoke_result, LLMResult):
+        if isinstance(invoke_result, LLMResult):
             content = invoke_result.message.content
             if content is None:
                 message_text = ""
--- a/api/core/workflow/nodes/llm/node.py
+++ b/api/core/workflow/nodes/llm/node.py
@@ -1,4 +1,4 @@
-import json
+_handle_invoke_resultimport json
 import logging
 from collections.abc import Generator, Mapping, Sequence
 from typing import TYPE_CHECKING, Any, Optional, cast
@@ -247,6 +247,11 @@ class LLMNode(BaseNode[LLMNodeData]):
 
     def _handle_invoke_result(self, invoke_result: LLMResult | Generator) -> Generator[NodeEvent, None, None]:
         if isinstance(invoke_result, LLMResult):
+            yield ModelInvokeCompletedEvent(
+                text=invoke_result.message.content,
+                usage=invoke_result.usage,
+                finish_reason=None,
+            )
             return
 
         model = None
--- a/api/core/workflow/nodes/llm/node.py
+++ b/api/core/workflow/nodes/llm/node.py
@@ -1,4 +1,4 @@
-_handle_invoke_resultimport json
+import json
 import logging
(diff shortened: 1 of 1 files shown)


crazywoola:
Please fix the lint errors

vasu-negi:
Sure, will do

vasu-negi:
retest

vasu-negi:
i think my IDE is incorrectly formatting the IDE, i am checking.

crazywoola:
Next time you can run this cmd `./dev/reformat` in root dir. :)

laipz8200 (approved):
LGTM

PR #10511, opened 2024-11-11
Revert "fix the error of unable to retrieve url from file"

crazywoola:
Reverts langgenius/dify#10498

Issue #10497, opened 2024-11-10
Unable to retrieve url from file


XiaoLey:
### Self Checks

- [X] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [X] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [X] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [X] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [X] Please do not modify this template :) and fill in all the required fields.

### Dify version

0.11.0

### Cloud or Self Hosted

Self Hosted (Docker), Self Hosted (Source)

### Steps to reproduce

![can_not_get_url](https://github.com/user-attachments/assets/4af2f118-1c06-4f3e-83d8-6b2b032d789d)

hjlarry:
this issue is same as https://github.com/langgenius/dify/issues/9567
now the file url of message history problem is already resolved ?

XiaoLey:
> this issue is same as #9567 now the file url of message history problem is already resolved ?

@hjlarry My recent commit was a bit messy, and I am resubmitting a PR.

hjlarry:
> > this issue is same as #9567 now the file url of message history problem is already resolved ?
> 
> @hjlarry My recent commit was a bit messy, and I am resubmitting a PR.

yeah, I mean if this PR resolved,   also can close  https://github.com/langgenius/dify/issues/9567  @crazywoola

PR #13603, opened 2025-02-12
fix the error of unable to retrieve url from file

csyourui:
The issue was initially addressed in #10497, then fixed in #10498. However, it was subsequently reverted in #10511. Unfortunately, the problem still persists.


crazywoola:
@laipz8200

laipz8200 (approved):

PR #13640, opened 2025-02-13
Delete the thinking process in the chat history.

abandonever:
Delete the thinking process in the chat history.


crazywoola:
Please do not modify the pr template, you need to open an issue and we might need to know why you need to remove this.

PR #13648, opened 2025-02-13
fix: github tool NoneType has no len()

pwddd:
# Summary

解决github工具在搜索列表时，如果仓库description为空时报错NoneType has no len()问题。


crazywoola:
Do not modify the PR template, all the description should be in English. 

Please be aware that we are going to roll out the v1.0.0 changes related to models runtimes and tools will not take effect above v1.0, please make sure adapt the changes to https://github.com/langgenius/dify-official-plugins as well. :)

PR #15049, opened 2025-03-06
chore: bump pydantic to 2.11 and pydantic-settings to 2.9

### LINKED ISSUE #15048: Bump pydantic to 2.11
### Self Checks

- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）

- pydantic 2.11 enhances performance improvements since 2.9: https://pydantic.dev/articles/pydantic-v2-11-release
- bump pydantic dependencies:

pydantic = "~2.10.6"
pydantic-settings = "~2.8.1"
pydantic_extra_types = "~2.10.2"

_No response_

bowenliang123:
# Summary

- to close #15048
- pydantic 2.11 enhances 2x performance improvements and less memory usage since 2.9: https://pydantic.dev/articles/pydantic-v2-11-release
- bump pydantic dependencies:
```
pydantic = "~2.11.2"
pydantic-settings = "~2.9.1"
pydantic_extra_types = "~2.10.3"
```

image

> [!Tip]

# Screenshots

| Before | After |
|--------|-------|
| ...    | ...   |

# Checklist

> [!IMPORTANT]  


api/configs/middleware/__init__.py
```diff
@@ -171,7 +171,7 @@ def SQLALCHEMY_DATABASE_URI(self) -> str:
 
     RETRIEVAL_SERVICE_EXECUTORS: NonNegativeInt = Field(
         description="Number of processes for the retrieval service, default to CPU cores.",
-        default=os.cpu_count(),
+        default_factory=lambda: NonNegativeInt(str(os.cpu_count())),
```
Viicos: `NonNegativeInt` is defined as an annotated form:

```python
NonNegativeInt = Annotated[int, annotated_types.Ge(0)]
```

So in theory this shouldn't be necessary?

bowenliang123: Yes, but all the changes are minimum required changes for passing the type checks.

bowenliang123: Oh, I checked again the errors from pymy. Changed to use `default=os.cpu_count() or 4,` in this case. The violation does not come from type wrapper , but form the optional None from `os_count()` instead. Thanks!

bowenliang123:
cc @laipz8200

api/configs/middleware/__init__.py
```diff
@@ -171,7 +171,7 @@ def SQLALCHEMY_DATABASE_URI(self) -> str:
 
     RETRIEVAL_SERVICE_EXECUTORS: NonNegativeInt = Field(
         description="Number of processes for the retrieval service, default to CPU cores.",
-        default=os.cpu_count(),
+        default=os.cpu_count() or 8,
```
laipz8200: why use 8 here?

api/core/plugin/entities/endpoint.py
```diff
@@ -23,8 +23,8 @@ class EndpointProviderDeclaration(BaseModel):
     declaration of an endpoint group
     """
 
-    settings: list[ProviderConfig] = Field(default_factory=list)
-    endpoints: Optional[list[EndpointDeclaration]] = Field(default_factory=list)
+    settings: list[ProviderConfig] = Field(default=[])
```
laipz8200: Can you explain this change?

bowenliang123: After second check, this line change is not necessary and it has been  reverted. Thanks.

api/configs/middleware/__init__.py
bowenliang123: The `os.cpu_count()` returns `Optional[int]` which violates `mypy` type checks for NonNegativeInt. Adding an default int to pass the checks. In general most cases , `os.cpu_count()` does return valid int and it woun't fall to the default value.

```
configs/middleware/__init__.py:172: error: Incompatible types in assignment (expression has type "int | None", variable has type "int")  [assignment]
```

laipz8200:
I don't think changing `default_factory=list` to `default=[]` is a good idea. Even though Pydantic can handle it correctly, it is inconsistent with Python's behavior of using default values. I believe this modification is unnecessary.

To ensure the program runs correctly, if `os.cpu_count()` cannot be obtained, the default value should be set to 1.

The other parts LGTM.

bowenliang123:
The 8 for defualt CPU value is referrring the similar hardcoded value in `anyio` lib. And now it's set to 1, as you suggested.

As for the changes to `default=[]`, this could be responding to `Optional[T]` type in mypy checkes. Or shall we leave `default_factory` as it is and set `default=None` in response?

bowenliang123:
After checking the source code of Pydantic's code, we could see `default_factory` and `default` has the same result in our cases.  If `default_factory` method is set, it will be called with no argumenets as `default_factory()`, otherwise, it will use the `default` value.

image

So it's the same for us. I'm sure that it's the right way to change `default_factory=list` to `default=[]` with no side effects and passing the mypy type checks.

api/configs/middleware/__init__.py
bowenliang123: Changed the default value to 1. Solved.

bowenliang123:
Alright, I have figured a better way to pass the mypy checks and satisfied your preference of keeping `default_factory`.

Original code:
```
    loop_variables: Optional[list[LoopVariableData]] = Field(default_factory=list)
```

Getting Error in checks:
```
core/workflow/nodes/loop/entities.py:29: error: Argument "default_factory" to "Field" has incompatible type "type[list[_T]]"; expected "Callable[[], Never] | Callable[[dict[str, Any]], Never]"  
```

Fixed in  default_factory with [_T]:
```
    loop_variables: Optional[list[LoopVariableData]] = Field(default_factory=list[LoopVariableData])
```

laipz8200:
Using `default_factory` can pass mypy checks in the current version. If this change was introduced by an upstream update, wouldn't waiting for the upstream to fix it be a better solution?

bowenliang123:
I don't think the Pydantic will regard this a blocking issue or have a fix in it, as it occurs only in mypy checkings of static types. As I'm also reported this regression via adding Dify as third-party tests in Pydantic (https://github.com/pydantic/pydantic/pull/11719). @Viicos from Pydantic team has also reviewed this PR as well.

bowenliang123:
The change the type for `default_factory` in Pydantic's Field is reasonable to me. And I feel it's alright that mypy requests the implementation of default_factory align to the type definition. So just let us to decide the possible workaround for it , or just accept these minor changes as update `list` to `list[T]` for this argument.

api/core/workflow/utils/condition/entities.py
```diff
@@ -39,7 +39,7 @@ class SubCondition(BaseModel):
 
 class SubVariableCondition(BaseModel):
     logical_operator: Literal["and", "or"]
-    conditions: list[SubCondition] = Field(default=list)
+    conditions: list[SubCondition] = Field(default=[])
```
laipz8200: Can we also use default_factory here?

bowenliang123: Sure. Changed to use `default_factory=list` here.

laipz8200 (approved):
LGTM

PR #15068, opened 2025-03-06
fix: improve latex preprocess

pvp166:
This one to improve Latex process


crazywoola:
There is a similar issue here https://github.com/langgenius/dify/pull/14910 and please do not modify the template.

Issue #15189, opened 2025-03-07
When deploying dify locally, the tool icon cannot be displayed in workflow and agent. The solution is to modify the local nginx configuration to achieve it.


zhangxl01:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.0.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Modify the default.conf.template file under /dify/docker/nginx/conf.d，Add two rules：

```
# Rule 1: Handle application entry (keep /app/{id})
location ~ ^/app/[a-f0-9-]+$ {
# Forward the full path directly (without modifying the URL)
proxy_pass http://api:5001/;
include proxy.conf;
}
```
```
# Rule 2: When processing static resource requests (such as images), remove the /app/{id} prefix
location ~ ^/app/[a-f0-9-]+/(console/api/.*)$ {
# Rewrite path: remove /app/{id}, keep the following part
    rewrite ^/app/[a-f0-9-]+/(.*)$ /$1 break;
    proxy_pass http://api:5001;
    include proxy.conf;
}
```
<img width="473" alt="Image" src="https://github.com/user-attachments/assets/65200492-f593-4519-bf91-799980c7769f" />

Restart nginx service

problem solving

<img width="853" alt="Image" src="https://github.com/user-attachments/assets/66dd94bc-c12a-4473-a09f-c438791069a6" />

### ✔️ Expected Behavior

Help everyone solve problems

### ❌ Actual Behavior

May be solved

zhangxl01:
Do you have a screenshot? You can take a look at the situation you mentioned, I don't have this problem here.

crazywoola:
@zhangxl01 I tried, it works as expected. I can open a pr for this or you prefer do it yourself? Since this solution is proposed by you.

zhangxl01:
Okay, thank you for helping to submit a pr for this.

kurokobo:
[...]

Issue #14835, opened 2025-03-04
After upgrading to 1.0.0, the tool node icon and the tool icon in the add node pop-up box do not display


BannyLon:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

V1.0.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

I am a community version of Docker Compose deployed locally. I upgraded from V0.15.3 to V1.0.0, but the tool node icon sent for installation and the tool icon in the add node pop-up box cannot be displayed properly. Click on the menu tool or plugin, and the icon will display normally. As shown in the picture:
我是Docker Compose 部署本地的社区版，从V0.15.3版本升级到V1.0.0，发信安装的工具节点图标以及添加节点弹出框里的工具图标都不能正常显示。点击菜单工具或插件，图标能正常显示。如图：

![Image](https://github.com/user-attachments/assets/46ec9bb7-abe2-40ca-b627-e7843a74e030)

junjiem:
Whether the parameter CONSOLE_API_URL is configured?

BannyLon:
[...]

Issue #14668, opened 2025-03-02
Tool icon not displayed


zhangxl01:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

dify1.0.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

<img width="1172" alt="Image" src="https://github.com/user-attachments/assets/d6ac41e9-cc65-41ac-819b-dae0e10a4cd0" />

When using an intelligent agent to create an application, select the tool and its icon will not be displayed

crazywoola:
You need to install the tools from the marketplace if you upgrade from an older version.

zhangxl01:
> 如果您从旧版本升级，则需要从市场安装工具。

I am a brand new installation, not an upgrade from an old version

crazywoola:
I think you can just install the missing tools from the marketplace.

zhangxl01:
> I think you can just install the missing tools from the marketplace.

<img width="1381" alt="Image" src="https://github.com/user-attachments/assets/0fa3dffd-9c61-42b6-ac28-38db339118b8" />
http://127.0.0.1/app/5c7cd73b-2147-48bf-8ee9-944c00c1b854/console/api/workspaces/current/plugin/icon?tenant_id=69483186 -3613-40d3-9e6e-faf641721d9f&filename=1c5871163478957bac64c3fe3d72d003f767497d921c74b742aad27a8344a74.svg 
remove the/ app/5c7cd73b-2147-48bf-8ee9-944c00c1b854 from the address, and the image can be accessed normally. This indicates that there is a problem with the configuration in Nginx, caused by inconsistent output paths between the front-end and back-end.

kevintsai1202:
+1
https://github.com/langgenius/dify/issues/14578

backrunner:
Have same issue here, just upgrade from v0.15.3, the old plugins and the new installed plugins all cannot display the icons, return 404.

Issue #14578, opened 2025-02-28
The icon does not appear when adding a tool.


kevintsai1202:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

When adding a tool to the workflow, the icon does not appear. This issue only occurs on the community board, while it displays correctly on Cloud.

![Image](https://github.com/user-attachments/assets/1716a95e-80bb-47f5-8010-c9bc48726a0d)

kevintsai1202:
The installed plugin can display the icon, but it fails to show up when adding a new node. It’s probably not a settings issue.

![Image](https://github.com/user-attachments/assets/97b36411-8f4a-45a7-926e-206fb73480c0)

crazywoola:
You can install those missing packages back.

kevintsai1202:
> You can install those missing packages back.

After uninstalling and reinstalling the tool, the icon still doesn't appear when adding a process.

kevintsai1202:
@crazywoola 
> You can install those missing packages back.

I checked the icon URL, and the Docker icon URL is as follows:
http://localhost/app/22bb5036-a40e-4387-8ebc-4138c950eb91/console/api/workspaces/current/tool-provider/builtin/audio/icon

If I manually enter the URL below, it displays correctly. The URI below is consistent with the Cloud version. Could there be an error with the URI in the community version?
http://localhost/console/api/workspaces/current/tool-provider/builtin/audio/icon

![Image](https://github.com/user-attachments/assets/e481a7ed-3dc1-4f59-afeb-d736ae34aab4)

kevintsai1202:
@dosu 

Can you change the issue status to "open"? I found that the icon isn’t displaying because, in the community version, the icon’s URI in the workflow or agent includes an extra "/app/22bb5036-a40e-4387-8ebc-4138c950eb91," which causes it to fail to display.

PR #15279, opened 2025-03-08
fix: return absolute path as the icon url if CONSOLE_API_URL is empty

kurokobo:
# Summary

This is follow-up for #15241 by @crazywoola and @zhangxl01.

- Reverts #15241 
- Change API to solve the issue
- Add test cases for `yarl.URL`

The PR #15241 solves the issue by Nginx, but I think the issue should be solved by the API side.

The issue is fundamentally caused by the API returning a relative path (without leading `/`) as the icon URL when the `CONSOLE_API_URL` is empty.

Image

If the specified URL does not start with `/`, the browser treats it as a relative path from the current URL. This causes 404 for all icons (See the screen shot below).

Simply ensuring that the API returns a path starting with `/` as the icon URL will resolve the issue, as the browser will interpret it as an absolute path from the root of the current domain.

Related #15189, #14668, #14835, #14570, #13982, #14835, #14578

# Screenshots

| Before | After |
|--------|-------|
| Image | Image |

# Checklist

> [!IMPORTANT]  

### DIFF
--- a/api/core/tools/tool_manager.py
+++ b/api/core/tools/tool_manager.py
@@ -765,17 +765,22 @@ class ToolManager:
 
     @classmethod
     def generate_builtin_tool_icon_url(cls, provider_id: str) -> str:
-        return (
-            dify_config.CONSOLE_API_URL
-            + "/console/api/workspaces/current/tool-provider/builtin/"
-            + provider_id
-            + "/icon"
+        return str(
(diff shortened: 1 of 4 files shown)


crazywoola (approved):

crazywoola:
Cool, this seems much better to me.

PR #16083, opened 2025-03-18
chore: to expand compatibility with Docker versions.

binglun30:
# Summary

When migrating from Docker 23.0.3 to 20.10.9, it was discovered that three containers(worker, sandbox, api) encountered the following error:

`runtime/cgo: pthread_create failed: Operation not permitted
SIGABRT: abort`

As a result, the seccomp parameter was added to the runtime settings of these three containers to ensure they could also run in other versions of the Docker environment..

> [!Tip]

# Screenshots

| Before | After |
|--------|-------|
| error1error2| ...   |

# Checklist

> [!IMPORTANT]  

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -434,6 +434,8 @@ services:
   api:
     image: langgenius/dify-api:1.0.1
     restart: always
+    security_opt:
+      - seccomp=unconfined
     environment:
       # Use the shared environment variables.
       <<: *shared-api-worker-env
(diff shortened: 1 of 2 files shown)


kurokobo:
In my opinion, it doesn't seem very reasonable to completely disable the security mechanisms in the kernel by default just to support an old version of Docker.

I would recommend using an override file for individual customizations.

binglun30:
> In my opinion, it doesn't seem very reasonable to completely disable the security mechanisms in the kernel by default just to support an old version of Docker.
> 
> I would recommend using an override file for individual customizations.

Thank you for your suggestion, I agree with it and will make the necessary revisions。

PR #16227, opened 2025-03-19
fix: fix Account Model incorect

lcgash:
# Summary

when i try to create workspace by inner api, found error: AttributeError: type object 'Account' has no attribute 'query'

> [!Tip]

i can not found any relation issue

# Screenshots

| Before | After |
|--------|-------|
| AttributeError: type object 'Account' has no attribute 'query'| image |

# Checklist

> [!IMPORTANT]  

- [√ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [√] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [√] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [√] I've updated the documentation accordingly.
- [√] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods

### DIFF
--- a/api/models/account.py
+++ b/api/models/account.py
@@ -19,7 +19,7 @@ class AccountStatus(enum.StrEnum):
     CLOSED = "closed"
 
 
-class Account(UserMixin, Base, db.Model):
+class Account(UserMixin, Base, db.Model):  # type: ignore[name-defined]
     __tablename__ = "accounts"
     __table_args__ = (db.PrimaryKeyConstraint("id", name="account_pkey"), db.Index("account_email_idx", "email"))
 
--- a/api/models/account.py
+++ b/api/models/account.py
@@ -19,7 +19,7 @@ class AccountStatus(enum.StrEnum):
     CLOSED = "closed"
 
 
-class Account(UserMixin, Base):
+class Account(UserMixin, Base, db.Model):
     __tablename__ = "accounts"
     __table_args__ = (db.PrimaryKeyConstraint("id", name="account_pkey"), db.Index("account_email_idx", "email"))
 


crazywoola:
lease open an issue or link an existing issue in the description.

lcgash:
> lease open an issue or link an existing issue in the description.

done

DavideDelbianco:
@crazywoolala
The issue the user is trying to fix is related to line: 23 of api/controllers/inner_api/workspace/workspace.py
image

The fix the user proposes is to add  `db.Model` to the table Account, probably because that's the suggested fix on Stackoverflow for users that extends UserMixin using sqlalchemy

Looking at the blame graph, it seems like @Yeuoly intentionally replaced `db.Model` with `Base` during the v1.0.0 update

Now, using simple logic, I supposed there are other parts of the application that query the Account table.
Doing a search in the codebase for `db.session.query(Account).filter_by`

Yields these results:
- api/libs/login.py
image

- api/services/account_service.py
image
image

The most notable is the last one, which is doing the same filter on the email column of the Account table
It seems to be the authentication by user / password, a method I supposed is called a lot and there should have been many other tickets about this problem.

This makes me believe that maybe the fix, even if it works, could be not the right one and this piece PR should be reviewed, probably by @Yeuoly

Yeuoly (changes requested):
Please change to use `session.query(Account).filter` instead of `Account.query` as `db.Model` is not type friendly, IDEs like vscode could not hint its fields as expected

PR #16744, opened 2025-03-25
fix: fix retrieve input restriction.

### LINKED ISSUE #16724: When I use /datasets/uuid:dataset_id/retrieve(/datasets/%s/retrieve), it prompts "Query is required and cannot exceed 250 characters"
### Self Checks

### Dify version

V1.1.2

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Hi,

I have a requirement to retrieve the dataset, but I've encountered a problem. When using /retrieve, I received this error. My understanding is that there shouldn't be such a restriction in retrieve. thanks~

If possible, I can submit code to fix this issue.

Image

Image

### ✔️ Expected Behavior

I can retrieve my dataset

### ❌ Actual Behavior

{
    "code": "invalid_param",
    "message": "Query is required and cannot exceed 250 characters",
    "status": 400
}

gt-yuanmeijie:
When calling the interface, only check for the existence of the query without considering character limitations.

Closes https://github.com/langgenius/dify/issues/16724

# Summary

Resolves #<16724>

# Screenshots

| Before | After |

# Checklist

> [!IMPORTANT]  

### DIFF
--- a/api/controllers/console/datasets/hit_testing_base.py
+++ b/api/controllers/console/datasets/hit_testing_base.py
@@ -43,6 +43,11 @@ class DatasetsHitTestingBase:
         HitTestingService.hit_testing_args_check(args)
 
     @staticmethod
+    def retrieval_args_check(args):
+        HitTestingService.retrieval_args_check(args)
+
+
+    @staticmethod
     def parse_args():
         parser = reqparse.RequestParser()
 
--- a/api/controllers/console/datasets/hit_testing_base.py
+++ b/api/controllers/console/datasets/hit_testing_base.py
@@ -46,7 +46,6 @@ class DatasetsHitTestingBase:
     def retrieval_args_check(args):
         HitTestingService.retrieval_args_check(args)
 
-
     @staticmethod
     def parse_args():
         parser = reqparse.RequestParser()
--- a/api/controllers/service_api/dataset/hit_testing.py
+++ b/api/controllers/service_api/dataset/hit_testing.py
@@ -9,7 +9,7 @@ class HitTestingApi(DatasetApiResource, DatasetsHitTestingBase):
 
         dataset = self.get_and_validate_dataset(dataset_id_str)
         args = self.parse_args()
-        self.hit_testing_args_check(args)
+        self.retrieval_args_check(args)
 
         return self.perform_hit_testing(dataset, args)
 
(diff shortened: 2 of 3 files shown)


api/controllers/console/datasets/hit_testing_base.py
```diff
@@ -42,6 +42,11 @@ def get_and_validate_dataset(dataset_id: str):
     def hit_testing_args_check(args):
         HitTestingService.hit_testing_args_check(args)
 
+    @staticmethod
+    def retrieval_args_check(args):
```
crazywoola, line 46: `hit-testing` is the same with `retrieve`.

- `retrieve` is the new endpoint
- We keep the `hit-testing` mainly because we do not want to break existing applications.

crazywoola (changes requested):
May I ask how many characters do you need, it seems 250 is enough for most of the cases.
Are there any examples?
- Dead code should also be removed
- Please fix the lint errors as well.

gt-yuanmeijie:
Sorry for the delayed response. I had some matters to attend to today.
My main development language is Java, and yesterday I submitted the code using IntelliJ IDEA, which caused some formatting issues.
I might need to use these two scenarios:

Use a knowledge base for dynamic semantic deduplication.
Dynamically use different knowledge bases based on different conditions, and then call an agent to respond.
The approach for the first scenario is as follows: based on the user's query, call the retrieve interface. If there are results, return the data with the highest match. If not, call the semantic analysis agent and insert the agent's result as a segment of a dataset into the knowledge base.

The reason for the second scenario is that we need to choose the knowledge base from three dimensions, one of which increases over time. We do not want to frequently modify the application (modifying the application carries risks). My idea is to store these configuration relationships in the database. When a user asks a question, retrieve the desired knowledge base from the database based on different dimensions, and then call the app with the result returned by the retrieve interface as a parameter.

The user's questions cannot be controlled, and currently, the limit of 250 characters is indeed a bit short.

I originally intended to write a separate RetrieveApi to differentiate between retrieve and hit_test, but I am not that familiar with Python web development, so I did not make any changes. Additionally, hit_testing_args_check is used in the api/controllers/datasets/hit_testing.py file. This time, I have only made modifications to the api/controllers/service_api/dataset/hit_testing.py file.

PR #17877, opened 2025-04-11
feat: structured output

Nov1c444:
# Summary

Adding structured output capability to the LLM node, users can specify an output schema to make the LLM node produce JSON output with the specified structure.

# Screenshots

| Before | After |
|--------|-------|
| ...    | image|

# Checklist

> [!IMPORTANT]  


juniorsereno:
excellent implementation, looking forward to using it, congratulations.

api/controllers/console/app/generator.py
```diff
@@ -85,5 +85,37 @@ def post(self):
...
+    @login_required
+    @account_initialization_required
+    def post(self):
+        parser = reqparse.RequestParser()
+        parser.add_argument("instruction", type=str, required=True, nullable=False, location="json")
+        parser.add_argument("model_config", type=dict, required=True, nullable=False, location="json")
+        args = parser.parse_args()
+
+        account = current_user
+        structured_output_max_tokens = int(os.getenv("STRUCTURED_OUTPUT_MAX_TOKENS", "1024"))
```
QuantumGhost: I recommend adding a `STRUCTURED_OUTPUT_MAX_TOKENS` field to the `DifyConfig` class or another appropriate configuration model, rather than directly reading and parsing the environment variable within the API handler.

This approach ensures consistency with our existing configuration management practices and makes it easier to catch misconfigurations early, rather than only discovering them during API runtime.

api/core/llm_generator/llm_generator.py
```diff
@@ -340,3 +341,43 @@ def generate_qa_document(cls, tenant_id: str, query, document_language: str):
 
         answer = cast(str, response.message.content)
         return answer.strip()
+
+    @classmethod
+    def generate_structured_output(cls, tenant_id: str, instruction: str, model_config: dict, max_tokens: int):
```
QuantumGhost: I noticed that the `max_tokens` parameter isn’t utilized within the function body. Could you clarify if this is intentional, or if something might be missing?

api/core/llm_generator/llm_generator.py
```diff
@@ -340,3 +341,43 @@ def generate_qa_document(cls, tenant_id: str, query, document_language: str):
...
+
+        model_manager = ModelManager()
+        model_instance = model_manager.get_model_instance(
+            tenant_id=tenant_id,
+            model_type=ModelType.LLM,
+            provider=model_config.get("provider", ""),
+            model=model_config.get("name", ""),
+        )
+
+        prompt_messages = [UserPromptMessage(content=prompt)]
```
QuantumGhost: I suggest using the system role to provide instructions for JSON schema generation, and the user role to deliver the actual user input. Avoid embedding user input directly into the model's instructions through templating.

This approach helps mitigate prompt injection vulnerabilities.

api/core/model_runtime/entities/model_entities.py
```diff
@@ -197,6 +198,18 @@ class AIModelEntity(ProviderModel):
...
+    @model_validator(mode="after")
+    def validate_model(self):
+        supported_schema_keys = ["json_schema"]
+        schema_key = next((rule.name for rule in self.parameter_rules if rule.name in supported_schema_keys), None)
+        if schema_key:
+            if self.features is None:
+                self.features = [ModelFeature.STRUCTURED_OUTPUT]
+            else:
+                if ModelFeature.STRUCTURED_OUTPUT not in self.features:
+                    self.features = [*self.features, ModelFeature.STRUCTURED_OUTPUT]
```
QuantumGhost: I suggest using `self.features.append(ModelFeature.STRUCTURED_OUTPUT)` instead.

api/core/model_runtime/entities/model_entities.py
```diff
@@ -197,6 +198,18 @@ class AIModelEntity(ProviderModel):
     parameter_rules: list[ParameterRule] = []
     pricing: Optional[PriceConfig] = None
 
+    @model_validator(mode="after")
+    def validate_model(self):
+        supported_schema_keys = ["json_schema"]
+        schema_key = next((rule.name for rule in self.parameter_rules if rule.name in supported_schema_keys), None)
+        if schema_key:
```
QuantumGhost: I recommend using an early return to simplify the logic and reduce nested if statements:

```py
if not schema_key:
    return self

# Update self.features here
...

return self
```

api/core/workflow/nodes/llm/node.py
```diff
@@ -192,7 +201,20 @@ def _run(self) -> Generator[NodeEvent | InNodeEvent, None, None]:
                     self.deduct_llm_quota(tenant_id=self.tenant_id, model_instance=model_instance, usage=usage)
                     break
             outputs = {"text": result_text, "usage": jsonable_encoder(usage), "finish_reason": finish_reason}
-
+            if self.node_data.structured_output_enabled and self.node_data.structured_output:
```
QuantumGhost: Deeply nested if statements can be hard to read and maintain. 

I suggest extracting this section into a dedicated local function, such as `process_structured_output`, and placing it at the beginning of the `_run` method. You can then call this function here for improved clarity and organization.

Additionally, leveraging an early-return approach within `process_structured_output` can further simplify the logic.

api/core/workflow/nodes/agent/entities.py
```diff
@@ -24,3 +24,18 @@ class AgentInput(BaseModel):
 class ParamsAutoGenerated(Enum):
     CLOSE = 0
     OPEN = 1
+
+
+class AgentOldVersionModelFeatures(Enum):
```
QuantumGhost, line 29: Is `AgentOldVersionModelFeatures` essentially a duplicate of `ModelFeatures` without the `STRUCTURED_OUTPUT` type?

Duplicating `ModelFeatures` and renaming it as `OldVersion` seems like a questionable approach, as it can complicate long-term code maintenance.

Nov1c444: This is because there is a dependency on this code in the Python SDK. If the plugin hasn't been upgraded to the latest version, but Dify is updated to the version in this PR, it will cause the Agent node to become unavailable. Therefore, a copy of the old version of the Python SDK content is kept here. This part is fixed and will not be added to in the future.

api/core/workflow/utils/structured_output/entities.py
```diff
@@ -0,0 +1,16 @@
+from enum import StrEnum
+
+
+class ResponseFormat(StrEnum):
```
QuantumGhost, line 4: I suggest enhancing the documentation for the `ResponseFormat` enum.  

For instance, provide a brief description for each enum value, highlighting their distinctions, and include examples to illustrate their usage.

api/core/workflow/nodes/llm/node.py
```diff
@@ -926,6 +967,165 @@ def _handle_list_messages(
...
+            raise LLMNodeError("structured_output_schema is not valid JSON format")
+
+    def _check_model_structured_output_support(self) -> Optional[bool]:
+        """
+        Check if the current model supports structured output.
+
+        Returns:
+            Optional[bool]:
+                - True if model supports structured output
+                - False if model exists but doesn't support structured output
```
QuantumGhost: I'm curious about the rationale behind returning `None` here.

Is it essential to differentiate between a model lacking structured output support and a scenario where structured output is explicitly disabled?

api/core/workflow/nodes/llm/node.py
```diff
@@ -926,6 +967,165 @@ def _handle_list_messages(
...
+            dict[str, Any]: The structured output schema
+        """
+        if not self.node_data.structured_output:
+            raise LLMNodeError("Please provide a valid structured output schema")
+        structured_output_schema = json.dumps(self.node_data.structured_output.get("schema", {}), ensure_ascii=False)
+        if not structured_output_schema:
+            raise LLMNodeError("Please provide a valid structured output schema")
+
+        try:
+            schema = json.loads(structured_output_schema)
```
QuantumGhost, line 1101: Could you clarify the purpose of the `json.dumps` and `json.loads` roundtrip in this code? 

Is it intended to ensure that all elements in `structured_output["schema"]` are serializable?

api/core/workflow/nodes/llm/node.py
```diff
@@ -926,6 +967,165 @@ def _handle_list_messages(
...
+        """
+        # Convert schema to string format
+        schema_str = json.dumps(self._fetch_structured_output_schema(), ensure_ascii=False)
+
+        # Find existing system prompt with schema placeholder
+        system_prompt = next(
+            (prompt for prompt in prompt_messages if isinstance(prompt, SystemPromptMessage)),
+            None,
+        )
+        structured_output_prompt = STRUCTURED_OUTPUT_PROMPT.replace("{{schema}}", schema_str)
```
QuantumGhost, line 1015: It appears we cannot fully ensure that `schema_str` is not user-controllable. The current templating approach still seems susceptible to prompt injection attacks.

api/core/workflow/nodes/llm/node.py
```diff
@@ -926,6 +967,165 @@ def _handle_list_messages(
...
+        system_prompt_content = (
+            structured_output_prompt + "\n\n" + system_prompt.content
+            if system_prompt and isinstance(system_prompt.content, str)
+            else structured_output_prompt
+        )
+        system_prompt = SystemPromptMessage(content=system_prompt_content)
+
+        # Extract content from the last user message
+
+        filtered_prompts = [prompt for prompt in prompt_messages if not isinstance(prompt, SystemPromptMessage)]
```
QuantumGhost, line 1026: It seems that if structured output is enabled and the model doesn't natively support `json_schema`, the system prompts provided by the user are silently discarded. Should we consider explicitly highlighting this behavior in the relevant documentation?

api/core/workflow/nodes/agent/agent_node.py
```diff
@@ -251,7 +251,15 @@ def _generate_agent_parameters(
                                 prompt_message.model_dump(mode="json") for prompt_message in prompt_messages
                             ]
                     value["history_prompt_messages"] = history_prompt_messages
-                    value["entity"] = model_schema.model_dump(mode="json") if model_schema else None
+                    if model_schema:
+                        # remove structured output feature to support old version agent plugin
+                        if model_schema.features:
```
QuantumGhost: I recommend extracting this logic into a separate function (or perhaps a method on `ProviderModel`) named `remove_unsupported_model_features_for_old_version`.

```python
if model_schema.features:
    for feature in model_schema.features:
        if feature.value not in AgentOldVersionModelFeatures:
            model_schema.features.remove(feature)
```

Doing so will help simplify the compatibility logic within `AgentNode` and improve code readability and reusability.

api/core/workflow/nodes/llm/node.py
Nov1c444: You can see the notification in the interface.
image

api/core/workflow/nodes/llm/node.py
Nov1c444: Yeah, we cannot completely guarantee it, as we need to use the user's input as the basis for generation, but we will implement validation for this format on the frontend, which can partially mitigate this issue.

api/core/workflow/nodes/llm/node.py
Nov1c444: This step is to parse the output from a string type into an object with accessible internal attributes, to enable variable references later on.

QuantumGhost (approved):
LGTM

api/core/workflow/nodes/agent/agent_node.py
```diff
@@ -348,3 +353,10 @@ def _fetch_model(self, value: dict[str, Any]) -> tuple[ModelInstance, AIModelEnt
         )
         model_schema = model_type_instance.get_model_schema(model_name, model_credentials)
         return model_instance, model_schema
+
+    def _remove_unsupported_model_features_for_old_version(self, model_schema: AIModelEntity) -> AIModelEntity:
+        if model_schema.features:
+            for feature in model_schema.features:
+                if feature.value not in AgentOldVersionModelFeatures:
```
eightHundreds, line 360: Here, feature is an enum type, and feature.value is of string type.

The code snippet:

```
feature.value not in AgentOldVersionModelFeatures,
```
results in the following error: TypeError: unsupported operand type(s) for 'in': 'str' and 'EnumType'.

eightHundreds: This issue occurs in Python 3.11

Nov1c444: Thanks for your feedback, would you like to contribute to  fix the bug ?

QuantumGhost: Updating the base class of `AgentOldVersionModelFeatures` and its associated enumerations to `StrEnum` can resolve the issue.

PR #18837, opened 2025-04-26
fix: ensure proper conversation role alternation for vLLM 

### LINKED ISSUE #18810: Question Classifier doesn't work completely with vLLM
### Self Checks

### Dify version

1.3.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Launch a model with vLLM (in this case, google/gemma3-1b-it)

   ```bash
   $ vllm serve "google/gemma-3-1b-it" --device cuda
   ```
2. In Dify's settings, go to "Model Provider" and set up OpenAI-API-compatible plugin for the vLLM endpoint.
3. Create a chatflow application and workflow with a question classifier using the vLLM model.
4. Run the workflow.

My workflow for reference:

Image

It seems like this issue was addressed before (#17773), but was closed as it's "This is fixed in later version. Please upgrade to the latest version instead." So I assume this is a regression and the issue is happening again.

I think the issue is caused by this part: [question_classifier_node.py#L82-L92](https://github.com/langgenius/dify/blob/main/api/core/workflow/nodes/question_classifier/question_classifier_node.py#L82-L92). There are two successive user prompts in the messages because both `self._get_prompt_template` and `self._fetch_prompt_messages` append a user prompt at the end of the list. Since vLLM only allows alternating user/assistant messages, the question classifier fails to execute.

You can set `sys_query=None` when calling the `self._fetch_prompt_messages` method to address the issue. If needed, I can make a pull request.

### ✔️ Expected Behavior

Prerequisite:
- vLLM model is configured in Question Classifier block 

Expected Behavior:
- Question Classifier block works without an error. (= user/assistant prompts alternate in the request messages)

### ❌ Actual Behavior

Prerequisite:
- vLLM model is configured in Question Classifier block 

Actual Behavior:
- Question Classifier block fails with an error.

> [openai_api_compatible] Error: PluginInvokeError: {"args":{"description":"[models] Error: API request failed with status code 400: {\"object\":\"error\",\"message\":\"Conversation roles must alternate user/assistant/user/assistant/...\",\"type\":\"BadRequestError\",\"param\":null,\"code\":400}"},"error_type":"InvokeError","message":"[models] Error: API request failed with status code 400: {\"object\":\"error\",\"message\":\"Conversation roles must alternate user/assistant/user/assistant/...\",\"type\":\"BadRequestError\",\"param\":null,\"code\":400}"}

Image

StellaContrail:
# Summary
This PR resolves an issue where the Question Classifier returned an error when working with vLLM models. Fixes #18810 

# Screenshots

| Before | After |
|--------|-------|
| image | image |

# Checklist

> [!IMPORTANT]  


laipz8200 (changes requested):
You're trying to modify a general method to fix the vLLM. How can you confirm that this change won't break other providers?

StellaContrail:
Thanks for the review and for raising that point.

Since it’s impractical to test against every possible provider, here’s my reasoning:
- The change lives entirely within the QuestionClassifier’s `_run` method so it doesn't affect the other nodes at least.
- The user query already exists in the prompt template, so removing the duplicate "user" message won’t alter the prompt’s content semantically.
- I’ve confirmed that both vLLM (google/gemma-3-1b-it) and OpenAI (gpt-4o-mini) run and successfully choosing the right class.

If you still have concerns, the workarounds would be to patch it on the plugin side or something. I'd appreciate your thoughts on this.

crazywoola (approved):

crazywoola (dismissed):

laipz8200:
I believe implementing this as a plugin is the better option. However, if you need to make this change directly in Dify, you should remove the code related to `sys_query` from the `_fetch_prompt_messages` function and add sufficient comments to explain the reason for the modification.

StellaContrail:
@laipz8200
Thank you for the feedback.

The real issue isn’t with `_fetch_prompt_messages` alone, but with how it combines with `_get_prompt_template` in the Question Classifier both add a user message, causing two in a row. 
Removing all `sys_query` logic from `_fetch_prompt_messages` feels too broad since other parts use it. Handling this just in the Question Classifier is safer and avoids breaking other code.

Also, after checking again, the user/assistant alternation rule comes from the model’s chat template, not vLLM itself. This means other providers with similar templates might have the same problem in the future.

StellaContrail:
If there are still concerns about this change, I’m okay with closing this issue and PR for now. I think the most important thing is that everyone is comfortable with the changes :)

laipz8200:
> If there are still concerns about this change, I’m okay with closing this issue and PR for now. I think the most important thing is that everyone is comfortable with the changes :)

I think setting it to empty in the QuestionClassifier is good, but I hope you can add some comments in the code to explain why we do it.

StellaContrail:
@laipz8200 
I’ve added a comment in the code to explain the reason.

laipz8200 (approved):
LGTM

api/core/workflow/nodes/question_classifier/question_classifier_node.py
```diff
@@ -79,9 +79,13 @@ def _run(self):
             memory=memory,
             max_token_limit=rest_token,
         )
+        # Some models (e.g. Gemma, Mistral) force roles alternation (user/assistant/user/assistant...).
+        # If both self._get_prompt_template and self._fetch_prompt_messages append a user prompt,
+        # two consecutive user prompts will be generated, causing model's error.
+        # To avoid this, set sys_query to an empty string so that only one user prompt is appended at the end.
         prompt_messages, stop = self._fetch_prompt_messages(
             prompt_template=prompt_template,
-            sys_query=query,
```
AssadAlhawamdeh, line 84: Sys_query='',

AssadAlhawamdeh: Sys_query='',

AssadAlhawamdeh (commented):
api/core/workflow/nodes/question_classifier/question_classifier_node.py

PR #19353, opened 2025-05-07
unify plugin base api parms for data structure,

zhanluxianshen:


zhanluxianshen:
> Please link an existing issue to describe why you want to change this.

Although `_request_with_model` has no reference called.
But it is the same with `_request_with_plugin_daemon_response` and `_stream_request`, both method call `_request()`. But the method's parms define is not the same.

PR #19397, opened 2025-05-08
clean rag word_extractor.

zhanluxianshen:
1 image_folder has already create, before the method call.
2 `if x_child is None` is oddly.


crazywoola (approved):

zhanluxianshen:
Thanks.

PR #19470, opened 2025-05-09
sort extensions for review,

zhanluxianshen:
sort extensions for review, Its so good like that. 
image

in future ,will add one extension , hope keep the orders.


crazywoola (approved):

zhanluxianshen:
Thanks

bowenliang123:
Hey, the order of extensions is set by design and may cause unexpected consequences. Please do not sort it simply in alphabetical order just for good looking. This PR should be reverted unless full evidence is provided that all the dependent relations are clarified. cc @crazywoola @zhanluxianshen

zhanluxianshen:
@bowenliang123 thanks for the detail explain..

PR #19687, opened 2025-05-14
Support for copying nodes between workflows (This feature is unrelated to remove functions. When using the copy function, the browser will permanently retain the last copied node)."

FyhSky:
# Summary

This pull request updates the clipboard functionality in the `workflow-slice.ts` file to persist clipboard elements in `localStorage`. The most important change involves initializing `clipboardElements` from `localStorage` and ensuring updates to it are saved back to `localStorage`.

Close https://github.com/langgenius/dify/issues/17566

### Clipboard persistence updates:

* `clipboardElements` now initializes by retrieving data from `localStorage` (`clipboard_elements` key) or defaults to an empty array if no data is found.
* `setClipboardElements` has been updated to save the provided clipboard elements to `localStorage` (`clipboard_elements` key) whenever it is called.

Support for copying nodes between workflows. Resolves #19545, see https://github.com/langgenius/dify/issues/19545

PR #19708, opened 2025-05-15
Revert "Support for copying nodes between workflows (This feature is unrelated to remove functions. When using the copy function, the browser will permanently retain the last copied node).""

crazywoola:
Reverts langgenius/dify#19687

### DIFF
--- a/web/app/components/workflow/store/workflow/workflow-slice.ts
+++ b/web/app/components/workflow/store/workflow/workflow-slice.ts
@@ -37,13 +37,8 @@ export type WorkflowSliceShape = {
 export const createWorkflowSlice: StateCreator<WorkflowSliceShape> = set => ({
   workflowRunningData: undefined,
   setWorkflowRunningData: workflowRunningData => set(() => ({ workflowRunningData })),
-  clipboardElements: (() => {
-    const storedElements = localStorage.getItem('clipboard_elements')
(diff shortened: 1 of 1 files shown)


zxhlyh (approved):

crazywoola:
@FyhSky I believe we should keep the previous logic and only add the localStorage update in places where actual changes occur. The current implementation causes the copy behavior to stop working after the first time, which breaks expected functionality.

@Kevin9703 Thanks for the heads up. :)

FyhSky:
> @FyhSky I believe we should keep the previous logic and only add the localStorage update in places where actual changes occur. The current implementation causes the copy behavior to stop working after the first time, which breaks expected functionality.
> 
> @Kevin9703 Thanks for the heads up. :)



> @FyhSky I believe we should keep the previous logic and only add the localStorage update in places where actual changes occur. The current implementation causes the copy behavior to stop working after the first time, which breaks expected functionality.
> 
> @Kevin9703 Thanks for the heads up. :)



sorry for this issue. Maybe this code will execute ok

setClipboardElements: (clipboardElements) => {
set(() => ({ clipboardElements })) // forget this code
localStorage.setItem('clipboard_elements', JSON.stringify(clipboardElements))
},

If this issue is resolved, can this feature be submitted again?

tuo-yang:
will it support in chatflow?

Issue #19735, opened 2025-05-15
Automatically Persist Output Variables for Seamless Debugging in Workflow Editing


QuantumGhost:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:)
- [x] Please do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing? Tell me about your story.

Currently, Dify provides support for single-stepping through individual nodes during workflow editing, aiding in streamlining the development process. However, certain node types — such as `code`, `template`, and `variable_assigner` — require input variables to be manually provided for execution.

This manual input step is both tedious and prone to user errors, creating unnecessary friction in the debugging workflow.

To address this, I propose a feature that automatically persists the output variables of executed nodes. These persisted variables can then be reused when debugging subsequent nodes in the workflow. By eliminating the need for manual input, this enhancement would significantly simplify the process, leading to a smoother and more efficient user experience.

### 3. Can you help us with this feature?

- [x] I am interested in contributing to this feature.

PR #19737, opened 2025-05-15
feat(api): Introduce `WorkflowDraftVariable` Model

QuantumGhost:
# Summary

This PR introduces the `WorkflowDraftVariable` model along with the corresponding migration, laying the groundwork for implementing the solution to issue #19735.

# Screenshots

N/A

# Checklist

> [!IMPORTANT]  


api/models/enums.py
```diff
@@ -14,3 +14,10 @@ class UserFrom(StrEnum):
 class WorkflowRunTriggeredFrom(StrEnum):
     DEBUGGING = "debugging"
     APP_RUN = "app-run"
+
+
+class DraftVariableType(StrEnum):
+    # node means that the correspond variable
+    node = "node"
```
laipz8200:
```suggestion
    NODE = "node"
```

api/models/workflow.py
```diff
@@ -797,3 +804,204 @@ def from_variable(cls, *, app_id: str, conversation_id: str, variable: Variable)
...
+def _naive_utc_datetime():
+    return datetime.now(UTC).replace(tzinfo=None)
+
+
+class WorkflowDraftVariable(Base):
+    UNIQUE_INDEX_APP_ID_NODE_ID_NAME = "workflow_draft_variables_app_id_key"
+
+    __tablename__ = "workflow_draft_variables"
+    __table_args__ = (
+        PrimaryKeyConstraint("id", name="workflow_draft_variables_pkey"),
```
laipz8200: The `PrimaryKeyConstraint` can be removed because you used `primary_key=True`.

api/models/workflow.py
```diff
@@ -797,3 +804,204 @@ def from_variable(cls, *, app_id: str, conversation_id: str, variable: Variable)
...
+            raise ValueError("invalid selector.")
+        return selector
+
+    def _set_selector(self, value: list[str]):
+        self.selector = json.dumps(value)
+
+    def get_value(self) -> Segment | None:
+        return build_segment(json.loads(self.value))
+
+    def set_name(self, name: str):
```
laipz8200, line 910: Consider using `@property.setter` here?

api/models/workflow.py
```diff
@@ -797,3 +804,204 @@ def from_variable(cls, *, app_id: str, conversation_id: str, variable: Variable)
...
+        variable.app_id = app_id
+        variable.node_id = node_id
+        variable.name = name
+        variable.app_id = app_id
+        variable.set_value(value)
+        variable._set_selector(list(variable_utils.to_selector(node_id, name)))
+        return variable
+
+    @classmethod
+    def create_conversation_variable(
```
laipz8200: I prefer to use methods starting with "new" to create objects, just a suggestion.

api/models/workflow.py
QuantumGhost: I did consider using `@property.setter` for `value` and `selector` columns, and even tried it. However, I found that it reduced readability in this particular case.

One issue is that it becomes less clear whether we should interact with the raw database column (e.g., `_selector`, assuming we renamed it for clarity) or the wrapped property (`selector`). Additionally, renaming columns to something like `_selector` makes the database column name less intuitive and aesthetically unpleasing when working with ORM operations. Consequently, I felt that sticking to the current approach provides better clarity and usability.

api/models/workflow.py
QuantumGhost: Ah, good point! I was mostly following the definition of another model and hadn’t given it much thought.

laipz8200 (approved):

ke-feng:
@laipz8200  After upgrading Dify from 1.4.3 to 1.5.0, i get a new issue of "workflow_draft_variables" does not exist, this issue should be introduced from this commit, can you please guide how to fix it?

image

PR #20910, opened 2025-06-11
refactor: define the Dify project version in pyproject.toml

bowenliang123:
> [!IMPORTANT]
>

## Summary

- reraise #19178
- Move the dify version code from Python source code to project's pyproject.toml file.
- It helps to utilise the version feature in the uv cli commands to show or bump the version. e.g. `uv version` , `uv version bump minor`. doc: https://docs.astral.sh/uv/reference/cli/#uv-version
  - Before:
     image
  - After:
image

- The version code displays correctly on the about modal.

image

## Screenshots

| Before | After |
|--------|-------|
| ...    | ...   |

## Checklist

### DIFF
--- a/api/configs/app_config.py
+++ b/api/configs/app_config.py
@@ -106,7 +106,7 @@ class DifyConfig(
             TomlConfigSettingsSource(
                 settings_cls=settings_cls,
                 toml_file=search_file_upwards(
-                    base_dir_path=os.path.dirname(Path(__file__)),
+                    base_dir_path=Path(__file__).parent,
                     target_file_name="pyproject.toml",
                     max_search_parent_depth=2,
                 ),
--- a/api/configs/app_config.py
+++ b/api/configs/app_config.py
@@ -1,4 +1,6 @@
 import logging
+import os
+from pathlib import Path
 from typing import Any
 
 from pydantic.fields import FieldInfo
@@ -104,6 +106,7 @@ class DifyConfig(
             TomlConfigSettingsSource(
                 settings_cls=settings_cls,
                 toml_file=search_file_upwards(
+                    base_dir_path=os.path.dirname(Path(__file__)),
                     target_file_name="pyproject.toml",
                     max_search_parent_depth=2,
                 ),
--- a/api/configs/app_config.py
+++ b/api/configs/app_config.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from pathlib import Path
 from typing import Any
 
--- a/api/configs/app_config.py
+++ b/api/configs/app_config.py
@@ -2,7 +2,9 @@ import logging
 from typing import Any
 
 from pydantic.fields import FieldInfo
-from pydantic_settings import BaseSettings, PydanticBaseSettingsSource, SettingsConfigDict
+from pydantic_settings import BaseSettings, PydanticBaseSettingsSource, SettingsConfigDict, TomlConfigSettingsSource
+
+from libs.file_utils import search_file_upwards
 
 from .deploy import DeploymentConfig
 from .enterprise import EnterpriseFeatureConfig
@@ -99,4 +101,11 @@ class DifyConfig(
             RemoteSettingsSourceFactory(settings_cls),
             dotenv_settings,
             file_secret_settings,
+            TomlConfigSettingsSource(
+                settings_cls=settings_cls,
+                toml_file=search_file_upwards(
+                    target_file_name="pyproject.toml",
+                    max_search_parent_depth=2,
+                ),
+            ),
         )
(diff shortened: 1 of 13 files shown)


api/configs/app_config.py
```diff
@@ -99,4 +99,5 @@ def settings_customise_sources(
             RemoteSettingsSourceFactory(settings_cls),
             dotenv_settings,
             file_secret_settings,
+            TomlConfigSettingsSource(settings_cls, toml_file="pyproject.toml"),
```
QuantumGhost: The `pyproject.toml` parameter uses a relative path that resolves relative to the current working directory. If Dify is not started from the `api` directory, 
`TomlConfigSettingsSource` will fail to locate `"pyproject.toml"` correctly, potentially causing configuration loading issues.

Recommendation: Construct the path to `pyproject.toml` explicitly using `pathlib` to ensure it always points to the correct location regardless of the working directory:

```python
import pathlib

_CURRENT_FILE_PATH = pathlib.Path(__file__)
_PYPROJECT_TOML_PATH = _CURRENT_FILE_PATH.parent.parent / "pyproject.toml"

# Then use in the settings source:
TomlConfigSettingsSource(settings_cls, toml_file=_PYPROJECT_TOML_PATH),
```

This approach ensures the configuration file is always found relative to the location of the `app_config.py` file itself, making the application more robust when started from different working directories.

QuantumGhost (commented):
Other changes LGTM. 

cc @laipz8200 What's your idea about putting `version` back to `pyproject.toml`?

api/configs/app_config.py
bowenliang123: Thanks for the review. It's a very important hint and a persuasive sample for me. It helps me to revert some unncessary changes as well. Now it is changed to use relative path style as you suggested. Thanks!

api/configs/app_config.py
```diff
@@ -99,4 +100,7 @@ def settings_customise_sources(
             RemoteSettingsSourceFactory(settings_cls),
             dotenv_settings,
             file_secret_settings,
+            TomlConfigSettingsSource(
+                settings_cls=settings_cls, toml_file=Path(__file__).parent.parent / "pyproject.toml"
```
laipz8200: Could we consider implementing a search method to locate the nearest `pyproject.toml` file?

bowenliang123: Are you looking for a common method that could be reused in dify python plugin SDK?

bowenliang123: OK, extacted a common method to search specific file in current folder and it's parent folder.

QuantumGhost: I have considered this approach, and thought it may not worthy the effort. ;)

api/libs/file_utils.py
```diff
@@ -0,0 +1,27 @@
+import os
+from pathlib import Path
+from typing import Optional
+
+
+def search_file_upwards(
+    base_dir_path: str = os.path.dirname(Path(__file__)),
```
QuantumGhost: I suggest removing the default value for `base_dir_path` since the current default `os.path.dirname(Path(__file__))` is misleading. The [`__file__`](https://docs.python.org/3/reference/datamodel.html#module.__file__) variable always resolves to the path of the module where `search_file_upwards` is **defined** (`api/libs/file_utils.py`), not the path of the module where the function is **called**.

This means the function will always search from `api/libs/` regardless of where it's invoked, which is likely not the intended behavior. Callers should explicitly provide the starting directory to ensure predictable search behavior.

api/configs/app_config.py
bowenliang123: Either way is good for me. Any further preference from @laipz8200 ?

api/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "dify-api"
-dynamic = ["version"]
+version = "1.4.3"
```
QuantumGhost, line 3: The version number should be updated if #21415 is merged before this PR.

QuantumGhost (dismissed):
Other change LGTM.

api/pyproject.toml
bowenliang123: cc @laipz8200

api/libs/file_utils.py
bowenliang123: Great suggestion. The base_dir_path is now set explictly in calling. Resolved.

api/configs/app_config.py
bowenliang123: An error will be thrown when the target file is not found.

image

QuantumGhost (dismissed):
LGTM.

Thank you for your contribution!

bowenliang123:
Conflicts resolved. Please review. cc @laipz8200 @QuantumGhost

laipz8200 (approved):
LGTM

bowenliang123:
🎉 Glad to see this PR was merged. Thanks for the detailed reviews. @laipz8200 @QuantumGhost

Issue #20449, opened 2025-05-29
Proposal: Removing `db.session` Usage from core/workflow Module


laipz8200:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:)
- [x] Please do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing? Tell me about your story.

Part of #19429

## Background

The `core/workflow` module currently relies on Flask-SQLAlchemy's `db.session` for database operations. This approach has several drawbacks:

1. **Hard to manage session lifecycle**: Flask-SQLAlchemy's session is tied to the Flask request context, making it difficult to control in background tasks or multi-threaded environments.

2. **Causes bugs in complex workflows**: The shared session can lead to unexpected behavior when multiple operations are performed concurrently, especially in the workflow engine which uses threading.

3. **Tight coupling to Flask**: Direct usage of `db.session` creates a dependency on Flask, making the core workflow module less portable and harder to test.

## Current Implementation Analysis

After examining the codebase, I found:

1. The `core/workflow` module defines repository interfaces that abstract database operations:
   - `WorkflowExecutionRepository`
   - `WorkflowNodeExecutionRepository`

2. The implementations of these interfaces are in `core/repositories`:
   - `SQLAlchemyWorkflowExecutionRepository`
   - `SQLAlchemyWorkflowNodeExecutionRepository`

3. These implementations already use a better approach - they accept a `session_factory` or `engine` in their constructor and create sessions when needed:
   ```python
   def __init__(self, session_factory: sessionmaker | Engine, ...):
       # ...
   
   def save(self, execution: NodeExecution) -> None:
       # ...
       with self._session_factory() as session:
           session.merge(db_model)
           session.commit()
   ```

4. However, there are still places in the workflow module that import and potentially use `db.session` directly, such as in `graph_engine.py`.

## Proposed Solution

1. **Complete the Repository Pattern Implementation**:
   - Ensure all database operations in the workflow module go through the repository interfaces
   - Remove any direct imports of `db` from `extensions.ext_database`

2. **Session Factory Injection**:
   - Modify any remaining components that use `db.session` directly to accept a session factory
[...]

PR #21061, opened 2025-06-16
feat: add a flask_context_manager.

laipz8200:
Signed-off-by: -LAN- <laipz8200@outlook.com>

> [!IMPORTANT]
>

## Summary

part of #20449

Replaces manual context management and user restoration logic with a
unified `flask_context_manager` for better code maintainability and
clarity. Removes redundant import of `has_request_context` from
multiple modules. Updates `.gitignore` to exclude `.roo/` directory,
presumably related to AI Assistant outputs.

## Screenshots

| Before | After |
|--------|-------|
| ...    | ...   |

## Checklist


api/core/helper/__init__.py
```diff
@@ -0,0 +1,3 @@
+from core.helper.flask_context import flask_context_manager
```
QuantumGhost: I think we should avoid creating broad packages named `helper` or `utils`, as they tend to become catch-all directories that grow unwieldy over time.

Instead, I'd recommend placing `flask_context_manager` in a more specific location like `libs/flask_utils.py`. This approach offers several benefits:

1. **Better semantic grouping** - Flask-related utilities are clearly organized together
2. **Cleaner imports** - Direct imports are more explicit and easier to track
3. **Improved maintainability** - Avoids the "junk drawer" problem common with generic helper packages

The import would then be:

```python
from libs.flask_utils import flask_context_manager
```

This follows the principle of organizing code by domain/functionality rather than by generic categories.

api/core/helper/flask_context.py
```diff
@@ -0,0 +1,61 @@
...
+    with flask_app.app_context():
+        try:
+            # Restore user in new app context if it was saved
+            if saved_user is not None:
+                g._login_user = saved_user
+
+            # Yield control back to the caller
+            yield
+        finally:
+            # Any cleanup can be added here if needed
```
QuantumGhost: Should we reset context vars at the end of `flask_context_manager`? 

If it's not necessary, I suggest adding a comment to explain why we don't need to do so, and adding appropriate comments about the prerequisites to the docstring. For example, the `context_vars` should be a copy of the original `Context`, so no resetting is necessary.

api/core/helper/flask_context.py
```diff
@@ -0,0 +1,65 @@
...
+from contextlib import contextmanager
+from typing import TypeVar
+
+from flask import Flask, g, has_request_context
+
+T = TypeVar("T")
+
+
+@contextmanager
+def flask_context_manager(
```
QuantumGhost: I think it would be better to rename this function to `preserve_flask_contexts`. This name more clearly conveys the function's primary purpose of preserving Flask application context and user state across context boundaries.

Besides,  I strongly recommend adding comprehensive tests for this utility.

QuantumGhost (dismissed):
Other changes LGTM.

api/core/helper/flask_context.py
```diff
@@ -0,0 +1,65 @@
...
+    1. flask-login's UserProxy copy
+    2. ContextVars copy
+    3. flask_app.app_context()
+
+    This context manager ensures that the Flask application context is properly set up,
+    the current user is preserved across context boundaries, and any provided context variables
+    are set within the new context.
+
+    Note:
+        This manager aims to allow use current_user cross thread and app context,
```
QuantumGhost: 👍

api/core/helper/__init__.py
laipz8200: SGTM

QuantumGhost (approved):
LGTM

PR #21153, opened 2025-06-18
bug: fix sequence number may be duplicated when multi-threads running the same workflow #21047

NeatGuyCoding:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 2. Ensure there is an associated issue and you have been assigned to it
> 3. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

This pull request fixes a bug where sequence numbers could be duplicated when multiple threads were running the same workflow concurrently. The issue was resolved by modifying the SQL query used to fetch the next sequence number. Specifically:

- The query now uses `with_for_update()` to ensure that concurrent executions are properly synchronized.
- The `coalesce` function is used to safely calculate the next sequence number, defaulting to 0 if no records exist.

This change ensures that sequence numbers remain unique, even under high concurrency.

Close #21047

## Screenshots

| Before | After |
|--------|-------|
| Sequence numbers could be duplicated in concurrent workflows | Sequence numbers are now unique in all workflows |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `dev/reformat` (backend) and `cd web && npx lint-staged` (frontend) to appease the lint gods.

PR #21207, opened 2025-06-19
Revert "bug: fix sequence number may be duplicated when multi-threads running the same workflow #21047"

QuantumGhost:
The changes introduced in #21153 are problematic because the `FOR UPDATE` clause cannot be used with aggregate functions such as `MAX()` or `COALESCE()`. PostgreSQL does not allow locking rows when the query involves computed aggregate results instead of actual table rows. This limitation stems from the database's inability to determine which specific rows should be locked under such circumstances.

Here is an excerpt from the error log:

```bash
2025-06-19 02:16:27,552 ERROR [_internal.py:97]  Error on request:
Traceback (most recent call last):
  File "/Users/novicelee/pro/dify/api/.venv/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1967, in _exec_single_context
    self.dialect.do_execute(
  File "/Users/novicelee/pro/dify/api/.venv/lib/python3.12/site-packages/sqlalchemy/engine/default.py", line 941, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.FeatureNotSupported: FOR UPDATE is not allowed with aggregate functions
```

Reverts langgenius/dify#21153
Reopens #21047.

### DIFF
--- a/api/core/repositories/sqlalchemy_workflow_execution_repository.py
+++ b/api/core/repositories/sqlalchemy_workflow_execution_repository.py
@@ -151,17 +151,12 @@ class SQLAlchemyWorkflowExecutionRepository(WorkflowExecutionRepository):
             existing = session.scalar(select(WorkflowRun).where(WorkflowRun.id == domain_model.id_))
             if not existing:
                 # For new records, get the next sequence number
-                # in case multiple executions are created concurrently, use for update
-                stmt = (
-                    select(func.coalesce(func.max(WorkflowRun.sequence_number), 0) + 1)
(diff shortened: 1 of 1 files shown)


crazywoola (approved):

NeatGuyCoding:
Sorry for this

QuantumGhost:
@NeatGuyCoding No worries, these things are part of the process! Thank you for your effort and contribution—it’s always appreciated. Let’s continue refining and improving together! 😊

Issue #20147, opened 2025-05-23
Proposal: Add in-memory repository to improve workflow api perf


xinlmain:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:)
- [x] Please do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing? Tell me about your story.

# Core Objective

The default workflow-run api is very slow on simple tasks, making it unacceptable for real time interactions. The api calls spend most of the time interacting with DB. In a scenario where the api is called by another service, there is not much sense to save each call into DB.

![Image](https://github.com/user-attachments/assets/9a9ddc8f-db04-40db-b499-8c9322863ccc)

# Approach to Achieve the Objective
Add an in-memory repository so the execution won't interact with DB too much.

# Implementation Plan
1. Add a new WorkflowNodeExecutionRepository implementation with pure in-memory underneath. This is viable as workflows are each executed in 1 process only.
2. Add a toggle to switch the repositories.

# Expected Outcomes
Improved performance: Workflows api can become faster, and thus enables deploying a dedicated api service for serving external requests.

# Tradeoffs
Not able to see execution detail in application logs from the web page.

### 2. Additional context or comments

none.

### 3. Can you help us with this feature?

- [x] I am interested in contributing to this feature.

jzymx50:
I found for each api call, Dify will update the last updated time for api token in db, making high concurrent requests very slow. Could you also help to improve that? The function name is validate_and_get_api_token @xinlmain

zzy-life:
They don't think about high concurrency, they want to make money

zzy-life:
The redis container is started, but redis is not even used now

laipz8200:
Hi @xinlmain.

Thank you for your patience. In #21458, I have made two repositories configurable. Could you please take a look and consider implementing this feature based on the new structure? I believe that migrating persistence operations to Celery tasks is a better approach than abandoning persistence. I would appreciate hearing your thoughts.

PR #21458, opened 2025-06-25
feat(api/repo): Allow to config repository implementation

laipz8200:
> [!IMPORTANT]
>

## Summary

related #20147 

## Screenshots

| Before | After |
|--------|-------|
| ...    | ...   |

## Checklist


laipz8200:
**Update:** Nearly finished, with only some code related to `WorkflowRun` remaining.

laipz8200:
Hi @QuantumGhost. Do you have some time to take a look at this?

api/core/app/apps/common/workflow_response_converter.py
```diff
@@ -63,8 +64,10 @@ def __init__(
         self,
         *,
         application_generate_entity: Union[AdvancedChatAppGenerateEntity, WorkflowAppGenerateEntity],
+        workflow_execution_repository: WorkflowExecutionRepository,
     ) -> None:
         self._application_generate_entity = application_generate_entity
+        self._workflow_execution_repository = workflow_execution_repository
```
QuantumGhost: The `workflow_execution_repository` parameter is added to the `WorkflowResponseConverter` constructor but the stored `_workflow_execution_repository` instance variable appears to be unused throughout the class. Consider either implementing its usage or removing this unused dependency to maintain code clarity.

api/core/memory/token_buffer_memory.py
```diff
@@ -36,28 +42,18 @@ def get_history_prompt_messages(
...
-                Message.workflow_run_id,
-                Message.parent_message_id,
-                Message.answer_tokens,
-            )
-            .filter(
-                Message.conversation_id == self.conversation.id,
-            )
-            .order_by(Message.created_at.desc())
+        stmt = (
+            select(Message).where(Message.conversation_id == self.conversation.id).order_by(Message.created_at.desc())
```
QuantumGhost, line 46: The original query only selected specific fields of the `Message` model, but the new query selects all fields from the database. Since the downstream logic only uses the originally selected fields, this change may introduce performance degradation by fetching unnecessary data.

I've verified that the code below indeed only uses the originally selected fields, so the original implementation appears to be an optimization. Consider maintaining this optimization by using field-specific selection in the new SQLAlchemy 2.0 style query. 

```python
# the code is suggested by AI.
stmt = select(
    Message.id,
    Message.query, 
    Message.answer,
    Message.created_at,
    Message.workflow_run_id,
    Message.parent_message_id,
    Message.answer_tokens
).where(Message.conversation_id == self.conversation.id).order_by(Message.created_at.desc())
```

api/core/memory/token_buffer_memory.py
```diff
@@ -74,18 +70,20 @@ def get_history_prompt_messages(
...
-                    if message.workflow_run_id:
-                        workflow_run = (
-                            db.session.query(WorkflowRun).filter(WorkflowRun.id == message.workflow_run_id).first()
-                        )
-
-                        if workflow_run and workflow_run.workflow:
-                            file_extra_config = FileUploadConfigManager.convert(
-                                workflow_run.workflow.features_dict, is_vision=False
-                            )
+                    raise ValueError(f"Invalid app mode: {self.conversation.mode}")
```
QuantumGhost: Consider using `AssertionError` instead of `ValueError` for the unreachable code path at line 86:

```python
raise AssertionError("this statement is unreachable")
```

`ValueError` is commonly caught in various places throughout the codebase, which could lead to this exception being silently ignored. `AssertionError` better indicates that this represents a programming error rather than invalid input data, making debugging easier if this code path is unexpectedly reached.

api/core/prompt/utils/get_thread_messages_length.py
```diff
@@ -8,19 +10,9 @@ def get_thread_messages_length(conversation_id: str) -> int:
...
-            Message.id,
-            Message.parent_message_id,
-            Message.answer,
-        )
-        .filter(
-            Message.conversation_id == conversation_id,
-        )
-        .order_by(Message.created_at.desc())
-    )
+    stmt = select(Message).where(Message.conversation_id == conversation_id).order_by(Message.created_at.desc())
```
QuantumGhost, line 13: The [suggestion above](https://github.com/langgenius/dify/pull/21458/files#r2194435264) about selecting specific fields applies to here, too.

api/core/memory/token_buffer_memory.py
QuantumGhost: I'll check the schema and average size of the `messages` table later, to provide more data about whether we should keep the optimization.

api/core/repositories/factory.py
```diff
@@ -0,0 +1,215 @@
...
+        """
+        try:
+            module_path, class_name = class_path.rsplit(".", 1)
+            module = importlib.import_module(module_path)
+            return getattr(module, class_name)
+        except (ValueError, ImportError, AttributeError) as e:
+            raise RepositoryImportError(f"Cannot import repository class '{class_path}': {e}") from e
+
+    @staticmethod
+    def _validate_repository_interface(repository_class: Any, expected_interface: Any) -> None:
```
QuantumGhost: I suggest changing the type annotation for `repository_class` from `Any` to `type` in this module to better indicate that `repository_class` should be a class. 
Additionally, `expected_interface` could be changed to `type` or `type[Protocol]`. 

(I have already made this change in my fork and can push it to the source branch if we agree on the change.)

api/core/repositories/factory.py
```diff
@@ -0,0 +1,215 @@
...
+            expected_interface: The expected interface/protocol
+
+        Raises:
+            RepositoryImportError: If the class doesn't implement the interface
+        """
+        # Check if the class has all required methods from the protocol
+        required_methods = [
+            method
+            for method in dir(expected_interface)
+            if not method.startswith("_") and callable(getattr(expected_interface, method, None))
```
QuantumGhost, line 79: Consider marking required methods in the Protocol with `@abstractmethod`. This would enable more reliable validation of required methods by leveraging the [`__isabstractmethod__ `](https://docs.python.org/3/library/abc.html#abc.ABCMeta:~:text=as%20abstract%20using-,__isabstractmethod__,-.%20In%20general%2C%20this) magic attribute instead of the current introspection approach.

There's also an undocumented magic attribute [`__abstractmethods__`](https://discuss.python.org/t/can-abstractmethods-be-made-documented-public/95164) that could be used to retrieve required methods directly. However, since it's undocumented, I suggest we stick with `__isabstractmethod__` for better reliability and maintainability.

api/core/repositories/factory.py
```diff
@@ -0,0 +1,215 @@
...
+        Validate that a repository class constructor accepts required parameters.
+
+        Args:
+            repository_class: The class to validate
+            required_params: List of required parameter names
+
+        Raises:
+            RepositoryImportError: If the constructor doesn't accept required parameters
+        """
+        import inspect
```
QuantumGhost: The `import inspect` statement should be moved to top level.

api/core/repositories/factory.py
```diff
@@ -0,0 +1,215 @@
...
+
+        try:
+            signature = inspect.signature(repository_class.__init__)
+            param_names = list(signature.parameters.keys())
+
+            # Remove 'self' parameter
+            if "self" in param_names:
+                param_names.remove("self")
+
+            missing_params = [param for param in required_params if param not in param_names]
```
QuantumGhost, line 121: I considered defining `__init__` with a required signature in the repository protocol classes. However, MyPy [does not enforce compatibility for constructors](https://github.com/python/mypy/issues/3823), which would make such changes less effective for type checking purposes.

api/core/repositories/factory.py
```diff
@@ -0,0 +1,215 @@
...
+            return repository_class(  # type: ignore[no-any-return]
+                session_factory=session_factory,
+                user=user,
+                app_id=app_id,
+                triggered_from=triggered_from,
+            )
+        except RepositoryImportError:
+            # Re-raise our custom errors as-is
+            raise
+        except Exception as e:
```
QuantumGhost, line 174: I believe exceptions raised during repository instantiation should not be caught and converted to `RepositoryImportError`. 

These errors are typically caused by incorrect implementation details (e.g., constructor issues, dependency injection problems). Converting them to `RepositoryImportError` doesn't provide meaningful value from an error handling perspective. Consider allowing these exceptions to propagate with their original context for better debugging.

api/core/repositories/factory.py
```diff
@@ -0,0 +1,215 @@
...
+            if "self" in param_names:
+                param_names.remove("self")
+
+            missing_params = [param for param in required_params if param not in param_names]
+            if missing_params:
+                raise RepositoryImportError(
+                    f"Repository class '{repository_class.__name__}' constructor does not accept required parameters: "
+                    f"{missing_params}. Expected parameters: {required_params}"
+                )
+        except Exception as e:
```
QuantumGhost, line 127: [The comment below about error handling](https://github.com/langgenius/dify/pull/21458/files#r2194585083) applies to here, too.

api/core/repositories/factory.py
```diff
@@ -0,0 +1,215 @@
...
+            return repository_class(  # type: ignore[no-any-return]
+                session_factory=session_factory,
+                user=user,
+                app_id=app_id,
+                triggered_from=triggered_from,
+            )
+        except RepositoryImportError:
+            # Re-raise our custom errors as-is
+            raise
+        except Exception as e:
```
QuantumGhost, line 220: ditto

api/repositories/api_workflow_node_execution_repository.py
```diff
@@ -0,0 +1,196 @@
...
+    def get_execution_by_id(
+        self,
+        execution_id: str,
+        tenant_id: Optional[str] = None,
+    ) -> Optional[WorkflowNodeExecutionModel]:
+        """
+        Get a workflow node execution by its ID.
+
+        This method retrieves a specific execution by its unique identifier.
+        Tenant filtering is optional for cases where the execution ID is globally unique.
```
QuantumGhost, line 96: I recommend adding the following comment about the optional `tenant_id`:

```python
# When `tenant_id` is None, it's the caller's responsibility to ensure proper data isolation between tenants.
# If the `execution_id` comes from untrusted sources (e.g., retrieved from an API request), the caller should 
# set `tenant_id` to prevent horizontal privilege escalation.
```

api/repositories/api_workflow_node_execution_repository.py
```diff
@@ -0,0 +1,196 @@
...
+        Delete workflow node executions by their IDs.
+
+        This method deletes specific executions by their IDs,
+        typically used after backing up the data.
+
+        Args:
+            execution_ids: List of execution IDs to delete
+
+        Returns:
+            The number of executions deleted
```
QuantumGhost, line 195: Notes about hard deletion should be added to here too.

api/repositories/factory.py
```diff
@@ -0,0 +1,103 @@
...
+            repository_class = cls._import_class(class_path)
+            cls._validate_repository_interface(repository_class, DifyAPIWorkflowNodeExecutionRepository)
+            # Service repository requires session_maker parameter
+            cls._validate_constructor_signature(repository_class, ["session_maker"])
+
+            return repository_class(session_maker=session_maker)  # type: ignore[no-any-return]
+        except RepositoryImportError:
+            # Re-raise our custom errors as-is
+            raise
+        except Exception as e:
```
QuantumGhost, line 63: [The comment below about error handling](https://github.com/langgenius/dify/pull/21458/files#r2194585083) applies to here, too.

api/repositories/factory.py
```diff
@@ -0,0 +1,103 @@
...
+        logger.debug(f"Creating APIWorkflowRunRepository from: {class_path}")
+
+        try:
+            repository_class = cls._import_class(class_path)
+            cls._validate_repository_interface(repository_class, APIWorkflowRunRepository)
+            # Service repository requires session_maker parameter
+            cls._validate_constructor_signature(repository_class, ["session_maker"])
+
+            return repository_class(session_maker=session_maker)  # type: ignore[no-any-return]
+        except RepositoryImportError:
```
QuantumGhost, line 98: ditto

api/repositories/sqlalchemy_api_workflow_node_execution_repository.py
```diff
@@ -0,0 +1,286 @@
...
+            execution_id: The execution identifier
+            tenant_id: Optional tenant identifier for additional filtering
+
+        Returns:
+            The WorkflowNodeExecutionModel if found, or None if not found
+        """
+        stmt = select(WorkflowNodeExecutionModel).where(WorkflowNodeExecutionModel.id == execution_id)
+
+        # Add tenant filtering if provided
+        if tenant_id is not None:
```
QuantumGhost, line 141: The [comment above about tenant_id](https://github.com/langgenius/dify/pull/21458/files#r2194602462) applies to here too.

api/repositories/sqlalchemy_api_workflow_run_repository.py
```diff
@@ -0,0 +1,218 @@
...
+            if last_id:
+                # Get the last workflow run for cursor-based pagination
+                last_run_stmt = base_stmt.where(WorkflowRun.id == last_id)
+                last_workflow_run = session.scalar(last_run_stmt)
+
+                if not last_workflow_run:
+                    raise ValueError("Last workflow run not exists")
+
+                # Get records created before the last run's timestamp
+                workflow_runs = session.scalars(
```
QuantumGhost: The logic here could be simplified to 

```py
base_stmt = base_stmt.where(
    WorkflowRun.created_at < last_workflow_run.created_at,
    WorkflowRun.id != last_workflow_run.id,
)
```

and remove the `else` branch.

api/repositories/sqlalchemy_api_workflow_run_repository.py
```diff
@@ -0,0 +1,218 @@
...
+                    .limit(limit)
+                ).all()
+            else:
+                # First page - get most recent records
+                workflow_runs = session.scalars(base_stmt.order_by(WorkflowRun.created_at.desc()).limit(limit)).all()
+
+            # Check if there are more records for pagination
+            has_more = False
+            if len(workflow_runs) == limit:
+                current_page_last_run = workflow_runs[-1]
```
QuantumGhost: A better approach would be querying `limit + 1` record, and check if the number of returned records is `limit +1`.

(It's ok to keep the logic as is here.)

api/services/workflow_draft_variable_service.py
```diff
@@ -116,8 +118,13 @@ def load_variables(self, selectors: list[list[str]]) -> list[Variable]:
 class WorkflowDraftVariableService:
     _session: Session
 
-    def __init__(self, session: Session) -> None:
+    def __init__(self, session: Session, session_maker: sessionmaker | None = None) -> None:
```
QuantumGhost: This API design feels awkward and unintuitive.

While I understand the evolutionary reasons that led to this current implementation, I believe we should explore a cleaner approach in the future. 

However, it's perfectly fine to leave it as-is in this PR since it's outside the scope of the current changes.

api/tasks/remove_app_and_related_data_task.py
```diff
@@ -175,30 +177,32 @@ def del_workflow(workflow_id: str):
...
-    def del_workflow_run(workflow_run_id: str):
-        db.session.query(WorkflowRun).filter(WorkflowRun.id == workflow_run_id).delete(synchronize_session=False)
-
-    _delete_records(
-        """select id from workflow_runs where tenant_id=:tenant_id and app_id=:app_id limit 1000""",
-        {"tenant_id": tenant_id, "app_id": app_id},
-        del_workflow_run,
-        "workflow run",
+    """Delete all workflow runs for an app using the service repository."""
+    session_maker = sessionmaker(bind=db.engine, expire_on_commit=False)
```
QuantumGhost: The `expire_on_commit=False` argument appears to be unnecessary here since no ORM models have been loaded into the session.

QuantumGhost (dismissed):
LGTM

QuantumGhost:
By the way, now that PostgreSQL is available in the CI pipeline, you may write integration tests for the repository implementations to ensure they behave as expected.

api/core/memory/token_buffer_memory.py
laipz8200: I believe that utilizing the entire message object in this context will enhance the clarity of the code. We can focus on optimizing performance in a future PR.

api/core/repositories/factory.py
laipz8200: > (I have already made this change in my fork and can push it to the source branch if we agree on the change.)

SGTM.

api/core/repositories/factory.py
laipz8200: It's not a concern to identify the original error since we utilize `raise from` in the conversation.

QuantumGhost (approved):
LGTM

luosenxinxi:
Internal Server Error

Issue #13419, opened 2025-02-08
Think tags have destroyed the original returned content of the API, inserting HTML tags.


HuChundong:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

0.15.3

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Think tags should only be related to web display and should not include inserted HTML when calling the API. Furthermore, this change has disrupted the return format of the LLM node, preventing users from obtaining the original return values. If a user wants to display the thinking process, they should use another node or add a setting option to the LLM node.

HuChundong:
@dosu plz turn to humen ^_^

crazywoola:
reasoning model unified think tag is <think></think> (#13392)

You can checkout this pull request, we have reverted the changes in backend, instead we choose to render it in the frontend so it will fix your problem.

Issue #18556, opened 2025-04-22
UI ISSUE：show “[ENDTHINKFLAG]” , unable to distinguish between "think" and "response"


huo131313:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:)
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.2.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

![Image](https://github.com/user-attachments/assets/32c3b483-e62f-4bfb-980d-2ddcb7006857)

wubucheng:
same problem,how to solve?

voyager0003:
same problem when using vllm pulgin

### my environment:
**server:**
dify version 1.3.1 (docker compose) on a server whose OS is ubuntu 22.04


**client:**
chrome 135.0.7049.115

![Image](https://github.com/user-attachments/assets/a800b3fa-4f13-4c09-b177-c8713c7e6b6d)

HZJprince:
If you are deploying using VLLM, you can add the parameter '--enable-reasoning --reasoning-parser deepseek_r1' to avoid the occurrence of [ENDTHINKFLAG].

![Image](https://github.com/user-attachments/assets/b29c688e-d0ab-45f6-b72a-867d23020ced)

voyager0003:
thx @HZJprince , I find the official doc of vllm: https://docs.vllm.ai/en/stable/features/reasoning_outputs.html

wukai123123123:
This happens occasionally when I use DeepSeek-R1-Distill-Qwen-32B model, is it the same problem? But it works fine on v1.1.3

<img width="1300" alt="Image" src="https://github.com/user-attachments/assets/e11f9963-8f11-47f2-8413-a51f63abbe71" />

cutiechi:
> This happens occasionally when I use DeepSeek-R1-Distill-Qwen-32B model, is it the same problem? But it works fine on v1.1.3
> 
[...]

Issue #13107, opened 2025-01-29
Add Support for Think Tags in Dify


blogdefotsec:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing? Tell me about your story.

Dify currently ignores and removes any "think" tags from its output, which prevents users like you from showcasing their thought processes effectively. We would greatly appreciate it if your guys could add support for these tags. (This function might belongs to models in DeepSeek family,  and LM Studio already support this)
Be like:
---
User: Hello.
Agent:
<think>
The user says hello to me. I should reply a hello.
</think>
Hello. How can I help you today?
---
I believe that users would appreciate the transparency of AI's thought process, as it can enhance their patience by showing that the system is working through complex tasks step-by-step. Additionally, seeing a structured and detailed reasoning behind AI decisions might increase user trust and encourage them to continue using the platform.
;) Keep on keeping on

### 3. Can you help us with this feature?

- [ ] I am interested in contributing to this feature.

crazywoola:
feat: add agent thinking content display of deepseek R1 (#12949)

PR #21897, opened 2025-07-03
feat: enhance LLMNode with selective <think> tag and reasoning block removal

### LINKED ISSUE #21896: Support optional removal of <think> tags and  reasoning blocks via LLM_NODE_THINKING_TAGS_ENABLED
### Self Checks

I'm building a chat-flow where multiple LLMs are chained together.
One model (e.g. Qwen-3.0) outputs its reasoning inside
&lt;think&gt;reasoning process...&lt;/think&gt;
blocks, while another model (e.g. Gemini-2.0) does not.

When the flow continues to the next node (code-executor, tool parser, etc.),
the raw &lt;think&gt;&lt;/think&gt; tags and reasoning blocks cause parsing errors and unpredictable behavior.
Maintaining ad-hoc regex cleansers after every LLM node is fragile and error-prone.

I propose adding a single environment variable called LLM_NODE_THINKING_TAGS_ENABLED.
If you leave it at its default value (true), Dify keeps behaving exactly as it does today and the <think> blocks remain visible.
If you set the variable to false, the LLMNode quietly strips out every &lt;think&gt;&lt;/think&gt; reasoning segment before the text reaches the next node, so mixed-model flows parse cleanly.

It addresses both sides of the long-running discussion: people who need the reasoning can keep it, and anyone who just wants a clean answer can switch it off with no code changes.

Fixes #13419  
Related: #13107, #18556

ultramancode:
> [!IMPORTANT]
>

## Summary

Fixes #21896
Related: #13107, #13419, #18556  

reasoning-enabled models (e.g. **Qwen-2.5**) emit `<think> … </think>` blocks,  
while others (e.g. **Gemini-1.5**) do not.  
When these tags reach downstream nodes (code-executor, JSON parser, etc.) they break
parsing and produce runtime errors.

This PR introduces an environment flag:

| Variable | Default | Behavior |
|----------|---------|-----------|
| `LLM_NODE_THINKING_TAGS_ENABLED` | `true` | `true` → keep tags and reasoning blocks(current behavior) <br>`false` → strip tags reasoning blocks inside `LLMNode` |

### Deployment notes
* `.env.sample` now includes **LLM_NODE_THINKING_TAGS_ENABLED=true**  
* `docker-compose.yaml` exposes the variable:
  ```yaml
  LLM_NODE_THINKING_TAGS_ENABLED: ${LLM_NODE_THINKING_TAGS_ENABLED:-true}
  ``` 

This PR adds a safe fallback mechanism for mixed-model flows.

While it's occasionally possible to manually suppress <think> output via prompt tricks in ASSISTANT mode, such behavior is not reliable across all models or input patterns. 

That causes issues when strict JSON or code output parsing is involved.
The new env‑flag ensures the `<think>` block is stripped entirely at the LLMNode level.

> For example, with ASSISTANT Mode, the output can sometimes include reasoning like:  
> "Okay, the user is asking, ... The capital of South Korea is **Seoul**."
> 
> In contrast, with this PR’s behavior, the output becomes:  
> "The capital of South Korea is **Seoul**."

**Note**: While Qwen3 supports prompt-based toggles like `/no_think`, they **sometimes leave behind empty `<think></think>` tags or retain full reasoning text**, and their behavior is not always reliable. This means they don’t guarantee clean output.

스크린샷 2025-07-11 오전 1 11 21

In contrast, setting `LLM_NODE_THINKING_TAGS_ENABLED=false` **unconditionally removes** the entire `<think>…</think>` block—including reasoning content—**at the code level**, ensuring that downstream components always receive a **clean, parser-safe final answer**.

*Only the `text` field returned by **LLMNode** changes when the flag is `false`;  
the collapsible “thinking” panel itself remains exactly as before.*

## Screenshots

| Before (`true`, think tag visible) | After (`false`, think tag removed) |
|--------|-------|
| before | after|

## Checklist


ultramancode:
If you have any feedback or suggestions, I’m happy to adjust.   @crazywoola

XiaoCC:
企业微信截图_17515969026370
You can try adding one in the llm node.
image
ASSISTANT   Content input
image
You can turn off the thinking of the thought model.

ultramancode:
@XiaoCC, thanks again for taking the time to review!

I experimented with the approach of injecting content via “ASSISTANT Mode” as you suggested.

While ASSISTANT Mode visually hides the <think> block, the reasoning is still included in the backend output—  
which breaks strict JSON parsing and executor steps.

Even if you combine ASSISTANT Mode with prompt-level tricks (e.g. `/no_think`), such methods remain unreliable across different models or input patterns—  
which causes issues when downstream systems require strict JSON or code output parsing.

The new env‑flag ensures the <think> block is stripped entirely at the LLMNode level.

This new environment flag in this PR takes things one step further: when `LLM_NODE_THINKING_TAGS_ENABLED=false`, `LLMNode` completely strips out the `<think>…</think>` reasoning block before emitting the text. If set to `true` (the default), behavior remains unchanged so existing users who rely on the visible reasoning panel aren’t affected.

> For example, with ASSISTANT Mode,  the output can sometimes include reasoning like:
> "Okay, the user is asking, ... The capital of South Korea is **Seoul**."
> 
> In contrast, with this PR’s behavior, the output becomes:  
> "The capital of South Korea is **Seoul**."

Also, just to avoid any misunderstandings, I’ve updated the PR title and description to make the goal even clearer.

Thanks again for your time and thoughtful feedback—it’s hugely appreciated!

ultramancode:
@QuantumGhost 
Hope you're doing well!
If you're not too busy, would you mind taking a look at this PR when you get a chance?
Happy to make any changes if needed!

QuantumGhost:
@ultramancode Thanks for your contribution!

I'm reviewing this PR now. Will share my feedback soon.

QuantumGhost:
## A More Thorough Approach

I have examined the responses generated by Gemini and DeepSeek, and it appears that these models naturally separate reasoning content from the actual output. The Dify plugin SDK wraps the reasoning content with `<think>` tags in the 
[`_wrap_thinking_by_reasoning_content`](https://github.com/langgenius/dify-plugin-sdks/blob/21e77bd043bc5e1c893dc80de779e1b46ae81a7a/python/dify_plugin/interfaces/model/large_language_model.py#L523) function.

In my opinion, a more robust approach would be to separate reasoning content from output processing entirely by transmitting them as different fields when returning from the plugin daemon. For example, we could add a field named `reasoning_content` to the [`LLMResult`](https://github.com/langgenius/dify-plugin-sdks/blob/main/python/dify_plugin/entities/model/llm.py#L126) and [`LLMResultChunk`](https://github.com/langgenius/dify-plugin-sdks/blob/main/python/dify_plugin/entities/model/llm.py#L84) models. This would allow Dify to process reasoning content without manually parsing the output generated by the plugin daemon.

However, this would be a breaking change because:

1. The frontend code currently displays reasoning by matching the `<think>` tag. Migrating to a different approach would require frontend code changes.
2. More importantly, existing workflows may be parsing the `<think>` tag and extracting reasoning content for further processing. This change would break those workflows.

Given these constraints, it would require time to deprecate the `<think>` tag format and migrate to the new approach.

I'm uncertain whether other LLM models return `<think>` tags for their reasoning content. It appears that Qwen models do so. I believe different formats should be handled by the model plugin itself, not by the Dify application. This would simplify the logic inside Dify and provide better separation of concerns.

If we implement the changes described above, we should also add the `reasoning_content` field to the outputs of LLM nodes to facilitate the use of reasoning content in workflows.

I have discussed this with @laipz8200, and we both agree that this approach should be preferred.

## Review of the Current Implementation

Regarding the current implementation, the removal of `<think>` tags should be configured at the node level, not as a process-wide configuration. Multiple workflows may be running in the same process, and only some of them may want to remove the `<think>` tags.

Note that adding configuration options to the LLM node requires ensuring compatibility, which introduces additional maintenance burden.

## Conclusion

Based on the analysis above, we recommend not merging this PR in its current form. While the PR addresses a legitimate issue with `<think>` tag handling in mixed-model workflows, the proposed solution introduces process-wide configuration that lacks the granularity needed for complex deployment scenarios. We believe the architectural approach outlined above—separating reasoning content at the plugin level and transmitting it as distinct fields—would provide a more robust and maintainable long-term solution.

ultramancode:
@QuantumGhost @laipz8200

Thank you so much for the incredibly thoughtful and thorough feedback — I truly appreciate the time you both took to dive deep into the architectural concerns and help guide the direction forward.

Based on your suggestions, I've outlined a revised plan that preserves existing workflows while aligning with your recommendation to shift reasoning content out of `<think>` tags and into a structured field, with node-level granularity.

---

### Proposed node parameter: `reasoning_format`

| value  | behavior |
|--------|----------|
| `legacy` | Maintain the current `<think>`-tag format (reasoning + answer in `text`) |
| `field`  | Plugin/SDK sends a separate `reasoning_content`; `text` holds only the final answer |
| `auto` *(default)* | If `reasoning_content` exists, treat as `field`; otherwise, fallback to `legacy` |

---

### Scope of changes

1. **dify-plugin-sdks** – Modify `_wrap_thinking_by_reasoning_content` to pass through `reasoning_content` directly, instead of wrapping with `<think>` tags  
2. **dify (backend)** – Add `reasoning_format` parameter to LLM nodes; update parsing logic and frontend variable exposure (`{{#node_id.reasoning_content#}}`)
3. **dify (frontend)** –  update reasoning display logic to use `reasoning_content` instead of `<think>` tags
4. **dify-official-plugins** – Remove custom `<think>` tag generation; rely on SDK's `reasoning_content` field

---

Does this direction align with your vision?  
I'm absolutely open to adjustments or feedback if there's a better or simpler path.

Thank you again for your time and thoughtful guidance — it's been incredibly helpful.

QuantumGhost:
I truly appreciate your willingness to work on this improvement and the thoughtful approach you've outlined.

The approach you mentioned in https://github.com/langgenius/dify/pull/21897#issuecomment-3070341287 indeed aligns with our vision. However, there are several important considerations we need to address:

- As I mentioned earlier, I'm uncertain whether other LLMs may generate `<think>` tags to wrap reasoning content as part of their response. I attempted to test Qwen3 with the OpenAI client, but the reasoning content doesn't appear to be returned. This likely requires further investigation to understand the behavior across different models.

- The chatflow application type requires support for streaming LLM output to the frontend, which means:

  1. The SSE wire protocol used to send updates to the browser needs to be updated to properly ensure the streaming feature works as expected for reasoning output.
  2. Both [`LLMResult`](https://github.com/langgenius/dify-plugin-sdks/blob/main/python/dify_plugin/entities/model/llm.py#L126) and [`LLMResultChunk`](https://github.com/langgenius/dify-plugin-sdks/blob/main/python/dify_plugin/entities/model/llm.py#L84) need to include fields for reasoning content generated by LLMs.

cc @laipz8200 - what are your thoughts on this plan?

joebnb:
will some make this work in dify, the think block are surplus

Issue #18812, opened 2025-04-25
Agent node and LLM node cannot read multi-modal content at a correct turn


fdb02983rhy:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:)
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.3.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

![Image](https://github.com/user-attachments/assets/5b6b1ea5-569c-42c8-ab68-17a8f59e37a9)

![Image](https://github.com/user-attachments/assets/de518f78-6045-448d-a88e-3921640c3cc8)

AAlexDing:
It seems that the Dify team has not tested the multimodal logic of the agent node. Although the agent node has been launched, it remains in an unusable state.

fdb02983rhy:
@laipz8200 Hi, is this feature on the roadmap by any change?

fdb02983rhy:
> Hi, [@fdb02983rhy](https://github.com/fdb02983rhy). I'm [Dosu](https://dosu.dev), and I'm helping the Dify team manage their backlog and am marking this issue as stale.
> 
> **Issue Summary:**
> 
> * You reported that in Dify v1.3.0 self-hosted via Docker, the agent node cannot read multimodal content.
> * The expected and actual behavior details were not fully provided.
> * I suggested checking model settings, file types, error logs, and known issues.
> * A contributor mentioned that the multimodal logic for the agent node has not been tested and is currently unusable.
> * You asked if support for this feature is planned for future development.
> 
> **Next Steps:**
> 
> * Please let me know if this issue is still relevant with the latest version of Dify by commenting here.
> * If I do not hear back within 15 days, this issue will be automatically closed.
> 
> Thank you for your understanding and contribution!

Yes, and the issue is Agent node and LLM node cannot read multi-modal content at a correct turn, Agent Node Cannot Read MultiModal Content is just one of the symptoms.

fdb02983rhy:
[...]

Issue #12952, opened 2025-01-22
Chatflow LLM node unwantedly include sys.files in LLM requests


fdb02983rhy:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

0.15.1

### Cloud or Self Hosted

Cloud, Self Hosted (Docker)

### Steps to reproduce

[File Analysis.yml.zip](https://github.com/user-attachments/files/18503326/File.Analysis.yml.zip)

https://github.com/user-attachments/assets/b65b5540-375e-4e9e-88d6-ec73630cff32

crazywoola:
Duplicated, https://github.com/langgenius/dify/pull/7955 check this out this is the reason why it contains the pdf file you have uploaded.

fdb02983rhy:
> Duplicated, [#7955](https://github.com/langgenius/dify/pull/7955) check this out this is the reason why it contains the pdf file you have uploaded.

Please notice that it does nothing to the inputs field before users enter the chat interface. 
I was talking about the files that user input in the message input field. The error caused by the LLM node unwantedly attached the sys.files in previous conversation, which is a docx file that Claude does not supports so it raised error.

Wu-Jiayang:
I don't think this is related to #7955; it seems like a different bug.

fdb02983rhy:
@crazywoola Just confirm the same issue remain even https://github.com/langgenius/dify/pull/7955 is merged

fdb02983rhy:
> Hi, [@fdb02983rhy](https://github.com/fdb02983rhy). I'm [Dosu](https://dosu.dev), and I'm helping the Dify team manage their backlog. I'm marking this issue as stale.
> 
> **Issue Summary:**
> 
> * The issue involves the Chatflow LLM node in Dify 0.15.1 unintentionally including `sys.files` in LLM requests.
> * This affects both cloud and self-hosted versions of Dify.
[...]

PR #21938, opened 2025-07-04
Fix: Agent node and LLM node cannot read multi-modal content at a correct turn

fdb02983rhy:
## Summary
Use a table to track which node uses which file, so that history messages can be constructed with the corresponding files. This allows developers to build agents that let users freely send files via sys.files.

Fixes 
- https://github.com/langgenius/dify-official-plugins/issues/1311
- https://github.com/langgenius/dify/issues/18812
- https://github.com/langgenius/dify/issues/12952

## Screenshots

https://github.com/user-attachments/assets/9c2c7d92-2757-41fc-9da5-b7d97795f8b9

https://github.com/user-attachments/assets/45789c75-8dd4-4f4a-a088-fc48399f2b09

## Checklist


fdb02983rhy:
cc @laipz8200 @Nov1c444

api/core/memory/token_buffer_memory.py
```diff
@@ -66,8 +70,43 @@ def get_history_prompt_messages(
...
+        #   • allowed_node_id == ""       → strip ALL attachments
+        #   • allowed_node_id == <node id> → keep only files this node used
+
+        allowed_upload_ids: set[str] | None = None
+
+        if allowed_node_id == "":
+            allowed_upload_ids = set()  # keep none → strip all attachments
+        elif allowed_node_id is None:
+            # Shared memory: allow files used by any node
+            usage_rows = (
```
Nov1c444, line 85: Is this table really necessary? I think we might be able to avoid creating this new table by using join queries instead.

fdb02983rhy: Currently, there's no way to track the relationship between nodes and files.
I believe creating a new table is the simplest and cleanest solution.
Alternatives like adding node_id to message_files or message_id to workflow_node_executions would works as well, but they introduce denormalization or tighter coupling.
Also, since this table would be queried frequently—especially whenever a conversation includes sys.files—it's better to avoid relying on joins every time.

Please feel free to correct me if I'm wrong.

laipz8200:
I’d like to know what will happen if I remove a node from the graph. Will the corresponding row remain in the database permanently?

I think what you actually need is a node-level memory, and we are currently designing that feature.

In my opinion, this pull request isn’t quite a mature design yet.

fdb02983rhy:
> I’d like to know what will happen if I remove a node from the graph. Will the corresponding row remain in the database permanently?
> 
> I think what you actually need is a node-level memory, and we are currently designing that feature.
> 
> In my opinion, this pull request isn’t quite a mature design yet.

Thanks for clarifying—deleting a node only updates the graph JSON, so the related rows stay in the DB. 
Glad to hear node-level memory is on the way. I’ll close this PR, and I hope file handling for Agent nodes can be included so they’re fully functional.

PR #22621, opened 2025-07-18
feat(api): Implement workflow suspension

QuantumGhost:
## Summary

This PR adds a model to save workflow state during execution suspension, serving as the foundation for the Human Input feature.

Changes

- Introduces a new model `WorkflowSuspension` to save workflow state during execution suspension
- Adds a `suspension_id` field to the `WorkflowExecutionModel` to record the current suspension state of the execution
- Adds corresponding database migration

Related issue: #21455 

## Screenshots

N/A

## Checklist


laipz8200:
How about saving this state data in `workflow_runs`?

QuantumGhost:
The design rationale behind using a separate `workflow_suspensions` table is:

**Multiple Suspension Support**: A single workflow execution can be suspended multiple times throughout its lifecycle (e.g., requiring approvals from different stakeholders at various stages). Storing suspension state directly in `workflow_runs` would overwrite previous suspension records, losing valuable historical data about the workflow's suspension history.

**Data Integrity and Schema Design**: The suspension feature requires at least three tightly coupled fields:

- State version (for consistency tracking)
- Serialized state data (for resumption)  
- Suspension timestamp (for audit trails)

These fields form a cohesive unit with a clear invariant: they must all be `NULL` (when not suspended) or all be non-`NULL` (when suspended). Managing multiple correlated columns directly in `workflow_runs` introduces complexity in maintaining this invariant and makes the schema less intuitive.

**Separation of Concerns**: The `workflow_runs` table serves as the primary execution record, while `workflow_suspensions` handles the specialized concern of state persistence during interruptions. This separation follows the single responsibility principle and makes the codebase more maintainable.

api/core/app/entities/queue_entities.py
```diff
@@ -326,6 +327,13 @@ class QueueWorkflowStartedEvent(AppQueueEvent):
     graph_runtime_state: GraphRuntimeState
 
 
+class QueueWorkflowSuspendedEvent(AppQueueEvent):
+    event: QueueEvent = QueueEvent.WORKFLOW_SUSPENDED
+    # next_node_id records the next node to execute after resuming
+    # workflow.
+    next_node_id: str
```
laipz8200, line 334: There may be multiple nodes in the future. Should we consider recording the ID of the previous (or current) node?

api/core/workflow/graph_engine/_engine_utils.py
```diff
@@ -0,0 +1,15 @@
+import time
+
+
+def get_timestamp() -> float:
```
laipz8200:
```suggestion
def get_current_timestamp() -> float:
```

api/core/workflow/graph_engine/entities/graph_runtime_state.py
```diff
@@ -1,18 +1,32 @@
 from typing import Any
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, PrivateAttr
 
 from core.model_runtime.entities.llm_entities import LLMUsage
 from core.workflow.entities.variable_pool import VariablePool
+from core.workflow.graph_engine._engine_utils import get_timestamp
```
IthacaDream, line 7: It looks like a dependency module function `get_timestamp` is missing here. @QuantumGhost

IthacaDream: It should be `get_current_timestamp` in moduel `_engine_utils`

Issue #19910, opened 2025-05-19
psycopg2.errors.DuplicateTable: relation "task_id_sequence" already exists


omegazeng:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:)
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.4.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

# install
```bash
helm repo add douban https://douban.github.io/charts/
helm repo update douban
helm upgrade --install dify douban/dify -f value.yaml -n dify --version 0.6.2

kubectl exec -it -n dify dify-api-64ccd4f5bf-gml5q -- bash
flask db upgrade
```

# init db failed
```bash
flask db upgrade
```

## log
```
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade  -> 64b051264f32, init
Traceback (most recent call last):
  File "/app/api/.venv/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1967, in _exec_single_context
    self.dialect.do_execute(
  File "/app/api/.venv/lib/python3.12/site-packages/sqlalchemy/engine/default.py", line 941, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.DuplicateTable: relation "task_id_sequence" already exists


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/app/api/.venv/bin/flask", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/app/api/.venv/lib/python3.12/site-packages/flask/cli.py", line 1129, in main
    cli.main()
  File "/app/api/.venv/lib/python3.12/site-packages/click/core.py", line 1363, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
[...]

PR #22750, opened 2025-07-22
feat(k8s): Add pure migration option for `api` component

BorisPolonsky:
> [!IMPORTANT]
>

## Summary
This PR ease maintenance in k8s scenario by adding pure migration mode that execute data migration logic and exit normally without launching the web server. This is useful when upgrading `dify` version, as it enable the maintainer to launch a separate container that execute the `flask upgrade-db` logic via `entrypoint.sh` instead of tampering with the actual `api` or `worker`

Without this mode, this logic will be executed in both `api` and `worker` component instead of a standalone, single replica logic. This will case dead lock in case there are multiple replica of `api` or `worker`, or if the `worker` starts before `api` and tries to execute the migration logic that end up with issues like https://github.com/langgenius/dify/issues/19910. With this migration only mode, it's possible to disable auto migration for `api` and `worker` and bring up a standalone container that does the migration beforehand (via a `Job` defined with pre-installation hook annotation that recognized by `Helm`).

## Screenshots

| Before | After |
|--------|-------|
| Server will launch after `flask upgrade-db` | Run `flask upgrade-db` only without launching server when `MODE=migration`   |

## Checklist

### DIFF
--- a/api/docker/entrypoint.sh
+++ b/api/docker/entrypoint.sh
@@ -5,12 +5,11 @@ set -e
 if [[ "${MIGRATION_ENABLED}" == "true" ]]; then
   echo "Running migrations"
   flask upgrade-db
-fi
-
-# Pure migration mode
-if [[ "${MODE}" == "migration" ]]; then
+  # Pure migration mode
+  if [[ "${MODE}" == "migration" ]]; then
   echo "Migration completed, exiting normally"
   exit 0
+  fi
 fi
 
 if [[ "${MODE}" == "worker" ]]; then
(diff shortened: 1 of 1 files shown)


BorisPolonsky:
As alternative solution, we may also change the `else` statement (at line 29) below and enforce explicit `MODE=api` condition (i.e. `elif [[ "${MODE}" == "api" ]]; then`) for the same purpose.
https://github.com/langgenius/dify/blob/e9c9c5d8f1129301f808552230ad6253af44310b/api/docker/entrypoint.sh#L29-L41

crazywoola:
cc @laipz8200

laipz8200 (approved):
Although we've added a lock at the https://github.com/langgenius/dify/blob/a4ef9009167ed7cf22b0b8bd3a903e8ba0a74567/api/commands.py#L677 to manage migration issues across multiple instances, I believe it would be beneficial to include an additional solution as well.

PR #23160, opened 2025-07-30
fix: improve boolean field handling in plugin configuration forms

AlanBustamanteRemerge:
Summary

This PR fixes an Internal Server Error that occurs when configuring endpoints in the Slack bot extension. The error was caused by improper handling of boolean fields in plugin configuration forms, where boolean values were being sent to the backend as strings or numbers instead of proper boolean types.

Problem

When users try to configure the Slack bot endpoint, they encounter an "Internal Server Error" message because:

Form fields with boolean type are being submitted as strings ("true", "false") or numbers (1, 0) instead of actual boolean values
The backend validation rejects these malformed boolean data types, causing the server to return a 500 error
This prevents users from successfully configuring Slack bot endpoints

Solution

The fix implements proper boolean field handling in two key areas:

Form Schema Processing (web/app/components/tools/utils/to-form-schema.ts):

Added boolean field conversion logic in addDefaultValue function
Converts string values ('true', '1', 'True') to proper boolean true
Converts number values (1) to boolean true, others to false

Preserves existing boolean values

Endpoint Configuration (web/app/components/plugins/plugin-detail-panel/endpoint-modal.tsx):
Added boolean field processing in handleSave function before calling onSaved
Creates a processed copy of form data with correct boolean types
Ensures all boolean fields are properly typed before sending to backend

Impact

✅ Fixes: Internal Server Error when configuring Slack bot endpoints
✅ Improves: Plugin configuration form reliability and user experience
✅ Maintains: Backward compatibility with existing functionality
✅ Enables: Users to successfully configure Slack bot endpoints without errors

Testing

Tested with Slack bot endpoint configuration forms
Verified boolean fields are properly converted from string/number to boolean
Confirmed no more Internal Server Error when saving endpoint configuration
Validated that existing functionality remains intact

### DIFF
--- a/web/app/components/plugins/plugin-detail-panel/endpoint-modal.tsx
+++ b/web/app/components/plugins/plugin-detail-panel/endpoint-modal.tsx
@@ -47,19 +47,18 @@ const EndpointModal: FC<Props> = ({
         return
       }
     }
-    
+
     // Fix: Process boolean fields to ensure they are sent as proper boolean values
     const processedCredential = { ...tempCredential }
     formSchemas.forEach((field) => {
       if (field.type === 'boolean' && processedCredential[field.name] !== undefined) {
         const value = processedCredential[field.name]
-        if (typeof value === 'string') {
+        if (typeof value === 'string')
           processedCredential[field.name] = value === 'true' || value === '1' || value === 'True'
-        } else if (typeof value === 'number') {
+         else if (typeof value === 'number')
           processedCredential[field.name] = value === 1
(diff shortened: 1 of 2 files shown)


crazywoola (commented):
These code seems identical to me, please consider extracting them into a util function.

crazywoola:
Thanks for the PR, please link an issue in the description as well :)

crazywoola (dismissed):

crazywoola (approved):

PR #23629, opened 2025-08-08
Fix: Unable to reload page effectively when token expires

joeyhacker:


crazywoola:
Please do not modify the PR template, also you need to describe why you need this changes, you need to link an existing issue or create a new one in the description.

PR #23712, opened 2025-08-11
Fix/plugin upload hardening

### LINKED ISSUE #23716: Arbitrary file type upload allowed via /files/upload/for-plugin (client mimetype trusted)
### Self Checks

The plugin upload endpoint accepts any file type because it trusts the client-supplied mimetype and filename. This allows uploading files with non-allowlisted or dangerous types (e.g., .php, .jsp, scripts) by spoofing the mimetype and/or suffix. Downstream consumers could be exposed to RCE, stored XSS, or data exfiltration if these files are rendered/executed/viewed elsewhere.

Affected area
api/controllers/files/upload.py – class PluginUploadFileApi.post

Impact / Risk
	•	Upload of server-side script files or binaries under benign extensions.
	•	Potential execution/rendering by other services or users (depending on deployment), leading to RCE/XSS.
	•	At minimum, policy violation for accepted file types.

Root cause
Server trusts request.files["file"].mimetype and the filename extension. There is no server-side allowlist nor content sniffing before persisting.

Steps to reproduce (high level)
	1.	Obtain a valid signed upload URL (existing flow).
	2.	POST a multipart form with a dangerous file (e.g., evil.jpg.php) while declaring type=image/png and filename=evil.png.
	3.	Upload succeeds; the file is stored and can be accessed via preview URL.

Expected
Uploads restricted to a small allowlist; content must match the declared/expected type.

CedricHwong:
Fixes #23716
> [!IMPORTANT]
>

## Summary

Harden /files/upload/for-plugin without breaking existing clients.
	•	Add server-side validation module core/file/safe_upload.py
	•	Suffix allowlist + dangerous suffix blocking
	•	Limit excessive compound suffixes
	•	Lightweight header sniffing (no new deps)
	•	Env-configurable: PLUGIN_UPLOAD_ALLOWED_EXTS, PLUGIN_UPLOAD_BLOCKED_SUFFIXES
	•	Keep signature verification unchanged (still uses client-declared mimetype for compatibility)
	•	Persist/return canonical server mimetype (derived from validated suffix), never trust client mimetype
	•	Log declared_mimetype vs server_mimetype mismatch for observability
	•	Minimal controller changes; main flow preserved

Rationale

The endpoint allowed arbitrary file types by trusting client mimetype/filename. This adds strict, configurable server-side checks with near-zero perf cost and no extra dependencies.

Compatibility / Risk
	•	Signature path unchanged → existing plugins work
	•	Allowlist is adjustable via env; rollout can be tuned per environment
	•	No schema or data migrations

Configuration

Add to your environment (or docker-compose, k8s, systemd, etc.):

```yaml
PLUGIN_UPLOAD_ALLOWED_EXTS=.png,.jpg,.jpeg,.gif,.pdf,.txt,.csv,.json
PLUGIN_UPLOAD_BLOCKED_SUFFIXES=.php,.jsp,.exe,.sh,.bat,.js,.html,.htm
```

Offline verification (no service needed)
The script below loads safe_upload.py by file path (bypasses app imports), generates sample files, and checks them with the validator.

Set MOD_PATH to the real path of your module (e.g. api/core/file/safe_upload.py).
```shell
# === path to safe_upload.py ===
MOD_PATH="api/core/file/safe_upload.py"

# --- common checker (no app deps) ---
check() {
  FILE="$1" MOD_PATH="$MOD_PATH" python3 - <<'PY'
import os, importlib.util, os.path as p
MOD_PATH = os.environ["MOD_PATH"]
spec = importlib.util.spec_from_file_location("safe_upload", MOD_PATH)
m = importlib.util.module_from_spec(spec); spec.loader.exec_module(m)
path = os.environ["FILE"]
name = m.normalize_filename(p.basename(path))
_, suff = m.split_suffixes(name)
ok_suffix = m.is_safe_suffixes(suff)
ext = suff[-1] if suff else ""
with open(path, "rb") as f:
    ok_sniff = m.sniff_ok(f, ext) if ok_suffix else False
print(f"{path} -> suffixes={suff} ok_suffix={ok_suffix} ok_sniff={ok_sniff} mime={m.canonical_mimetype(ext)} RESULT={'OK' if ok_suffix and ok_sniff else 'BLOCK'}")
PY
}

# ===== Samples that SHOULD PASS =====
printf '\xFF\xD8\x00\x00\x00\x00\x00\x00' > /tmp/good.jpg            # JPEG
printf 'GIF89a\x00\x00\x00\x00\x00\x00\x00' > /tmp/good.gif          # GIF89a
printf '%%PDF-1.7\nrest' > /tmp/good.pdf                             # PDF
printf 'hello,world\n' > /tmp/good.txt                               # UTF-8 text
printf 'a,b\n1,2\n' > /tmp/good.csv                                  # CSV
printf '{"x":1}\n' > /tmp/good.json                                  # JSON
check /tmp/good.jpg
check /tmp/good.gif
check /tmp/good.pdf
check /tmp/good.txt
check /tmp/good.csv
check /tmp/good.json

# upper-case suffix & JSON with BOM (still OK)
printf '\xFF\xD8\x00\x00\x00\x00\x00\x00' > /tmp/upper.JPEG
python3 - <<'PY'
open('/tmp/json_bom.json','wb').write(b'\xEF\xBB\xBF' + b'{"a":1}\n')
PY
check /tmp/upper.JPEG
check /tmp/json_bom.json

# ===== Samples that SHOULD BLOCK =====
# Fake PNG/PDF (wrong magic)
printf '<?php echo 1; ?>' > /tmp/evil.png
printf 'not-a-pdf' > /tmp/evil.pdf
# Dangerous double suffix
printf '<?php echo 1; ?>' > /tmp/evil.jpg.php
# Excessive compound suffixes
printf 'dummy' > /tmp/many && cp /tmp/many /tmp/a.report.backup.json.zip && rm /tmp/many
# Non-allowlisted
printf '{"x":1}\n' > /tmp/data.jsonl
# GIF with invalid header
printf 'GIF88a\x00\x00\x00\x00\x00' > /tmp/fake_gif.gif
# No suffix
touch /tmp/noext
# Text file that is not valid UTF-8 (fails text sniff)
printf '\xFF\xFE\x00\x00BINARY' > /tmp/binary.txt
# .tar.gz (not in allowlist)
tmpd=$(mktemp -d); printf 'abc' > "$tmpd/a"; tar -czf /tmp/a.tar.gz -C "$tmpd" a; rm -rf "$tmpd"

check /tmp/evil.png
check /tmp/evil.pdf
check /tmp/evil.jpg.php
check /tmp/a.report.backup.json.zip
check /tmp/data.jsonl
check /tmp/fake_gif.gif
check /tmp/noext
check /tmp/binary.txt
check /tmp/a.tar.gz

# Optional: allow .jsonl temporarily via env for this single run
PLUGIN_UPLOAD_ALLOWED_EXTS=".png,.jpg,.jpeg,.gif,.pdf,.txt,.csv,.json,.jsonl" \
FILE=/tmp/data.jsonl MOD_PATH="$MOD_PATH" python3 - <<'PY'
import os, importlib.util, os.path as p
spec = importlib.util.spec_from_file_location("safe_upload", os.environ["MOD_PATH"])
m = importlib.util.module_from_spec(spec); spec.loader.exec_module(m)
path = os.environ["FILE"]; name = m.normalize_filename(p.basename(path))
_, suff = m.split_suffixes(name); ok = m.is_safe_suffixes(suff); ext = suff[-1] if suff else ""
with open(path,'rb') as f: sniff = m.sniff_ok(f, ext) if ok else False
print("jsonl allowed via env ->", "OK" if ok and sniff else "BLOCK")
PY
```

Expected: all “GOOD” samples return RESULT=OK; all “BLOCK” samples return RESULT=BLOCK.

close https://github.com/langgenius/dify/issues/23716

## Screenshots

| Before | After |
|--------|-------|
| ...    | ...   |

## Checklist


crazywoola:
BTW, please link an existing issue or create a new one.

kurokobo:
I haven't been able to do a detailed review or thorough testing yet, but I'm concerned about the lack of extensibility, as well as the possibility of breaking existing workflows.

With the current implementation, `sniff_ok` only checks eight specific file extensions, and it seems that any other extensions will return False, even if they're listed in `PLUGIN_UPLOAD_ALLOWED_EXTS`. In addition, there doesn’t seem to be a way to extend `sniff_ok`. Similarly, `EXT_TO_MIME` treats any extension other than these eight as `application/octet-stream` (which also isn’t extensible).

Plugins can generate a wide variety of legitimate files, such as docx, pptx, xlsx, wav, mp3, and zip files. If the checks introduced by this PR are enforced, I’m worried that many existing workflows will stop working, even if the admins customize PLUGIN_UPLOAD_ALLOWED_EXTS .
 
Is my concern valid? What do you think?

crazywoola:
> Plugins can generate a wide variety of legitimate files, such as docx, pptx, xlsx, wav, mp3, and zip files. If the checks introduced by this PR are enforced, I’m worried that many existing workflows will stop working, even if the admins customize PLUGIN_UPLOAD_ALLOWED_EXTS .

I second this.

CedricHwong:
@kurokobo  Your concern is valid. In the current patch, if an admin adds an extension to PLUGIN_UPLOAD_ALLOWED_EXTS but we don’t have a sniffer for it, sniff_ok() would return False and block it. Also, EXT_TO_MIME isn’t extensible.

CedricHwong:
I’ve updated the PR so it’s extensible by default and doesn’t break existing workflows:

What changed
	•	Sniffing now uses the filetype library (header-based detection). We no longer maintain a hard-coded list of 8 extensions or a custom MIME table. filetype covers a wide range (zip/docx/xlsx/pptx/mp3/wav/webp/mp4, etc.) out of the box.
	•	Allowlist gate remains the primary control: admins only need to add an extension to PLUGIN_UPLOAD_ALLOWED_EXTS to enable it. No code changes, no custom sniffers, no MIME maintenance.
	•	“Strict” verification is limited to risky types only: by default we strictly validate just image/PDF types (.png/.jpg/.jpeg/.gif/.pdf). If the detected type doesn’t match the extension, we reject (blocks “evil.php renamed to .png”).
Non-strict types (e.g., .docx/.pptx/.xlsx/.zip/.mp3/.wav) are allowed when allowlisted, and we still normalize server MIME from filetype (fallback to mimetypes only if unknown).
	•	We do not trust client-declared mimetype for persistence (signature verification is unchanged for backward compatibility). We log detected-vs-declared mismatches for auditing.
	•	No schema changes, minimal I/O: we read ~560 bytes for detection; perf impact is negligible.

Why this addresses extensibility & breakage
	•	Extensibility: Any new format becomes usable by simply adding its suffix to the env allowlist; detection uses filetype — no need to extend sniff_ok or maintain MIME maps.
	•	Compatibility: Workflows producing docx/pptx/xlsx/wav/mp3/zip keep working once allowlisted. Only obviously risky masquerades in image/PDF are blocked.
	•	Operations: If you want an even stricter posture later, we can trivially expose a flag to expand the strict set (e.g., via env). Default remains safe and non-breaking.

Housekeeping
	•	Added filetype>=1.2.0 to the API dependencies.
	•	Synced env templates so the config-parity test passes.

kurokobo:
I'm just a contributor, so I can't judge whether this PR should move forward or not, but here are a few comments from my opinion.
I still haven't fully tested this PR at all on my side, so they may not necessarily be correct.

- The `filetype` is specialized for handling binary files, and I don't think it can detect text files. The `txt`, `csv`, and `json`, which are included in `STRICT_EXTS`, might not have their extensions or MIME types detected correctly.
- The extensions that `filetype` supports are limited, so I'm not sure if this is sufficient: https://github.com/h2non/filetype.py?tab=readme-ov-file#supported-types
- If this check is enabled by default, it means that, for Dify SaaS, The Dify Team will need to list all valid file types generated by every plugin, which might be practically difficult. For example, you might want to set a wildcard like `.*` as the default to disable the check, and then only apply strict checking for self-hosted environments with special requirements by specifying environment variable.
- If you implement this setting via environment variables, it's customary to centrally manage them using `dify_config`, so it might be better to avoid using os.getenv directly.
- Some comments are still in Chinese.

PR #24093, opened 2025-08-18
fix  pg_vector extension requires SUPERUSER, but not available on Huawei Cloud RDS

ElvisWai:
https://support.huaweicloud.com/rds-pg_faq/rds_faq_0075.html
Fixes #24094

### DIFF
--- a/api/configs/middleware/vdb/pgvector_config.py
+++ b/api/configs/middleware/vdb/pgvector_config.py
@@ -8,6 +8,7 @@ class PGVectorConfig(BaseSettings):
     """
     Configuration settings for PGVector (PostgreSQL with vector extension)
     """
+
     PGVECTOR_IS_OWNER: Optional[bool] = Field(
         description="Whether the PGVector is the owner of the database",
         default=True,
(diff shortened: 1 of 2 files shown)


crazywoola (approved):

api/configs/middleware/vdb/pgvector_config.py
```diff
@@ -9,6 +9,11 @@ class PGVectorConfig(BaseSettings):
     Configuration settings for PGVector (PostgreSQL with vector extension)
     """
 
+    PGVECTOR_IS_OWNER: Optional[bool] = Field(
+        description="Whether the PGVector is the owner of the database",
+        default=True,
```
crazywoola, line 14:
```suggestion
        default=False,
```

crazywoola:
@ElvisWai Hello thanks for the fix, however it seem the docker compose environments variables are missing, please add those variables in related section and submit a PR again.

api/configs/middleware/vdb/pgvector_config.py
ElvisWai: The default value of True is reasonable, as most cloud platforms support the use of super permissions.

Issue #24067, opened 2025-08-18
[Chore/Refactor] Queue-based Graph Engine Architecture


laipz8200:
## Description

part of #19429

This refactor transforms the workflow graph engine from a monolithic architecture to a queue-based, worker pool architecture with improved modularity and scalability.

The refactoring introduces:
- **Command-based orchestration**: Commands flow through channels for better control and testability
- **Worker pool management**: Dynamic scaling of workers based on workload with configurable thresholds
- **Event-driven architecture**: Clear event routing and handling with separation of graph events from node events
- **Domain-driven design**: Clear separation between domain logic, infrastructure, and orchestration layers
- **Improved error handling**: Multiple recovery strategies (retry, default value, fail branch, abort)
- **Better state management**: Dedicated managers for node states, edge states, and execution tracking
- **Response coordination**: Centralized handling of streaming responses and output management
- **Layer-based processing**: Support for extensible processing layers (e.g., debug logging)

## Motivation

The original monolithic graph engine had several limitations:
- Tight coupling between components made testing and maintenance difficult
- Limited scalability due to single-threaded execution model
- Complex state management spread across multiple components
- Difficult to extend with new features or debugging capabilities
- Hard to reason about the flow of execution and data

This refactoring addresses these issues by:
- Introducing clear boundaries between different concerns (command processing, event management, state management)
- Enabling parallel execution through worker pools with dynamic scaling
- Centralizing state management with dedicated managers
- Providing extension points through layers and strategies
- Making the execution flow more explicit through commands and events

## Additional Context

The refactoring maintains backward compatibility with existing workflows while providing a more robust foundation for future enhancements. Configuration for the worker pool is now managed through environment variables via dify_config, allowing for easy tuning of performance parameters.

Key configuration parameters include:
- `GRAPH_ENGINE_MIN_WORKERS`: Minimum number of workers in the pool
- `GRAPH_ENGINE_MAX_WORKERS`: Maximum number of workers
- `GRAPH_ENGINE_SCALE_UP_THRESHOLD`: Queue size threshold for scaling up
- `GRAPH_ENGINE_SCALE_DOWN_IDLE_TIME`: Idle time before scaling down

kinoooolu:
Will this feature support providing an asynchronous HTTP API for callers?

laipz8200:
No, it won't.

PR #24116, opened 2025-08-18
feat: add execution limit layer

Nov1c444:
## Summary

#24067
add execution limit layer

## Checklist

### DIFF
--- a/api/core/app/apps/workflow_app_runner.py
+++ b/api/core/app/apps/workflow_app_runner.py
@@ -49,6 +49,7 @@ from core.workflow.graph_events import (
     NodeRunStreamChunkEvent,
     NodeRunSucceededEvent,
 )
+from core.workflow.graph_events.graph import GraphRunAbortedEvent
 from core.workflow.nodes import NodeType
 from core.workflow.nodes.node_factory import DifyNodeFactory
 from core.workflow.nodes.node_mapping import NODE_TYPE_CLASSES_MAPPING
@@ -365,6 +366,8 @@ class WorkflowBasedAppRunner:
             )
         elif isinstance(event, GraphRunFailedEvent):
             self._publish_event(QueueWorkflowFailedEvent(error=event.error, exceptions_count=event.exceptions_count))
+        elif isinstance(event, GraphRunAbortedEvent):
+            self._publish_event(QueueWorkflowFailedEvent(error=event.reason or "Unknown error", exceptions_count=0))
         elif isinstance(event, NodeRunRetryEvent):
             node_run_result = event.node_run_result
             inputs = node_run_result.inputs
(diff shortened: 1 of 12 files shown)


api/core/workflow/graph_engine/layers/execution_limits.py
```diff
@@ -0,0 +1,139 @@
...
+    Layer that enforces execution limits for workflows.
+
+    Monitors:
+    - Step count: Tracks number of node executions
+    - Time limit: Monitors total execution time
+
+    Automatically aborts execution when limits are exceeded.
+    """
+
+    def __init__(self, max_steps: PositiveInt, max_time: PositiveInt) -> None:
```
laipz8200: IMO, using pydantic's type here is not a good idea.

api/core/workflow/graph_engine/layers/execution_limits.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        if not self._execution_started or self._execution_ended or self._abort_sent:
+            return
+
+        # Track step count for node execution events
+        if isinstance(event, NodeRunStartedEvent):
+            self._handle_node_execution_event(event)
+
+        # Check limits after each event
+        self._check_limits()
```
laipz8200: For better readability.

```suggestion
        if self._reached_time_limitation():
            self._send_abort_command()
```

api/core/workflow/graph_engine/layers/execution_limits.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Monitors execution progress and enforces limits.
+        """
+        if not self._execution_started or self._execution_ended or self._abort_sent:
+            return
+
+        # Track step count for node execution events
+        if isinstance(event, NodeRunStartedEvent):
+            self._handle_node_execution_event(event)
+
```
laipz8200, line 93:
```suggestion

    if isinstance(event, NodeRunSucceededEvent | NodeRunFailedEvent) and self._if_reached_step_limitation():
        self._sent_abort_command()
        return
```

api/core/workflow/graph_engine/layers/execution_limits.py
```diff
@@ -0,0 +1,139 @@
...
+        # Check limits after each event
+        self._check_limits()
+
+    def on_graph_end(self, error: Optional[Exception]) -> None:
```
laipz8200, line 97: I think we can remove this method.

laipz8200 (changes requested):

api/core/workflow/graph_engine/layers/execution_limits.py
Nov1c444: This is implementing an abstract method from the base class, so we need to keep it

api/core/app/apps/workflow_app_runner.py
```diff
@@ -365,6 +366,8 @@ def _handle_event(self, workflow_entry: WorkflowEntry, event: GraphEngineEvent)
             )
         elif isinstance(event, GraphRunFailedEvent):
             self._publish_event(QueueWorkflowFailedEvent(error=event.error, exceptions_count=event.exceptions_count))
+        elif isinstance(event, GraphRunAbortedEvent):
+            self._publish_event(QueueWorkflowFailedEvent(error=event.reason or "Unknown error", exceptions_count=0))
```
Nov1c444, line 370: This isn't a good implementation, I need to refactor it.

PR #22596, opened 2025-07-17
fix text splitter

jubinsoni:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 2. Ensure there is an associated issue and you have been assigned to it
> 3. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

close https://github.com/langgenius/dify/issues/22046

## Screenshots

| Before | After |
|--------|-------|
| ...    | ...   |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods

PR #24366, opened 2025-08-22
fix chunks get space removed

jubinsoni:
> [!IMPORTANT]
>

## Summary

to fix https://github.com/langgenius/dify/issues/24251

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

below screenshot shows that space is retained when chunk size is selected as 256
Screenshot 2025-08-22 at 10 58 22 PM

tested with chunk size 512 and space is retained for this as well
Screenshot 2025-08-22 at 10 58 42 PM

## Checklist


DavideDelbianco:
You have completely removed the handling of chunk overlap, the pseudocode flow I have proposed in the issue is a trace to how it should work in case it surpasses the chunk length.

This PR should be thoroughly reviewed to avoid damaging the General Embed variant in favor of just the Parent Child (that does not use overlap)

DavideDelbianco:
@jubinsoni
1. You should completely revert your last commit.
That latest version is not good. It removes the overlap mechanism, which is crucial.

2. In the previous version it worked fine, but it was not the right fix, because written that way, it doubles the number of iterations during `_merge_splits`

That would be caused by the fact you appended a SPACE character instead of concatenating it, resulting in:
```python
["word", " ", "another", " ", "more", " "]
```
If done this way, it also adds an extra space at the end.

3. You had rewritten the condition
```python
        _separator = "" if self._keep_separator else separator
```
as:
```python
        # When keep_separator=True and separator is space, preserve it for merging
        if self._keep_separator and separator == " ":
            _separator = ""
        else:
            _separator = "" if self._keep_separator else separator
```

But in spite of the comment stating that you want to "save" the SPACE, the code does the exact same thing as before.

In a previous commit by you, in an attempt to fix #22596
You added the concatenation of the separator, that was gone missing by doing this:
image

Now the question is:
Why the SPACE character is treated differently?
Why not split on the `separator` also for the SPACE?
What's the difference?
image

If the fix is to add a space during that phase, then the best way to fix it would be change
```python
        if separator:
            if separator == " ":
                splits = text.split()
            else:
                splits = text.split(separator)
                splits = [item + separator if i < len(splits) else item for i, item in enumerate(splits)]
        else:
```
into:
```python
        if separator:
                splits = text.split(separator)
                splits = [item + separator if i < len(splits) else item for i, item in enumerate(splits)]
        else:
```
To avoid adding extra elements to the array of splits just to add the SPACE

Going back one more commit, we find that @JohnJyong probably noticed that chunk overlap stopped working by his improvements to the chunking and fixed it this way:
image

I suppose the issue lies with `_separator` and not `separator` but he didn't notice, by testing the algorithm with properly formatted documents and the default 1024/512 check sizes.

Based purely on logic and what I read in this file... why the line that calculates the `_separator` says:
image
I read it as: "If you want to KEEP the separator, then the `_separator` is empty, otherwise it is separator"
Shouldn't it be the other way around?

@JohnJyong is probably best suited to answer this question, being the author of the change and the main developer of the RAG part.

The issue of #24251 lies in the following lines.
The issue happens on EVERY line of a document that exceeds the `chunk_size` when the `fixed_separator` is NOT present in the text.
In that case the fallback separators are used, and since SPACE is always present in the initial call (I haven't found a piece of code that does not pass the separators): 
image

It will always end (at least one time) in the IF branch of the post-split code.
It may, eventually, reach also the overlap code in the ELSE branch (something that happens more often in well-formatted documents that, at some point, have a sentence that, after being split, is longer than `chunk_size` (that's the reason the overlap logic MUST be preserved and your latest commit is completely wrong).

Now, what happens in the IF branch?
When splitting on the SPACE character: 99.99% of the splits are considered shorter than the `chunk_size`, because the min `chuck_size` is forced to 50 chars and the splits are single words:
image

This means that every word will be appended to the `_good_split` Array.
At the end, the `_good_split` will be merged, but the _separator character, that we want to keep was set to "" by:

```python
_separator = "" if self._keep_separator else separator
```
image


I'm no python expert, but I see no value in treating SPACE differently during the split:
```python
        if separator:
            if separator == " ":
                splits = text.split()  # <-- WHY??
            else:
                splits = text.split(separator) # <--- THIS WOULD WORK THE SAME WAY
                splits = [item + separator if i < len(splits) else item for i, item in enumerate(splits)]
        else:
            splits = list(text)
```

----------------

I've tested all the possible fixes while waiting for @JohnJyong review

1. What happens now:
image

2. What happens when the fix is done this way:
```python
        if separator:
            splits = text.split(separator)
            splits = [item + separator if i < len(splits) else item for i, item in enumerate(splits)]
        else:
            splits = list(text)
```
image

3. What happens if the fix is done by removing old @jubinsoni fix and invert the logic on `keep_separator`:
```python
        if separator:
            if separator == " ":
                splits = text.split()
            else:
                splits = text.split(separator)
        else:
            splits = list(text)
```
```python
        _separator = separator if self._keep_separator else ""
```
image

The separator is kept, but seems to be just the last one (SPACE)

4. What happens when mixing the two:
```python
        if separator:
            splits = text.split(separator)
            splits = [item + separator if i < len(splits) else item for i, item in enumerate(splits)]
        else:
            splits = list(text)
```
```python
        _separator = separator if self._keep_separator else ""
```
image


Personally I will go with n° 2,  but I think there's something broken in how the `keep_separator` is handled

@crazywoola do not approve this PR in its current state, you will break RAG :)
convince @JohnJyong to take a look at this, PR / comment because RAG 2.0 is not behind the corner.

jubinsoni:
hi @DavideDelbianco , 
I'll go through the analysis

as stated 
1) I'll revert to previous commit
2) invert the logic on keep_separator 
3) fix  the code post above 2 steps

Had a long day at work ...  just now came back from gym and saw your analysis
Tomorrow I have holiday coz of Ganesh Chaturthi festival
Happy Ganesh Chaturthi to you

I'll try to update pr by tomorrow evening

DavideDelbianco:
I wouldn't invert the logic until @JohnJyong reviews this PR
It will add a second " " when merging doc parts as shown in example fix n° 4

My analysis is long just because I wanted to include explanations and test all the possible fixes, it's not meant to waste your time :)

jubinsoni:
@DavideDelbianco  updated
Screenshot 2025-08-29 at 5 20 12 AM

this will ensure space not added at the end

PR #24310, opened 2025-08-21
try flask_restful -> flask_restx

asukaminato0721:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

part of #24173

## Summary

This is only a test, to try to update something.

this can have auto swagger doc generation. https://flask-restx.readthedocs.io/en/latest/swagger.html

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods

PR #24412, opened 2025-08-24
GPT-5-high fix #24310 bug.

Eric-Guo:
## Summary

Another pure GPT-5-high and cursor auto fix.

ExternalApi.handle_error no longer accesses self.errors directly (Flask-RESTX lacks that attribute). It now checks for an errors dict via getattr before using it.

it’s related to “try flask_restful -> flask_restx (#24310)”. Flask-RESTful Api exposes errors; Flask-RESTX does not. Your error handler assumed it existed and raised AttributeError.

Also fixed Flask-RESTX marshalling crashes by updating custom fields:

* AppIconUrlField.output(self, key, obj, ordered: bool = False)
* AvatarUrlField.output(self, key, obj, ordered: bool = False)

### DIFF
--- a/api/libs/external_api.py
+++ b/api/libs/external_api.py
@@ -78,9 +78,11 @@ class ExternalApi(Api):
         data = getattr(e, "data", default_data)
 
         error_cls_name = type(e).__name__
-        if error_cls_name in self.errors:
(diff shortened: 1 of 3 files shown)


Eric-Guo:
Close to prefer #24416

Issue #23981, opened 2025-08-15
[Umbrella] Dify Trigger — Event-driven Workflow Foundation


Yeuoly:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] Please do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing? Tell me about your story.

1

### 3. Can you help us with this feature?

- [ ] I am interested in contributing to this feature.

lyzno1:
1

CathyL0:
2

ACAne0320:
3

bowenliang123:
4

ZeroZ-lab:
5

hjlarry:
6
please assign Webhook Node to me

Yeuoly:
> 6 please assign Webhook Node to me

Of course, please create a sub-issue like #23995 so that I can link it to issues tracker

lyzno1:
Created sub-issue for Schedule Node frontend: #24001
  @Yeuoly  please link this to the task tracker

CathyL0:
I've created a sub-issue for the WebHook node frontend: https://github.com/langgenius/dify/issues/24020.
Please help add it to the task tracker. Thx! @Yeuoly

bowenliang123:
I would like to join and share some parts of Unified Trigger Entry #23995 or Trigger Plugin Specification. How could I create sub tasks for #23995 ?

wbkys:
Great

ZeroZ-lab:
Created sub-issue for Backend: Single-Step Debug for Trigger [#24163](https://github.com/langgenius/dify/issues/24163)

lyzno1:
Created sub-issue for frontend: #24196 #24197 #24198 #24199 #24200

Yeuoly:
Created sub-issue for Plugin specification #24241

lyzno1:
## Schedule Trigger Frontend Implementation Complete

### Implemented Features

**1. Complete UI Component System**
- Configuration panel supporting both visual and cron expression modes
[...]

PR #24428, opened 2025-08-24
feat: add workflow schedule trigger support

### LINKED ISSUE #24021: [☂️] Dify Trigger — Backend: Schedule System & Node
### Self Checks

1

_No response_

ACAne0320:
> [!IMPORTANT]
>

## Summary

part of #23981
close #24021

This PR implements workflow schedule trigger support for Dify, enabling users to execute workflows automatically based on cron expressions with timezone support.

### Features Implemented

**1. Schedule Trigger Node**

- New trigger node type trigger-schedule for workflow configuration
- Support for standard cron expressions (e.g., 0 9 * * 1-5 for weekdays at 9 AM)
- Passes current time as input to triggered workflows

**2. Custom Celery Beat Scheduler**

- Database-driven dynamic scheduler replacing default Beat scheduler
- High availability support using PostgreSQL's `SELECT FOR UPDATE SKIP LOCKED`
- Configurable tick intervals and batch processing

**3. Automatic Schedule Management**

- Schedules plan automatically created/updated when publishing workflows
- Schedules plan delete when workflow has no schedule trigger node
- One-to-one mapping between app and schedule plan

**4. Performance Optimizations**

- Batch database updates to reduce round trips

### How to Test

**1. Migrate database**
```bash
 uv run flask db upgrade
```

**2. Setup Schedule Trigger in Workflow:**
  - Create a new workflow
  - Add a "Schedule Trigger" node as the start node
  - Configure cron expression
  - Add workflow logic after the trigger
  - Publish the workflow

**3. Start the Scheduler Service:**
```bash
# Start celery worker
dev/start-worker

# Start celery beat
dev/start-beat
```
**3. Verify Execution:**
  - Check workflow run history in the UI

#### Schedule Trigger
- [x] Use [UUIDv7](https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-7) as PK for workflow_schedule_plans table
- [x] Use beat_schedule instead of Custom Scheduler
- [x] Adjust `workflow_schedule_plans` table structure
- [x] Stop dispatching tasks when the daily limit is reached
- [ ] Add a check when exporting DSL files
- [ ] Add a failure log when the daily limit is reached

#### Test
- [x] Unit Tests
- [ ] Integration Tests

## Checklist


dev/start-workflow-scheduler
```diff
@@ -0,0 +1,75 @@
...
+cd "$SCRIPT_DIR/.."
+
+echo "Starting Workflow Scheduler..."
+echo "  Log Level: ${LOGLEVEL}"
+if [[ -n "${PIDFILE}" ]]; then
+  echo "  PID File: ${PIDFILE}"
+fi
+
+# Build celery beat command with custom scheduler
+CMD="uv --directory api run celery -A app.celery beat"
```
Yeuoly: A beat service already exists, see PluginUpgradeService for details

ACAne0320: The `workflow-scheduler` is a dedicated Celery Beat instance that uses a custom scheduler class `WorkflowScheduler` to handle dynamic workflow schedules.
It runs **independently** from the system beat service which handles static periodic tasks.

Yeuoly: Yes, but I intent to add this schedule task to ext_celery.py, beat_schedule handles this.

api/schedule/workflow_schedule_task.py
```diff
@@ -0,0 +1,117 @@
...
+    dispatched = 0
+
+    for schedule in schedules:
+        next_run_at = ScheduleService.calculate_next_run_at(
+            schedule.cron_expression,
+            schedule.timezone,
+        )
+
+        if next_run_at:
+            schedule.next_run_at = next_run_at
```
vincentx11: What if we take this operation after the execution of the scheduled async task?

Consider the following scenario for reference:
1. The scheduled poller runs per 1min.
2. The scheduled trigger of a certain workflow is set to every 5min.
3. Due to some non-declarative reasons, the workflow task takes 15 minutes to complete within a certain period.

There will be two different outcomes depending on where this code is placed:
1. If we reset next_run_at before the task execution, the task will run at least 3 times within 15 minutes, with 1 run completing successfully and the other 2 remaining in a running state.
2. If we reset next_run_at after the task execution, the task will only run once within 15 minutes.

In combination with the business scenario, I recommend adopting the second outcome. By the way, the above case isn't the worst-case scenario. The workflow task might also hang for an extended period. If this happens, the more times the task runs, the higher load of the service will be, which could eventually lead to the whole service unhealthy.

dev/start-beat
```diff
@@ -0,0 +1,60 @@
...
+  echo ""
+  echo "Options:"
+  echo "  --loglevel LEVEL       Log level (default: INFO)"
+  echo "  --scheduler SCHEDULER  Scheduler class (default: celery.beat:PersistentScheduler)"
+  echo "  -h, --help             Show this help message"
+  echo ""
+  echo "Examples:"
+  echo "  $0"
+  echo "  $0 --loglevel DEBUG"
+  echo "  $0 --scheduler django_celery_beat.schedulers:DatabaseScheduler"
```
Yeuoly, line 17: what is django_celery

api/schedule/workflow_schedule_task.py
```diff
@@ -0,0 +1,117 @@
...
+            processed = _process_schedules(session, due_schedules)
+            logger.info(
+                "Processed %d/%d schedules (%d tenants rate-limited)",
+                processed,
+                len(due_schedules),
+                len(rate_limited_tenants),
+            )
+
+
+def _get_rate_limited_tenants(session: Session) -> set[str]:
```
Yeuoly: It may lead a huge array, I intent to do the check before here, while InvokeDailyRateLimitError was raised  during invoking trigger_workflow_async, the tenant can be marked as `RATE_LIMITED` today, it's already been implemented by `check_daily_quota`

api/schedule/workflow_schedule_task.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        .where(
+            WorkflowSchedulePlan.next_run_at <= now,
+            WorkflowSchedulePlan.next_run_at.isnot(None),
+            AppTrigger.status == AppTriggerStatus.ENABLED,
+        )
+    )
+
+    if exclude_tenants:
+        query = query.where(WorkflowSchedulePlan.tenant_id.notin_(exclude_tenants))
```
Yeuoly: Here is what problem is, query to WorkflowSchedulePlan was executed multiple times. 

That said, redis may reduce the queries.

api/services/schedule_service.py
```diff
@@ -0,0 +1,313 @@
...
+            next_run_at=next_run_at,
+        )
+
+        session.add(schedule)
+        session.flush()
+
+        return schedule
+
+    @staticmethod
+    def update_schedule(
```
Yeuoly, line 59: It's not recommended to implement updating method using `updates` dict

api/services/schedule_service.py
```diff
@@ -0,0 +1,313 @@
...
+from core.workflow.nodes import NodeType
+from models.account import Account, TenantAccountJoin
+from models.workflow import Workflow, WorkflowSchedulePlan
+
+logger = logging.getLogger(__name__)
+
+
+class ScheduleService:
+    @staticmethod
+    def calculate_next_run_at(
```
Yeuoly: What about make it a util, as it has no reference to `ScheduleService` and it's a helper function.

api/services/schedule_service.py
```diff
@@ -0,0 +1,313 @@
...
+
+    @staticmethod
+    def update_schedule(
+        session: Session,
+        schedule_id: str,
+        updates: dict,
+    ) -> Optional[WorkflowSchedulePlan]:
+        schedule = session.get(WorkflowSchedulePlan, schedule_id)
+        if not schedule:
+            return None
```
Yeuoly: We should raise exception here I think, return None is confused to contributors, what's more, it means any calling to this method should check nullable result like error handing Golang
```
if _, err := foo(); err != nil {
//TODO
}
```

api/services/schedule_service.py
```diff
@@ -0,0 +1,313 @@
...
+        return schedule
+
+    @staticmethod
+    def delete_schedule(
+        session: Session,
+        schedule_id: str,
+    ) -> bool:
+        schedule = session.get(WorkflowSchedulePlan, schedule_id)
+        if not schedule:
+            return False
```
Yeuoly: Also, for non-exists WorkflowSchedulePlan, we should raise exception.

api/events/event_handlers/sync_workflow_schedule_when_app_published.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+
+        if not schedule_config:
+            if existing_plan:
+                logger.info("No schedule node in workflow for app %s, removing schedule plan", app_id)
+                session.delete(existing_plan)
+                session.commit()
+            return None
+
+        node_id = schedule_config["node_id"]
```
Yeuoly: Is it possible to create a pydantic model to parse dict?

api/services/schedule_service.py
```diff
@@ -0,0 +1,313 @@
...
+                    )
+
+                if cron_expression:
+                    return {
+                        "node_id": node_id,
+                        "cron_expression": cron_expression,
+                        "timezone": timezone,
+                    }
+                else:
+                    logger.warning("Invalid schedule configuration in node %s", node_id)
```
Yeuoly: We should not log the message here, raising exception instead.

api/services/schedule_service.py
```diff
@@ -0,0 +1,313 @@
...
+            schedule.cron_expression,
+            schedule.timezone,
+        )
+
+        schedule.next_run_at = next_run_at
+        session.flush()
+        return next_run_at
+
+    @staticmethod
+    def extract_schedule_config(workflow: Workflow) -> Optional[dict]:
```
Yeuoly: A little bit hard to understand the response of `extract_schedule_config`, as it's a nullable dict, I'm confused if I'm a newbee to Dify schedule

api/services/schedule_service.py
```diff
@@ -0,0 +1,313 @@
...
+                return f"{minute} {hour} {','.join(sorted_days)} * *"
+
+            else:
+                return None
+
+        except Exception:
+            return None
+
+    @staticmethod
+    def parse_time(time_str: str) -> tuple[Optional[int], Optional[int]]:
```
Yeuoly: Also, nullable response is not a good idea.

Yeuoly (changes requested):

dev/start-beat
ACAne0320: This script doesn't use django_celery.
This example follows the official Celery documentation's instructions for custom scheduler classes:
https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html#using-custom-scheduler-classes

api/schedule/workflow_schedule_task.py
ACAne0320: Oh, I see! You mean we should add the rate-limited marking directly in the `check_daily_quota` method itself?

like this:

```python
# api/services/workflow/queue_dispatcher.py

def check_daily_quota(self, tenant_id: str) -> bool:
    remaining = self.rate_limiter.get_remaining_quota(
        tenant_id=tenant_id,
        max_daily_limit=self.get_daily_limit()
    )

    if remaining <= 0:
        self.rate_limiter.mark_tenant_rate_limited(tenant_id)
        return False

    return True
```

```python
# api/services/workflow/rate_limiter.py

class TenantDailyRateLimiter:

    def mark_tenant_rate_limited(self, tenant_id: str) -> None:
        date_str = datetime.utcnow().strftime("%Y-%m-%d")
        key = f"workflow:tenants:rate_limited:{date_str}"
        self.redis.sadd(key, tenant_id)
        self.redis.expire(key, self._get_ttl_seconds())
```

If this is correct, I'd be happy to implement it.

Yeuoly: `check_daily_quota` already returns if the tenant hit the limit, I guess it's no need to mark again.
like this:
```python
# api/services/workflow/xxx_service.py
if not self.dispatcher.check_daily_quota(tenant_id):
    raise
```

api/schedule/workflow_schedule_task.py
ACAne0320: Thank you for raising this question.

Under the current architecture, regardless of where next_run_at is updated, the task will execute 3 times within 15 minutes.
The reason for this is that the execution of `trigger_workflow_async()` is non-blocking.

Our scheduling system is intentionally designed to be fully decoupled:

1. Scheduler: Responsible for triggering tasks based on cron expressions
2. Executor: Handles workflow execution asyn via [Unified Trigger Entry](https://github.com/langgenius/dify/issues/23995)
3. Non-blocking: `trigger_workflow_async()` immediately returns after successfully entering the queue without waiting for completion

While your concern about multiple executions is valid from a scheduling perspective, our architecture has a critical safety mechanism that prevents the system overload you're worried about: the workflow execution queue.

When `trigger_workflow_async()` is called, it doesn't immediately execute the workflow. Instead:

1. Creates a trigger log entry with status QUEUED
2. Dispatches to tier-specific queues (professional/team/sandbox)
3. Returns immediately without waiting

The actual workflow execution happens in dedicated worker pools with concurrency limits per workspace.

like this:
Schedule Trigger → Unified Trigger Entry → Workflow Queue → [Concurrency Control] → Actual Execution

api/schedule/workflow_schedule_task.py
```diff
@@ -0,0 +1,98 @@
...
+            .join(
+                AppTrigger,
+                and_(
+                    AppTrigger.app_id == WorkflowSchedulePlan.app_id,
+                    AppTrigger.node_id == WorkflowSchedulePlan.node_id,
+                    AppTrigger.trigger_type == "trigger-schedule",
+                ),
+            )
+            .where(
+                WorkflowSchedulePlan.next_run_at <= now,
```
Yeuoly, line 83: Is there a possible for the schedules to be triggered twice?
1. _fetch_due_schedules was invoked but `next_run_at` was not updated, since _process_schedules may take some times to process all the plans, this would lead delays.
2. when the next cycle occurs, if schedule plan already been fetched before but not been updated will been triggered again.

api/services/schedule_service.py
```diff
@@ -0,0 +1,278 @@
...
+                frequency = node_data.get("frequency")
+                visual_config_dict = node_data.get("visual_config", {})
+                visual_config = VisualConfig(**visual_config_dict)
+                cron_expression = ScheduleService.visual_to_cron(frequency, visual_config)
+            else:
+                raise ScheduleConfigError(f"Invalid schedule mode: {mode}")
+
+            return ScheduleConfig(node_id=node_id, cron_expression=cron_expression, timezone=timezone)
+
+        return None
```
Yeuoly: I believe this unnecessary return of None could be minimized.

api/tasks/workflow_schedule_tasks.py
```diff
@@ -0,0 +1,69 @@
...
+from models.enums import WorkflowRunTriggeredFrom
+from models.workflow import WorkflowSchedulePlan
+from services.async_workflow_service import AsyncWorkflowService
+from services.schedule_service import ScheduleService
+from services.workflow.entities import TriggerData
+
+logger = logging.getLogger(__name__)
+
+
+@shared_task(queue="schedule")
```
Yeuoly: It may be beneficial for `run_schedule_trigger` and `poll_workflow_schedules` to operate on separate queues to prevent them from blocking one another.

api/schedule/workflow_schedule_task.py
```diff
@@ -0,0 +1,98 @@
...
+                WorkflowSchedulePlan.next_run_at.isnot(None),
+                AppTrigger.status == AppTriggerStatus.ENABLED,
+            )
+        )
+        .with_for_update(skip_locked=True)
+        .limit(dify_config.WORKFLOW_SCHEDULE_POLLER_BATCH_SIZE)
```
Yeuoly, line 90: This approach may lead to unbounded queuing, for example, in this turn the limit size is 100, but there are actually 1000 items to process, by the next turn, the queued items may grow to 1800, maybe a parallel and streaming process is what we expected.

api/schedule/workflow_schedule_task.py
vincentx11: That sounds interesting. With this solution, whether tasks of a specified workflow can run in parallel (with multiple instances over a period) is controlled by the workflow itself—specifically through the "Parallel Setting" in the trigger, which wasn't mentioned in this feature scenario.

This is acceptable for now. However, looking ahead, I strongly recommend adding a `Parallel Setting` to the trigger node with three modes:

1. `Parallel` – this is the current solution
2. `Serial Waiting` – only one task instance exists per workflow; new tasks wait if another is already running
3. `Serial Rejection` – new tasks get rejected if one is already running, ensuring only one instance at a time

Following the design approach you described, if we considered this solution, this check could be implemented before `trigger_workflow_async` within `run_schedule_trigger`.

api/tasks/workflow_schedule_tasks.py
```diff
@@ -0,0 +1,69 @@
...
+
+        tenant_owner = ScheduleService.get_tenant_owner(session, schedule.tenant_id)
+        if not tenant_owner:
+            raise TenantOwnerNotFoundError(f"No owner or admin found for tenant {schedule.tenant_id}")
+
+        try:
+            current_utc = datetime.now(UTC)
+            schedule_tz = ZoneInfo(schedule.timezone) if schedule.timezone else UTC
+            current_in_tz = current_utc.astimezone(schedule_tz)
+            inputs = {"current_time": current_in_tz.isoformat()}
```
vincentx11, line 51: Additionally, I have a question regarding `current_time`—the only parameter that the trigger node passes to the workflow. This seems more like a discussion at the requirement level rather than a technical communication. If this is not the appropriate place to discuss this, please point me to the right channel.

I’m curious about how we handle the input variables defined in the original `Start` node. Given that the trigger does not add these user inputs to the `variable_pool`, should the subsequent nodes that depend on these inputs for execution throw an error, or should they not depend on these input variables in the first place?

Furthermore, if the bussiness scenario must set default values for the variables defined in the "Start" node during scheduled tasks, how should I implement this under the current requirement design? Are there any practices or solutions I can refer to?

Yeuoly: Schedule features an entry-type node, rather than a time-based call to `Start node`, additionally, each entry-type branch can only have a single entry node, these kind of nodes we now called it `start` kind, which means the original `Start node` was now renamed to `User Input node`.

The variables you defined in `User Input` will never occur in the same entry branch, so NVMD about it.

api/schedule/workflow_schedule_task.py
ACAne0320: I'm glad you were able to provide this solution.
Maybe we need to discuss with the Dify team whether we should add this feature to the schedule trigger.
I'll get back to you if there's any result.

api/schedule/workflow_schedule_task.py
ACAne0320: Under the current implementation, scheduling is not normally double-triggered.
1. The scheduled plan is obtained via a `SELECT ... FOR UPDATE SKIP LOCKED` statement, preventing the same row from being reselected during processing. 
2. Within the same transaction, we calculate and set `next_run_at`, schedule the task, and then commit. After commit, the plan no longer expires and will not be reselected.

Yeuoly: Great!

Yeuoly (approved):
LGTM

api/tasks/workflow_schedule_tasks.py
vincentx11: Thank you for your explaination, and sorry for the late reply. 

This is acceptable from the perspective of decoupled design. However, what if I want to reuse existing workflows that depend on the variables from `User Input node`? Specifically, how can I enable the `auto-schedule trigger feature` for those existing workflows created prior to this version? What is the recommendation solution about this scenario?

ACAne0320: I think you might be able to achieve this by adding some `Environment Variables`.
image

vincentx11: Yes, `Env Variables` works. But this means the variables cannot be changed when there were created. And, it doesn't coopreate with `User Input` node well to acheive the `dynamic input variables' value` when executing by user and the` default value` by trigger.

ACAne0320: Simply add a `Code` node after the `User Input` and `Schedule Trigger`, then properly initialize the variable values you need, like this:

image

When you enter `input_arg1` via `User Input`
- `arg1=input_arg1`

When you leave `User Input` blank
- `arg1=env_arg1`

How does that sound?

vincentx11: Big awards! :Thumbup

This seems to be a complete integration solution that integrates all nodes to work collaboratively, though it incorporates an additional node maintenance task. However, this is actually the best solution we have so far, in my opinion. 

Thank you very much for your professional solution and the patience you've devoted.

PR #24440, opened 2025-08-25
api: support to config disabling Swagger UI in api service

### LINKED ISSUE #24439: support config switch of Swagger UI in api service
### Self Checks

- support config the switch and the url path of Swagger UI by flask-restx   in api service
- 

_No response_

bowenliang123:
> [!IMPORTANT]
>

## Summary

- to close #24439
- This PR helps to disable the swagger UI page, and the `/swagger.json` API entry as the Swagger data source. (known in https://github.com/python-restx/flask-restx/issues/116)

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


laipz8200:
I think we can simply use DEBUG to determine whether the API documentation is enabled.

bowenliang123:
> I think we can simply use DEBUG to determine whether the API documentation is enabled.

DEBUG config may not be suitable in this case, as it also enforces more detailed logging in all cases. In practice, a dedicated config for enabling Swagger UI helps Dify to be deployed in UAT deployment without debugging enabled but with Swagger UI enabled for downstream system integration development usage.

crazywoola (dismissed):

bowenliang123:
Do you think we should also unified the URL path for swagger UI page into `/swagger-ui.html` or `/swagger` in this PR? @laipz8200

crazywoola:
> Do you think we should also unified the URL path for swagger UI page into `/swagger-ui.html` or `/swagger` in this PR? @laipz8200

+1

bowenliang123:
Maybe let's do the path unificiation next time. With the same configs code, it works on `http://127.0.0.1:5001/files/swagger-ui.html` but not the `http://127.0.0.1:5001/mcp/swagger-ui.html`


image


image

laipz8200:
> Maybe let's do the path unificiation next time. With the same configs code, it works on `http://127.0.0.1:5001/files/swagger-ui.html` but not the `http://127.0.0.1:5001/mcp/swagger-ui.html`
> image image

Also update ext_login.py

laipz8200:
If we want to allow using Swagger in production environments, we might need to configure authentication for it. My goal is simply to let developers access the API documentation. Of course, if you'd like to make it more comprehensive, go ahead and do so.

bowenliang123:
Agree that authentication should be a key enhancement for Swagger UI in future.
The original intention is to disable Swagger UI and its API. And I have some obstacles in using the [`add_specs` config] (https://flask-restx.readthedocs.io/en/latest/_modules/flask_restx/api.html) to disable the `/swagger.json` API, as we don't have a timing to run `Api.init_app(flask_app)` to apply this config.

bowenliang123:
> Also update ext_login.py

Thanks for the hint. It has been fixed.

bowenliang123:
It's ready for review now.

With the manual separate calls on construction and init_app to ensure configs in kwargs are effective, we can now disable the `/swagger.json` API as well.

api/libs/external_api.py
```diff
@@ -106,6 +107,14 @@ def handle_general_exception(e: Exception):
...
-        super().__init__(*args, **kwargs)
+    def __init__(self, bp, **kwargs):
+        kwargs["add_specs"] = dify_config.SWAGGER_UI_ENABLED
+        if not dify_config.SWAGGER_UI_ENABLED:
+            kwargs["doc"] = False
+        else:
+            kwargs["doc"] = dify_config.SWAGGER_UI_PATH
+
+        # manual separate call on construction and init_app to ensure configs in kwargs effective
+        super().__init__(**kwargs)
```
laipz8200: I don't think ignoring args is a good approach, if we use positional parameters anywhere, it might not work

laipz8200 (changes requested):
Please ensure all newly added environment variables are included in the .env.example file.

api/libs/external_api.py
bowenliang123: Agree. Got this fixed.

laipz8200 (changes requested):
Just make this change and see the system go down:

```
﻿﻿-api = ExternalApi(bp)
+api = ExternalApi(bp=bp)
```

Why not calling `super().__init__(*args, **kwargs)` directly?

bowenliang123:
> Just make this change and see the system go down:
> 
> ```
> ﻿﻿-api = ExternalApi(bp)
> +api = ExternalApi(bp=bp)
> ```
> 
> Why not calling `super().__init__(*args, **kwargs)` directly?

Our goal is to disable the `/swagger.json` API by using the `add_specs` config.
The only way to apply the config is to be used in `.init_app()` as :
```
api = restx.Api()
api.init_app(app, add_specs=False)
```
( refer to  https://github.com/python-restx/flask-restx/blob/ffb079c696e7901b0b27526810ff32c52beb8aa0/tests/test_api.py#L100 )


However, when using the restx.Api() constructor directly with args, all the configs in kwargs are missed and ignored in calling the `init_app` method. Therefore, the `add_specs` config is not working as expected.
image
(Refer to the Github's repo on 1.3.0 tag in https://github.com/python-restx/flask-restx/blob/1.3.0/flask_restx/api.py )

So, it has to be split into constuct the restx.Api() first without the Flask blueprint or app, and then call the `init_app` with configs mannually.

Furthermore, as the `*args` is positional oriented, if we use bp as well in calling the constructor, it will again falls into the `init_app` calling inside the __init__ constructor automatically. 

The current code is by design and should work as expected.

api/libs/external_api.py
```diff
@@ -107,5 +108,17 @@ def handle_general_exception(e: Exception):
 
 class ExternalApi(Api):
     def __init__(self, *args, **kwargs):
```
laipz8200: Let's change this line to 

```python
def __init__(self, app: Blueprint | Flask, *args, **kwargs)
```

laipz8200 (changes requested):

api/libs/external_api.py
bowenliang123: Sure. That's my previous available version for assigning the app instance. Adapted.

laipz8200 (approved):
LGTM

PR #24583, opened 2025-08-26
Refactor: replace count() > 0 check with exists()

hyongtao-code:
Fixes #24582

This PR refactors queries that check for the existence of a record from using `count() > 0` to `exists()`.
Considering cases where queries on `Message` use `conversation_id`, which is not a unique index, and queries on `AppDatasetJoin` rely solely on `dataset_id`, which cannot form a composite index, using `EXISTS` provides better performance than `COUNT()` since it avoids scanning the entire result set.

### Changes

- Replaced `db.session.query(...).count() > 0` with `db.session.scalar(select(exists().where(...)))` in multiple places.
- Updated to SQLAlchemy 2.0 style `select + session.scalar` instead of the legacy `query` API.

### Advantages of using exists()
- **Performance:** `count()` scans all matching rows to return the total count, which is unnecessary when we only need to know if any record exists.
- **Efficiency:** `exists()` generates a `SELECT EXISTS (SELECT 1 FROM ...)` query. The database can short-circuit as soon as it finds the first matching row.
- **Clarity:** The intent (check if record exists) is explicit and more readable than comparing a count with zero.
- **Modern API:** Aligns code with SQLAlchemy 2.0 best practices (select / scalar).

> [!IMPORTANT]
>

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/controllers/console/app/message.py
+++ b/api/controllers/console/app/message.py
@@ -32,6 +32,7 @@ from services.annotation_service import AppAnnotationService
 from services.errors.conversation import ConversationNotExistsError
 from services.errors.message import MessageNotExistsError, SuggestedQuestionsAfterAnswerDisabledError
 from services.message_service import MessageService
+from sqlalchemy import exists, select
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


asukaminato0721:
```
models/workflow.py:339: error: Incompatible return value type (got "bool | None", expected "bool")  [return-value]
services/dataset_service.py:658: error: Incompatible return value type (got "bool | None", expected "bool")  [return-value]
services/workflow_service.py:90: error: Incompatible return value type (got "bool | None", expected "bool")  [return-value]
```

hjlarry (approved):
LGTM

crazywoola (approved):

PR #24696, opened 2025-08-28
fix(datasource): load default fileExtensions for file node earlier

kurokobo:
## Summary

This is for the ✨ **feat/rag-2** branch.

Closes #25304

### Issues to be solved

With specific steps, the `fileExtensions` option will be null and the `Browse` link will disappear. See the `Before` in `Screenshots` section.

- Create a new pipeline
- Place the File datasource node
- **IMPORTANT: From now on, do not open the panel for this node**
- Connect and configure the extractor, chunker, and vdb
- Ensure there are no validation errors
- Invoke `Test Run`

You will notice that there is no `Browse` link, and if you export the DSL, there is no `fileExtensions` property. You can still try uploading files by drag and drop, but it won't be accepted.

In the current implementation, the default `fileExtensions` are loaded **when the panel for this node is opened**.
This means that if the user never opens the panel until they publish the pipeline, `fileExtensions` has no value and the pipeline cannot accept any file extensions.
This is a possible scenario that can actually occur when the users want to leave all of the file node settings at their default values.

### Changes in this PR

This PR makes the default `fileExtensions` value load **when placing the node**, instead of when opening the panel.
I'm not sure if this is the best solution, but I wanted to let you know about the issue and suggest an idea to solve it.

Of course it's okay to close this if there is an appropriate alternative solution.

F.Y.I. @crazywoola 

## Screenshots

| Before | After |
|--------|-------|
| Image Image | Image Image |

## Checklist

### DIFF
--- a/web/app/components/workflow/block-selector/data-sources.tsx
+++ b/web/app/components/workflow/block-selector/data-sources.tsx
@@ -10,13 +10,14 @@ import type {
   OnSelectBlock,
   ToolWithProvider,
 } from '../types'
-import type { ToolDefaultValue } from './types'
+import type { DataSourceDefaultValue, ToolDefaultValue } from './types'
 import Tools from './tools'
 import { ViewType } from './view-type-select'
 import cn from '@/utils/classnames'
 import type { ListRef } from '@/app/components/workflow/block-selector/market-place-plugin/list'
 import { getMarketplaceUrl } from '@/utils/var'
 import { useGlobalPublicStore } from '@/context/global-public-context'
+import { DEFAULT_FILE_EXTENSIONS_IN_LOCAL_FILE_DATA_SOURCE } from './constants'
 
 type AllToolsProps = {
   className?: string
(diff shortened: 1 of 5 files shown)


kurokobo:
Conflict resolved.

web/app/components/workflow/block-selector/data-sources.tsx
```diff
@@ -37,14 +38,22 @@ const DataSources = ({
...
       plugin_id: toolDefaultValue?.provider_id,
       provider_type: toolDefaultValue?.provider_type,
       provider_name: toolDefaultValue?.provider_name,
       datasource_name: toolDefaultValue?.tool_name,
       datasource_label: toolDefaultValue?.tool_label,
       title: toolDefaultValue?.title,
-    })
+    }
+    // Update defaultValue with fileExtensions if this is the local file data source
+    if (toolDefaultValue?.provider_id === 'langgenius/file' && toolDefaultValue?.provider_name === 'file') {
```
WTW0313, line 50: Using `provider_type === 'local_file'` would be more precise here.

kurokobo:
@WTW0313 
This PR was created before the beta release, but I've confirmed that the issue still occurs with 2.0.0-beta.1, so I created an Issue https://github.com/langgenius/dify/issues/25304 and rebased my branch.
As I mentioned in my previous comment, I'm not sure if this is the best way to implement it, so if there's a better way to fix it, feel free to close this PR.

Also, please let me know if the PR for fixes to beta should not be sent to `feat/rag-2` branch.

WTW0313 (approved):

Issue #24876, opened 2025-09-01
[Chore/Refactor] Enum to StrEnum / IntEnum


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

StrEnum is designed for str.

### Motivation

https://sandeepnarayankv.medium.com/python-3-11s-game-changing-strenum-and-intenum-say-goodbye-to-value-forever-778bcf5b8034

### Additional Context

we can also use auto() in them

example #24877

Kritoooo:
I'd like to work on it

laipz8200:
Using `auto()` will use uppercase for the value. I think this is a breaking change for enums that will be serialized and passed as a response.

cc @crazywoola

Kritoooo:
The string generated by auto() seems to be a lowercase version of the variable name? I agree that simply replacing everything with auto() isn't appropriate when there's no clear correspondence. In some cases, this causes tests to fail, and the relevant configurations and tests need to be updated accordingly.

laipz8200:
> The string generated by auto() seems to be a lowercase version of the variable name? I agree that simply replacing everything with auto() isn't appropriate when there's no clear correspondence. In some cases, this causes tests to fail, and the relevant configurations and tests need to be updated accordingly.

You're right. I tested it and got a lowercase result. But the referenced article has this example:(

```python
from enum import StrEnum, auto

class ApiEndpoint(StrEnum):
    USERS = auto()      # Automatically becomes "USERS"
    PRODUCTS = auto()   # Automatically becomes "PRODUCTS"
    ORDERS = auto()     # Automatically becomes "ORDERS"
    
# String operations work directly
endpoint = ApiEndpoint.USERS
print(f"Calling /{endpoint.lower()}")  # Calling /users
# No .value needed for comparisons!
[...]

PR #25129, opened 2025-09-04
chore: adopt StrEnum and auto() for some string-typed enums

Kritoooo:
> [!IMPORTANT]
>

## Summary
part of #24876 

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


api/models/model.py
```diff
@@ -62,7 +62,7 @@ def value_of(cls, value: str) -> "AppMode":
         raise ValueError(f"invalid mode value {value}")
 
 
-class IconType(Enum):
+class IconType(StrEnum):
     IMAGE = "image"
```
asukaminato0721: here :)

Kritoooo: Thanks for the heads-up! The latest commit contain it.

api/core/app/entities/task_entities.py
```diff
@@ -51,7 +51,7 @@ class WorkflowTaskState(TaskState):
     answer: str = ""
 
 
-class StreamEvent(Enum):
+class StreamEvent(StrEnum):
```
asukaminato0721, line 53: here

api/core/entities/embedding_type.py
```diff
@@ -1,7 +1,7 @@
-from enum import Enum
+from enum import StrEnum
 
 
-class EmbeddingInputType(Enum):
+class EmbeddingInputType(StrEnum):
     """
     Enum for embedding input type.
     """
```
asukaminato0721, line 7: below

api/core/helper/model_provider_cache.py
```diff
@@ -1,20 +1,20 @@
...
+from enum import StrEnum
 from json import JSONDecodeError
 from typing import Optional
 
 from extensions.ext_redis import redis_client
 
 
-class ProviderCredentialsCacheType(Enum):
+class ProviderCredentialsCacheType(StrEnum):
     PROVIDER = "provider"
```
asukaminato0721, line 10: here

api/core/entities/embedding_type.py
Kritoooo: Thank you for the careful check! Besides the ones you pointed out, I’ve also switched PromptType and RetrieveStrategy in this PR to auto(). Feel free to flag anything I may have missed.

api/core/helper/model_provider_cache.py
Kritoooo: This one can’t be switched to auto(), because the auto-generated values would differ from the explicitly assigned strings, and those exact strings are referenced elsewhere in the codebase—right?

asukaminato0721: I guess if they are only in the codebase, not "outside", like in the database, config files.

We can replace them.

> those exact strings

Actually, awful early design, I am keeping fixing them.

asukaminato0721 (dismissed):

Kritoooo:
@crazywoola Could you review this PR? Thanks a lot.

api/core/helper/tool_parameter_cache.py
```diff
@@ -1,12 +1,12 @@
...
-from enum import Enum
+from enum import StrEnum
 from json import JSONDecodeError
 from typing import Optional
 
 from extensions.ext_redis import redis_client
 
 
-class ToolParameterCacheType(Enum):
+class ToolParameterCacheType(StrEnum):
```
asukaminato0721, line 9: here

asukaminato0721 (dismissed):

Kritoooo:
Hi, my original plan was to batch-replace all StrEnum members that map one-to-one with the strings auto() generates, then deal with the non-matching cases separately.
I’ve been swamped these two days, so I’ve set the PR to draft; this doesn’t mean I’ve dropped it—if anyone has bandwidth to finish it sooner, that’s perfectly fine.
If the current part can be merged first, I’ll open a follow-up PR when I’m free, and I’ll keep an eye on any progress so no duplicate work happens.

Kritoooo:
It looks like changes in some test files are causing the previous modifications to fail the tests; I'll investigate and fix the issue within this week.

crazywoola (approved):

PR #25270, opened 2025-09-05
chore(docker): set default FILES_URL value for Docker deployment

### LINKED ISSUE #25269: [Chore/Refactor] Set default FILES_URL value for Docker deployment
## Description
Set a sensible default value for FILES_URL environment variable in Docker deployment configuration to improve out-of-the-box experience.

## Motivation
The FILES_URL environment variable is required for file processing functionality in Dify. Currently, it's left empty in the .env.example file, requiring users to manually configure it. Setting a default value of `http://api:5001` provides a working configuration for Docker deployments, reducing setup friction for new users while still allowing customization when needed.

## Additional Context
This change affects the Docker deployment configuration files:
- Updates .env.example with default FILES_URL value
- Regenerates docker-compose.yaml to reflect the new default

laipz8200:
## Summary
- Set default FILES_URL value to `http://api:5001` in Docker configuration
- Regenerate docker-compose.yaml with updated defaults

## Motivation
Improves the out-of-the-box experience for Docker deployments by providing a sensible default value for the FILES_URL environment variable, which is required for file processing functionality.

Fixes #25269


kurokobo:
I'm not really trying to argue strongly against it, though... `FILES_URL` is also used as part of the URL that allows users to download files, but usually, `api:5001` can't be resolved from a user's browser, and I think that port is closed as well.

So, wouldn't it be better to set `api:5001` as the default for `INTERNAL_FILES_URL` instead? I think this is used when exchanging files between containers. If there are any spots where file transfers between containers fail because `INTERNAL_FILES_URL` is not being used, shouldn't we fix the internal implementation?

For example:

- `FILES_URL`: `http://localhost` or `http://host.docker.internal`
- `INTERNAL_FILES_URL`: `http://api:5001`

laipz8200:
> I'm not really trying to argue strongly against it, though... `FILES_URL` is also used as part of the URL that allows users to download files, but usually, `api:5001` can't be resolved from a user's browser, and I think that port is closed as well.
> 
> So, wouldn't it be better to set `api:5001` as the default for `INTERNAL_FILES_URL` instead? I think this is used when exchanging files between containers. If there are any spots where file transfers between containers fail because `INTERNAL_FILES_URL` is not being used, shouldn't we fix the internal implementation?
> 
> For example:
> 
>     * `FILES_URL`: `http://localhost` or `http://host.docker.internal`
> 
>     * `INTERNAL_FILES_URL`: `http://api:5001`

You are right, I need to re-examine the communication during the tolerance period and confirm they are using `INTERNAL_FILES_URL`. Thank you for the reminder.

PR #25416, opened 2025-09-09
refactor: centralize default end user session ID constant

Yeuoly:
> [!IMPORTANT]
>

## Summary

This PR refactors the handling of the default end user session ID by centralizing it as an enum in the models module where the `EndUser` model is defined. This improves code organization and makes the relationship between the constant and the model clearer.

### Changes Made:
- **Moved constant to models module**: Relocated `DEFAULT_SERVICE_API_USER_ID` from `core/file/constants.py` to `models/model.py` as `DefaultEndUserSessionID` enum
- **Updated all references**: Changed all imports and usages to use the new enum value
- **Fixed plugin file upload**: Corrected user_id handling in `controllers/files/upload.py` to use actual `user.id` instead of raw user_id parameter
- **Simplified file helpers**: Removed unnecessary None checks for user_id in file helpers since it's now always required
- **Improved code consistency**: Ensured consistent handling of default session ID across the codebase

This refactoring is a follow-up to PR #25377 which addressed the security issue with tenant isolation. This PR focuses on improving the code organization of the constants used in that fix.

## Screenshots

N/A - This is a code refactoring with no visual changes.

## Checklist

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/api/models/model.py
+++ b/api/models/model.py
@@ -1459,6 +1459,14 @@ class OperationLog(Base):
     updated_at = mapped_column(sa.DateTime, nullable=False, server_default=func.current_timestamp())
 
 
+class DefaultEndUserSessionID(StrEnum):
+    """
+    End User Session ID enum.
+    """
+
+    DEFAULT_SESSION_ID = "DEFAULT-USER"
+
+
 class EndUser(Base, UserMixin):
     __tablename__ = "end_users"
     __table_args__ = (
--- a/api/controllers/files/upload.py
+++ b/api/controllers/files/upload.py
@@ -86,7 +86,7 @@ class PluginUploadFileApi(Resource):
             filename=filename,
             mimetype=mimetype,
             tenant_id=tenant_id,
-            user_id=user_id,
+            user_id=user.id,
             timestamp=timestamp,
             nonce=nonce,
             sign=sign,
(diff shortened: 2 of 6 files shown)


api/models/model.py
```diff
@@ -1459,6 +1459,14 @@ class OperationLog(Base):
     updated_at = mapped_column(sa.DateTime, nullable=False, server_default=func.current_timestamp())
 
 
+class DefaultEndUserSessionID(StrEnum):
```
QuantumGhost, line 1462: IMO, it's better to make the default session ID a constant:


```py
DEFAULT_END_USER_SESSION_ID = "DEFAULT-USER"
```

This also simplify its usage, avoid `.value` at call sites.

Is there a specific reason for not using a string constant in this case?

Yeuoly: I agree, but just to maintain consistency with models in `account.py`, it dose not function as an enum, but just because of a single constant between classes looking weird.

On the other hand, I'm not sure whether a session id like `DEFAULT-TRIGGER-USER` is needed in the future.

QuantumGhost: OK. Keep the `DefaultEndUserSessionID` as is for now.

QuantumGhost (approved):
LGTM

PR #25447, opened 2025-09-09
Weaviate update version

DhruvGorasiya:
This pull request updates the Weaviate client dependency to a newer major version and ensures compatibility in both the Python package configuration and the Docker development environment. The most important changes are as follows:

**Dependency Updates:**

* Updated the `weaviate-client` dependency in `api/pyproject.toml` from version `~3.24.0` to accept any `4.x.x` version, ensuring compatibility with the latest Weaviate features and bug fixes.

**Development Environment Configuration:**

* Modified `docker/docker-compose.override.yml` to mount the `weaviate_vector.py` file as read-only into the `api` service and to explicitly install the `weaviate` package version `>=4.0.0` at container startup, ensuring the development environment matches the new dependency requirements.to support the latest Python client (Weaviate Integration)
close https://github.com/langgenius/dify/pull/12258


crazywoola:
Hello, please resolve the conflicts. Most of the code LGTM. :)

crazywoola (approved):

kurokobo:
@DhruvGorasiya @crazywoola 
Has this been tested thoroughly?

- Instead of modifying the Dockerfile or entrypoint.sh, you’re taking a rather forceful and illogical approach by placing a docker-compose.override.yml file.
- Has it actually been tested that existing Dify users with Weaviate 1.19 can migrate their data without issues just by switching the image to 1.27?

At the very least, there’s a side effect reported in #26938.

DhruvGorasiya:
Let me take a look at this and get back to you ASAP @kurokobo

kurokobo:
@DhruvGorasiya 
Thanks for your rapid response, I understand that the override file was added in order to run Weaviate 1.27 even with the Dify 1.9.1 image, and that this is just a temporary workaround until the next version of Dify is released. Is that correct?

- However, override files are listed in .gitignore so that users can add services or customize configurations as needed. The override file added in this PR could lead to duplicate override files (`yml` and `yaml`), and it may unintentionally disable user-provided override files (because yml has priority), causing unforeseen side effects.
- Also, the contents of the override file are not appropriate. You should specify the `entrypoint` instead of the `command`, otherwise `pip install` won’t be triggered.
- In addition, you should be installing `weaviate-client`, not `weaviate`. The version specified also differs from what’s in `pyproject.toml`.
- Moreover, the command does not consider virtual environment for Python at all. Even if you install weaviate-client with `pip install`, it only installs the module as system-wide and not in the virtual environment where Dify runs. You should be using `uv` instead.
- If possible, it would be better to add these modification (volumes and entrypoint) directly to the docker-compose template file, rather than using an override file. However, please make sure to add a TODO comment to remove these additions when the next version is released.
- Weaviate is accessed not only by the api but also by worker.

That said, as a prerequisite, changing the Weaviate version should only be done after testing that migrating data from 1.19 to 1.27 can be done automatically without user intervention (I haven't tested this myself, so maybe I'm just worrying too much).
Note, according to the Weaviate migration guide, it says to update only one minor version at a time, so even if it does work, this wouldn't be a supported migration path.

DhruvGorasiya:
@kurokobo Thanks for the detailed feedback! Yes, your understanding is correct, this was meant as a temporary workaround until the next Dify release. I’m currently working on fixing the issues you pointed out (override behavior, entrypoint usage, correct package install, and virtual environment handling) and will update the PR accordingly soon.

DhruvGorasiya:
Hey @kurokobo I have created another PR with correct changes, I have removed the docker-compose-override file and I have updated the docker-compose-template and I have also added TODO message just as you said. Let me know if there are any changes that needs to be made.

https://github.com/langgenius/dify/pull/26964

PR #25518, opened 2025-09-11
security(api): fix privilege escalation vulnerability in model config and chat message APIs

QuantumGhost:
## Summary

The `ChatMessageApi` (`POST /console/api/apps/{app_id}/chat-messages`) and 
`ModelConfigResource` (`POST /console/api/apps/{app_id}/model-config`) 
endpoints do not properly validate user permissions, allowing users without `editor` 
permission to access restricted functionality.

This PR addresses this issue by adding proper permission check.

Closes #25516.

## Screenshots

N/A

## Checklist

### DIFF
--- a/api/models/account.py
+++ b/api/models/account.py
@@ -7,6 +7,7 @@ import sqlalchemy as sa
 from flask_login import UserMixin  # type: ignore[import-untyped]
 from sqlalchemy import DateTime, String, func, select
 from sqlalchemy.orm import Mapped, Session, mapped_column, reconstructor
+from typing_extensions import deprecated
 
 from models.base import Base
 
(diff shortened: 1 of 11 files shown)


api/models/account.py
```diff
@@ -188,6 +188,16 @@ def is_admin(self):
 
     @property
     def is_editor(self):
+        # TODO: Rename this property to `has_editing_permission`.
```
Nov1c444: Would it be better to write it this way?
```python
from typing_extensions import deprecated


@deprecated(reason="Use has_editing_permission instead.")
@property
def is_editor(self):
    return self.has_editing_permission


@property
def has_editing_permission(self):
    return TenantAccountRole.is_editing_role(self.role)

```

Nov1c444 (dismissed):

api/models/account.py
QuantumGhost: Sounds good.

Nov1c444 (dismissed):

Nov1c444 (approved):

Nov1c444 (approved):

PR #26052, opened 2025-09-22
improve: pooling httpx clients for requests to code sandbox and ssrf

### LINKED ISSUE #26047: [Chore/Refactor]Performance issue: redundant TLS context creation
### Self Checks

- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).

## Kubernetes Spec

Image

## Summary
- Each outbound request rebuilt an SSL context, related files:
    - api/core/helper/code_executor/code_executor.py
    - api/core/helper/ssrf_proxy.py
- Recreating the context on every call added latency and extra CPU allocations under load.
- The cost compounded across SSRF-proxied requests and the code execution proxy, degrading throughput.

## Impact
- during workflow load testing, we observed slow executions for http-node and code-execution-node

## Steps to Reproduce
1. workflow configurations are shown below, one for http calls, one for sandbox calls:

Image

Image

2. use locust to release many load to these two workflows, the result are shown below:
http node:

Image

code executor:

Image

3. use py-spy to inspect the cpu usage, the result suggests the issue is related to the ssl verification context.

Image

## Proposed Fix
- Introduce a shared helper that caches the `ssl.SSLContext` instance and centralises env parsing logic.
- Reuse this helper wherever SSL verification settings are consumed so the context is created once and reused safely.

we have gone ahead and prepared a pull request to this issue that **doubled the performance**.

### Motivation

_No response_

### Additional Context

_No response_

Blackoutta:
# Summary
Closes #26047
- refactor SSL verification parsing into `core/helper/ssl_verification.py` with a cached `ssl.SSLContext`
- reuse the helper in SSRF proxy and code executor modules to eliminate redundant context construction
- add `CODE_EXECUTION_SSL_VERIFY` config + env sample so sandbox traffic can tune verification independently

# Performance Impact
- for http nodes, the RPS went from 60 to 170
- for code execution nodes, the RPS went from 110 to 170

# workflow setup
## Http Node
image

## Code Executor Node
image

# Before-Afters
### Before
image

### After
image

### Before
image

### After
image

### Before
image

### After
image


api/core/helper/ssrf_proxy.py
```diff
@@ -51,9 +45,10 @@
         )
 
     if "ssl_verify" not in kwargs:
-        kwargs["ssl_verify"] = http_request_node_ssl_verify
+        kwargs["ssl_verify"] = HTTP_REQUEST_NODE_SSL_VERIFY
 
-    ssl_verify = kwargs.pop("ssl_verify")
+    ssl_verify_bool = kwargs.pop("ssl_verify")
+    ssl_verify = HTTP_REQUEST_NODE_SSL_VERIFY if ssl_verify_bool else False
```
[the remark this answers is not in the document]
Blackoutta: good point, merged this change to my code!

api/core/helper/ssl_verification.py
```diff
@@ -0,0 +1,45 @@
...
+    if isinstance(value, bool):
+        return get_cached_ssl_context() if value else False
+
+    lowered_value = str(value).strip().lower()
+    if lowered_value == "true":
+        return get_cached_ssl_context()
+    if lowered_value == "false":
+        return False
+
+    raise ValueError(f"Invalid value for {config_name}. Expected 'True' or 'False'.")
```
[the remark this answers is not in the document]
Blackoutta: I'll stay with the explict check style

api/core/helper/ssl_verification.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+
+    if isinstance(value, ssl.SSLContext):
+        return value
+
+    if value is None:
+        return get_cached_ssl_context()
+
+    if isinstance(value, bool):
+        return get_cached_ssl_context() if value else False
```
bowenliang123: Try to prefer splitting into named parameters for this method. Prevent mixing up the value in different types and usages.

api/core/helper/ssl_verification.py
```diff
@@ -0,0 +1,45 @@
...
+
+def parse_ssl_verify_config(value: Any, *, config_name: str) -> SslVerifySetting:
+    """Parse a configuration value into an httpx-compatible verify setting.
+
+    Accepts booleans, strings, or ssl.SSLContext instances. Truthy strings resolve
+    to the cached SSL context while falsy strings disable verification.
+    """
+
+    if isinstance(value, ssl.SSLContext):
+        return value
```
bowenliang123: If an `ssl.SSLContext` instance returned here, how it's guranateed as a instance of `SslVerifySetting` in method singature as well ? It's sounds tricky.

api/core/helper/ssrf_proxy.py
```diff
@@ -8,23 +8,16 @@
...
-    if http_request_node_ssl_verify_lower == "true":
-        http_request_node_ssl_verify = True
-    elif http_request_node_ssl_verify_lower == "false":
-        http_request_node_ssl_verify = False
-    else:
-        raise ValueError("Invalid value. HTTP_REQUEST_NODE_SSL_VERIFY should be 'True' or 'False'")
-except NameError:
-    http_request_node_ssl_verify = True
+HTTP_REQUEST_NODE_SSL_VERIFY = parse_ssl_verify_config(
+    getattr(dify_config, "HTTP_REQUEST_NODE_SSL_VERIFY", None),
```
bowenliang123: Or just direct usage as `dify_config.HTTP_REQUEST_NODE_SSL_VERIFY`?

api/core/helper/ssl_verification.py
```diff
@@ -0,0 +1,45 @@
...
+import ssl
+from functools import lru_cache
+from typing import Any, Union
+
+import certifi
+
+SslVerifySetting = Union[bool, ssl.SSLContext]
+
+
+@lru_cache(maxsize=1)
```
bowenliang123: Why is the max size set to 1 for all accesses in the global? Is the singleton good enough to handle it?

api/core/helper/ssl_verification.py
```diff
@@ -0,0 +1,45 @@
...
+def _cached_ssl_context() -> ssl.SSLContext:
+    """Return a cached default SSL context backed by certifi's CA bundle."""
+    return ssl.create_default_context(cafile=certifi.where())
+
+
+def get_cached_ssl_context() -> ssl.SSLContext:
+    """Expose the cached SSL context so callers reuse the same instance."""
+    return _cached_ssl_context()
```
bowenliang123: Another transparent method?

bowenliang123 (commented):
A good PR to boost HTTP request performance, though some comments from me for details.

crazywoola:
Hello I am banana from Dify, please add my wechat crazyphage.

api/core/helper/ssl_verification.py
Blackoutta: - parse_ssl_verify_config declares -> SslVerifySetting, and SslVerifySetting is just a type alias for Union[bool, ssl.SSLContext]; returning an ssl.SSLContext therefore satisfies the annotation because that class
  is one of the union members (api/core/helper/ssl_verification.py:30).
  - When the function hits return get_cached_ssl_context(), the helper itself is annotated to return ssl.SSLContext, so both static type checkers and readers know the value fits the alias.
  - There’s no runtime casting happening—Python annotations aren’t enforced at runtime—but type checkers (e.g., basedpyright per this repo) verify that every branch yields either a bool or an ssl.SSLContext.

api/core/helper/ssrf_proxy.py
Blackoutta: good point ;)

api/core/helper/ssl_verification.py
Blackoutta: - _cached_ssl_context has no inputs, so every call should return the same default context; setting maxsize=1 makes it a memoized singleton with the usual lru_cache ergonomics.
  - Building an ssl.SSLContext is relatively expensive, so caching the first instance and reusing it avoids repeated setup while keeping the function signature simple.
  - One cached instance is sufficient because callers don’t need different contexts—any configuration that requires a custom context can pass its own ssl.SSLContext and bypass the cache altogether.

api/core/helper/ssl_verification.py
Blackoutta: - _cached_ssl_context is intentionally private: it’s where the @lru_cache lives, and the leading underscore signals “implementation detail.”
- get_cached_ssl_context() is the public helper the rest of the code can import; if the caching strategy ever changes (different maxsize, reset hooks, metrics, config), callers stay untouched.
- The wrapper also makes testing easier—you can monkeypatch or stub it without touching the cached function, and you can add validation/logging later while keeping the existing API stable.

api/core/helper/ssl_verification.py
bowenliang123: hmmm,,, I would not prefer `SslVerifySetting = Union[bool, ssl.SSLContext]` as a good type definition. It's mixing varying types of inputs and outputs anyway and hard to understand.

api/core/helper/ssl_verification.py
bowenliang123: I got the point that caching over creating ssl context for callers. But the key point is the scope of `lru_cache` wrapper as it's global, and one for all?

api/core/helper/ssl_verification.py
bowenliang123: Sounds not insufficent to keep a dedicated wrapper here just for imports.

api/core/helper/ssl_verification.py
Blackoutta: good point, I'll refactor it tomorrow

api/core/helper/ssl_verification.py
Blackoutta: got it, will give this method an overhaul

api/core/helper/ssl_verification.py
Blackoutta: yes, it's a singleton, it's the same as:
```python
_cached_context: ssl.SSLContext | None = None

  def get_cached_ssl_context() -> ssl.SSLContext:
      global _cached_context
      if _cached_context is None:
          _cached_context = ssl.create_default_context(cafile=certifi.where())
      return _cached_context
```

api/core/helper/ssl_verification.py
Blackoutta: can remove the wrapper for simplicity if you will

Blackoutta:
# Http Pooling Summary

  - Introduced a reusable HTTP client pooling facility and wired it into the code execution sandbox and SSRF proxy flows to cut connection churn and improve throughput.
  - Expanded sandbox and SSRF configuration to cover SSL verification controls plus tunable connection/keepalive limits, surfacing them in *.env.example and compose defaults.

# Details
  - Code executor now builds its httpx client once per process via the new pooling helper, reusing tuned limits and honoring a configurable SSL verification flag when firing sandbox requests.
  - SSRF proxy routines share verified/unverified client pools, preserving existing proxy behavior while centralizing retry/backoff handling across pooled connections.

# Benchmark With Pooling
## Pooling Paramters
```
httpPooling:
  codeExecution:
    poolKeepaliveExpiry: 60 
    poolMaxConnections: 180 (poolMaxKeepaliveConnections * 2)
    poolMaxKeepaliveConnections: 90 (~=2000 concurrent request/3pod/7 gunicorn worker)
  ssrf:
    poolKeepaliveExpiry: 15
    poolMaxConnections: 180
    poolMaxKeepaliveConnections: 90
```
## HTTP Nodes
the RPS increased to 190-ish
image

## Code Executor
the RPS kind of stayed the same as the last commit
image

## py-spy
looks ok
image

Blackoutta:
@bowenliang123 Hi, I've migrated the previous design from caching-ssl-contexts to utilizing-multiple-isolated-http-client-pools, could you review it again if you have spared time, thank you ;)
Also, see above comments for newly generated benchmarks.

api/core/helper/code_executor/code_executor.py
```diff
@@ -13,9 +13,17 @@
 from core.helper.code_executor.jinja2.jinja2_transformer import Jinja2TemplateTransformer
 from core.helper.code_executor.python3.python3_transformer import Python3TemplateTransformer
 from core.helper.code_executor.template_transformer import TemplateTransformer
+from core.helper.http_client_pooling import get_pooled_http_client
 
 logger = logging.getLogger(__name__)
 code_execution_endpoint_url = URL(str(dify_config.CODE_EXECUTION_ENDPOINT))
+CODE_EXECUTION_SSL_VERIFY = bool(dify_config.CODE_EXECUTION_SSL_VERIFY)
```
bowenliang123: skip this line of redundant casting, use `dify_config.CODE_EXECUTION_SSL_VERIFY` in line.

api/configs/feature/__init__.py
```diff
@@ -404,6 +424,21 @@ def WEB_API_CORS_ALLOW_ORIGINS(self) -> list[str]:
...
     )
 
+    SSRF_POOL_MAX_CONNECTIONS: PositiveInt = Field(
+        description="Maximum number of concurrent connections for the SSRF HTTP client",
+        default=100,
+    )
+
+    SSRF_POOL_MAX_KEEPALIVE_CONNECTIONS: PositiveInt = Field(
+        description="Maximum number of persistent keep-alive connections for the SSRF HTTP client",
+        default=20,
```
bowenliang123, line 434: Maybe we could decrease the allowable keep-alive connections. There's no extra value for holding the connections when idle.

api/configs/feature/__init__.py
```diff
@@ -113,6 +113,21 @@ class CodeExecutionSandboxConfig(BaseSettings):
...
     )
 
+    CODE_EXECUTION_POOL_MAX_CONNECTIONS: PositiveInt = Field(
+        description="Maximum number of concurrent connections for the code execution HTTP client",
+        default=100,
+    )
+
+    CODE_EXECUTION_POOL_MAX_KEEPALIVE_CONNECTIONS: PositiveInt = Field(
+        description="Maximum number of persistent keep-alive connections for the code execution HTTP client",
+        default=20,
```
bowenliang123, line 123: ditto.

api/core/helper/ssrf_proxy.py
```diff
@@ -8,34 +8,70 @@
...
-    http_request_node_ssl_verify_lower = str(config_value).lower()
-    if http_request_node_ssl_verify_lower == "true":
-        http_request_node_ssl_verify = True
-    elif http_request_node_ssl_verify_lower == "false":
-        http_request_node_ssl_verify = False
-    else:
-        raise ValueError("Invalid value. HTTP_REQUEST_NODE_SSL_VERIFY should be 'True' or 'False'")
-except NameError:
-    http_request_node_ssl_verify = True
+HTTP_REQUEST_NODE_SSL_VERIFY = dify_config.HTTP_REQUEST_NODE_SSL_VERIFY
```
bowenliang123: skip this line and use dify_config.HTTP_REQUEST_NODE_SSL_VERIFY in place.

api/core/helper/ssrf_proxy.py
```diff
@@ -8,34 +8,70 @@
...
+
+
+def _get_ssrf_client(ssl_verify_enabled: bool) -> httpx.Client:
+    if not isinstance(ssl_verify_enabled, bool):
+        raise ValueError("SSRF client verify flag must be a boolean")
+
+    if ssl_verify_enabled:
+        return get_pooled_http_client(_SSL_VERIFIED_POOL_KEY, lambda: _build_ssrf_client(verify=True))
+
+    return get_pooled_http_client(_SSL_UNVERIFIED_POOL_KEY, lambda: _build_ssrf_client(verify=False))
```
bowenliang123:
```suggestion
    return get_pooled_http_client(_SSL_UNVERIFIED_POOL_KEY, lambda: _build_ssrf_client(verify= ssl_verify_enabled))
```

bowenliang123 (approved):
LGTM overall. Minor comments for code improvement.
Pooling httpx client instances is much better and safer than the previous version reusing the ssl context.
cc @QuantumGhost @laipz8200

bowenliang123:
Suggested PR title:
`improve: pooling httpx clients for requests to code sandbox and ssrf`
cc @Blackoutta

api/core/helper/ssrf_proxy.py
Blackoutta: I think the key is needed to retrieve the correct pool ( one will verify ssl, one will not)

api/configs/feature/__init__.py
Blackoutta: any suggestion on the default values?

api/core/helper/ssrf_proxy.py
Blackoutta: a valid one-liner could be:
```python
def _get_ssrf_client(ssl_verify_enabled: bool) -> httpx.Client:
    if not isinstance(ssl_verify_enabled, bool):
        raise ValueError("SSRF client verify flag must be a boolean")

    return get_pooled_http_client(
        _SSL_VERIFIED_POOL_KEY if ssl_verify_enabled else _SSL_UNVERIFIED_POOL_KEY,
        lambda: _build_ssrf_client(verify=ssl_verify_enabled),
```

api/core/helper/code_executor/code_executor.py
Blackoutta: solved.

api/core/helper/ssrf_proxy.py
Blackoutta: solved.

crazywoola (approved):

laipz8200:
I don't understand why we need to use such a complex pattern to build just three global httpxClient objects, and we also need to use a lock to access them.

wangrg:
Can plugin_daemon also be modified to use httpx clients pooling?

PR #26155, opened 2025-09-24
fix: variable not found #26144

sqewad:
1. variable not found when accessing a nested attr with empty value

> [!IMPORTANT]
>

## Summary

fix: #26144
Distinguish between the case where a variable exists but its value is empty and the case where the variable does not exist.

## Screenshots

| Before | After |
|--------|-------|
| Weixin Image_20250924084836_137 | Weixin Image_20250924084833_138 |

## Checklist

### DIFF
--- a/api/core/workflow/entities/variable_pool.py
+++ b/api/core/workflow/entities/variable_pool.py
@@ -197,7 +197,7 @@ class VariablePool(BaseModel):
                 The corresponding Segment built from the attribute value if the key exists,
                 otherwise None.
         """
-        if not isinstance(obj, dict) or not attr in obj:
+        if not isinstance(obj, dict) or attr not in obj:
             return None
         return variable_factory.build_segment(obj.get(attr))
 
--- a/api/core/workflow/entities/variable_pool.py
+++ b/api/core/workflow/entities/variable_pool.py
@@ -184,8 +184,19 @@ class VariablePool(BaseModel):
         """Extract the actual value from an ObjectSegment."""
         return obj.value if isinstance(obj, ObjectSegment) else obj
 
-    def _get_nested_attribute(self, obj: Mapping[str, Any], attr: str):
-        """Get a nested attribute from a dictionary-like object."""
+    def _get_nested_attribute(self, obj: Mapping[str, Any], attr: str) -> tuple[bool, Any]:
+        """
+        Get a nested attribute from a dictionary-like object.
+
+        Args:
+            obj: The dictionary-like object to search.
+            attr: The key to look up.
+
+        Returns:
+            tuple:
+                - A boolean indicating whether the attribute exists in the object.
+                - The value of the attribute if found, otherwise None.
+        """
         if not isinstance(obj, dict):
             return False, None
         return attr in obj, obj.get(attr)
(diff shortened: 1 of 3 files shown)


api/core/workflow/entities/variable_pool.py
```diff
@@ -187,8 +187,8 @@ def _extract_value(self, obj: Any):
     def _get_nested_attribute(self, obj: Mapping[str, Any], attr: str):
         """Get a nested attribute from a dictionary-like object."""
         if not isinstance(obj, dict):
-            return None
-        return obj.get(attr)
+            return False, None
+        return attr in obj, obj.get(attr)
```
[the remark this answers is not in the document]
sqewad: i think using "in" is more readable. there aren’t many performance requirements here, right?

api/core/workflow/entities/variable_pool.py
```diff
@@ -187,8 +187,8 @@ def _extract_value(self, obj: Any):
     def _get_nested_attribute(self, obj: Mapping[str, Any], attr: str):
```
QuantumGhost: A minor suggestion: instead of returning `tuple[bool, Any]`, consider changing the signature of this function to `tuple[Any, bool]`, and adding a short description about the second value in the tuple to the docstring.

QuantumGhost (commented):
Hi @sqewad, thank you for the contribution!

The overall changes look good to me. However, I'd suggest adding unit tests for both the `get` and `_get_nested_attribute` methods to cover the specific edge case you fixed. 
This will help prevent regression and ensure the nested attribute handling works correctly in various scenarios.

This would make the fix more robust and maintainable.

api/core/workflow/entities/variable_pool.py
sqewad: sure I'll do it, thanks for the suggestion

api/core/workflow/entities/variable_pool.py
```diff
@@ -184,11 +184,22 @@ def _extract_value(self, obj: Any):
...
+        """
+        Get a nested attribute from a dictionary-like object.
+
+        Args:
+            obj: The dictionary-like object to search.
+            attr: The key to look up.
+
+        Returns:
+            tuple:
+                - A boolean indicating whether the attribute exists in the object.
```
QuantumGhost: It seems that the order of elements in returned tuple is not updated.

IMO it's better to return the two values as 
`None, False` and `obj.get(attr), attr in obj`

sqewad: Thanks for pointing it out, I’ve updated the tuple order as suggested.

api/core/workflow/entities/variable_pool.py
```diff
@@ -184,11 +184,22 @@ def _extract_value(self, obj: Any):
...
+
+        Returns:
+            tuple:
+                - The value of the attribute if found, otherwise None.
+                - A boolean indicating whether the attribute exists in the object.
+        """
         if not isinstance(obj, dict):
-            return None
-        return obj.get(attr)
+            return None, False
```
laipz8200: return a tuple is hard to read, please consider returning a NoneSegment if the key exists but value is None.

you can use `variable_factory.build_segment(None)` to get one.

laipz8200 (changes requested):

api/core/workflow/entities/variable_pool.py
sqewad: good idea, woking on it

laipz8200 (approved):

PR #26585, opened 2025-10-05
feat(graph_engine): Support pausing workflow graph executions

laipz8200:
## Summary

Fixes #26584

### Main features

- Support pausing execution of `GraphEngine` via `NodeEvent` or `Command`.
- Move necessary states to `GraphRuntimeState` and provide serialization and deserialization methods.
- Delete `WorkflowCycleManager` and replace it with a new persistent Layer.
- Move `VariablePool`, `GraphRuntimeState`, and related classes to the new `runtime` package
- Added a based `HumanInput` node for testing.
- Supported GraphEngine's TestRunner with Factory pattern.
- Added a Builder to `Graph` to support quick Graph construction via code for testing.

### Others

- Introduce `pytest-timeout` to avoid test in dead loop.
- Add RAG Pipeline queue to `dev/start-worker`.
- Add a `get_by_prefix` method for `VariablePool`.
- Remove some outdated attrs from queue events.

## Screenshots

| Before | After |
|--------|-------|
| n/a | n/a |

## Checklist


laipz8200:
This PR is ready to be merged. I haven't actually handled the persistence of `GraphRuntimeState` in this PR, but it should be easy to implement by capturing the engine's pause events in the new persistence layer. Afterwards, we can simply reassemble the `GraphRuntimeState` to allow the new engine to continue execution from where it last stopped. 

cc @QuantumGhost @crazywoola

crazywoola (dismissed):

crazywoola (dismissed):

api/core/app/apps/advanced_chat/generate_task_pipeline.py
```diff
@@ -305,35 +315,38 @@ def _handle_error_event(self, event: QueueErrorEvent, **kwargs) -> Generator[Err
...
             message = self._get_message(session=session)
             if not message:
                 raise ValueError(f"Message not found: {self._message_id}")
 
-            message.workflow_run_id = workflow_execution.id_
-            workflow_start_resp = self._workflow_response_converter.workflow_start_to_stream_response(
-                task_id=self._application_generate_entity.task_id,
-                workflow_execution=workflow_execution,
-            )
+            message.workflow_run_id = run_id
```
QuantumGhost, line 303: The `message` modified here is not added to the `session` and the modification will not be persisted.

api/core/app/apps/advanced_chat/generate_task_pipeline.py
```diff
@@ -368,14 +376,12 @@ def _handle_node_succeeded_event(
...
             )
 
-        workflow_node_execution = self._workflow_cycle_manager.handle_workflow_node_execution_success(event=event)
         node_finish_resp = self._workflow_response_converter.workflow_node_finish_to_stream_response(
             event=event,
             task_id=self._application_generate_entity.task_id,
-            workflow_node_execution=workflow_node_execution,
         )
 
-        self._save_output_for_event(event, workflow_node_execution.id)
```
QuantumGhost, line 377: I'm not sure the two fields (`workflow_node_execution.id` and `event.node_execution_id`) are interchangeable. The name `node_execution_id` is is very confusing. In some situations it means the database primary key of `WorkflowNodeExecutionModel`. In other scenarios it means the unique identifier of a specific `Node` instance.

api/core/app/apps/advanced_chat/generate_task_pipeline.py
```diff
@@ -386,16 +392,13 @@ def _handle_node_failed_events(
...
-
         node_finish_resp = self._workflow_response_converter.workflow_node_finish_to_stream_response(
             event=event,
             task_id=self._application_generate_entity.task_id,
-            workflow_node_execution=workflow_node_execution,
         )
 
         if isinstance(event, QueueNodeExceptionEvent):
-            self._save_output_for_event(event, workflow_node_execution.id)
+            self._save_output_for_event(event, event.node_execution_id)
```
QuantumGhost, line 371: ditto

api/core/app/apps/common/workflow_response_converter.py
```diff
@@ -39,93 +38,209 @@
...
+        self._workflow_execution_id: str | None = None
+        self._workflow_started_at: datetime | None = None
+
+    # ------------------------------------------------------------------
+    # Workflow lifecycle helpers
+    # ------------------------------------------------------------------
+    def _prepare_workflow_inputs(self) -> Mapping[str, Any]:
+        inputs = dict(self._application_generate_entity.inputs)
+        for field_name, value in self._system_variables.to_dict().items():
+            if field_name == SystemVariableKey.CONVERSATION_ID:
```
QuantumGhost, line 97: I suggest adding a comment for the reason why `CONVERSATION_ID` is omitted from inputs mapping here.

api/core/app/apps/common/workflow_response_converter.py
```diff
@@ -39,93 +38,209 @@
...
 
 
+@dataclass(slots=True)
+class _NodeSnapshot:
+    """In-memory cache for node metadata between start and completion events."""
+
+    title: str
+    index: int
+    start_at: datetime
+    iteration_id: str = ""
```
QuantumGhost, line 66: I suggest adding comments for iteration_id and loop_id to clarify that an empty string (`""`) indicates the node is not within a loop or iteration."

api/core/app/apps/common/workflow_response_converter.py
```diff
@@ -39,93 +38,209 @@
...
         application_generate_entity: Union[AdvancedChatAppGenerateEntity, WorkflowAppGenerateEntity],
         user: Union[Account, EndUser],
+        system_variables: SystemVariable,
     ):
         self._application_generate_entity = application_generate_entity
         self._user = user
+        self._system_variables = system_variables
+        self._workflow_inputs = self._prepare_workflow_inputs()
         self._truncator = VariableTruncator.default()
+        self._node_snapshots: dict[str, _NodeSnapshot] = {}
```
QuantumGhost: There should be a comment for `_node_snapshots` to clarify what its keys represent. For example:

```python
# _node_snapshots: Dict mapping node_execution_id to the corresponding node's metadata snapshot
```

api/core/app/apps/advanced_chat/generate_task_pipeline.py
```diff
@@ -789,39 +752,31 @@ def _process_stream_response(
...
-        # Initialize graph runtime state
-        graph_runtime_state: GraphRuntimeState | None = None
-
         for queue_message in self._base_task_pipeline.queue_manager.listen():
             event = queue_message.event
 
             match event:
                 case QueueWorkflowStartedEvent():
-                    graph_runtime_state = event.graph_runtime_state
+                    self._resolve_graph_runtime_state()
```
QuantumGhost, line 730: The modification of `self._graph_runtime_state` is obscure. It's hard to grasp when and how `self._graph_runtime_state` is modified.

api/core/app/apps/common/workflow_response_converter.py
```diff
@@ -39,93 +38,209 @@
...
                 "email": user.email,
             }
-        elif isinstance(user, EndUser):
+        else:
             created_by = {
                 "id": user.id,
                 "user": user.session_id,
             }
-        else:
-            raise NotImplementedError(f"User type not supported: {type(user)}")
```
QuantumGhost, line 103: I suggest keep the `raise NotImplementedError` to simplify debugging when new user types are added.

api/core/app/apps/common/workflow_response_converter.py
```diff
@@ -39,93 +38,209 @@
...
-                error=workflow_execution.error_message,
-                elapsed_time=workflow_execution.elapsed_time,
-                total_tokens=workflow_execution.total_tokens,
-                total_steps=workflow_execution.total_steps,
+                id=run_id,
+                workflow_id=workflow_id,
+                status=status.value,
+                outputs=encoded_outputs,
+                error=error,
+                elapsed_time=elapsed_time,
```
QuantumGhost, line 245: IMO the `elapsed_time` should be defined as total time spent **on execution**. It should not include time  elapsed during pause.

Consider the following scenario:

1. Workflow w1 started executing at timestamp t0.
2. w1 paused at timestamp t1.
3. w1 continued execution at t2.
4. w1 finished at t3.

The `elapsed_time` for the example above should be `(t1 - t0) + (t3 - t2)`.

It means that the execution time must be persisted as part of state, instead of kept as a in-memory value in `WorkflowResponseConverter`.

Another definition of `elapsed_time` is the difference between the first time workflow started executing and the time workflow finished. In this definition, the  `started_at` should be  `GraphRuntimeState.start_at`, which is the time the workflow started executing.

api/core/app/apps/common/workflow_response_converter.py
```diff
@@ -189,41 +294,54 @@ def workflow_node_finish_to_stream_response(
...
+                process_data=process_data,
+                process_data_truncated=process_data_truncated,
+                outputs=outputs,
+                outputs_truncated=outputs_truncated,
+                status=status,
+                error=error_message,
+                elapsed_time=elapsed_time,
+                execution_metadata=metadata,
+                created_at=int(start_at.timestamp()),
+                finished_at=int(finished_at.replace(tzinfo=UTC).timestamp()),
```
QuantumGhost: `.replace(tzinfo=UTC)` is unnecessary.

api/core/app/apps/common/workflow_response_converter.py
```diff
@@ -234,44 +352,45 @@ def workflow_node_retry_to_stream_response(
...
-        if not workflow_node_execution.finished_at:
+    ) -> NodeRetryStreamResponse | None:
+        if event.node_type in {NodeType.ITERATION, NodeType.LOOP}:
             return None
+        run_id = self._ensure_workflow_run_id()
 
-        json_converter = WorkflowRuntimeTypeConverter()
+        snapshot = self._get_snapshot(event.node_execution_id)
+        if snapshot is None:
+            raise ValueError("node retry event arrived without a stored snapshot")
```
QuantumGhost: I suggest use `AssertionError` to better indicate that this is a violation of precondition.

api/core/app/apps/workflow/generate_task_pipeline.py
```diff
@@ -261,14 +237,41 @@ def _database_session(self):
 
     def _ensure_workflow_initialized(self):
         """Fluent validation for workflow state."""
-        if not self._workflow_run_id:
+        if not self._workflow_execution_id:
             raise ValueError("workflow run not initialized.")
 
-    def _ensure_graph_runtime_initialized(self, graph_runtime_state: GraphRuntimeState | None) -> GraphRuntimeState:
+    def _ensure_graph_runtime_initialized(
```
QuantumGhost: The `_ensure_graph_runtime_initialized`, `_extract_workflow_run_id` and `_resolve_graph_runtime_state` methods are identical to those in `AdvancedChatAppGenerateTaskPipeline`. Is it better to share those code between the two classes?

api/core/app/apps/workflow/generate_task_pipeline.py
```diff
@@ -331,14 +327,12 @@ def _handle_node_succeeded_event(
...
     ) -> Generator[StreamResponse, None, None]:
         """Handle node succeeded events."""
-        workflow_node_execution = self._workflow_cycle_manager.handle_workflow_node_execution_success(event=event)
         node_success_response = self._workflow_response_converter.workflow_node_finish_to_stream_response(
             event=event,
             task_id=self._application_generate_entity.task_id,
-            workflow_node_execution=workflow_node_execution,
         )
 
-        self._save_output_for_event(event, workflow_node_execution.id)
```
QuantumGhost, line 341: ditto

api/core/app/apps/workflow/generate_task_pipeline.py
```diff
@@ -349,17 +343,13 @@ def _handle_node_failed_events(
...
-        )
         node_failed_response = self._workflow_response_converter.workflow_node_finish_to_stream_response(
             event=event,
             task_id=self._application_generate_entity.task_id,
-            workflow_node_execution=workflow_node_execution,
         )
 
         if isinstance(event, QueueNodeExceptionEvent):
-            self._save_output_for_event(event, workflow_node_execution.id)
+            self._save_output_for_event(event, event.node_execution_id)
```
QuantumGhost, line 320: ditto

api/core/workflow/graph_engine/graph_engine.py
```diff
@@ -237,47 +225,68 @@ def run(self) -> Generator[GraphEngineEvent, None, None]:
...
-            # Start execution
-            self._graph_execution.start()
+            is_resume = self._graph_execution.started
+            if not is_resume:
+                self._graph_execution.start()
+            else:
+                self._graph_execution.paused = False
+                self._graph_execution.pause_reason = None
+
             start_event = GraphRunStartedEvent()
```
QuantumGhost, line 238: Should we distinguish between `start` and `resume` events here?

api/core/workflow/graph_events/graph.py
```diff
@@ -26,3 +26,10 @@ class GraphRunAbortedEvent(BaseGraphEvent):
 
     reason: str | None = Field(default=None, description="reason for abort")
     outputs: dict[str, object] = Field(default_factory=dict, description="partial outputs if any")
+
+
+class GraphRunPausedEvent(BaseGraphEvent):
+    """Event emitted when a graph run is paused by user command."""
+
+    reason: str | None = Field(default=None, description="reason for pause")
+    outputs: dict[str, object] = Field(default_factory=dict, description="partial outputs if any")
```
QuantumGhost: I'm wondering the purpose of  the `outputs` fields here. Could you explain it or add a short comment to the `outputs` field?

api/core/workflow/runtime/graph_runtime_state.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def register(self, response_node_id: str) -> None: ...
+
+    def loads(self, data: str) -> None: ...
+
+    def dumps(self) -> str: ...
+
+
+class GraphProtocol(Protocol):
+    """Marker protocol for Graph runtime state attachment."""
```
QuantumGhost: `Marker protocol` seems weird to me. I guess this is equivalent to `Any` in the type checker's perspective. Is there any special reason to introduce this marker protocol?

api/core/workflow/runtime/graph_runtime_state.py
```diff
@@ -0,0 +1,353 @@
...
+        """Retrieve and clear the list of paused nodes awaiting resume."""
+
+        nodes = list(self._paused_nodes)
+        self._paused_nodes.clear()
+        return nodes
+
+    # ------------------------------------------------------------------
+    # Builders
+    # ------------------------------------------------------------------
+    def _build_ready_queue(self) -> ReadyQueueProtocol:
```
QuantumGhost, line 375: Is there any special reason to use `import_module` to import and create those classes?

api/core/workflow/runtime/graph_runtime_state.py
```diff
@@ -0,0 +1,353 @@
...
+from core.model_runtime.entities.llm_entities import LLMUsage
+from core.workflow.runtime.variable_pool import VariablePool
+
+
+class ReadyQueueProtocol(Protocol):
+    """Structural interface required from ready queue implementations."""
+
+    def put(self, item: str) -> None: ...
+
+    def get(self, timeout: float | None = None) -> str: ...
```
QuantumGhost: I suggest adding some documentation for methods in those protocols, to help reader understand their purposes, preconditions and postconditions.

crazywoola (dismissed):

api/core/app/apps/advanced_chat/generate_task_pipeline.py
laipz8200: This `message` will auto linked to session in `_get_message` and auto commited when exit the `with` statement, check `_database_session`.

api/core/app/apps/advanced_chat/generate_task_pipeline.py
laipz8200: Thanks for flagging this! The engine now generates a single UUID for each node execution when `Node.run()` yields its `NodeRunStartedEvent`. That ID is persisted directly in `WorkflowPersistenceLayer` as both the domain `WorkflowNodeExecution.id` and its `node_execution_id`. When the runner emits `QueueNodeSucceededEvent`, it reuses that same value in `event.node_execution_id`, so the queue payload already contains the database primary key. Because of that wiring we no longer need to fetch the execution again just to read its id, and this continues to hold across pause/resume flows because the persisted execution anchors on that UUID.

api/core/app/apps/advanced_chat/generate_task_pipeline.py
laipz8200: Echoing my note above: the queue event already carries the persisted execution UUID, so we can keep using `event.node_execution_id` without pulling the record again.

api/core/app/apps/common/workflow_response_converter.py
laipz8200: Added an inline note explaining that we skip `sys.conversation_id` so workflow inputs stay reusable without binding subsequent reruns to the original conversation.

api/core/app/apps/common/workflow_response_converter.py
laipz8200: Documented the `iteration_id` and `loop_id` fields in `_NodeSnapshot` to clarify that the empty string represents the no-iteration/no-loop case.

api/core/app/apps/common/workflow_response_converter.py
laipz8200: Introduced a `NodeExecutionId` NewType so `_node_snapshots` now advertises its key semantics via typing, rather than needing an inline comment.

api/core/app/apps/common/workflow_response_converter.py
laipz8200: This modification is because the type checker reports that 'user' has only two types, and I believe we will not easily add a new 'User' type. The modification can avoid a single 'isinstance' check.

api/core/app/apps/advanced_chat/generate_task_pipeline.py
laipz8200: Dropped the cached `_graph_runtime_state` altogether and replaced it with `_require_graph_runtime_state(...)`, which always reads from the queue manager and raises if the engine never attached a state. Passing an override now writes straight through to the queue manager, so the data flow is explicit and there is only one place where the shared state can change.

laipz8200: I split the lifecycle out into `_seed_graph_runtime_state_from_queue_manager()` during init and `_update_graph_runtime_state()` for explicit refreshes, so `_resolve_graph_runtime_state()` now just reads the cached value unless we hand it a new snapshot. That should make the write sites obvious.

api/core/app/apps/common/workflow_response_converter.py
laipz8200: Thanks for the detailed scenario. For now we prefer to keep the current elapsed time semantics and debate the alternatives before changing persisted state. We will revisit once we align on the definition.

api/core/app/apps/common/workflow_response_converter.py
laipz8200: Good catch! Dropped the redundant tz conversion so we now serialize the naive UTC timestamp directly.

api/core/app/apps/common/workflow_response_converter.py
laipz8200: Switched this precondition violation to raise AssertionError as suggested.

api/core/app/apps/workflow/generate_task_pipeline.py
laipz8200: Refactored both pipelines to reuse a shared GraphRuntimeStateSupport mixin and added focused unit coverage to keep the helper behaviour consistent.

api/core/workflow/graph_engine/graph_engine.py
laipz8200: Thanks for raising it. For now we’ll keep using the existing 'start' event to cover both fresh runs and resume flows while we evaluate whether separate event types buy us enough signal.

api/core/workflow/graph_events/graph.py
laipz8200: Added explicit descriptions on the run-level events clarifying that the outputs payload carries the final or partial workflow outputs so downstream consumers know what is available.

This event is just like GraphRunSucceededEvent, because Engine will stop after this event, we need an outputs for blocking api.

api/core/workflow/runtime/graph_runtime_state.py
laipz8200: Replaced the empty marker with a structural protocol that spells out the graph surface (nodes, edges, root node, and outgoing-edge traversal) required by the runtime.

api/core/workflow/runtime/graph_runtime_state.py
laipz8200: Swapped the dynamic import_module calls for direct imports so the runtime builders instantiate the concrete default classes without reflection.

api/core/workflow/runtime/graph_runtime_state.py
laipz8200: Expanded the protocol docstrings so each method now spells out its intent, inputs, and expected behaviour for implementors.

api/core/workflow/runtime/graph_runtime_state.py
laipz8200: Kept the lazy import_module approach but documented why we need it to satisfy the import-linter boundaries.

laipz8200: I'll try to move them out.

api/core/app/apps/advanced_chat/generate_task_pipeline.py
QuantumGhost: #TIL

api/core/workflow/runtime/graph_runtime_state.py
laipz8200: To maintain our current layered design, it would require a lot of additional work to avoid using import_module in this PR, let's complete it in the next PR.

QuantumGhost (approved):

dlmu-lq:
hello, is this feature not adapted in the frontend? @laipz8200

Ayyappa44488:
Could you please share any screenshots regarding this feature usage? @laipz8200.

laipz8200:
@dlmu-lq @Ayyappa44488 

This feature is designed to enhance the basic capabilities of the workflow engine, and it will support the subsequent HITL functionality, but we have not released this feature yet.

PR #26787, opened 2025-10-12
feat: run with params from logs

wellCh4n:
> [!IMPORTANT]
>

close #26786 

## Screenshots

| Before | After |
|--------|-------|
| ... | imageimage |

## Checklist


web/app/components/workflow-app/index.tsx
```diff
@@ -47,6 +53,29 @@ const WorkflowAppWithAdditionalContext = () => {
...
+    if (!runUrl)
+      return
+    fetchRunDetail(runUrl).then((res) => {
+      const { setInputs, setShowInputsPanel, setShowDebugAndPreviewPanel } = workflowStore.getState()
+      if (typeof res.inputs === 'object') {
+        const userInputs = Object.fromEntries(
+          Object.entries(res.inputs as Record<string, any>).filter(([key]) => !key.startsWith('sys.')),
+        )
+        setInputs(userInputs)
+        setShowInputsPanel(true)
```
lyzno1: - [P1] Guard against null inputs when replaying 
    When fetchRunDetail returns a run whose inputs field is null (happens for runs started without user parameters), typeof res.inputs === 'object' still evaluates to true,
  so we enter the block and call Object.entries(res.inputs). Because res.inputs is null, this throws TypeError: Cannot convert undefined or null to object, leaving the new
  replay flow broken for those runs and producing an unhandled rejection. Please guard with a truthy check (e.g. if (res.inputs && typeof res.inputs === 'object')) before
  calling Object.entries.

wellCh4n: I fixed this. It looks better now.

DavideDelbianco:
It would partially resolve this issue too #22507

lyzno1 (approved):

wellCh4n:
@lyzno1 Looks better, thanks!

PR #26811, opened 2025-10-13
fix: use account id in workflow app log filter

hj24:
> [!IMPORTANT]
>

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


zyssyz123 (dismissed):
LGTM

api/services/workflow_app_service.py
```diff
@@ -86,12 +86,14 @@ def get_paginate_workflow_app_logs(
                 ),
             )
         if created_by_account:
+            account = session.scalar(select(Account).where(Account.email == created_by_account))
+            account_id = account.id if account else None
```
laipz8200: It's better to abort here if you can't get an account.

laipz8200 (changes requested):

api/services/workflow_app_service.py
hj24: updated, add fast path

laipz8200: If we can't get an account, it's better to raise an error instead of returning an empty result. Because it's better to say here is an error about data that we need to check.

crazywoola (dismissed):

crazywoola (approved):

laipz8200 (approved):

api/services/workflow_app_service.py
```diff
@@ -86,12 +86,16 @@ def get_paginate_workflow_app_logs(
                 ),
             )
         if created_by_account:
+            account = session.scalar(select(Account).where(Account.email == created_by_account))
+            if not account:
+                raise ValueError(f"Account not found: {created_by_account}")
```
laipz8200, line 91: It's better to put a log here, but we can do it later.

PR #26851, opened 2025-10-13
feat(stream): added a small tool to clean up LLM stream output

DammyBot:
Hey there 👋

I added a small Python class called `StreamingHTMLStripper` that cleans up streamed LLM responses by removing extra tags like 

```html
<details><summary>Thinking...</summary>
```

. It works even when those tags are split across chunks in the stream.

This helps keep the output clean and readable while the stream is still going. I also added a quick demo script and a few tests to make sure it works as expected.

How to test:
- Run `python tools/stream_filter_demo.py` to see how it behaves
- Or run `pytest tests/test_stream_filter.py` to check the tests

Happy to adjust anything if needed 🙂

### DIFF
--- a/backend/utils/stream_filter.py
+++ b/backend/utils/stream_filter.py
@@ -1,4 +1,3 @@
-# backend/utils/stream_filter.py
 from typing import Iterator
 
 class StreamingHTMLStripper:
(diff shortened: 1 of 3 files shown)


crazywoola:
Please do not modify the PR template, you need to open an issue first and link it in the description first.

PR #26964, opened 2025-10-15
Fixed Weaviate no module found issue (issue #26938)

### LINKED ISSUE #26938: No module named 'weaviate.classes'
### Self Checks

### Dify version

1.9.1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Image

### ✔️ Expected Behavior

success

### ❌ Actual Behavior

No module named 'weaviate.classes'

DhruvGorasiya:
> [!IMPORTANT]
>

## Summary

close https://github.com/langgenius/dify/issues/26999
close https://github.com/langgenius/dify/issues/26938
## Screenshots

| Changed | Before | After |
|--------|--------|-------|
| Docker Compose Template |  No entrypoint overrides | Added entrypoint overrides for API and Worker services |
| API and worker Service Entrypoint | Default entrypoint only | Custom entrypoint that installs weaviate-client 4.17.0 |
| Weaviate Client Version | weaviate-client==3.24.2 | weaviate-client==4.17.0 |
| Weaviate Support |Weaviate 1.19.0 and below | Weaviate 1.27.0 |

## Checklist


kurokobo:
I haven't been able to test anything on my end, but here are a few comments:

- In the end, did you test whether existing user data can be migrated from 1.19 to 1.27 without any issues?
- There aren't any bind mounts added to the `volumes` section in the compose file. Is that alright? If you've confirmed that it works with the old code, then that's fine. By the way if the next version is coming out soon, maybe we don't need to update the compose file at all.
- Providing overly detailed explanations increases future maintenance costs. I mean, personally, I don't think we need a migration guide. Even if we decide to keep it, its file location and name feel a bit out of place compared to other files.

As with your previous PR, based on the code and comments, I get the impression that many of your changes seem to be generated by AI.
Using generative AI is okay, but before submitting a PR, please make sure to thoroughly review and test your changes to ensure they're appropriate, of sufficient quality for future maintenance, and free from side effects.

crazywoola (approved):

quicksandznzn:
**Concern about backward compatibility:**

I'm currently using a self-hosted Weaviate instance at version 1.19.0. If I update Dify with this PR but keep my Weaviate server at v1.19.0, will this cause compatibility issues?

From what I understand:
- This PR upgrades the Weaviate Python client from v3 to v4
- The target server version is now 1.27.0
- Client v4 may not be compatible with server v1.19.0

**If there are compatibility issues, this would be a breaking change that should be clearly documented**, including:
1. Minimum required Weaviate server version
2. Migration path for users running older Weaviate versions
3. Whether users must upgrade their Weaviate server before updating Dify

DhruvGorasiya:
@quicksandznzn You’re absolutely right, the Weaviate Python client v4 works only with Weaviate server v1.24.0 and above. Running it with an older version (like 1.19.0) will cause compatibility issues because of API and schema changes between versions.

To clarify:
	•	This PR does introduce a breaking change, so users on older Weaviate versions will need to upgrade their server to at least v1.24.0.
	•	For self-hosted setups, data might need to be re-indexed, and that can be done using Weaviate’s Cursor API.
	•	We’ll also update the Dify documentation for the Weaviate integration to include the minimum server version, migration steps, and guidance for local users.

DhruvGorasiya:
Thanks for the feedback, @kurokobo!

I’ve tested the migration, the new setup works fine with Weaviate 1.27.0 (or any version >1.24.0). Since the v4 client isn’t compatible with server 1.19.0, users will need to upgrade their Weaviate instance to 1.24+ when updating Dify. For older setups, data migration can be handled via the Cursor API or standard backup/restore tools.

Also no bind mounts are required. The container installs weaviate-client v4 at startup and works with the existing volumes. That said, if an upcoming Dify release already includes v4 in the base image, we could skip updating the compose file, what’s your preference?

Totally agree on the migration guide point, having it as a separate README doesn’t really make sense. We’ll include the migration details directly in the official Dify docs instead so everything stays in one place.

PR #27153, opened 2025-10-20
 fix Version 2.0.0-beta.2: Chat annotations Api Error #25506

Cursx:
> [!IMPORTANT]
>

## Summary

fix  #25506
I think the reasons may be as follows:
1.Management endpoints (/v1/apps/annotations) require Account context.
2.The annotation list uses @ validate_app_token (without fetch_user_arg), while the annotation service layer internally calls Current_Countw_ith_tenant(), forcing that Current_user is Account and has already loaded Current_tenant. This is the root cause of 500.
3. using @ validate_app_token (fetch_user_arg=...), it will set the current user to EndUser based on the user parameter, and the service layer does not require an Account.

GET/v1/apps/annotations originally had 500 because it required an Account, but using only @ validate_app_token would not automatically set the Account; I have fixed it to automatically log in to the owner account before calling, now it won't be 500.

## Screenshots

| Before | After |
|
微信图片_20251020152055_301_32
|
微信图片_20251017150711_299_32
|
| ... | ... |

## Checklist

### DIFF
--- a/api/controllers/service_api/app/annotation.py
+++ b/api/controllers/service_api/app/annotation.py
@@ -6,28 +6,28 @@ from flask_restx.api import HTTPStatus
 from werkzeug.exceptions import Forbidden
 
 from controllers.service_api import service_api_ns
-from controllers.service_api.wraps import validate_app_token
+from controllers.service_api.wraps import FetchUserArg, WhereisUserArg, validate_app_token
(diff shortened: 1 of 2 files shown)


Cursx:
I only analyzed the get request and did not carefully check everything else. Now it seems that I need to carefully check it

PR #27298, opened 2025-10-22
feat(api): Introduce workflow pause state management

### LINKED ISSUE #27297: State Management for workflow pause
### Self Checks

Introduce a mechanism for recording runtime state of paused workflows and their metadata.

The runtime state serialization has already been implemented in PR ##26584. The main purpose of this issue
is saving the serialized state to object storage, and associated them with `WorkflowRun` records.

_No response_

QuantumGhost:
## Summary

This PR introduces a new model `WorkflowPause` model for recording metadata associated with a paused workflow and its runtime state.

The serialized runtime states are saved on object storage, the metadata of those objects are recorded with `UploadFile` model.

Introduce tests for the state management code (Most of the tests are generated by AI, though. I have examined the tests and the test logic is roughly correct)

Resolves #27297.

## Screenshots

N/A

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


api/core/workflow/entities/pause_reason.py
```diff
@@ -0,0 +1,49 @@
+from enum import StrEnum
+from typing import Annotated, Any, ClassVar, TypeAlias
+
+from pydantic import BaseModel, Discriminator, Tag
+
+
+class _PauseReasonType(StrEnum):
+    HUMAN_INPUT_REQUIRED = "human_input_required"
+    SCHEDULED_PAUSE = "schduled_pause"
```
laipz8200: typo here.

```suggestion
    HUMAN_INPUT_REQUIRED = auto()
    SCHEDULED_PAUSE = auto()
```

api/core/app/layers/persist_layer.py
```diff
@@ -0,0 +1,72 @@
...
+        This method receives all events generated during graph execution, including:
+        - Graph lifecycle events (start, success, failure)
+        - Node execution events (start, success, failure, retry)
+        - Stream events for response nodes
+        - Container events (iteration, loop)
+
+        Args:
+            event: The event emitted by the engine
+        """
+        if isinstance(event, GraphEngineEvent):
```
laipz8200: All income events here are GraphEngineEvent.

api/core/app/layers/pause_state_persist_layer.py
laipz8200, line 1: I think we need a better file name here, for example, pause_state_persist_layer.py

api/core/workflow/graph_engine/command_processing/command_handlers.py
```diff
@@ -25,4 +27,7 @@ class PauseCommandHandler(CommandHandler):
     def handle(self, command: GraphEngineCommand, execution: GraphExecution) -> None:
         assert isinstance(command, PauseCommand)
         logger.debug("Pausing workflow %s: %s", execution.workflow_id, command.reason)
-        execution.pause(command.reason)
+        # Convert string reason to PauseReason if needed
+        reason = command.reason or "unknown reason"
```
laipz8200: The processing logic is separated from the command. Consider making 'unknown reason' the default value for command.reason?

api/core/workflow/graph_events/node.py
```diff
@@ -54,4 +55,4 @@ class NodeRunRetryEvent(NodeRunStartedEvent):
 
 
 class NodeRunPauseRequestedEvent(GraphNodeEventBase):
-    reason: str | None = Field(default=None, description="Optional pause reason")
+    reason: PauseReason = Field(..., description="Optional pause reason")
```
laipz8200:
```suggestion
    reason: PauseReason = Field(..., description="pause reason")
```

api/core/workflow/node_events/node.py
```diff
@@ -43,4 +44,4 @@ class StreamCompletedEvent(NodeEventBase):
 
 
 class PauseRequestedEvent(NodeEventBase):
-    reason: str | None = Field(default=None, description="Optional pause reason")
+    reason: PauseReason = Field(..., description="Optional pause reason")
```
laipz8200:
```suggestion
    reason: PauseReason = Field(..., description="pause reason")
```

api/extensions/ext_storage.py
```diff
@@ -85,7 +85,7 @@ def create_clickzetta_volume_storage():
             case _:
                 raise ValueError(f"unsupported storage type {storage_type}")
 
-    def save(self, filename, data):
+    def save(self, filename, data: bytes):
```
laipz8200:
```suggestion
    def save(self, filename: str, data: bytes):
```

api/models/base.py
```diff
@@ -13,3 +19,34 @@ class TypeBase(MappedAsDataclass, DeclarativeBase):
     """
 
     metadata = metadata
+
+
+class ModelMixin:
```
laipz8200:
```suggestion
class DefaultFieldsMixin:
```

api/models/workflow.py
```diff
@@ -549,6 +564,15 @@ class WorkflowRun(Base):
     finished_at: Mapped[datetime | None] = mapped_column(DateTime)
     exceptions_count: Mapped[int] = mapped_column(sa.Integer, server_default=sa.text("0"), nullable=True)
 
+    pause: Mapped[Optional["WorkflowPause"]] = orm.relationship(
```
laipz8200, line 567: I suggest avoiding the use of ORM's relationship pattern, as this may cause trouble when we switch to asyncio (possibly) in the future, handling it in the Repository might be more appropriate.

api/models/workflow.py
```diff
@@ -1583,3 +1606,86 @@ class WorkflowDraftVariableFile(Base):
...
+    # `app_id` represents the application identifier associated with this state.
+    # It corresponds to the `id` field in the `App` model.
+    #
+    # While this field is technically redundant (as the corresponding app can be
+    # determined by querying the `Workflow`), it is retained to simplify data
+    # cleanup and management processes.
+    app_id: Mapped[str] = mapped_column(
+        StringUUID,
+        nullable=False,
+    )
```
laipz8200: Since this model and WorkflowRun have a composite relationship, I think we don't need to add tenant_id and app_id here. It should always be queried through workflow_run_id.

api/models/workflow.py
```diff
@@ -1583,3 +1606,86 @@ class WorkflowDraftVariableFile(Base):
...
+        nullable=True,
+    )
+
+    # state_object_key stores the object key referencing the serialized runtime state
+    # of the `GraphEngine`. This object captures the complete execution context of the
+    # workflow at the moment it was paused, enabling accurate resumption.
+    state_object_key: Mapped[str] = mapped_column(String(length=255), nullable=False)
+
+    # Relationship to WorkflowRun
+    workflow_run: Mapped["WorkflowRun"] = orm.relationship(
```
laipz8200, line 1666: same as https://github.com/langgenius/dify/pull/27298/files#r2471819646

api/repositories/api_workflow_run_repository.py
laipz8200, line 1: Make WorkflowPause a part of WorkflowRun and process them together in the existing persistence layer. How does that sound? This way, we will be able to avoid adding methods to the Repo and also make the relationships of the data model closer.

laipz8200 (changes requested):

api/repositories/sqlalchemy_api_workflow_run_repository.py
```diff
@@ -301,6 +312,283 @@ def delete_runs_by_app(
...
+                        pause.state_object_key,
+                    )
+                except Exception:
+                    logger.exception(
+                        "Failed to delete state file for pause, pause_id=%s, object_key=%s",
+                        pause.id,
+                        pause.state_object_key,
+                    )
+                    continue
+                session.delete(pause)
```
laipz8200, line 580: Fetch and delete should in the same session here.

laipz8200 (changes requested):

api/models/workflow.py
QuantumGhost: This won't cause any problems even if we switch to asyncio, because the `pause` relationship is declared with `lazy="raise"`. This requires users of the `pause`  property explicitly preload its value in select statement, like [this](https://github.com/langgenius/dify/pull/27298/files#diff-cc321641e7f661da7a190c9424ff01f0b89751e0a8ab04c315b7de11cdf99a66R408):

```py
stmt = select(WorkflowRun).options(selectinload(WorkflowRun.pause)).where(WorkflowRun.id == workflow_run_id)
```

If the `pause` property is not explicitly loaded, accessing it would raise an exception.

This is also [recommended by SQLAlchemy](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#:~:text=If%20not%20using%20AsyncAttrs%2C%20relationships%20can%20be%20declared%20with%20lazy%3D%22raise%22%20so%20that%20by%20default%20they%20will%20not%20attempt%20to%20emit%20SQL.%20In%20order%20to%20load%20collections%2C%20eager%20loading%20would%20be%20used%20instead.)

ref: SQLAlchemy's documentation on [`lazy="raise"`](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#summary-of-relationship-loading-styles:~:text=Eager%20Loading.-,raise%20loading,-%2D%20available%20via%20lazy)

If you're worrying about misusing the `orm.relationship` attribute, maybe we can introduce a dedicated linter for this? IMO it should be possible to build such a lint rule with ast-grep / semgrep.

api/repositories/sqlalchemy_api_workflow_run_repository.py
QuantumGhost: Ideally the same session should be used for both `fetch` and `delete` here. However, doing so would making the transaction long lasting and cause additional overhead. Here I use a separate session for each deletion to ensure all transactions are short-lived.

api/models/workflow.py
QuantumGhost: Fixed

api/repositories/api_workflow_run_repository.py
QuantumGhost: This would be challenging to implement cleanly, as the API/service layers currently use the `WorkflowRun` database model directly as the domain model. Introducing `WorkflowPause` as part of the `WorkflowRun` model would create significant coupling and architectural complexity.

The current `WorkflowPauseEntity` provides good abstraction with minimal implementation detail leakage. I believe it's architecturally sound to maintain `WorkflowRun` in its current form and use a dedicated model for pause state and related metadata, which better adheres to the separation of concerns principle.

Yeuoly (approved):
LGTM

PR #27771, opened 2025-11-03
Refactor: centralize node data hydration

laipz8200:
> [!IMPORTANT]
>

## Summary

Fixes #27770

- turn `Node` into a generic base that hydrates `BaseNodeData` once and exposes typed accessors
- remove redundant `init_node_data`/accessor overrides across workflow node subclasses and adjust factory wiring
- add a pytest-only local fallback for code execution and cover the new behavior with targeted unit tests

## Screenshots

| Before | After |
|--------|-------|
| N/A | N/A |

## Checklist


QuantumGhost (commented):
After discussing with @laipz8200, we've reached a consensus that while this PR introduces complexity, it's a necessary evolution for the graph runtime architecture.

Specifically:

1. **Generic type parameters are essential** - They enable future direct construction of `Node` instances from structured `NodeData` subclasses, rather than passing unstructured `config: Mapping[str, Any]` to `Node.__init__`. This represents a significant architectural improvement.

2. **The `__init_subclass__` implementation, while complex, provides substantial benefits** - It significantly simplifies the creation of custom nodes by eliminating boilerplate code. With comprehensive documentation and adequate test coverage, the maintenance burden should be manageable.

3. **This approach represents the culmination of multiple iterations** - Each previous version encountered specific limitations that this implementation addresses, particularly around type system integrity with multi-level inheritance.

The complexity is primarily contained within the base class implementation, while node subclasses become significantly simpler and more maintainable. This represents a worthwhile trade-off between implementation complexity and developer experience.

api/tests/unit_tests/core/workflow/nodes/base/test_base_node.py
```diff
@@ -34,3 +43,84 @@ def test_ensure_subclasses_of_base_node_has_node_type_and_version_method_defined
...
+
+def test_init_subclass_raises_without_generic_or_explicit_type():
+    """A subclass must either use Node[T] or explicitly set _node_data_type."""
+    with pytest.raises(TypeError, match="must inherit from Node\\[T\\] with a BaseNodeData subtype"):
+
+        class _InvalidNode(Node):
+            pass
+
+
+def test_init_subclass_allows_explicit_node_data_type():
```
QuantumGhost: While this test case contains valid Python code that would execute successfully, it appears to violate the intended type system design. The test demonstrates bypassing the generic type parameter mechanism by explicitly setting `_node_data_type`, which undermines the type safety guarantees the implementation aims to provide.

Is there a specific reason this bypass mechanism should be allowed? If not, we should consider:

1. Removing this test case if the behavior is unintended
2. Adding validation to prevent explicit `_node_data_type` overrides when using generic parameters
3. Documenting this as an escape hatch only for specific edge cases with clear justification

The current implementation creates an inconsistency between the declared generic type and the actual runtime type, which could lead to confusing type errors in downstream code.

api/core/workflow/nodes/variable_aggregator/variable_aggregator_node.py
```diff
@@ -1,40 +1,17 @@
...
-from core.workflow.enums import ErrorStrategy, NodeType, WorkflowNodeExecutionStatus
+from core.workflow.enums import NodeType, WorkflowNodeExecutionStatus
 from core.workflow.node_events import NodeRunResult
-from core.workflow.nodes.base.entities import BaseNodeData, RetryConfig
 from core.workflow.nodes.base.node import Node
 from core.workflow.nodes.variable_aggregator.entities import VariableAssignerNodeData
 
 
-class VariableAggregatorNode(Node):
+class VariableAggregatorNode(Node[VariableAssignerNodeData]):
```
QuantumGhost, line 10: The name of `VariableAssignerNodeData` seems pretty misleading.

QuantumGhost (approved):
LGTM

api/core/workflow/nodes/variable_aggregator/variable_aggregator_node.py
laipz8200: Right, we can rename it in another PR.

PR #28126, opened 2025-11-12
fix: correct scoring in WeightRerankRunner for hybrid search

myml:
> [!IMPORTANT]
>

## Summary

Fixes #28125 

- Prevent text search scores from being mistakenly treated as vector search scores during deduplication, which caused inaccurate ranking
- Fix usage of document.vector for cosine similarity, as document does not have this field

## Screenshots

| Before | After |
|--------|-------|
| Image |
图片 |

## Checklist


api/core/rag/rerank/weight_rerank.py
```diff
@@ -1,14 +1,9 @@
 import math
 from collections import Counter
```
asukaminato0721, line 4:
```suggestion
from collections import Counter, defaultdict
```

api/core/rag/rerank/weight_rerank.py
```diff
@@ -35,35 +30,32 @@ def run(
...
-                    unique_documents.append(document)
-
-        documents = unique_documents
 
         query_scores = self._calculate_keyword_score(query, documents)
-        query_vector_scores = self._calculate_cosine(self.tenant_id, query, documents, self.weights.vector_setting)
-
+        query_vector_scores = self._get_documents_score(documents)
+
+        scoreMap = {}
```
asukaminato0721:
```suggestion
        score_map =defaultdict(float)
```

api/core/rag/rerank/weight_rerank.py
```diff
@@ -35,35 +30,32 @@ def run(
...
+        for document, q_score, q_vector_score in zip(documents, query_scores, query_vector_scores):
+            if document.provider == "dify" and document.metadata is not None and "doc_id" in document.metadata:
+                doc_id = document.metadata["doc_id"]
+                if doc_id not in scoreMap:
+                    scoreMap[doc_id] = 0
```
asukaminato0721:
```suggestion
```

api/core/rag/rerank/weight_rerank.py
```diff
@@ -35,35 +30,32 @@ def run(
...
+                scoreMap[doc_id] += self.weights.keyword_setting.keyword_weight * q_score
```
asukaminato0721:
```suggestion
                score_map[doc_id] += self.weights.keyword_setting.keyword_weight * q_score
```

api/core/rag/rerank/weight_rerank.py
```diff
@@ -35,35 +30,32 @@ def run(
...
+                scoreMap[doc_id] += self.weights.vector_setting.vector_weight * q_vector_score
```
asukaminato0721:
```suggestion
                score_map[doc_id] += self.weights.vector_setting.vector_weight * q_vector_score
```

api/core/rag/rerank/weight_rerank.py
```diff
@@ -35,35 +30,32 @@ def run(
...
-                + self.weights.keyword_setting.keyword_weight * query_score
-            )
-            if score_threshold and score < score_threshold:
-                continue
-            if document.metadata is not None:
-                document.metadata["score"] = score
+        for document in documents:
+            if document.provider == "dify" and document.metadata is not None and "doc_id" in document.metadata:
+                doc_id = document.metadata["doc_id"]
+                score = scoreMap[doc_id]
```
asukaminato0721:
```suggestion
                score = score_map[doc_id]
```

crazywoola:
Thanks for the PR, please resolve the lint errors :)

JohnJyong:
The root cause is that the full-text search score was incorrectly used when calculating the similarity distance. Therefore, we should simply ignore the full-text search score, since we will calculate the full-text similarity ourselves during weight-rerank. @myml

myml:
> The root cause is that the full-text search score was incorrectly used when calculating the similarity distance. Therefore, we should simply ignore the full-text search score, since we will calculate the full-text similarity ourselves during weight-rerank. @myml

You raise a valid point. When we deduplicate documents from both full-text and vector search results before weight reranking, simply ignoring the full-text search scores could indeed cause some documents to lose their vector search scores. This happens because documents from the full-text search won't have vector scores, and those from vector search would lose their original scores if we discard them completely. 

So I've also modified weight_rerank.py. Now it will sum the vector score and the full-text score before deduplication and sorting. If a particular ID only exists in the full-text search results, it will only receive the full-text similarity score.

PR #28306, opened 2025-11-18
feat: add inputs parameter to MCP tool invocation methods

zinodynn:
> [!IMPORTANT]
>

## Summary

- **feat**: Forward chat input variables to MCP server via `_meta`
  - Enables passing additional variables from chat request `inputs` to MCP servers, facilitating use cases like identity verification and permission control.
  
- **feat**: Support MCP protocol-reserved `_meta` parameter in tool invocations
  - Adds the `_meta` field to `CallToolRequestParams` to comply with the MCP specification, enabling transmission of metadata such as `progressToken` or `identityToken`.

These changes improve flexibility for MCP-level business logic, allowing for more fine-grained control and context-aware processing based on upstream chat request details.

**References**:
- [MCP Spec - General Fields](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/index.mdx#general-fields)
- [#28076](https://github.com/langgenius/dify/discussions/28076), [#28167](https://github.com/langgenius/dify/discussions/28167)

## Screenshots

| Before | After |
|--------|-------|
| sql execution in MCP cannot be authenticated | The functions can be expanded by customizing parameters |
||32765fe0b941c70ffb5cad795c14777|

## Checklist


zinodynn (commented):
format code

zinodynn:
Reference my approach to pass input parameters from chat requests to MCP tools, as outlined below:

1. Create an `identityToken` using the app's `input` parameters, then save and publish it.  
image


2. Modify the MCP server code, and receive the `identityToken` params:
```typescript
export async function executeSqlToolHandler(
  { sql }: { sql: string }, 
  _extra: any // Additional parameters from chat request
) {
  const connector = ConnectorManager.getCurrentConnector();
  // Retrieve `input` params from the request via:
  // const identityToken = extra._meta?.identityToken;   <---  receive params
  try {
    // ... Use the `sql` parameter ...
  }
  // ...
}
```

3. Send a chat request (example curl command):
```bash
curl --location 'http://localhost:5001/v1/chat-messages' \
--header 'Authorization: Bearer app-xxxxxx' \
--data '{
    "inputs": {"identityToken":"eyJhbG...."},
    "query": "Query today'\''s sales orders?",
    "response_mode": "streaming",
    "user": "abc-123"
}'
```

zinodynn:
@Nov1c444, to prevent every MCP from receiving the parameters, would it be a good idea to distinguish parameters in `inputs` using dot-separated notation (e.g., `.`)? For example, create a parameter like `mysql-mcp-server.identityToken` in `inputs` to pass it specifically to the `mysql-mcp-server` MCP. This way, other MCPs won’t receive the `identityToken` parameter. If this makes sense, I’ll proceed to work on it.

Nov1c444:
Hey there! 👋
I understand this PR aims to solve a core problem: allowing MCP tools to receive request-level dynamic context information, including:
- User identity information (roles, IDs, etc.) for business logic differentiation (ref #28167)
- Authentication credentials (access tokens, etc.) for MCP server authentication (ref #28076)
The current implementation passes this data to MCP servers as `_meta` via the inputs parameter, which indeed solves the problem in Agent Apps 👍
However, there are still a few issues we need to optimize:
1. Workflow/Chatflow cannot use the same mechanism
The current solution only works in Agent Apps. If users want to invoke MCP tools based on user roles in Workflows, or pass dynamic tokens, there's no corresponding solution. The same MCP tool has inconsistent capabilities across different application types, which creates a fragmented user experience.
2. Using reflection mechanism lacks clarity
The base class uses `inspect.signature()` to detect if subclasses accept the inputs parameter:
  ```python
   sig = inspect.signature(self._invoke)   
   if "inputs" in sig.parameters:       
      invoke_kwargs["inputs"] = inputs
   ```
While this works, the interface contract is unclear, IDEs can't provide type hints, and it's not very developer-friendly for tool authors.
3. Cannot directly pass to HTTP Headers (for #28076)
Currently `_meta` is passed in the request body. If users want to put tokens in HTTP request headers (like IDToken: xxx), they still need to manually handle it on the MCP server side, which isn't straightforward.
I need to think through the specific implementation approach and will update later. Welcome everyone to discuss.

Nov1c444:
I have some thoughts about this requirement. Add a meta configuration to the MCP Server, like this. Does this design solve the issue? If so, we can discuss more details.:

1. When clicking the MCP tool settings, you could see two tabs: Parameters and Meta
2. Parameters are LLM-generated, while Meta is user-defined
3. You could add some meta key-value pairs in the Meta tab. The meta values could be variables that you define yourself, and they will be passed into the MCP meta field.
4. Agent apps, workflows, and chatflows would work similarly.
image

| Before  | After |
| --- | --- |
| image | image |

zinodynn:
> * When clicking the MCP tool settings, you could see two tabs: Parameters and Meta
> * Parameters are LLM-generated, while Meta is user-defined
> * You could add some meta key-value pairs in the Meta tab. The meta values could be variables that you define yourself, and they will be passed into the MCP meta field.
> * Agent apps, workflows, and chatflows would work similarly.

Totally agree with the new changes you proposed! Honestly, I didn’t realize the specificity of the `inputs` parameter in chat requests before. I kind of had a gut feeling that MCP-related configurations should be set in the MCP tool’s own properties, but I still went with `inputs` at the time—since I thought this parameter persists throughout the entire chat flow, making it easier to fetch values.
If we go with the Meta tab design you mentioned, I’m curious how we’d pass these values to the actual HTTP requests? From what I know, request parameters often get filtered. For example, would something like IDkey still need to be included in `inputs` to be sent in the request?

Nov1c444:
I mean, you could pass the IDKey variable through the inputs, like this:image
image

zinodynn:
So does that mean we’ll still pass parameters via inputs, but the configuration and metadata will be handled differently? @Nov1c444

Nov1c444:
> So does that mean we’ll still pass parameters via inputs, but the configuration and metadata will be handled differently? @Nov1c444

Yeah, it's the best way I could think of to pass the variable in the Agent App's MCP Tool. Do you have any better ideas?

zinodynn:
@Nov1c444 Hi , Here are my understandings and follow-up questions:  

1. If we stick with passing parameters via `inputs`, after configuring the key in MCP Meta, will the Workflow fetch the specific value from `inputs` using that key? Due to Dify's parameter filtering mechanism, it seems like this would require two configurations: first, adding the corresponding field in `inputs`, and second, creating a field in MCP Meta where the key matches the one in `inputs` (the value can either be passed from `inputs` or set as a default). Is that correct?  

2. For passing parameters like `IDKey` via HTTP headers—what would the workflow look like? Do we just need to create a field in MCP Meta? Will Dify filter out these header parameters, or will they be forwarded directly to the MCP server?

zinodynn:
image

In this example, the `Field value` should be `inputs.IDkey`. right?   `id_key` hasn't been defined.

zinodynn:
Headers and _meta are two separate channels: headers work at the HTTP layer, while _meta operates at the MCP protocol layer. I’m wondering if defining something like `IDkey` in headers is appropriate—after all, in the current code, headers are mainly used for OAuth authentication. I think it’s better to keep it within the MCP protocol layer, i.e., defining it in `inputs`. Looking forward to everyone's thoughts/discussions!

cion-J:
If I make a request through the API, such as `/workflows/run`, and customize the header in the request, can it be attached to the MCP request? `inputs` will disrupt the parameter model of the mcp service.

YQP-2511:
Is it the difference between active reception and passive reception parameters of the MCP service? In Field VALUE, we can use / to select parameters defined in AGENT/CHATFLOW/WORKFLOW just like in INSTRUCTIONS. Are these applications required to use that MCP in order to select through / in the _meta of that MCP?

cion-J:
The situation I encountered is that there is no need to configure or adjust each MCP individually. If this global _meta or variable has been determined at the beginning, it can be automatically passed down to each MCP, just like tracking. This will reduce the workload when configuring multiple MCPs.

tvp100:
I need this feature to set meta data, so I can get it from McpMeta.
this is my java code, use spring ai mcp
```java
    @McpTool(name = "getAreaTree", description = "获取区域相关的信息, 能管的设备分4个, 电、水、气、秤, 返回值的areaIdChain表示从顶级区域开始到当前区域的ID链, 用'/'分割")
    public List<AreaTreeNodeDto> getAreaTree(McpSyncRequestContext context, AiListAreaTreeParam param, McpMeta meta) {
        return remoteSystemOpenService.getAreaTree(param).getData();
    }
```

and this is another tool which has the feature I need.
image


is there anyway I cloud pass it through Agent in dify? or will this feature be merged?



> image
> In this example, the `Field value` should be `inputs.IDkey`. right? `id_key` hasn't been defined.

zxy665:
Hello, according to your solution, how can an MCP server retrieve parameters from _meta? @zinodynn

CrabSAMA:
@Nov1c444 hello, this feature is so important with me, in actully work we need transfer header to mcp dynamic, can you help me to push this pr be merged?
Thanks a lot!

crazywoola:
Hi @zinodynn, thank you for taking the time to contribute to Dify.

### Why this is being closed

This PR was opened by a first-time contributor before February 2026 and does not link to an associated issue. Per our contribution workflow, PRs should have a related issue opened and linked before review so the scope, expected behavior, and ownership are clear.

### Next steps

Please open an issue first, or link an existing issue that describes the problem or feature this PR addresses. After the issue is confirmed, you are welcome to open a fresh PR that references it. Thanks again for your contribution and understanding.

PR #28188, opened 2025-11-13
feat:mysql adaptation for metadb

longbingljw:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

This pull request introduces multi-database support (PostgreSQL, MySQL, and OceanBase), improves database configuration flexibility, and refactors several SQL query implementations for better compatibility and maintainability. It also updates environment and workflow files to support these changes, and refactors metadata filtering logic for datasets.

**Database support and configuration:**

* Added support for selecting the database type (`DB_TYPE`) in `DatabaseConfig` and `.env.example`, enabling PostgreSQL, MySQL, and OceanBase as backend options. The SQLAlchemy URI scheme and engine options now adjust dynamically based on the selected database type. [[1]](diffhunk://#diff-0aeb1a877a7ce604c8cca16fff2336c5d8297b313ac42832af471f210592fc5bR108-R113) [[2]](diffhunk://#diff-0aeb1a877a7ce604c8cca16fff2336c5d8297b313ac42832af471f210592fc5bL143-R152) [[3]](diffhunk://#diff-0aeb1a877a7ce604c8cca16fff2336c5d8297b313ac42832af471f210592fc5bL207-L214) [[4]](diffhunk://#diff-2fdd31c8eaeb05db626d15ffa6f31b34be3044212e22d6831e976afbab90cc86L75-R83)
* Updated documentation and example environment files to clarify database configuration and added/removal of OceanBase vector settings. [[1]](diffhunk://#diff-c7b1c23e00a1fd0378f6ea41a7dfe1c9e043d766f738dba26860c783b82e8c6aL18-R19) [[2]](diffhunk://#diff-2fdd31c8eaeb05db626d15ffa6f31b34be3044212e22d6831e976afbab90cc86R180-R190) [[3]](diffhunk://#diff-2fdd31c8eaeb05db626d15ffa6f31b34be3044212e22d6831e976afbab90cc86L342-L350)

**CI/CD and workflow changes:**

[...]

PR #28351, opened 2025-11-18
feat: add new table of end user oauth

CourTeous33:
> [!IMPORTANT]
>

## Summary

Added a new table for end user auth:

id: Unique identifier for each authentication entry (auto-generated)
name: A human-readable name for the credentials (defaults to "API KEY 1")
tenant_id: Which tenant/organization this belongs to
end_user_id: Which specific end user owns these credentials
provider: The name of the tool/service provider (e.g., "github", "google", etc.)
encrypted_credentials: The actual credentials stored in encrypted JSON format
credential_type: What kind of authentication this is (defaults to "api-key", but could be "oauth2", etc.)
created_at: When the credentials were first added
updated_at: When they were last modified
expires_at: When the credentials expire (defaults to -1, meaning no expiration)

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


api/models/tools.py
```diff
@@ -114,6 +114,59 @@ def credentials(self) -> dict[str, Any]:
...
+    __tablename__ = "tool_enduser_authentication_providers"
+    __table_args__ = (
+        sa.PrimaryKeyConstraint("id", name="tool_enduser_authentication_provider_pkey"),
+        sa.UniqueConstraint("tenant_id", "provider", "end_user_id", "name", name="unique_enduser_authentication_provider"),
+        sa.Index("tool_enduser_authentication_provider_tenant_id_idx", "tenant_id"),
+        sa.Index("tool_enduser_authentication_provider_end_user_id_idx", "end_user_id"),
+    )
+
+    # id of the authentication provider
+    id: Mapped[str] = mapped_column(StringUUID, server_default=sa.text("uuid_generate_v4()"), init=False)
```
laipz8200: uuidv7 should be used here to avoid using server_default.

api/models/tools.py
```diff
@@ -114,6 +114,59 @@ def credentials(self) -> dict[str, Any]:
...
 
+class EndUserAuthenticationProvider(TypeBase):
+    """
+    This table stores the authentication credentials for end users in tools.
+    Mimics the BuiltinToolProvider structure but for end users instead of tenants.
+    """
+
+    __tablename__ = "tool_enduser_authentication_providers"
+    __table_args__ = (
+        sa.PrimaryKeyConstraint("id", name="tool_enduser_authentication_provider_pkey"),
```
laipz8200: Use primary_key=True below to avoid specifying the name of the primary key yourself.

api/models/tools.py
```diff
@@ -114,6 +114,59 @@ def credentials(self) -> dict[str, Any]:
...
+        sa.Index("tool_enduser_authentication_provider_tenant_id_idx", "tenant_id"),
+        sa.Index("tool_enduser_authentication_provider_end_user_id_idx", "end_user_id"),
+    )
+
+    # id of the authentication provider
+    id: Mapped[str] = mapped_column(StringUUID, server_default=sa.text("uuid_generate_v4()"), init=False)
+    name: Mapped[str] = mapped_column(
+        String(256),
+        nullable=False,
+        server_default=sa.text("'API KEY 1'"),
```
laipz8200: Avoid hard coding string; if necessary, change it to a constant.

api/models/tools.py
```diff
@@ -114,6 +114,59 @@ def credentials(self) -> dict[str, Any]:
...
+        default="API KEY 1",
+    )
+    # id of the tenant
+    tenant_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    # id of the end user
+    end_user_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    # name of the tool provider
+    provider: Mapped[str] = mapped_column(String(256), nullable=False)
+    # encrypted credentials for the end user
+    encrypted_credentials: Mapped[str | None] = mapped_column(sa.Text, nullable=True, default=None)
```
laipz8200: Can an empty string be used as a default value to avoid optional?

api/models/tools.py
```diff
@@ -114,6 +114,59 @@ def credentials(self) -> dict[str, Any]:
...
+        nullable=False,
+        server_default=sa.text("'API KEY 1'"),
+        default="API KEY 1",
+    )
+    # id of the tenant
+    tenant_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    # id of the end user
+    end_user_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    # name of the tool provider
+    provider: Mapped[str] = mapped_column(String(256), nullable=False)
```
laipz8200: Is there a specific reason to limit the length of this field to 256? If not, text should be applicable here.

api/models/tools.py
```diff
@@ -114,6 +114,59 @@ def credentials(self) -> dict[str, Any]:
...
+    """
+    This table stores the authentication credentials for end users in tools.
+    Mimics the BuiltinToolProvider structure but for end users instead of tenants.
+    """
+
+    __tablename__ = "tool_enduser_authentication_providers"
+    __table_args__ = (
+        sa.PrimaryKeyConstraint("id", name="tool_enduser_authentication_provider_pkey"),
+        sa.UniqueConstraint("tenant_id", "provider", "end_user_id", "name", name="unique_enduser_authentication_provider"),
+        sa.Index("tool_enduser_authentication_provider_tenant_id_idx", "tenant_id"),
```
laipz8200: Similarly, please use index=True below to automatically create the index and avoid manually specifying the name.

laipz8200 (changes requested):

api/models/tools.py
CourTeous33: Fixed!

api/models/tools.py
CourTeous33: Fixed!

api/models/tools.py
CourTeous33: This part is from BuiltinToolProvider() (line 84), should I change the function there as well?

api/models/tools.py
CourTeous33: Changed type to sa.Text

api/models/tools.py
CourTeous33: Added index=Ture for tenant_id and end_user_id to speed up queries.

api/models/tools.py
CourTeous33: Fixed!

api/models/tools.py
```diff
@@ -114,6 +115,55 @@ def credentials(self) -> dict[str, Any]:
...
+    """
+
+    __tablename__ = "tool_enduser_authentication_providers"
+    __table_args__ = (
+        sa.UniqueConstraint("tenant_id", "provider", "end_user_id", "name", name="unique_enduser_authentication_provider"),
+    )
+
+    # id of the authentication provider
+    id: Mapped[str] = mapped_column(StringUUID, primary_key=True, default=uuidv7, init=False)
+    name: Mapped[str] = mapped_column(
```
QuantumGhost, line 127: IMO, enforcing name uniqueness for a `(provider, end_user_id)` pair creates more overhead than value.

Ideally, we should rely on IDs as the primary identifier and treat names merely as user-facing hints. Therefore, strict uniqueness shouldn't be required. If a user creates duplicate names, they should bear the responsibility for any ambiguity.

Enforcing name uniqueness has caused implementation issues in the past, such as with credential management in the EE version. Furthermore, it complicates name generation.

That said, I suggest discussing this with the PM. I’ve skimmed the PRD and found no explicit requirement for name uniqueness.

api/models/tools.py
```diff
@@ -114,6 +115,55 @@ def credentials(self) -> dict[str, Any]:
...
+        String(256),
+        nullable=False,
+        default="API KEY 1",
+    )
+    # id of the tenant
+    tenant_id: Mapped[str] = mapped_column(StringUUID, nullable=False, index=True)
+    # id of the end user
+    end_user_id: Mapped[str] = mapped_column(StringUUID, nullable=False, index=True)
+    # name of the tool provider
+    provider: Mapped[str] = mapped_column(sa.Text, nullable=False)
```
QuantumGhost: You should use `models.types.LongText` for MySQL compatibility.

check the PR as an example: https://github.com/langgenius/dify/pull/28188/files#diff-218a854b432725d374615d6c2083936fa2a3d7d3b3f406a4cf4a5f4a57dcea0d.

Besides, I'm not sure whether `LONGTEXT` is necessary for the identifier of provider? The content of `provider` column is something like `langgenius/openai:0.2.6@e2665624a156f52160927bceac9e169bd7e5ae6b936ae82575e14c90af390e6e`, I guess.

api/models/tools.py
```diff
@@ -114,6 +115,55 @@ def credentials(self) -> dict[str, Any]:
...
+    # encrypted credentials for the end user
+    encrypted_credentials: Mapped[str] = mapped_column(sa.Text, nullable=False, default="")
```
QuantumGhost: ditto

api/models/tools.py
```diff
@@ -114,6 +115,55 @@ def credentials(self) -> dict[str, Any]:
...
+    )
+    updated_at: Mapped[datetime] = mapped_column(
+        sa.DateTime,
+        nullable=False,
+        default=datetime.now,
+        onupdate=func.current_timestamp(),
+        init=False,
+    )
+    # credential type, e.g., "api-key", "oauth2"
+    credential_type: Mapped[str] = mapped_column(
```
QuantumGhost: Defining an `StrEnum` for `credential_type` and use `EnumText` for column type would improve readability.

api/models/tools.py
```diff
@@ -114,6 +115,55 @@ def credentials(self) -> dict[str, Any]:
...
+        String(32), nullable=False, default="api-key"
+    )
+    expires_at: Mapped[int] = mapped_column(sa.BigInteger, nullable=False, default=-1)
```
QuantumGhost, line 155: I assume `expires_at` is an absolute timestamp (e.g., `2026-01-01T00:00:00+08:00`) rather than a relative duration (e.g., 3600 seconds). If so, consider using `sa.DateTime` as the column type for clarity.

api/models/tools.py
```diff
@@ -114,6 +115,55 @@ def credentials(self) -> dict[str, Any]:
...
 
+class EndUserAuthenticationProvider(TypeBase):
+    """
+    This table stores the authentication credentials for end users in tools.
+    Mimics the BuiltinToolProvider structure but for end users instead of tenants.
+    """
+
+    __tablename__ = "tool_enduser_authentication_providers"
+    __table_args__ = (
+        sa.UniqueConstraint("tenant_id", "provider", "end_user_id", "name", name="unique_enduser_authentication_provider"),
```
QuantumGhost: How about changing the unique index to `(end_user_id, provider, name)`? (Assuming we still want the name to be unique for a specific `(end_user_id, provider)` pair.)

Since the `EndUser` model is already bound to a specific `Tenant`, the query will likely imply the tenant context. Therefore, including `tenant_id` in the index doesn't add any extra selectivity.

Furthermore, the selectivity of `end_user_id` is likely higher than that of `provider`. Moving `end_user_id` to the left would improve the index's effectiveness.

Note that if there are query patterns filtering **only** by `tenant_id` and `provider`, the adjustment above might not apply. (In that case, we could use multiple indexes: one on `(end_user_id, provider, name)` and another on `(tenant_id, provider)`.)

If enforcing name uniqueness is not necessary, I suggest simply using a standard `sa.Index(end_user_id, provider, name)`.

Last but not least, the `name` parameter should be removed, as SQLAlchemy will automatically generate a proper name for the index.

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "tool_enduser_authentication_providers",
+        sa.Column(
+            "id",
+            models.types.StringUUID(),
+            server_default=sa.text("uuid_generate_v4()"),
```
QuantumGhost: This would break MySQL support.

QuantumGhost (changes requested):

api/models/tools.py
QuantumGhost: LOL ;) we have different ideas on this.

api/models/tools.py
CourTeous33: I agree the tenant_id should be removed from the unique constraint.

For the uniqueness, the PRD mentioned that 1 enduser can have multiple oauth tokens from the one provider. That means we do **not** want to be unique for a specific (end_user_id, provider) pair. In this case, I think the unique ness can bound to `(end_user_id, provider, name)` group? 

For `sa.Index(end_user_id, provider, name)`, should I add `Index=True` in these field instead (as @laipz8200 suggested above)

Other suggestions has been applied

api/models/tools.py
CourTeous33: Agreed, should we remove the unique constraint at all? Since ID is enough for uniqueness

api/models/tools.py
CourTeous33: Fixed!

api/models/tools.py
CourTeous33: Fixed!

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
CourTeous33: Will edit the script at once after all comments got resolved at the new provider

api/models/tools.py
CourTeous33: I grabbed this from BuiltinToolPlugin. Looked like the exsiting code handles expire_at as an integer (-1 represents never expire.) So I think the type as BigInteger should be fine

api/models/tools.py
CourTeous33: Re-used exsiting enum CredentialType

api/models/tools.py
QuantumGhost: > For `sa.Index(end_user_id, provider, name)`, should I add `Index=True` in these field instead (as @laipz8200 suggested above)

Adding `index=True` to those columns individually won't create a composite index. IMO, a composite index is more efficient than separate indexes here. (That said, it depends on your query patterns—is there any scenario where you need to query `tool_enduser_authentication_providers` without filtering by `end_user_id`?)

api/models/tools.py
QuantumGhost: I suggest adding a comment specifying that `expires_at` is a timestamp, including details about its epoch and resolution.

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
```diff
@@ -0,0 +1,91 @@
...
+        sa.Column(
+            "name",
+            sa.String(length=256),
+            server_default="API KEY 1",
+            nullable=False,
+        ),
+        sa.Column("tenant_id", models.types.StringUUID(), nullable=False),
+        sa.Column("end_user_id", models.types.StringUUID(), nullable=False),
+        sa.Column("provider", sa.Text(), nullable=False),
+        sa.Column("encrypted_credentials", sa.Text(), server_default="", nullable=False),
```
laipz8200: I suppose to remove `server_default` here.

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
```diff
@@ -0,0 +1,91 @@
...
+        sa.Column(
+            "created_at",
+            sa.DateTime(),
+            server_default=sa.text("CURRENT_TIMESTAMP(0)"),
```
laipz8200: Please avoid using the pg specifically grammar like `CURRENT_TIMESTAMP(0)`

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
```diff
@@ -0,0 +1,91 @@
...
+            nullable=False,
+        ),
+        sa.Column(
+            "updated_at",
+            sa.DateTime(),
+            server_default=sa.text("CURRENT_TIMESTAMP(0)"),
```
laipz8200: ditto

laipz8200 (changes requested):

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
CourTeous33: Fixed!

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
CourTeous33: Fixed!

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
CourTeous33: Fixed!

api/models/tools.py
CourTeous33: Added!

Stream29 (approved):

api/models/tools.py
```diff
@@ -109,6 +111,59 @@ def credentials(self) -> dict[str, Any]:
...
 
+class EndUserAuthenticationProvider(TypeBase):
+    """
+    This table stores the authentication credentials for end users in tools.
+    Mimics the BuiltinToolProvider structure but for end users instead of tenants.
+    """
+
+    __tablename__ = "tool_enduser_authentication_providers"
+    __table_args__ = (
+        sa.UniqueConstraint("end_user_id", "provider"),
```
QuantumGhost: Should this be changed to `sa.Index("end_user_id", "provider")`?

api/models/tools.py
```diff
@@ -109,6 +111,59 @@ def credentials(self) -> dict[str, Any]:
...
+    id: Mapped[str] = mapped_column(StringUUID, primary_key=True, default=lambda: str(uuid4()), init=False)
+    name: Mapped[str] = mapped_column(
+        String(256),
+        nullable=False,
+        default="API KEY 1",
+    )
+    # id of the tenant
+    tenant_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    # id of the end user
+    end_user_id: Mapped[str] = mapped_column(StringUUID, nullable=False, index=True)
```
QuantumGhost: This index is already covered by the above index `(end_user_id, provider)` .

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
```diff
@@ -0,0 +1,91 @@
...
+        "tool_enduser_authentication_providers",
+        ["provider"],
+        unique=False,
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index(
```
QuantumGhost, line 80: Dropping the `tool_enduser_authentication_providers` table also drops indexes associated with it, so I suggest removing the two `op.drop_index` calls.

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
```diff
@@ -0,0 +1,91 @@
...
+        sa.Column("expires_at", sa.BigInteger(), server_default=sa.text("-1"), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint(
+            "end_user_id",
+            "provider",
+        ),
+    )
+    op.create_index(
+        op.f("ix_tool_enduser_authentication_providers_end_user_id"),
+        "tool_enduser_authentication_providers",
```
QuantumGhost, line 65: This index is already covered by the above index `(end_user_id, provider)`.

QuantumGhost (approved):
Other changes LGTM.

api/models/tools.py
CourTeous33: Fixed! Removed both index statement

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
CourTeous33: Removed both drop index

api/migrations/versions/2025_11_18_1400-a7b4e8f2c9d1_add_enduser_authentication_provider.py
CourTeous33: removed both create index statement!

PR #28704, opened 2025-11-26
Add the return_direct capability to tools: supports skipping the LLM and directly returning the final answer

Cursx:
> [!IMPORTANT]
>

## Summary

After the tool is called, Dify's Agent will inject the tool response as an Observation into LLM, which will then output the final answer; When the tool returns excessively long text/JSON (such as ECharts configuration), it will significantly increase context load and latency, for example, my workflow may involve Echart charts for 2000 sets of data
Learn from Langchain to enable agents to skip LLM when calling certain tools.

-Add a return signal (strictly Boolean True) to pass from the tool message to the Agent Runner.
-FC Runner adopts the strategy of "short circuit only when all tools return directly": when all tools in the same round are return-direct=True, after executing all tools, aggregate the text as the final answer and return directly; Otherwise, continue back to LLM. Framework similar to Langgraph.

Provided some screenshots of the tests

## Checklist


Cursx:
I am currently conducting more tests

PR #29004, opened 2025-12-02
Refactor apps service toward TanStack Query

lyzno1:
## Summary
- add TanStack Query hooks for apps domain (lists, stats, voices, API keys) with stable query keys and invalidation helpers
- replace SWR usage in app surfaces with the new hooks (overview charts, apps list/nav, plugin app selector, account page, TTS voice config, demo)
- keep promise-based service helpers and drop leftover SWR fetcher typing


zhsama (approved):
LGTM

PR #29028, opened 2025-12-02
refactor/marketplace react query

lyzno1:
**Summary**
- wrap marketplace collection/plugin fetches with React Query (keep X-Dify-Version + no-store, support abort signals, safe empty fallback)
- migrate marketplace search to useInfiniteQuery with cached pages, proper loading flags, and defer rendering until data arrives
- update data source/model settings hooks to reuse cached pinned collections; keep UI loading only for initial fetch


zhsama (approved):

zhsama:
ref #29087

PR #29049, opened 2025-12-03
integrate Amplitude analytics into the application

CodingOnStar:
integrate Amplitude analytics into the application

### DIFF
--- a/web/app/components/base/amplitude/AmplitudeProvider.tsx
+++ b/web/app/components/base/amplitude/AmplitudeProvider.tsx
@@ -1,0 +1,42 @@
+'use client'
+
+import type { FC } from 'react'
+import React, { useEffect } from 'react'
+import * as amplitude from '@amplitude/analytics-browser'
+
+export type IAmplitudeProps = {
+  apiKey?: string
+}
+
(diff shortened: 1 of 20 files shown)


web/app/components/base/amplitude/AmplitudeProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const AmplitudeProvider: FC<IAmplitudeProps> = ({
+  apiKey = process.env.NEXT_PUBLIC_AMPLITUDE_API_KEY ?? '',
+  sessionReplaySampleRate = 1,
+}) => {
+  useEffect(() => {
+    // Only enable in non-CE edition
+    if (IS_CE_EDITION) {
+      console.warn('[Amplitude] Amplitude is disabled in CE edition')
```
iamjoel: There is not necessary to warning. Just Remove it. 
And the if condition should change to if it's not SaaS, then return.

iamjoel: Only enable in SaaS.

iamjoel (approved):

PR #29115, opened 2025-12-04
Feat/support multimodal embedding

JohnJyong:
## Summary

Why Build a Multimodal Knowledge Base

As enterprises increasingly rely on internal knowledge systems, the need to retrieve information from large volumes of heterogeneous files continues to grow. These materials often span multiple formats, including text, images, documents, videos, and audio—for example: product photos, illustrated manuals, or reports containing mixed text and graphics.

Traditional embedding models can vectorize certain types of data, but they are usually limited to a single modality. This limitation forces organizations into one of two sub-optimal solutions:
	•	Building complex cross-modality pipelines, embedding each modality separately and then manually fusing results;
	•	Restricting applications to a single modality, leaving most of the data’s value untapped.

Furthermore, for content that naturally includes multiple modalities—such as documents containing both text and images—traditional models struggle to capture the deep relationships between modalities, resulting in incomplete understanding.

For these reasons, multimodal embeddings have become essential for enterprises seeking to enhance data comprehension, unify data processing workflows, and overcome the constraints of single-modality systems.

⸻

Supported Multimodal Embedding & Rerank Models (First Release)

AWS Bedrock
	1.	nova-2-multimodal-embeddings-v1:0

Google Vertex AI
	1.	multimodalembedding@001

Jina
	1.	jina-embedding-v4
	2.	jina-clip-v1
	3.	jina-clip-v2
	4.	jina-reranker-m0 (rerank model)

Tongyi (Alibaba Cloud)
	1.	multimodal-embedding-v1

## Screenshots

image

image

image

image

image

## 

PR for plugin: langgenius/dify-plugin-daemon#503
PR for SDK: langgenius/dify-plugin-sdks#237

## Checklist


docker/.env.example
crazywoola, line 1: Missing `IMAGE_FILE_BATCH_LIMIT`

api/.env.example
crazywoola, line 1: Missing `IMAGE_FILE_BATCH_LIMIT`

crazywoola (commented):
See comments

api/models/dataset.py
```diff
@@ -866,6 +867,47 @@ def get_sign_content(self) -> str:
 
         return text
 
+    @property
+    def attachments(self) -> list[dict[str, Any]]:
```
QuantumGhost, line 871: I suggest avoiding wrapping database IO inside Python `@property`. This would make it really easy to write code leading to [N+1 queries](https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping). For example, without reading the implementation of `attachments`, a contributor or a new employee may write the following code:

```py
for doc in documents:
    process_attachments(doc.attachments)
```

which is highly inefficient compared to batch loading.

One alternative approach is to preload required attachments with SQLAlchemy's [`relationship`](https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html) feature. If this is not applicable (E.G. the attachments should only be loaded in certain circumstances), I would suggest changing the `attachments` property to a function, and rename it to `load_attachments` to signal the user that this function loads data from databases.

For ORM relationship, I suggest setting [`lazy="raise"`](https://docs.sqlalchemy.org/en/20/orm/relationship_api.html#sqlalchemy.orm.relationship.params.lazy) to avoid the N+1 problems above.

A even better approach is providing a batch loading function with the following signature:

```py
@classmethod
def batch_load_attachments_for_document(cls, documents: Sequence[Document]): ...
```

and use it to load attachments if needed. However, this may require significant code changes.

api/models/dataset.py
```diff
@@ -963,6 +1005,38 @@ class DatasetQuery(TypeBase):
...
     )
 
+    @property
+    def queries(self) -> list[dict[str, Any]]:
+        try:
+            queries = json.loads(self.content)
+            if isinstance(queries, list):
+                for query in queries:
+                    if query["content_type"] == QueryType.IMAGE_QUERY:
+                        file_info = db.session.query(UploadFile).filter_by(id=query["content"]).first()
```
QuantumGhost, line 1015: Here is an example of N+1 query. Consider batch it to improve performance.

api/models/dataset.py
```diff
@@ -1470,3 +1544,22 @@ class PipelineRecommendedPlugin(TypeBase):
...
+    __tablename__ = "segment_attachment_bindings"
+    __table_args__ = (
+        sa.PrimaryKeyConstraint("id", name="segment_attachment_binding_pkey"),
+        sa.Index("segment_attachment_binding_tenant_idx", "tenant_id"),
+        sa.Index("segment_attachment_binding_dataset_idx", "dataset_id"),
+        sa.Index("segment_attachment_binding_document_idx", "document_id"),
+        sa.Index("segment_attachment_binding_segment_idx", "segment_id"),
+        sa.Index("segment_attachment_binding_attachment_idx", "attachment_id"),
+    )
+    id: Mapped[str] = mapped_column(StringUUID, server_default=sa.text("uuid_generate_v4()"))
```
QuantumGhost: I suggest use `uuidv7` as the generation algorithm for the `id` field.
 
Besides, since we have started supporting MySQL since [v1.10.1](https://github.com/langgenius/dify/releases/tag/1.10.1) , the `server_default` argument should be removed, and id should be generated inside Python code.

api/models/dataset.py
```diff
@@ -1470,3 +1544,22 @@ class PipelineRecommendedPlugin(TypeBase):
         onupdate=func.current_timestamp(),
         init=False,
     )
+
+
+class SegmentAttachmentBinding(Base):
+    __tablename__ = "segment_attachment_bindings"
+    __table_args__ = (
+        sa.PrimaryKeyConstraint("id", name="segment_attachment_binding_pkey"),
+        sa.Index("segment_attachment_binding_tenant_idx", "tenant_id"),
```
QuantumGhost: For non-unique indexes with only one column, you should add `index=True` argument to the column definition, instead of defining those indexes inside `__table_args__` section. This also ensures that indexes naming are consistent.

api/models/dataset.py
```diff
@@ -1470,3 +1544,22 @@ class PipelineRecommendedPlugin(TypeBase):
         onupdate=func.current_timestamp(),
         init=False,
     )
+
+
+class SegmentAttachmentBinding(Base):
```
QuantumGhost, line 1549: I would suggest using `DefaultFieldsMixin` in models/base.py  and delete the `id` and `created_at` field below.

api/core/rag/datasource/vdb/vector_factory.py
```diff
@@ -203,6 +207,41 @@ def create(self, texts: list | None = None, **kwargs):
...
+                batch = file_documents[i : i + batch_size]
+                batch_start = time.time()
+                logger.info("Processing batch %s/%s (%s files)", i // batch_size + 1, total_batches, len(batch))
+                file_base64_list = []
+                real_batch = []
+                for document in batch:
+                    attachment_id = document.metadata["doc_id"]
+                    doc_type = document.metadata["doc_type"]
+                    # Query file info within db.session context to ensure thread-safe access
+                    upload_file = db.session.query(UploadFile).where(UploadFile.id == attachment_id).first()
```
QuantumGhost: The code here would query database `len(file_documents)` times.

api/core/rag/index_processor/processor/paragraph_index_processor.py
```diff
@@ -144,26 +163,70 @@ def index(self, dataset: Dataset, document: DatasetDocument, chunks: Any):
...
+                            "doc_type": DocType.IMAGE,
+                        }
+                        file_document = AttachmentDocument(
+                            page_content=file.filename or "image_file", metadata=file_metadata
+                        )
+                        attachments.append(file_document)
+                        all_multimodal_documents.append(file_document)
+                    doc.attachments = attachments
+                else:
+                    account = db.session.query(Account).where(Account.id == document.created_by).first()
```
QuantumGhost: There is a dedicated method `AccountService.load_user` for loading a user and the associated tenant from database. Consider using the existing method, instead of rolling another implementation.

api/models/dataset.py
QuantumGhost: From the query inside this PR, the following adjustment should be made to indexes:

- Add composite index on `segment_attachment_bindings(tenant_id, dataset_id, document_id, segment_id)` to cover the dominant lookups that filter by tenant/dataset, often narrowed to document/segment;  this lets queries in `DocumentSegment.attachments` and dataset/document-level attachment fetches hit a single index without extra sorts.
- Add unique composite index on `(segment_id, attachment_id)` to both speed up segment→attachments joins and prevent duplicate bindings for the same pair; it still supports the segment_id IN (...) deletions efficiently.
- Keep the single-column `attachment_id` index for the direct `attachment_id == upload_file.id` probe; 
  the new composites do not cover that access path.
- Consider dropping the standalone `dataset_id`, `document_id`, and `tenant_id` indexes after adding the multi-column scope index, unless you have other queries filtering solely on those single columns; 
  this reduces index bloat and write overhead.

QuantumGhost (approved):

api/services/dataset_service.py
```diff
@@ -2729,11 +2778,23 @@ def create_segment(cls, args: dict, document: Document, dataset: Dataset):
...
+                        tenant_id=current_user.current_tenant_id,
+                        dataset_id=document.dataset_id,
+                        document_id=document.id,
+                        segment_id=segment_document.id,
+                        attachment_id=attachment_id,
+                    )
+                    db.session.add(binding)
                 db.session.commit()
 
                 # save vector index
```
Chenyl-Sai, line 2800: This segment should not be included in the `if` section

Chenyl-Sai (commented):
please see this bug

PR #29456, opened 2025-12-11
fix: failed to delete model

crazywoola:
> [!IMPORTANT]
>

## Summary

Fix #29422 validation when delete a model credential.

## Screenshots

| Before | After |
|--------|-------|
| ... | image |

## Checklist

### DIFF
--- a/api/controllers/console/workspace/models.py
+++ b/api/controllers/console/workspace/models.py
@@ -230,7 +230,7 @@ class ModelProviderModelApi(Resource):
 
         return {"result": "success"}, 200
 
-    @console_ns.expect(console_ns.models[ParserDeleteModels.__name__], validate=True)
+    @console_ns.expect(console_ns.models[ParserDeleteModels.__name__])
     @setup_required
     @login_required
     @is_admin_or_owner_required


hjlarry (approved):

laipz8200:
Although this error has been fixed, I think the underlying reasons have not been thoroughly investigated. 

cc @asukaminato0721

asukaminato0721:
Did some research...

Solution: don't use restx 's validate, use pydantic's.

Context: we define pydantic model, but generate swagger doc, and use restx framework.

`definitions` is the keyword from JSON Schema Draft‑04/06/07. That’s what Swagger 2.0 and older tooling expect, and it’s the reason Flask‑RESTX uses #/definitions/... when it builds refs. Everything nested under definitions must be referenced with that path syntax.

`$defs` is the replacement introduced in JSON Schema 2019‑09 and retained in 2020‑12. It behaves the same conceptually—a container for reusable subschemas—but the `$` prefix signals that it’s part of the new vocabulary. Pydantic v2 targets the newer drafts, so even when change ref_template it still emits `$defs` internally for nested models.

Swagger 2.0 doesn’t know about $defs, so refs like `#/definitions/Item` point to a section that doesn’t exist, which is why RESTX validation breaks while doc generation (which just emits the JSON) still succeeds.

asukaminato0721:
fixed in https://github.com/langgenius/dify/pull/29289/commits/eb1f5f904734b48bd6fc6155d8e4bf35cde74e6f

PR #29503, opened 2025-12-11
refactor: split text generation item and add tests

lyzno1:
## Summary
- refactor GenerationItem into focused content/meta sections with shared hooks for workflow tabs and more-like-this state while preserving existing behaviors (including auto-switching to RESULT on new outputs)
- extract log formatting and keep handler memoization/guards clearer for copy, more-like-this, and log modal actions
- add GenerationItem tests covering more-like-this flow, log modal payload, copy behavior, feedback updates, child clearing, depth limits, and handler stability

## Testing
- Passed


web/app/components/app/text-generate/item/index.spec.tsx
```diff
@@ -0,0 +1,381 @@
...
+
+const createDeferred = <T,>() => {
+  let resolve!: (value: T) => void
+  const promise = new Promise<T>((res) => {
+    resolve = res
+  })
+  return { promise, resolve }
+}
+
+describe('GenerationItem', () => {
```
CodingOnStar, line 254: organize tests with describe sections and add a brief comment before each block

web/app/components/app/text-generate/item/index.spec.tsx
```diff
@@ -0,0 +1,381 @@
...
+    )
+  }
+  return {
+    __esModule: true,
+    default: Component,
+    __calls: calls,
+  }
+})
+
+jest.mock('./meta-section', () => {
```
CodingOnStar, line 100: Is the conditional logic in this component too heavy, and can the amount of mocking be reduced?

web/app/components/app/text-generate/item/index.spec.tsx
```diff
@@ -0,0 +1,381 @@
...
+        'button',
+        { 'data-testid': `feedback-${props.messageId}`, 'onClick': () => props.onFeedback?.({ rating: 'like' }) },
+        'feedback',
+      ),
+    )
+  }
+  return {
+    __esModule: true,
+    default: Component,
+    __calls: calls,
```
CodingOnStar: it is recommended to use the built-in capabilities of jest.fn()

web/app/components/app/text-generate/item/index.spec.tsx
CodingOnStar, line 1: 1、missing boundary condition testing
2、increase testing for ContentSection, MetaSection, useMoreLikeThisState, and useWorkflowTabs through layered testing

PR #29837, opened 2025-12-18
perf: decrease db query

### LINKED ISSUE #29816: Bug: message_cycle_manager N+1 query issues caused by commit 084dcd1a50
## 📋 Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language policy](https://github.com/langgenius/dify/issues/1542).

---

## 🌎 Environment

**Dify version:**
`1.11.1`

**Cloud or Self Hosted:**
Self Hosted (Source)

---

## 🐛 Bug Description

Commit `084dcd1a50` (#22988) introduced an inefficient query in `message_to_stream_response()` that causes:

1. **N+1 database queries**: A query is executed for every LLM chunk during streaming
2. **Dead logic**: The query condition is wrong and never returns `MESSAGE_FILE` type

### Problematic Code
```python
# api/core/app/task_pipeline/message_cycle_manager.py:226-227
def message_to_stream_response(...) -> MessageStreamResponse:
    with Session(db.engine) as session:
        message_file = session.scalar(select(MessageFile).where(MessageFile.id == message_id))
    event_type = StreamEvent.MESSAGE_FILE if message_file else StreamEvent.MESSAGE
    return MessageStreamResponse(..., event=event_type)
```

---

### 1. N+1 Query Problem
- During LLM streaming, `message_to_stream_response()` is called **per chunk**
- Each call executes: `SELECT FROM message_files WHERE id = ?`
- For 100 chunks = 100 wasted queries

### 2. Dead Logic (Always Fails)
- `message_id` parameter = `Message.id` (UUID)
- Query condition = `MessageFile.id == Message.id`
- But `MessageFile.id` is auto-generated, **never equals `Message.id`**
- **Result**: Always returns `StreamEvent.MESSAGE` anyway

### 3. "Fixing" It Creates Worse Bug
If we change to `MessageFile.message_id == message_id`:
- Message with files → all LLM chunks return `event: "message_file"`
- **Text chunks incorrectly labeled as files!**

---

## 🎯 Expected Behavior

- No unnecessary database queries during streaming
- No dead/unreachable code

---

## 🔄 Proposed Solution

**Revert commit `084dcd1a50`** to restore original behavior:

```python
def message_to_stream_response(
    self, answer: str, message_id: str, from_variable_selector: list[str] | None = None
) -> MessageStreamResponse:
    return MessageStreamResponse(
        task_id=self._application_generate_entity.task_id,
        id=message_id,
        answer=answer,
        from_variable_selector=from_variable_selector,
    )
```

**Remove:** The query, event type check, and event parameter.

---

## 📊 Impact

**Before commit `084dcd1a50`:**
- ✅ 0 extra queries
- ✅ Clean, working code

**After commit (current):**
- ❌ +N queries during streaming
- ❌ Logic that never works as intended

---

## 📎 Commit Info

- **Bad commit**: `084dcd1a50` - "using message_file event instead of message event" (#22988)
- **First appeared**: Version `1.7.2`
- **File**: `api/core/app/task_pipeline/message_cycle_manager.py:226-235`

fatelei:
> [!IMPORTANT]
>

## Summary

fix #29816

decrease unessary db query

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/core/app/task_pipeline/message_cycle_manager.py
+++ b/api/core/app/task_pipeline/message_cycle_manager.py
@@ -5,7 +5,7 @@ from threading import Thread
 from typing import Union
 
 from flask import Flask, current_app
-from sqlalchemy import select
+from sqlalchemy import exists, select
 from sqlalchemy.orm import Session
 
 from configs import dify_config
@@ -54,6 +54,22 @@ class MessageCycleManager:
     ):
         self._application_generate_entity = application_generate_entity
         self._task_state = task_state
+        self._message_has_file: set[str] = set()
+
+    def get_message_event_type(self, message_id: str) -> StreamEvent:
+        if message_id in self._message_has_file:
+            return StreamEvent.MESSAGE_FILE
+
+        with Session(db.engine, expire_on_commit=False) as session:
+            has_file = session.query(
+                exists().where(MessageFile.message_id == message_id)
+            ).scalar()
+
+        if has_file:
+            self._message_has_file.add(message_id)
+            return StreamEvent.MESSAGE_FILE
+
+        return StreamEvent.MESSAGE
 
     def generate_conversation_name(self, *, conversation_id: str, query: str) -> Thread | None:
         """
@@ -214,7 +230,11 @@ class MessageCycleManager:
         return None
 
     def message_to_stream_response(
-        self, answer: str, message_id: str, from_variable_selector: list[str] | None = None
+        self,
+        answer: str,
+        message_id: str,
+        from_variable_selector: list[str] | None = None,
+        event_type: StreamEvent | None = None,
     ) -> MessageStreamResponse:
         """
         Message to stream response.
@@ -222,16 +242,12 @@ class MessageCycleManager:
         :param message_id: message id
         :return:
         """
-        with Session(db.engine, expire_on_commit=False) as session:
-            message_file = session.scalar(select(MessageFile).where(MessageFile.id == message_id))
-        event_type = StreamEvent.MESSAGE_FILE if message_file else StreamEvent.MESSAGE
-
         return MessageStreamResponse(
             task_id=self._application_generate_entity.task_id,
             id=message_id,
             answer=answer,
             from_variable_selector=from_variable_selector,
-            event=event_type,
+            event=event_type or StreamEvent.MESSAGE,
         )
 
     def message_replace_to_stream_response(self, answer: str, reason: str = "") -> MessageReplaceStreamResponse:
--- a/api/core/app/task_pipeline/message_cycle_manager.py
+++ b/api/core/app/task_pipeline/message_cycle_manager.py
@@ -61,9 +61,7 @@ class MessageCycleManager:
             return StreamEvent.MESSAGE_FILE
 
         with Session(db.engine, expire_on_commit=False) as session:
-            has_file = session.query(
-                exists().where(MessageFile.message_id == message_id)
-            ).scalar()
+            has_file = session.query(exists().where(MessageFile.message_id == message_id)).scalar()
 
         if has_file:
             self._message_has_file.add(message_id)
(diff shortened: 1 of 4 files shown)


qiqizjl:
After this change, as long as this `message_id` exists in `MessageFile`, will subsequent streams all be `message_file`? Is this as expected? Also, `MessageFile` is written only in `save_messages`. Shouldn't data be unreadable here? My understanding is that `message_file` is defined as only being a `message_file` when this chunk is a file stream?

fatelei:
> After this change, as long as this `message_id` exists in `MessageFile`, will subsequent streams all be `message_file`? Is this as expected? Also, `MessageFile` is written only in `save_messages`. Shouldn't data be unreadable here? My understanding is that `message_file` is defined as only being a `message_file` when this chunk is a file stream?

the origin code is using self._message_id to query, in the same class the self._message_id stay the same, so the origin code is duplicate query the database using same self._message_id

qiqizjl:
> > After this change, as long as this `message_id` exists in `MessageFile`, will subsequent streams all be `message_file`? Is this as expected? Also, `MessageFile` is written only in `save_messages`. Shouldn't data be unreadable here? My understanding is that `message_file` is defined as only being a `message_file` when this chunk is a file stream?
> 
> the origin code is using self._message_id to query, in the same class the self._message_id stay the same, so the origin code is duplicate query the database using same self._message_id

Yes. However, I understand that this query might be meaningless, or could it introduce new bugs? Does the definition of `message_file` refer to whether the file exists in the current chunk or the file exists in the current request? If it's the former, is it a bug, or can't it be solved simply by modifying the SQL? If it's the latter, then it's OK. As I mentioned in the issue, the MessageFile returned by the workflow is written in `_save_message`, and no data is written when the chunk is output. If it returns `message_file` as long as the file exists in the current message, then it does seem to meet expectations. Therefore, I think we should first confirm the definition of the `message_file` event.

fatelei:
> > > After this change, as long as this `message_id` exists in `MessageFile`, will subsequent streams all be `message_file`? Is this as expected? Also, `MessageFile` is written only in `save_messages`. Shouldn't data be unreadable here? My understanding is that `message_file` is defined as only being a `message_file` when this chunk is a file stream?
> > 
> > 
> > the origin code is using self._message_id to query, in the same class the self._message_id stay the same, so the origin code is duplicate query the database using same self._message_id
> 
> Yes. However, I understand that this query might be meaningless, or could it introduce new bugs? Does the definition of `message_file` refer to whether the file exists in the current chunk or the file exists in the current request? If it's the former, is it a bug, or can't it be solved simply by modifying the SQL? If it's the latter, then it's OK. As I mentioned in the issue, the MessageFile returned by the workflow is written in `_save_message`, and no data is written when the chunk is output. If it returns `message_file` as long as the file exists in the current message, then it does seem to meet expectations. Therefore, I think we should first confirm the definition of the `message_file` event.

yes, i just modify a little, add a simple cache

api/core/app/task_pipeline/message_cycle_manager.py
```diff
@@ -54,6 +54,23 @@ def __init__(
     ):
         self._application_generate_entity = application_generate_entity
         self._task_state = task_state
+        self._message_has_file: set[str] = set()
+
+    def on_message_file_created(self, message_id: str):
```
QuantumGhost: The `on_message_file_created` seems not used.

QuantumGhost: The `on_message_file_created` method seems unused.

api/tests/unit_tests/core/app/task_pipeline/test_message_cycle_manager_simple.py
```diff
@@ -0,0 +1,232 @@
+"""Simple unit tests for the message cycle manager optimization."""
```
QuantumGhost: This test file seems not very helpful. It does not validate the behavior of the `MessageCycleManager` class.

QuantumGhost (commented):
Other changes LGTM

QuantumGhost (approved):

PR #29982, opened 2025-12-22
fix(schedule): support multiple schedule triggers in workflow

majiayu000:
## Summary
- Add `extract_all_schedule_configs()` method to return all schedule configs
- Refactor `sync_schedule_from_workflow()` to handle multiple schedules
- Follow same pattern as `update_app_triggers` handler for consistency

## Test plan
- [x] Added tests for `extract_all_schedule_configs()` with multiple schedules
- [x] Added tests for sync operations (add/remove/update)
- [ ] CI tests pass

Fixes #29945


Yeuoly:
1. IMO, If what you want is a more complex and flexible cron rules, we'd better support multiple cron expressions within a single schedule node, rather than using multiple schedule nodes, which means you may merge several cron rules into same db record, it's better for scheduling management, as you known, we just simply calculate the next execution time for each record.

2. if you want to design multiple flows within just one Dify workflow, we suggest you to use separate apps instead, allowing multiple schedule nodes in the same workflow is not a part of the current design, I believe the fact that allowing user to add multiple schedule nodes is actually a bug from frontend.

majiayu000:
Thanks for the feedback @Yeuoly!

I understand your points:
1. Supporting multiple cron expressions within a single schedule node is better for scheduling management
2. Multiple schedule nodes in the same workflow is not part of the current design

I'll close this PR and work on a new approach that supports multiple cron expressions within a single schedule node instead. This makes more sense for the architecture since it keeps the scheduling logic consolidated in one db record.

Also, should the frontend be updated to prevent users from adding multiple schedule nodes? I can file a separate issue for that if needed.

PR #30060, opened 2025-12-24
fix: fix loop node share the variable_pool

fatelei:
> [!IMPORTANT]
>

## Summary

fix #30037

in loop_node variable_pool is a share var, so in different loop this value is shared, so the
["llm_node", "text"] is shared, we can add a method to clear previous share data, make in indepent

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/core/workflow/nodes/loop/loop_node.py
+++ b/api/core/workflow/nodes/loop/loop_node.py
@@ -118,6 +118,8 @@ class LoopNode(LLMUsageTrackingMixin, Node[LoopNodeData]):
                 loop_count = 0
 
             for i in range(loop_count):
+                if i > 0:
+                    self._cleanup_sub_graph_variables_from_previous_iteration()
                 graph_engine = self._create_graph_engine(start_at=start_at, root_node_id=root_node_id)
 
                 loop_start_time = naive_utc_now()
@@ -256,7 +258,6 @@ class LoopNode(LLMUsageTrackingMixin, Node[LoopNodeData]):
             segment = self.graph_runtime_state.variable_pool.get(sel)
             self.node_data.outputs[key] = segment.value if segment else None
         self.node_data.outputs["loop_round"] = current_index + 1
-
         return reach_break_node
 
     def _append_loop_info_to_event(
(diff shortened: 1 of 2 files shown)


api/tests/unit_tests/core/workflow/nodes/loop/test_loop_node_variable_cleanup.py
```diff
@@ -0,0 +1,250 @@
...
+
+from core.app.entities.app_invoke_entities import InvokeFrom
+from core.workflow.entities import GraphInitParams
+from core.workflow.nodes.loop.loop_node import LoopNode
+from core.workflow.runtime import GraphRuntimeState
+from core.workflow.runtime.variable_pool import VariablePool
+from models.enums import UserFrom
+
+
+class TestLoopNodeVariableCleanup:
```
QuantumGhost, line 20: IMO the tests here should not only validate the variable handling logic in LoopNode. but also validate the behavior of streaming responses.

api/core/workflow/nodes/loop/loop_node.py
```diff
@@ -437,3 +438,18 @@ def _create_graph_engine(self, start_at: datetime, root_node_id: str):
...
+
+        sub_graph_node_ids = self._extract_loop_node_ids_from_config(
+            graph_config=self.graph_config, loop_node_id=self._node_id
+        )
+        sub_graph_node_ids.discard(self._node_id)
+
+        for node_id in sub_graph_node_ids:
+            try:
+                self.graph_runtime_state.variable_pool.remove([node_id])
+            except Exception as e:
```
QuantumGhost, line 454: I suggest only catch exceptions related to variable deletion here.

api/core/workflow/nodes/loop/loop_node.py
```diff
@@ -118,6 +118,8 @@ def _run(self) -> Generator:
                 loop_count = 0
 
             for i in range(loop_count):
+                if i > 0:
```
QuantumGhost, line 121: Maybe I lacks some context, the @dosubot in the original issue suggests that the bug is related to `ResponseCoordinator`. This PR only manipulates variables in `VariablePool`. Does this PR fix the root cause of issue #30037

Nov1c444: @QuantumGhost 
This PR seems to duplicate #30059.
#30037 was caused by loop streaming showing the previous iteration's output again. For example, the first UUID appeared in the second iteration. This happened because the new response coordinator reused stale variables from the shared pool. Clearing subgraph variables before each loop run prevents old values from leaking into later iterations.

Issue #30069, opened 2025-12-24
Parameter Extraction node fails with Invalid text content type error


krap730:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/new).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report.
- [x] I have read the [README](https://github.com/langgenius/dify/blob/main/README.md) and [documentation](https://docs.dify.ai/).

### Dify version

Latest (main branch)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Create a workflow with a Parameter Extraction node
2. Use Gemini 2.5 Pro as the model
3. Configure input variable and parameters to extract (e.g., Array[String] type)
4. Run the workflow

### Expected Behavior

The Parameter Extraction node should successfully extract parameters from the input.

### Actual Behavior

The node fails with the following error:
```
Parameter extraction Error Invalid text content type: <class 'list'>. Expected str.
```

The input shows as `null` in the execution results.

### Root Cause Analysis

In `api/core/workflow/nodes/parameter_extractor/parameter_extractor_node.py` at line 284:

```python
text = invoke_result.message.content or ""
if not isinstance(text, str):
    raise InvalidTextContentTypeError(f"Invalid text content type: {type(text)}. Expected str.")
```

The `PromptMessage.content` field is typed as `str | list[PromptMessageContentUnionTypes] | None`. When certain models (like Gemini 2.5 Pro) return responses in multi-modal content format, `content` is a `list` instead of `str`, causing this error.

### Suggested Fix

The `PromptMessage` class already has a `get_text_content()` method that properly handles both `str` and `list` content types. The fix should replace:

```python
text = invoke_result.message.content or ""
```

with:

```python
text = invoke_result.message.get_text_content()
```

This will properly extract text content regardless of whether the model returns a string or a list of content items.

krap730:
## Additional Issue Found

While testing the fix for the Parameter Extraction node, I discovered a related but separate issue:

### Problem
When calling a workflow as a tool, if any output variable has a `null` value, the tool invocation fails with:
```
1 validation error for VariableMessage
Value error, Only basic types and lists are allowed.
```

### Root Cause
[...]

PR #30082, opened 2025-12-24
fix: allow None values in VariableMessage validation

krap730:
## Summary

Fix workflow-as-tool invocation failing when output variables contain `null` values.

### Problem

When calling a workflow as a tool, if any output variable has a `null` value, the tool invocation fails with:
```
1 validation error for VariableMessage
Value error, Only basic types and lists are allowed.
```

### Root Cause

In `api/core/tools/entities/tool_entities.py`, the `VariableMessage.transform_variable_value` validator only allows `dict | list | str | int | float | bool` types, but does not allow `None`.

When a workflow End node outputs contain `null` values (e.g., `{"result": "data", "optional_field": null}`), the `WorkflowTool._invoke()` method passes these `None` values to `create_variable_message()`, which fails the validation.

### Solution

Update the validation to allow `None` values by checking `if value is not None` before the isinstance check.

Related to #30069

## Test Plan

- [ ] Create a workflow with multiple output variables where some may be null
- [ ] Call the workflow as a tool from another workflow or agent
- [ ] Verify that null output values no longer cause validation errors

### DIFF
--- a/api/core/tools/entities/tool_entities.py
+++ b/api/core/tools/entities/tool_entities.py
@@ -153,11 +153,11 @@ class ToolInvokeMessage(BaseModel):
         @classmethod
         def transform_variable_value(cls, values):
             """
-            Only basic types and lists are allowed.
+            Only basic types, lists, and None are allowed.
             """
             value = values.get("variable_value")
-            if not isinstance(value, dict | list | str | int | float | bool):
-                raise ValueError("Only basic types and lists are allowed.")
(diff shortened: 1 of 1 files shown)


krap730:
Note: This issue can also be worked around by ensuring that workflow outputs don't contain `null` values (e.g., by providing default values in the workflow itself).

Whether to merge this fix is up to the maintainers' discretion - the workaround is a valid alternative if you prefer to keep the stricter validation.

crazywoola (approved):

PR #29101, opened 2025-12-03
perf(api): optimize tool provider list API with Redis caching

yangzheli:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

fix #29102

- Add ToolProviderListCache class with 5-minute TTL Redis caching
- Implement cache-first strategy for tool provider listings  
- Add automatic cache invalidation on CRUD operations across all provider types
- Optimize database queries by batching operations and reducing N+1 queries

### Before

- agent configuration page

https://github.com/user-attachments/assets/55b4ac33-d305-41c4-b6d7-3c23738dc740

- request using Postman

接口性能优化前

### After

- agent configuration page

https://github.com/user-attachments/assets/b917930e-0bd8-45a1-b953-b7af62728f2a

- request using Postman

性能优化后

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods

PR #30150, opened 2025-12-25
fix(api): move cache invalidation outside redis lock to prevent timeout

Mairuis:
## Summary

Fixes #30149

Move `ToolProviderListCache.invalidate_cache()` outside the Redis lock in `add_builtin_tool_provider` to prevent lock timeout errors.

## Problem

After #29101 introduced tool provider list caching, users are experiencing lock timeout errors:

```json
{"code":"invalid_param","message":"Cannot release a lock that's no longer owned","status":400}
```

The 20-second lock timeout can be exceeded when:
- `validate_credentials()` makes slow external API calls (e.g., OAuth token exchange)
- `invalidate_cache()` uses `scan_iter` which scans the entire Redis keyspace

## Solution

Move `invalidate_cache()` outside the lock since cache invalidation doesn't need to be atomic with the database write.

Also specify the `"builtin"` type to avoid unnecessary key scanning via `scan_iter`.

## Test plan

- [x] Verify OAuth credential flow completes without lock timeout errors
- [x] Verify tool provider list cache is properly invalidated after adding credentials

---


Mairuis:
@crazywoola  approve please～...

crazywoola (approved):

PR #29494, opened 2025-12-11
feat: clear free plan workflow run logs

hjlarry:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

fix https://github.com/langgenius/dify/issues/29776

related PR:
https://github.com/langgenius/dify/pull/29842
https://github.com/langgenius/dify/pull/29829
https://github.com/langgenius/dify/pull/29638

Todo:
rm `api/services/clear_free_plan_tenant_expired_logs.py` in the future.

#### 1. update db
to use the `created_at` index of `workflow_runs` table
```
flask db upgrade
```

#### 2. dry run 

```
flask clean-workflow-runs --start-from="2024-01-01" --end-before="2024-01-31" --dry-run
```

then you can get some logs like:
```
...
Dry run complete. Would delete 2024 workflow runs between 2014-01-01T00:00:00 and 2024-01-02T00:00:00; related records: node_executions 2022, offloads 6066, app_logs 2024, trigger_logs 2022, pauses 2022, pause_reasons 2022
```

you can compare with the sql result (maybe slow):
```
WITH params AS (
    SELECT 
        TIMESTAMP '2024-01-01 00:00:00' AS start_after,
        TIMESTAMP '2024-01-31 00:00:00' AS end_before
)
SELECT 'workflow_runs' AS table_name, COUNT(*) AS record_count 
FROM workflow_runs, params 
WHERE created_at >= start_after AND created_at < end_before

UNION ALL
SELECT 'workflow_node_executions', COUNT(*) 
FROM workflow_node_executions, params 
WHERE created_at >= start_after AND created_at < end_before

UNION ALL
SELECT 'workflow_node_execution_offload', COUNT(*) 
FROM workflow_node_execution_offload, params 
WHERE created_at >= start_after AND created_at < end_before

UNION ALL
SELECT 'workflow_app_logs', COUNT(*) 
FROM workflow_app_logs, params 
WHERE created_at >= start_after AND created_at < end_before

UNION ALL
SELECT 'workflow_trigger_logs', COUNT(*) 
FROM workflow_trigger_logs, params 
WHERE created_at >= start_after AND created_at < end_before

UNION ALL
SELECT 'workflow_pauses', COUNT(*) 
[...]

PR #30328, opened 2025-12-29
feat: archive workflow run logs

hjlarry:
> [!IMPORTANT]
>

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


api/libs/archive_storage.py
```diff
@@ -0,0 +1,342 @@
...
+                dify_config.ARCHIVE_STORAGE_SECRET_KEY,
+            ]
+        ):
+            raise ArchiveStorageNotConfiguredError(
+                "Archive storage configuration is incomplete. "
+                "Required: ARCHIVE_STORAGE_ENDPOINT, ARCHIVE_STORAGE_BUCKET, "
+                "ARCHIVE_STORAGE_ACCESS_KEY, ARCHIVE_STORAGE_SECRET_KEY"
+            )
+
+        self.bucket = dify_config.ARCHIVE_STORAGE_BUCKET
```
hjlarry: Should we support separate buckets for archive and download tasks?

api/repositories/sqlalchemy_api_workflow_node_execution_repository.py
```diff
@@ -290,3 +290,84 @@ def delete_executions_by_ids(
...
+
+    class RunContext(TypedDict):
+        run_id: str
+        tenant_id: str
+        app_id: str
+        workflow_id: str
+        triggered_from: str
+
+    @staticmethod
+    def delete_by_runs(session: Session, runs: Sequence[RunContext]) -> tuple[int, int]:
```
hjlarry, line 327: duplicated with https://github.com/langgenius/dify/pull/29494/
remove this method after 29494 merged

api/repositories/sqlalchemy_api_workflow_run_repository.py
```diff
@@ -314,6 +316,231 @@ def delete_runs_by_app(
...
+        session: Session,
+        pause_ids: Sequence[str],
+    ) -> Sequence[WorkflowPauseReason]:
+        if not pause_ids:
+            return []
+
+        stmt = select(WorkflowPauseReason).where(WorkflowPauseReason.pause_id.in_(pause_ids))
+        return list(session.scalars(stmt))
+
+    def delete_runs_with_related(
```
hjlarry, line 403: delete after #29494 merged

api/repositories/sqlalchemy_api_workflow_run_repository.py
```diff
@@ -314,6 +316,231 @@ def delete_runs_by_app(
         logger.info("Total deleted %s workflow runs for app %s", total_deleted, app_id)
         return total_deleted
 
+    def get_runs_batch_by_time_range(
```
hjlarry, line 316: almost same as #29494 but add `tenants_ids`

api/services/retention/workflow_run/archive_paid_plan_workflow_run.py
```diff
@@ -0,0 +1,570 @@
...
+        repo = self._get_workflow_run_repo()
+        return repo.get_runs_batch_by_time_range(
+            start_after=None,
+            end_before=self.cutoff_date,
+            last_seen=last_seen,
+            batch_size=self.batch_size,
+            tenant_ids=list(self.tenant_ids) if self.tenant_ids else None,
+        )
+
+    def _filter_paid_tenants(self, tenant_ids: set[str]) -> set[str]:
```
hjlarry, line 313: replace with public method

api/repositories/workflow_trigger_log_repository.py
```diff
@@ -109,3 +109,15 @@ def get_recent_logs(
             A sequence of recent WorkflowTriggerLog instances
         """
         ...
+
+    def delete_by_run_ids(self, run_ids: Sequence[str]) -> int:
```
hjlarry, line 113: delete after #29494 merged

api/repositories/sqlalchemy_workflow_trigger_log_repository.py
```diff
@@ -44,6 +46,21 @@ def get_by_id(self, trigger_log_id: str, tenant_id: str | None = None) -> Workfl
 
         return self.session.scalar(query)
 
+    def list_by_run_id(self, run_id: str) -> Sequence[WorkflowTriggerLog]:
+        """List trigger logs for a workflow run."""
+        query = select(WorkflowTriggerLog).where(WorkflowTriggerLog.workflow_run_id == run_id)
+        return list(self.session.scalars(query).all())
+
+    def delete_by_run_ids(self, run_ids: Sequence[str]) -> int:
```
hjlarry: delete after #29494 merged

api/models/workflow.py
```diff
@@ -841,6 +848,33 @@ def __table_args__(cls) -> Any:
     created_by: Mapped[str] = mapped_column(StringUUID)
     finished_at: Mapped[datetime | None] = mapped_column(DateTime)
 
+    def to_dict(self) -> dict[str, Any]:
```
hj24: Is it used only for archive jobs? If yes, maybe we should give it a more clear name like to_archive_dict and add a note here

api/models/workflow.py
```diff
@@ -1053,6 +1087,17 @@ class WorkflowNodeExecutionOffload(Base):
         primaryjoin="WorkflowNodeExecutionOffload.file_id == UploadFile.id",
     )
 
+    def to_dict(self) -> dict[str, Any]:
```
hj24: same here

api/services/retention/workflow_run/export_workflow_run.py
```diff
@@ -0,0 +1,314 @@
+"""
+Workflow Run Export Service.
```
hj24: could we add a sub dir here? like services/retention/workflow_runs/xxx.py, we may need a conversation dir later

api/libs/archive_storage.py
hj24: yes, we need one archive bucket and one download bucket, for download bucket, we should check if s3 or r2 can configure the lifecycle rule to delete the outdated temporarily uploaded objects.

api/libs/archive_storage.py
```diff
@@ -0,0 +1,336 @@
...
+            records: List of dictionaries to serialize
+
+        Returns:
+            Gzipped JSONL bytes
+        """
+        lines = []
+        for record in records:
+            # Convert datetime objects to ISO format strings
+            serialized = ArchiveStorage._serialize_record(record)
+            lines.append(json.dumps(serialized, ensure_ascii=False, default=str))
```
hj24: Is it possible to use orjson instead?

api/libs/archive_storage.py
```diff
@@ -0,0 +1,336 @@
...
+    Raises:
+        ArchiveStorageNotConfiguredError: If archive storage is not configured
+    """
+    global _archive_storage
+    if _archive_storage is None:
+        _archive_storage = ArchiveStorage()
+    return _archive_storage
+
+
+def build_workflow_run_prefix(
```
hj24: we should add the table_name as the prefix, and this logic should be added to archive service, not lib layer

api/controllers/console/app/workflow_run.py
```diff
@@ -20,14 +22,23 @@
...
+    reserve_task_for_run,
+    set_task_status,
+)
 from services.workflow_run_service import WorkflowRunService
+from tasks.workflow_run_export_task import export_workflow_run_task
 
 # Workflow run status choices for filtering
 WORKFLOW_RUN_STATUS_CHOICES = ["running", "succeeded", "failed", "stopped", "partial-succeeded"]
 
+EXPORT_TASK_RATE_LIMITER = RateLimiter("workflow_run_export_task", max_attempts=10, time_window=60)
```
hj24: should we add it to archive service layer? like the account service?

api/controllers/console/app/workflow_run.py
```diff
@@ -181,6 +206,69 @@ def get(self, app_model: App):
...
+    @login_required
+    @account_initialization_required
+    @get_app_model()
+    def post(self, app_model: App, run_id: str):
+        tenant_id = str(app_model.tenant_id)
+        app_id = str(app_model.id)
+        run_id_str = str(run_id)
+
+        # If a task already exists for this run, return its status to keep export idempotent.
+        existing_task_id = get_task_id_for_run(tenant_id, app_id, run_id_str)
```
hj24: should we pack those get_task_id_for_run/get_task_status/reserve_task_for_run methods into an export service layer?

api/extensions/ext_commands.py
```diff
@@ -21,6 +22,7 @@ def init_app(app: DifyApp):
         reset_email,
         reset_encrypt_key_pair,
         reset_password,
+        rollback_workflow_runs,
```
hj24: restore_workflow_runs?

api/controllers/console/app/workflow_run.py
```diff
@@ -93,6 +104,20 @@
     "WorkflowRunNodeExecutionList", workflow_run_node_execution_list_fields_copy
 )
 
+workflow_run_export_task_fields = console_ns.model(
+    "WorkflowRunExportTaskStatus",
+    {
+        "task_id": fields.String(description="Export task id"),
+        "status": fields.String(description="Task status: pending/running/success/failed"),
+        "storage_key": fields.String(description="Object storage key", required=False),
```
hj24: should we avoid to exposing these storage_key/checksum info to frontend?

api/commands.py
```diff
@@ -852,6 +853,126 @@ def clear_free_plan_tenant_expired_logs(days: int, batch: int, tenant_ids: list[
...
+        click.style(
+            f"Workflow run archiving completed. start={start_time.isoformat()} "
+            f"end={end_time.isoformat()} duration={elapsed}",
+            fg="green",
+        )
+    )
+
+
+@click.command(
+    "rollback-workflow-runs",
```
hj24: restore-workflow-runs

api/services/retention/workflow_run/restore_archived_workflow_run.py
```diff
@@ -0,0 +1,346 @@
...
+            return 0
+
+        # Convert datetime strings back to datetime objects
+        converted_records = []
+        for record in records:
+            converted = self._convert_datetime_fields(record, model)
+            converted_records.append(converted)
+
+        # Use INSERT ... ON CONFLICT DO NOTHING for idempotency
+        stmt = pg_insert(model).values(converted_records)
```
hj24, line 303: should we consider to do a schema mapping layer in case the schema has changed after the restore?

api/services/retention/workflow_run/export_workflow_run.py
```diff
@@ -0,0 +1,314 @@
...
+                raise WorkflowRunExportError(f"Workflow run {run_id} not found")
+
+            if run.tenant_id != tenant_id:
+                raise WorkflowRunExportError(f"Workflow run {run_id} does not belong to tenant {tenant_id}")
+
+            if run.is_archived:
+                return self._export_from_s3(session, run)
+            return self._export_from_db(session, run)
+
+    def export_to_storage(
```
hj24: we should check if s3 or r2 can configure the lifecycle rule to delete the outdated temporarily uploaded objects.

hj24 (changes requested):

api/models/workflow.py
hjlarry: Many models in Dify use `to_dict`, for example, [here](https://github.com/langgenius/dify/blob/0421387672d4fded149d1219d08540903d535710/api/services/clear_free_plan_tenant_expired_logs.py#L82). 

I previously used some code like below for model conversion, but I've found `to_dict` to be a more consistent and standard practice. I've therefore started adding the missing `to_dict` methods to some models. What are your thoughts on this approach?
```
for column in record.__table__.columns:
      record_dict[column.name] = getattr(record, column.name)
```

api/services/retention/workflow_run/export_workflow_run.py
hjlarry: sure,  we should also move the previous clear_sandbox_*** to the subdir

api/libs/archive_storage.py
hjlarry: Yes, both support object lifecycle rules. Different prefixes can be assigned different rules, but separating buckets seems better.

api/controllers/console/app/workflow_run.py
```diff
@@ -1,11 +1,13 @@
 from typing import Literal, cast
```
hj24, line 2: we should also check the service_api and do proper handling.

api/controllers/console/app/workflow_run.py
hjlarry: if we move it to service layer,  celery will check this limit ?  users seems can't get a alert message?

api/services/retention/workflow_run/archive_paid_plan_workflow_run.py
```diff
@@ -0,0 +1,546 @@
...
+    """Summary of the entire archive operation."""
+
+    total_runs_processed: int = 0
+    runs_archived: int = 0
+    runs_skipped: int = 0
+    runs_failed: int = 0
+    total_elapsed_time: float = 0.0
+
+
+class WorkflowRunArchiver:
```
hj24, line 87: if a workflow_run needs to be deleted, such as account deletion scenarios, those archived data should also be deleted.

hj24 (changes requested):

api/libs/archive_storage.py
hj24: we should separate buckets, and now we only need to set lifecycle rules for keys under export bucket, the lifecycle of those data should be very short

api/models/workflow.py
hj24: If session expired, will `for column in record.__table__.columns` cause implicit extra db query?
 
 If to_dict is command pattern, we can keep it.

api/libs/archive_storage.py
hjlarry: The `build_workflow_run_prefix` function is a helper common to the archive, restore, and export services for generating S3 prefixes from datetimes. The prefix also has a table_name. It has been moved to `api/libs/retention_utils` to improve the clarity of the `archive_storage` module.

api/services/retention/workflow_run/archive_paid_plan_workflow_run.py
hjlarry: ok, after #29494 merged, I will add this logic

api/controllers/console/app/workflow_run.py
hjlarry: when a service_api request to an archived data?  I will handle it when I test on it

api/controllers/console/app/workflow_run.py
hj24: Should we reduce the max_attempts? Like max_attempts=3, time_window=60, max_attempts seems still a little high

api/controllers/console/app/workflow_run.py
hj24: Yeah, maybe raise a BadRequest

hj24 (changes requested):

api/controllers/console/app/workflow_run.py
hjlarry: ok

api/services/retention/workflow_run/restore_archived_workflow_run.py
hjlarry: Added `_apply_schema_mapping` for the compatibility

api/services/retention/workflow_run/archive_paid_plan_workflow_run.py
hjlarry: add when delete a app, also delete its workflow archived logs and files

api/commands.py
```diff
@@ -852,6 +853,312 @@ def clear_free_plan_tenant_expired_logs(days: int, batch: int, tenant_ids: list[
...
+)
+@click.option(
+    "--tenant-ids",
+    required=False,
+    help="Tenant IDs (comma-separated).",
+)
+@click.option("--run-id", required=False, help="Workflow run ID to restore.")
+@click.option(
+    "--start-from",
+    type=click.DateTime(formats=["%Y-%m-%d", "%Y-%m-%dT%H:%M:%S"]),
```
hj24, line 1023: we should use days to calculate the datetime range everyday instead of pass a specific datetime? otherwise we will need to deploy the cronjob everyday?

api/models/workflow.py
```diff
@@ -1168,6 +1168,95 @@ def to_dict(self):
...
+    archived_at: Mapped[datetime] = mapped_column(
+        DateTime, nullable=False, server_default=func.current_timestamp(), init=False
+    )
+
+    @property
+    def created_from(self) -> str | None:
+        return self.log_created_from
+
+    @property
+    def created_at(self) -> datetime | None:
```
hj24: Is it possible to use created_at directly to represents the creation timestamp of the archive log? Seem you have already used other prefix like run_ to distinguish other created_at fields.

api/controllers/console/app/workflow_run.py
```diff
@@ -181,6 +196,56 @@ def get(self, app_model: App):
...
+                WorkflowArchiveLog.tenant_id == tenant_id,
+                WorkflowArchiveLog.app_id == app_id,
+                WorkflowArchiveLog.workflow_run_id == run_id_str,
+            )
+            .limit(1)
+        )
+        if not run_created_at:
+            return {"code": "archive_log_not_found", "message": "workflow run archive not found"}, 404
+
+        prefix = (
```
hj24, line 226: Although we can use presigned url directly now, can we keep the ratelimit? What's the cost on r2 if download actions happened?

api/models/workflow.py
```diff
@@ -1168,6 +1168,95 @@ def to_dict(self):
...
+    run_triggered_from: Mapped[str] = mapped_column(String(255), nullable=False)
+    run_error: Mapped[str | None] = mapped_column(LongText, nullable=True)
+    run_elapsed_time: Mapped[float] = mapped_column(sa.Float, nullable=False, server_default=sa.text("0"))
+    run_total_tokens: Mapped[int] = mapped_column(sa.BigInteger, server_default=sa.text("0"))
+    run_total_steps: Mapped[int] = mapped_column(sa.Integer, server_default=sa.text("0"), nullable=True)
+    run_created_at: Mapped[datetime] = mapped_column(DateTime, nullable=False)
+    run_finished_at: Mapped[datetime | None] = mapped_column(DateTime, nullable=True)
+    run_exceptions_count: Mapped[int] = mapped_column(sa.Integer, server_default=sa.text("0"), nullable=True)
+
+    trigger_metadata: Mapped[str | None] = mapped_column(LongText, nullable=True)
```
hj24, line 1219: Do we know the average size of this field? The row count of WorkflowArchiveLog is expected to be large (will store all workflow_runs records 90 days after creation), so we also need to control its size.

If trigger_metadata maybe large, I prefer to discuss with PM and remove this field

api/commands.py
hjlarry: yes, I removed the celery command, but forget add the days params here

api/models/workflow.py
```diff
@@ -1168,6 +1168,95 @@ def to_dict(self):
...
+            "elapsed_time": self.run_elapsed_time,
+            "total_tokens": self.run_total_tokens,
+            "total_steps": self.run_total_steps,
+            "created_at": self.run_created_at,
+            "finished_at": self.run_finished_at,
+            "exceptions_count": self.run_exceptions_count,
+        }
+
+    @property
+    def created_by_account(self):
```
hj24: Is it possible to avoid using db query in property, especially in list api scenario? Seems it will cause N+1 performance issue, we can get a batch of created_by values and get them in a single query, then make the final response in code level.

hj24 (changes requested):

api/services/retention/workflow_run/archive_paid_plan_workflow_run.py
```diff
@@ -0,0 +1,532 @@
...
+        trigger_repo = SQLAlchemyWorkflowTriggerLogRepository(session)
+        trigger_records = trigger_repo.list_by_run_id(run.id)
+        table_data["workflow_trigger_logs"] = [self._row_to_dict(row) for row in trigger_records]
+        trigger_metadata = trigger_records[0].trigger_metadata if trigger_records else None
+        return table_data, app_logs, trigger_metadata
+
+    @staticmethod
+    def _row_to_dict(row: Any) -> dict[str, Any]:
+        mapper = inspect(row).mapper
+        return {str(column.name): getattr(row, mapper.get_property_by_column(column).key) for column in mapper.columns}
```
hj24, line 467: Does it contain only db columns? Will the dynamic property be included?

api/services/retention/workflow_run/restore_archived_workflow_run.py
```diff
@@ -0,0 +1,474 @@
...
+
+            try:
+                created_at = run.created_at
+                prefix = (
+                    f"{run.tenant_id}/app_id={run.app_id}/year={created_at.strftime('%Y')}/"
+                    f"month={created_at.strftime('%m')}/workflow_run_id={run.id}"
+                )
+                archive_key = f"{prefix}/{ARCHIVE_BUNDLE_NAME}"
+                try:
+                    archive_data = storage.get_object(archive_key)
```
hj24: `repo.get_archived_run_ids` seems will implicitly open a transaction
Is it possible to move the storage operations out of the session to avoid the potential long transaction?

api/repositories/sqlalchemy_api_workflow_run_repository.py
```diff
@@ -314,6 +313,229 @@ def delete_runs_by_app(
...
+        - type in (workflow, rag-pipeline)
+        - status is an ended state
+        - optional tenant_id filter and cursor (last_seen) for pagination
+        """
+        with self._session_maker() as session:
+            stmt = (
+                select(WorkflowRun)
+                .where(
+                    WorkflowRun.created_at < end_before,
+                    WorkflowRun.type.in_([WorkflowType.WORKFLOW.value, WorkflowType.RAG_PIPELINE.value]),
```
hj24: We shouldn't hardcode the type in repo, please try pass it as args.

api/services/retention/workflow_run/archive_paid_plan_workflow_run.py
```diff
@@ -0,0 +1,532 @@
...
+            ├── manifest.json
+            ├── workflow_runs.jsonl
+            ├── workflow_app_logs.jsonl
+            ├── workflow_node_executions.jsonl
+            ├── workflow_node_execution_offload.jsonl
+            ├── workflow_pauses.jsonl
+            ├── workflow_pause_reasons.jsonl
+            └── workflow_trigger_logs.jsonl
+    """
+
```
hj24, line 103: We should also explicitly declare the ARCHIVED_TYPE here, otherwise the chatflow will be easily ignored by later maintainers
```suggestion
from core.workflow.enums import WorkflowExecutionStatus, WorkflowType

ARCHIVED_TYPE = [
    WorkflowType.WORKFLOW.value,
    WorkflowType.RAG_PIPELINE.value,
]
```

hj24 (changes requested):

api/controllers/console/app/workflow_run.py
hjlarry: 1. Presigned URLs are generated client-side with no communication with R2,  so once a user get a presigned url, he can download multiple times as he want until it expires, our ratelimit seems meanless here
2. The frontend page has a cache query, duplicated click will not request to this API

hjlarry: r2 is Zero Egress Fees,   $4.50 / million requests,  very cheap

api/models/workflow.py
hjlarry: It's only has value  when this workflow run is called by a trigger plugin, and it has these values
```
type: AppTriggerType = AppTriggerType.TRIGGER_PLUGIN
endpoint_id: str
plugin_unique_identifier: str
provider_id: str
event_name: str
icon_filename: str
icon_dark_filename: str
```

seems it under control ?

api/services/retention/workflow_run/archive_paid_plan_workflow_run.py
hjlarry: yes,  I  added more tests to the previous script to validate this
```
TEST 5: _row_to_dict column-only access
================================================================================
  [PROPERTY] inputs setter called
    [SETTER] Added internal metadata, stored data keys: ['data', '_internal_metadata']
   ✓ _row_to_dict keys: ['id', 'inputs']
   ✓ _row_to_dict['inputs']: {'data': {'key1': 'value1'}, '_internal_metadata': {'version': '1.0', 'set_via_property': True, 'timestamp': '2026-01-09T12:04:41.666923'}}
   Property call counts before: {'inputs_getter': 0, 'computed_value_getter': 0}
   Property call counts after : {'inputs_getter': 0, 'computed_value_getter': 0}
   ✓ Returned keys match mapped columns only
   ✓ inputs contains internal metadata (raw column value)
   ✓ Property getters were NOT called by _row_to_dict

✓ _row_to_dict column-only access verification completed

================================================================================
TEST 6: _row_to_dict skips relationship + property
================================================================================
   relationship loaded before: False
   relationship loaded after : False
   ✓ Relationship was not lazy-loaded by _row_to_dict
   ✓ Property getters were NOT called

✓ _row_to_dict relationship/property bypass verification completed
```

api/migrations/versions/2026_01_06_1718-9d77545f524e_.py
```diff
@@ -0,0 +1,64 @@
...
+
+"""
+from alembic import op
+import models as models
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = '9d77545f524e'
+down_revision = '7df29de0f6be'
```
hj24, line 15: check the migration version, maybe we need rebase or merge main and then regenerate the file

api/models/workflow.py
hj24: ok, maybe we can monitor its size when we do the canary and  decide if we need to remove it or not

hj24 (commented):
except the migration file version, other LGTM

PR #30333, opened 2025-12-29
fix(api): surface subscription deletion errors to users

Mairuis:
## Summary

This PR fixes a bug where trigger subscription deletion/unsubscribe errors were silently ignored during the rebuild process.

- Check `UnsubscribeResult.success` and propagate error messages to users when unsubscribe fails
- Simplify rebuild logic by removing unnecessary try/except wrapper around unsubscribe
- Refactor update API to use cleaner conditional logic with `any()`
- Remove redundant test cases that validated silent error handling behavior

Fixes #30332

## Test Plan

- [x] Verify that unsubscribe errors are now properly surfaced to users with clear error messages
- [x] Verify that successful subscription rebuilds continue to work as expected
- [x] Run existing unit tests to ensure no regressions


api/services/trigger/trigger_provider_service.py
```diff
@@ -868,111 +868,50 @@ def rebuild_trigger_subscription(
...
+            raise ValueError(f"Subscription {subscription_id} not found")
 
-                    # Clear subscription cache
-                    delete_cache_for_subscription(
-                        tenant_id=tenant_id,
-                        provider_id=subscription.provider_id,
-                        subscription_id=subscription.id,
-                    )
+        credential_type = CredentialType.of(subscription.credential_type)
+        if credential_type not in [CredentialType.OAUTH2, CredentialType.API_KEY]:
```
Yeuoly: I suggest to use `set` rather than `list`

api/controllers/console/workspace/trigger_providers.py
```diff
@@ -345,50 +344,35 @@ def post(self, subscription_id: str):
...
         try:
-            # rename only
-            if (
-                args.name is not None
-                and args.credentials is None
-                and args.parameters is None
-                and args.properties is None
-            ):
+            # for rename only, update the name
+            if args.name is not None and not any((args.credentials, args.parameters, args.properties)):
```
Yeuoly: I guess this branch can be merged into L357? They all just update the record in DB without interaction to thirdparty platform.

Yeuoly: Also, I suppose a better comments may help developers understand why `CredentialType.UNAUTHORIZED` is handled specifically.

Yeuoly (changes requested):

api/services/trigger/trigger_provider_service.py
```diff
@@ -868,111 +868,50 @@ def rebuild_trigger_subscription(
...
+        if not subscription:
+            raise ValueError(f"Subscription {subscription_id} not found")
 
-                    # Clear subscription cache
-                    delete_cache_for_subscription(
-                        tenant_id=tenant_id,
-                        provider_id=subscription.provider_id,
-                        subscription_id=subscription.id,
-                    )
+        credential_type = CredentialType.of(subscription.credential_type)
```
Yeuoly, line 878: There is no need to use `CredentialType.of` as `CredentialType` is a kind of `StrEnum`

Mairuis: I tried, but there are some CI errors 😭
image

Yeuoly: Emm, let's just keep it unchanged, this scenario is not so much fitting.

crazywoola (approved):

asukaminato0721 (approved):

PR #30397, opened 2025-12-30
fix(web): migrate es-toolkit/compat API (uniqueId & isEmpty)

### LINKED ISSUE #30245: Migrate es-toolkit/compat: API modifications (uniqueId & isEmpty)
Part of #30243

## Scope

Handle 4 files that require API modifications due to functions only available in compat.

### uniqueId (3 files)
- `web/app/components/workflow/nodes/http/hooks/use-key-value-list.ts`
- `web/app/components/workflow/nodes/http/components/edit-body/index.tsx`
- `web/app/components/workflow/nodes/question-classifier/components/class-item.tsx`

### isEmpty (1 file)
- `web/app/components/plugins/install-plugin/utils.ts`

### Option 1: uniqueId → React useId (Recommended)

**Before:**
```typescript
import { uniqueId } from 'es-toolkit/compat'
const [instanceId, setInstanceId] = useState(() => uniqueId())
```

**After:**
```typescript
import { useId } from 'react'
const instanceId = useId()
```

### Option 2: isEmpty → Custom implementation

**Before:**
```typescript
import { isEmpty } from 'es-toolkit/compat'
isEmpty(obj)
```

**After:**
```typescript
const isEmptyObject = (obj: any) => !obj || Object.keys(obj).length === 0
isEmptyObject(obj)
```

Or create a utility function in `web/utils/toolkit-helpers.ts`:
```typescript
export const isEmpty = (value: any): boolean => {
  if (value == null) return true
  if (Array.isArray(value) || typeof value === 'string') return value.length === 0
  if (value instanceof Map || value instanceof Set) return value.size === 0
  if (typeof value === 'object') return Object.keys(value).length === 0
  return false
}
```

## Checklist

- [ ] Replace uniqueId with React useId in 3 files
- [ ] Replace isEmpty with custom implementation in 1 file
- [ ] Run `pnpm type-check:tsgo`
- [ ] Run `pnpm test`
- [ ] Manual testing for affected components

## Estimated Effort

2-3 hours

majiayu000:
## Summary
- Replace `uniqueId` from es-toolkit/compat with native implementation
- Replace `isEmpty` from es-toolkit/compat with native implementation
- Add utility functions in `web/utils/toolkit-helpers.ts`

## Changes
- `uniqueId`: simple counter-based ID generator
- `isEmpty`: checks null, arrays, strings, Maps, Sets, objects

## Test plan
- [x] Add unit tests for `uniqueId` (3 tests)
- [x] Add unit tests for `isEmpty` (12 tests)
- [x] All 15 tests pass
- [x] `pnpm type-check:tsgo` passes
- [x] `pnpm lint` passes

Fixes #30245


zhsama:
Thank you for the contribution. Closing this PR as:

1. The project still heavily relies on `es-toolkit/compat` elsewhere (49+ files), so partial replacement provides minimal benefit
2. The custom implementations have semantic differences from the original (edge case behavior risks)

To avoid potential regressions, closing for now. A full migration with proper compatibility testing would be more appropriate.

PR #30572, opened 2026-01-05
fix: system model selector loading state flash

### LINKED ISSUE #30569: System Model Selector flashes 'not configured' state on initial load
The System Model Selector component in the Model Provider page initially displays a 'not configured' state (yellow warning) before the actual data is loaded, causing a visible flash. This is because the loading state from the data fetching hooks is ignored, and undefined data is interpreted as 'not configured'.

We need to:
1. Capture the isLoading state from useDefaultModel.
2. Display a proper loading indicator or defer the 'not configured' check until loading is complete.
3. Follow TanStack Query best practices.

lyzno1:
Fixes #30569 
## Screenshots

| Before | After |
|--------|-------|
|CleanShot 2026-01-05 at 17 14 47|CleanShot 2026-01-05 at 17 15 54|

## Checklist


web/app/components/header/account-setting/model-provider-page/system-model-selector/index.tsx
```diff
@@ -129,13 +131,16 @@ const SystemModel: FC<SystemModelSelectorProps> = ({
...
       }}
     >
-      <PortalToFollowElemTrigger onClick={() => setOpen(v => !v)}>
+      <PortalToFollowElemTrigger onClick={() => !isLoading && setOpen(v => !v)}>
         <Button
           className="relative"
           variant={notConfigured ? 'primary' : 'secondary'}
           size="small"
+          disabled={isLoading}
         >
```
hyoban, line 140: The disabled and click events are not on the same element.

lyzno1: Good catch. Since the trigger is a div wrapper, I added the `!isLoading` condition in its `onClick` handler to prevent the portal from opening when the button is disabled.

lyzno1: The trigger is a layout wrapper for the portal, so it's decoupled from the Button which handles the visual style.

lyzno1: https://github.com/langgenius/dify/pull/30572/changes/be935f9d1d18d01f63f807b007fcf91b44c8b3a6#diff-813a064d9bbc749c9cf6a5870b80df8c0e3458219f175fed8c599011783b2896R134

crazywoola (approved):

PR #30675, opened 2026-01-07
fix: signin page stuck on loading when refresh token valid but access token expired

### LINKED ISSUE #30674: Fix: Signin page stuck on loading when refresh token is valid but access token expired
## Bug Description

When a user visits the `/signin` page with an expired access token but a valid refresh token, the page gets stuck on the loading spinner indefinitely.

## Root Cause

Two issues were identified:

### Issue 1: `isLoading` includes `logged_in` state (`normal-form.tsx:31`)

```typescript
const isLoading = isCheckLoading || loginData?.logged_in || isInitCheckLoading
```

When `logged_in = true`, `isLoading` is always `true`. If the redirect (`router.replace`) fails or is delayed, the page remains stuck on loading forever.

### Issue 2: Non-401 errors treated as logged in (`use-common.ts:225-228`)

```typescript
catch (e: any) {
  if (e.status === 401)
    return { logged_in: false }
  return { logged_in: true }  // Bug: 500, 403, network errors are treated as logged in
}
```

## Steps to Reproduce

1. User logs in normally
2. User leaves the system for more than 60 minutes (access token expires) but less than 30 days (refresh token still valid)
3. User directly navigates to `/signin` (via bookmark, manual URL entry, or browser session restore)
4. Page gets stuck on loading spinner

## Expected Behavior

The signin page should either:
- Redirect successfully to `/apps` if the user is actually logged in
- Show the login form if authentication fails

## Actual Behavior

The page displays a loading spinner indefinitely with no user feedback.

## Affected Time Window

- Access token expires after: 60 minutes (default)
- Refresh token expires after: 30 days (default)
- Bug can be triggered: 60 minutes ~ 30 days after login

## Proposed Fix

1. Remove `loginData?.logged_in` from `isLoading` condition and use explicit `isRedirecting` state
2. Treat any error (not just 401) as "not logged in" in the login check

lyzno1:
## Summary

- Fix signin page getting stuck on loading when user has valid refresh token but expired access token
- Remove `loginData?.logged_in` from `isLoading` condition and use explicit `isRedirecting` state
- Fix error handling in `useIsLogin` hook to treat any error as "not logged in"

## Test plan

- [ ] Verify signin page loads correctly when not logged in
- [ ] Verify redirect works when user is already logged in
- [ ] Verify page doesn't get stuck when access token expires but refresh token is valid
- [ ] Verify error states (500, network errors) don't cause infinite loading

Fixes #30674

### DIFF
--- a/web/app/signin/normal-form.tsx
+++ b/web/app/signin/normal-form.tsx
@@ -28,7 +28,8 @@ const NormalForm = () => {
   const message = decodeURIComponent(searchParams.get('message') || '')
   const invite_token = decodeURIComponent(searchParams.get('invite_token') || '')
   const [isInitCheckLoading, setInitCheckLoading] = useState(true)
-  const isLoading = isCheckLoading || loginData?.logged_in || isInitCheckLoading
+  const [isRedirecting, setIsRedirecting] = useState(false)
+  const isLoading = isCheckLoading || isInitCheckLoading || isRedirecting
   const { systemFeatures } = useGlobalPublicStore()
   const [authType, updateAuthType] = useState<'code' | 'password'>('password')
   const [showORLine, setShowORLine] = useState(false)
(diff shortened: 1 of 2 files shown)


lyzno1:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

iamjoel (approved):

PR #30785, opened 2026-01-09
fix(plugins): support searching installed plugins by all locale labels issue #30784

Yevanchen:
before：
image
after：
image


zhsama:
It has fixed at #30501

PR #30796, opened 2026-01-09
feat: Support displaying non-image tool files in agent runs

IthacaDream:
> [!IMPORTANT]
>

## Summary

Fixes #30795

- This PR delivers a minimal viable improvement to surface non-image files produced by tools during agent runs.
- On the web, to minimize impact, I intentionally reuse and extend the existing `VisionFile` shape instead of introducing a new type. In the future, we can unify on `FileResponse` for a cleaner model and broader refactor.

### Changes:
- Backend:
    - Enrich `message_file` stream events with additional metadata (upload_file_id, transfer_method, mime_type, filename, size, related_id, extension) to support non-image files.
    - Extend tool engine handling for `BINARY_LINK` so non-image outputs with mime types are captured and persisted as `MessageFile` entries.
- Frontend:
    - Extend `VisionFile` with optional fields (mime_type, filename, size, related_id, extension) to consume the enriched stream event with minimal Web impact.
### Compatibility:
- The `message_file` event remains unchanged in shape for existing consumers; new fields are additive and optional on the frontend.
- Preserves historical data paths: when `ToolFile `metadata is unavailable, sensible defaults are provided, ensuring the UI continues to function.
### Testing:
- Verified single-file and multi-file scenarios, including historical messages, to ensure no regressions.
- Confirmed streaming behavior and rendering work for non-image files (e.g., .pptx, .docx, .pdf) and that image handling remains unaffected.
### Future Work:
- Consider unifying frontend file handling on `FileResponse` to reduce duplication and improve consistency across views.

## Screenshots

| Before | After |
|--------|-------|
| before| after
|

## Checklist

### DIFF
--- a/api/core/app/entities/task_entities.py
+++ b/api/core/app/entities/task_entities.py
@@ -153,6 +153,13 @@ class MessageFileStreamResponse(StreamResponse):
     type: str
     belongs_to: str
     url: str
+    upload_file_id: str | None = None
+    transfer_method: str
+    mime_type: str
+    filename: str
+    size: int
+    related_id: str | None = None
+    extension: str
 
 
 class MessageReplaceStreamResponse(StreamResponse):
--- a/api/core/tools/tool_engine.py
+++ b/api/core/tools/tool_engine.py
@@ -254,6 +254,8 @@ class ToolEngine:
                         ensure_ascii=False,
                     )
                 )
+            elif response.type == ToolInvokeMessage.MessageType.BINARY_LINK:
+                parts.append("A result with a file link has been returned, please tell user to check it.")
             else:
                 parts.append(str(response.message))
 
@@ -310,6 +312,16 @@ class ToolEngine:
                         else "application/octet-stream",
                         url=cast(ToolInvokeMessage.TextMessage, response.message).text,
                     )
+            elif response.type == ToolInvokeMessage.MessageType.BINARY_LINK:
+                # check if there is a mime type in meta
+                if response.meta and "mime_type" in response.meta:
+                    result = ToolInvokeMessageBinary(
+                        mimetype=response.meta.get("mime_type", "application/octet-stream")
+                        if response.meta
+                        else "application/octet-stream",
+                        url=cast(ToolInvokeMessage.TextMessage, response.message).text,
+                    )
+                    yield result
 
     @staticmethod
     def _create_message_files(
--- a/api/core/tools/tool_engine.py
+++ b/api/core/tools/tool_engine.py
@@ -316,9 +316,7 @@ class ToolEngine:
                 # check if there is a mime type in meta
                 if response.meta and "mime_type" in response.meta:
                     result = ToolInvokeMessageBinary(
-                        mimetype=response.meta.get("mime_type", "application/octet-stream")
-                        if response.meta
-                        else "application/octet-stream",
+                        mimetype=response.meta["mime_type"],
                         url=cast(ToolInvokeMessage.TextMessage, response.message).text,
                     )
                     yield result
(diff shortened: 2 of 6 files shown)


api/core/tools/tool_engine.py
```diff
@@ -310,6 +312,16 @@ def _extract_tool_response_binary_and_text(
...
                     )
+            elif response.type == ToolInvokeMessage.MessageType.BINARY_LINK:
+                # check if there is a mime type in meta
+                if response.meta and "mime_type" in response.meta:
+                    result = ToolInvokeMessageBinary(
+                        mimetype=response.meta.get("mime_type", "application/octet-stream")
+                        if response.meta
+                        else "application/octet-stream",
+                        url=cast(ToolInvokeMessage.TextMessage, response.message).text,
+                    )
```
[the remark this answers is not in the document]
IthacaDream: > medium
> 
> The `if response.meta` check within the `mimetype` assignment is redundant. The outer `if response.meta and "mime_type" in response.meta:` condition already ensures that `response.meta` is not `None`. Furthermore, since the key `"mime_type"` is guaranteed to exist, you can access it directly instead of using `.get()`. This simplification improves code readability and removes unreachable code.
> 
> ```python
>                     result = ToolInvokeMessageBinary(
>                         mimetype=response.meta["mime_type"],
>                         url=cast(ToolInvokeMessage.TextMessage, response.message).text,
>                     )
> ```

yes, this is a great advice.

IthacaDream:
@crazywoola PTAL

api/core/app/entities/task_entities.py
```diff
@@ -153,6 +153,13 @@ class MessageFileStreamResponse(StreamResponse):
     type: str
     belongs_to: str
     url: str
+    upload_file_id: str | None = None
```
QuantumGhost, line 156: The `MessageFileStreamResponse` is a part of the service API contract. I would suggest minimize  fields returned.

IMO only the following fields should be included:

- `mime_type`
- `filename`
- `size`

Fields not utilized by the frontend UI should also be removed from the response.

IthacaDream: @QuantumGhost thanks for your review.

I believe this PR is now staled, because I've noticed that PR #31293 introduced significant changes to the file structure. 

https://github.com/langgenius/dify/blob/24ebe2f5c6f31c5be430531ca1a02f83af35c51e/web/app/components/base/chat/chat/hooks.ts#L433-L443

Since I’m unsure of the impact of continuing with these changes, please do not merge this PR for now.

@fatelei Could you provide some help?  I don't understand the purpose of the `convertedFile` structure. I believe this can be achieved by adjusting the `message_file` event structure returned by the backend.

fatelei: const convertedFile: FileEntity is for multi modal image display

fatelei: it should test agent chat app with image display and media display in tool

IthacaDream:
Hi, all reviewers:

I've noticed that PR https://github.com/langgenius/dify/pull/31293 introduced significant changes to `onFile` processing logic.

https://github.com/langgenius/dify/blob/24ebe2f5c6f31c5be430531ca1a02f83af35c51e/web/app/components/base/chat/chat/hooks.ts#L427-L443

the file.type field changes semantics: it’s treated as a “simple category” (image|video|audio) and then rewritten/normalized into a MIME type string (e.g. image/png). 
This makes type ambiguous (sometimes category, sometimes MIME), and in some cases can incorrectly downgrade an already-MIME value to application/octet-stream and set supportFileType to a fallback.


This semantic mismatch can impact downstream logic that relies on a stable “file category”, especially `getProcessedFiles()` in：

https://github.com/langgenius/dify/blob/24ebe2f5c6f31c5be430531ca1a02f83af35c51e/web/app/components/base/file-uploader/utils.ts#L182

which sends type: fileItem.supportFileType to the backend. If supportFileType is derived from an inconsistent type, the request payload may report the wrong file type.


For the reasons outlined above, I need to explore alternative solutions and will temporarily close this PR

PR #30805, opened 2026-01-10
chore(web): upgrade pnpm to 10.28.0

lyzno1:
## Summary
- Upgrade pnpm package manager from 10.27.0 to 10.28.0 using `corepack use pnpm@latest`

## Test plan
- [x] Lockfile validation passed
- [x] Pre-commit hooks passed

Closes #30804

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -3,7 +3,7 @@
   "type": "module",
   "version": "1.11.2",
   "private": true,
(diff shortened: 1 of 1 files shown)


hyoban:
Can you reduce the frequency of upgrading pnpm? I think it's better to only upgrade major versions, critical fixes, or along with most other dependencies.

lyzno1:
> Can you reduce the frequency of upgrading pnpm? I think it's better to only upgrade major versions, critical fixes, or along with most other dependencies.

OK.

PR #25543, opened 2025-09-11
fix: delete get upload file endpoint

WTW0313:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods

PR #23320, opened 2025-08-03
feat: Add Download Button to UI for Knowledge Resource Source Files

somethingwentwell:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 2. Ensure there is an associated issue and you have been assigned to it
> 3. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

This pull request addresses issues #9547 and #10296 by introducing a download button in the chat UI for source files linked in the citations section of chat messages. It leverages the new API introduced in PR #11899 to download upload file. 

## Screenshots

| Before | After |
|--------|-------|
|Screenshot 2025-08-03 at 9 49 38 AM Screenshot 2025-08-03 at 9 49 44 AM | Screenshot 2025-08-03 at 9 42 20 AM Screenshot 2025-08-03 at 9 42 25 AM |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods

PR #31068, opened 2026-01-15
feat: Download the uploaded files

BeautyyuYanli:
solve #31085 

## Screenshot

**Download from the knowledge base documents list**

**Download from the konwledge base documents selections (bulk download as a zip file)**

**Download from the citation popup header**

## API

- download
- bulk download: limit the max files to zip to a constant 100. The zipped file is stored in disk when streaming downloading


kurokobo:
@BeautyyuYanli 
Thanks for the PR.
APIs and UI buttons for downloading documents existed in the past, but they were removed due to "security concerns."

- https://github.com/langgenius/dify/pull/23320
- https://github.com/langgenius/dify/pull/24459
- https://github.com/langgenius/dify/pull/25543

Does this PR implement the download feature in a way that addresses the security concerns LangGenius had at the time, which led to the removal of the download functionality?

If there are any discussions or background information, I’d appreciate it if you could share them. I’m also looking forward to seeing the download feature implemented, but I imagine there might be use cases where users must not be allowed to download files.

BeautyyuYanli:
hi @kurokobo , confirmed by @JohnJyong , I belive the problem is already covered by `cloud_edition_billing_rate_limit_check`
https://github.com/langgenius/dify/pull/31068/changes#diff-dc15435cf42211893f559907db1fab2da53a45fdd3a4874939056e6613be100eR997
https://github.com/langgenius/dify/pull/31068/changes#diff-dc15435cf42211893f559907db1fab2da53a45fdd3a4874939056e6613be100eR1020

kurokobo:
@BeautyyuYanli 
Thank you for checking! That's really great news for me.
I'm looking forward to the release, and I appreciate you moving this forward!

web/app/components/base/chat/chat/citation/popup.tsx
```diff
@@ -32,10 +34,39 @@ const Popup: FC<PopupProps> = ({
 }) => {
   const { t } = useTranslation()
   const [open, setOpen] = useState(false)
+  const [isDownloading, setIsDownloading] = useState(false)
```
hyoban: Please avoid manually defining and maintaining the loading state; you can use `useMutation`.

api/controllers/console/datasets/datasets_document.py
```diff
@@ -73,6 +95,67 @@ def _get_or_create_model(model_name: str, field_def):
     return existing
 
 
+def _get_upload_file_id_from_upload_file_document(
```
JohnJyong: `_get_upload_file_id_from_upload_file_document `, `_get_upload_file_from_upload_file_document `,  `_dedupe_zip_filename ` and `get_documents_and_upload_files_for_zip ` these methods are best written in a service module; for example, zip files can be added in a `fileservice`. @BeautyyuYanli

web/app/components/base/chat/chat/citation/popup.tsx
BeautyyuYanli: changed to `handleDownload` by a invisable <a> tag

api/controllers/console/datasets/datasets_document.py
BeautyyuYanli: zip related logic have been moved to the file service, while others are downloading specific logic and won't be reused, so move to the dataset service.

JohnJyong (approved):
backend logic is LGTM , @WTW0313 pls add your review suggestion , thanks

WTW0313 (approved):
LGTM

543385350:
Hello, could you please provide an externally accessible API interface in addition to the file download interface in the console? This function is of great necessity to me. Thank you.

BeautyyuYanli:
> Hello, could you please provide an externally accessible API interface in addition to the file download interface in the console? This function is of great necessity to me. Thank you.

GET /datasets/<uuid:dataset_id>/documents/<uuid:document_id>/download

entropy2333:
After migrating a knowledge base to the new pipeline mode, I observed that the `data_source_info_dict` field in the document object has been replaced by the `data_source_info` field. As a result, the `upload_file_id` is now populated from `data_source_info.real_file_id`. This change causes document downloads to fail because the function `_get_upload_file_id_for_upload_file_document` expects to find the upload file ID within the `data_source_info` dictionary.

**Steps to reproduce**:
1. Migrate an existing knowledge base to pipeline mode.
2. Retrieve documents from the migrated knowledge base.
3. Inspect the `data_source_info` and `data_source_info_dict` attributes.
4. Compare the original `data_source_info_dict.upload_file_id` with `data_source_info.real_file_id` – they are identical.

**Expected behavior**:
Documents should download successfully after migration, using the correct upload file ID regardless of the internal field changes.

**Actual behavior**:
Downloads fail because `_get_upload_file_id_for_upload_file_document` checks the `data_source_info` dictionary (or expects a specific key) and cannot retrieve the upload file ID due to the altered data structure.

PR #31075, opened 2026-01-15
fix: use configured max_retries in HTTP Request node error message

GlobalStar117:
## Summary

Fixes #31071 - The error message for HTTP Request node retries was showing incorrect retry count

## Problem

When the HTTP Request node's retry feature was enabled, the error message always showed "0 retries" because the `Executor` was initialized with a hardcoded `max_retries=0` instead of the configured value.

## Solution

Modified `_run` method to use the configured `max_retries` from `retry_config` when retry is enabled.

## Changes

- Modified `api/core/workflow/nodes/http_request/node.py`:
  - Read `max_retries` from `retry_config` when retry is enabled
  - Pass the correct value to `Executor`

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=146701565


kurokobo:
@GlobalStar117 
Thank you for the PR!
However, with the changes in this PR, the number of retries will unintentionally increase.
For example, if you set the retry count to 3 at the node level, it will actually retry 9 times, and if you set it to 5, it will retry 25 times. That's probably not the intended behavior.

Originally, retries were working correctly. Even though zero was hard-coded, that function including it gets retried specified times by a higher-level mechanism.
So, we only need to update the error message to avoid confusion, and it's not necessary to change the actual number of retries.

kurokobo:
@GlobalStar117 @asukaminato0721 
If we want to solve this issue with minimal changes, I think it’s fine to simply remove the message in the error output that might be misunderstood as “retried 0 times”.

For example, after reverting current modification, I think the simplest fix would be something like the following.
Here, in the case of a `max_retries_exceeded_error`, we're replacing the exception with our own custom error message instead of just using `str(e)`.

```diff
diff --git a/api/core/workflow/nodes/http_request/executor.py b/api/core/workflow/nodes/http_request/executor.py
index 429f8411a6..8763d4ffe9 100644
--- a/api/core/workflow/nodes/http_request/executor.py
+++ b/api/core/workflow/nodes/http_request/executor.py
@@ -366,7 +366,9 @@ class Executor:
                 **request_args,
                 max_retries=self.max_retries,
             )
-        except (self._http_client.max_retries_exceeded_error, self._http_client.request_error) as e:
+        except self._http_client.max_retries_exceeded_error as e:
+            raise HttpRequestNodeError(f"HTTP request failed for URL {self.url}") from e
+        except self._http_client.request_error as e:
             raise HttpRequestNodeError(str(e)) from e
         # FIXME: fix type ignore, this maybe httpx type issue
         return response
```

How do you think this approach?
Of course, ideally, we should display the exact number of retries, but implementing that is a bit costly.

image

kurokobo:
Since there has been a lack of activity on this PR for a month, I've made an alternative PR https://github.com/langgenius/dify/pull/32406 to address my comment above.

kurokobo:
This issue can be closed since my alternative PR #32406 has been merged.

@GlobalStar117 
Thanks for your contribution!

PR #31277, opened 2026-01-20
fix(workflow): prevent Start node from being deleted

GlobalStar117:
## Summary

This PR fixes a bug where the Start node could be accidentally deleted in Workflow mode by pressing the Delete key, breaking the entire workflow with no way to recover.

**Fixes:** #31275

## Problem

- Users could select the Start node and press Delete to remove it
- Once deleted, the Start node could not be added back
- Ctrl+Z (undo) didn't restore it
- The workflow becomes completely broken

## Root Cause

In `use-available-nodes-meta-data.ts`, the Start node was only marked as undeletable in **Chat mode**:

```typescript
isUndeletable: isChatMode, // start node is undeletable in chat mode
```

This meant in regular Workflow mode (`isChatMode = false`), the Start node could be deleted.

## Solution

Changed to always protect the Start node:

```typescript
isUndeletable: true, // start node is always undeletable - it's the workflow entry point
```

The Start node is the essential entry point for all workflows and should never be deletable.

## Testing

1. Create a new Workflow
2. Select the Start node
3. Press Delete key
4. ✅ Start node is NOT deleted

---

[Gittensor contribution - GlobalStar117]


lyzno1:
Thanks for the contribution. Your pull request contained other backend code. Secondly, the start node of the workflow cannot be set to be non-deletable. We'll temporarily close this pull request and consider fixing this issue once a solution is available.

PR #31292, opened 2026-01-20
fix(web): disable HTML escaping for form field validation messages

lyzno1:
Fixes #31290


hyoban (changes requested):
Try the following options for the i18n instance

https://react.i18next.com/guides/quick-start#configure-i18next

```
interpolation: {
  escapeValue: false // react already safes from xss
}
```

hyoban (approved):

Issue #31382, opened 2026-01-22
[Refactor/Chore] Define workflow external import boundaries


laipz8200:
### Description
Add an import-linter contract that explicitly defines which external modules the workflow package may depend on, while documenting existing exceptions.

### Motivation
Make workflow dependency boundaries explicit so future changes are easier to reason about and violations are caught early.

### Additional Context
N/A

PR #31385, opened 2026-01-22
refactor: Move workflow node factory to app workflow

### LINKED ISSUE #31384: Move workflow node factory under app workflow
## Summary
Please move the workflow node factory under the app workflow package so app-level workflow execution depends on a single app-owned factory.
Also align the workflow dependency rules so current external dependencies are explicitly allowed and linting stays consistent.

## Motivation
This clarifies ownership of the node factory used by app runners and workflow entry points.
Keeping the dependency rules aligned with current usage reduces lint noise and makes the intended boundaries explicit.

laipz8200:
## Summary
- relocate DifyNodeFactory under core.app.workflow and update workflow entry points, runners, and tests
- align the workflow node registry import guard with the new factory location
- document current workflow external dependency exceptions to keep linting consistent

Closes #31384

### DIFF
--- a/api/.importlinter
+++ b/api/.importlinter
@@ -27,7 +27,9 @@ ignore_imports =
     core.workflow.nodes.iteration.iteration_node -> core.workflow.graph_events
     core.workflow.nodes.loop.loop_node -> core.workflow.graph_events
 
-    core.workflow.nodes.node_factory -> core.workflow.graph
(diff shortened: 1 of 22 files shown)


fatelei (approved):

WH-2099:
Maybe #31382 should be closed.

PR #31566, opened 2026-01-26
feat: support opaque_body from plugin LLM calls

BeautyyuYanli:
## Summary

This is a part of langgenius/dify-official-plugins#2492

Adds end-to-end support for preserving assistant `opaque_body` produced by plugin LLM calls.

- Adds message-level and content-level `opaque_body` fields in model runtime prompt/message entities.
- Parses `opaque_body` from plugin LLM invocations (including streaming tool-call parsing).
- Captures assistant `opaque_body` during app run streaming, persists it into `message_metadata` as `assistant_opaque_body`, and restores it when rebuilding memory prompt history.

## Testing

- Added unit tests: `api/tests/unit_tests/core/model_runtime/__base/test_llm_invoke_opaque_body.py`


api/core/model_runtime/entities/message_entities.py
```diff
@@ -69,6 +69,7 @@ class PromptMessageContent(ABC, BaseModel):
     """
 
     type: PromptMessageContentType
+    opaque_body: JsonValue | None = None
```
QuantumGhost, line 72: Since the `PromptMessageContent` is the parent class for `*PromptMessageContent` classes and the `opaque_body` are shared by all subclasses, I would like to know what's the semantics for a specific subclass to have both `opaque_body` and other fields set?

For example, what's the semantics for a `TextPromptMessageContent` to have both `data` and `opaque_body` set. Should we send both data AND `opaque_body` or just ignore body and only sent `opaque_body`?

If other fields should be ignored when `opaque_body` is set. I'm wondering if introducing a subclass named `OpaqueMessageContent` to `PromptMessageContent` is better.

BeautyyuYanli: The opaque_body is supposed to be injected to the raw request to the upstream API, so the behavior of `TextPromptMessageContent(data="hello world", opaque_body={"foo": "bar"})` whould be sent to upstream like 
`{"messages": [{"role": "user", "content": "hello world"}], "foo": "bar"}`

QuantumGhost: I suggest documenting the semantics and behavior of `opaque_body`. 

Additionally, I noticed that `opaque_body` is annotated as  `JsonValue`,  which includes scalar types (e.g., `number`, `string`, `boolean`). Since these cannot be merged directly into the upstream JSON payload, should we refine the type of opaque_body to `Mapping[str, JsonValue]`? Or is there a specific use case I might be overlooking?

crazywoola:
Hi @BeautyyuYanli, thank you for taking the time to contribute to Dify.

### Why this is being closed

This PR was opened before February 2026 (`2026-01-26`) and touches more than 3 files (`4` files). Older broad PRs are very likely to conflict with the current codebase and are difficult to review safely after the surrounding code has moved on. Per our triage rules, we are closing old/conflicting PRs of this size to keep the review queue manageable.

### Next steps

If this change is still needed, please open a fresh PR from the latest `main`, keep the scope small, and link the related issue using GitHub's closing syntax such as `Fixes #123`, `Closes #123`, or `Resolves #123`. Thanks again for your work and understanding.

PR #31567, opened 2026-01-26
fix: clear all default providers in tenant for enterprise mode

GareArc:
## Summary
- Fixes EE-1511: When setting a new default tool credential in enterprise mode, properly clears ALL existing defaults for that provider in the tenant

## Problem
In enterprise mode, credentials synced from global have a different `user_id` than locally-created credentials. The existing code only cleared defaults for credentials with the same `user_id`, resulting in multiple credentials having `is_default=true` simultaneously.

This caused the credential usage check to incorrectly show workflows as "using" a credential that was no longer the effective default.

## Solution
- Enterprise mode: Clear all `is_default=true` for the provider in the tenant (regardless of `user_id`)
- Non-enterprise mode: Preserve original behavior (only clear defaults for current user)

## Related
- Companion PR in enterprise repo handles detecting default credential usage in workflows


GareArc:
@gemini-code-assist review

zhangx1n (approved):
LGTM

Issue #31627, opened 2026-01-28
[Refactor/Chore] Refactor plugin model schema cache to be process-global to prevent redundant Daemon API calls


NieRonghua:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

Currently, plugin_model_schemas is implemented using RecyclableContextVar . This causes the cache to be reset with every request (or thread recycle), rendering it ineffective for cross-request caching. As a result, the system re-fetches the model schema from the Plugin Daemon for every single API request, leading to unnecessary network overhead and latency.

This refactor changes plugin_model_schemas and its corresponding lock to be global process-level variables ( dict and Lock ). This ensures that once a model schema is fetched, it remains cached in memory for the lifetime of the process, significantly reducing internal API calls.

### Motivation

While debugging high latency and log volume, we observed that every request to the model API triggered a fetch to the Plugin Daemon:
```
INFO ... [base.py:157] - [Plugin Stream Response] Status: 200, URL: .../dispatch/model/schema
```
Tracing the code revealed that `RecyclableContextVar` was isolating the cache per request context. Since model schemas are relatively static and the cache key already includes `tenant_id` (handling multi-tenancy correctly), a global in-memory cache is the correct approach to improve performance.

### Additional Context

**Changes:**
1.  **`api/contexts/__init__.py`**:
    -   Changed `plugin_model_schemas` from `RecyclableContextVar` to a global `dict`.
    -   Changed `plugin_model_schema_lock` from `RecyclableContextVar` to a global `threading.Lock`.
2.  **`api/core/model_runtime/model_providers/model_provider_factory.py`** & **`api/core/model_runtime/model_providers/__base/ai_model.py`**:
    -   Removed `RecyclableContextVar` initialization logic.
    -   Updated code to access the global dictionary directly.

PR #31634, opened 2026-01-28
refactor(model): Refactor plugin model schema cache to be process-global to prevent redundant Daemon API calls

NieRonghua:
> [!IMPORTANT]
>

## Summary

Currently, `plugin_model_schemas` is implemented using `RecyclableContextVar`. This causes the cache to be reset with every request (or thread recycle), rendering it ineffective for cross-request caching. As a result, the system re-fetches the model schema from the Plugin Daemon for every single API request, leading to unnecessary network overhead and latency.

This refactor changes `plugin_model_schemas` and its corresponding lock to be global process-level variables (`dict` and `Lock`). This ensures that once a model schema is fetched, it remains cached in memory for the lifetime of the process, significantly reducing internal API calls.

## Screenshots

| Before | After |
|--------|-------|
| Clipboard_Screenshot_1769569761 | Clipboard_Screenshot_1769570075|

## Checklist

#31627

### DIFF
--- a/api/contexts/__init__.py
+++ b/api/contexts/__init__.py
@@ -29,11 +29,8 @@ plugin_model_providers_lock: RecyclableContextVar[Lock] = RecyclableContextVar(
     ContextVar("plugin_model_providers_lock")
 )
 
-plugin_model_schema_lock: RecyclableContextVar[Lock] = RecyclableContextVar(ContextVar("plugin_model_schema_lock"))
-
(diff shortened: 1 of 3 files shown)


laipz8200 (changes requested):
Hi @NieRonghua! Thanks so much for pointing this out.

I noticed that the current cache pattern could lead to unlimited memory growth. To address this, I think it would be better to:
1. Move the cache into Redis.
2. Set a TTL for the entries.

This approach also has the added benefit of avoiding a data race. Thanks again for your help!

laipz8200:
Moved to #31689

Issue #31680, opened 2026-01-28
Proposal: Add ty type checking to api/ 


BeautyyuYanli:
## Summary
Propose adding **`ty`** as an additional Python type checker for the `api/` codebase, rolled out incrementally so it is safe to adopt and easy to review.

Key points:
- Run via `uv`: `cd api && uv run ty check`
- Land in small PRs, grouped by dependency order (**A1 → A2 → B → C → D → E**) to avoid big-bang changes.
- When narrowing is required, prefer **runtime validation** (`isinstance`, Pydantic `BaseModel` / `TypeAdapter`) over `cast()`.

## Motivation
The `api/` layer touches many dynamic boundaries (request payloads, workflow graphs, external SDKs). `ty` provides another static check that can catch issues earlier and keep types consistent, but adopting it all-at-once would be disruptive.

This proposal introduces `ty` in a way that:
- doesn’t block developers on day 1
- provides a clear, reviewable path to increase coverage
- encourages safer runtime validation at boundaries (where static typing is inherently limited)

### PR0 (bootstrap): enable `ty` but exclude everything
**Goal:** Make `make type-check` pass trivially while introducing the tool + wiring.

Changes:
- Add `api/ty.toml` with `[src].exclude` covering all `api/` production code paths (effectively a no-op check initially).
- Update the root `Makefile` so **`make type-check`** includes `ty` (e.g. runs `cd api && uv run ty check`).
- (Optional) Add a CI step that runs `make type-check` to ensure consistency.

Acceptance:
- `make type-check` passes immediately after PR0.
- No developer workflow disruption.

### PR A1: foundational typing/contracts
Scope (examples): protocol boundaries, provider plumbing, low-level helpers.

Changes:
- Unexclude only the A1 paths from `api/ty.toml`.
- Fix `ty` errors with runtime checks where needed.

Acceptance:
- `cd api && uv run ty check` passes.

### PR A2: workflow engine & nodes
Scope: workflow execution, graph runtime, node factory, node implementations.

Changes:
- Unexclude A2 paths.
- Replace unsafe assumptions with runtime validation.

Acceptance:
- `ty check` passes.

### PR B: prompt + app runtime boundaries
Scope: prompt transforms and app runner boundaries.

Acceptance:
- `ty check` passes.

### PR C: services/controllers/fields/libs
Scope: request/service layers where untyped dict/list payloads enter.

Acceptance:
- `ty check` passes.

### PR D: observability + integrations
[...]

PR #31721, opened 2026-01-29
chore(typing): reduce ty excludes for A1

BeautyyuYanli:
Relates to #31680.

- Fix ty typing issues in foundational API modules (A1).
- Remove A1 paths from `api/ty.toml` excludes and drop temporary ignore rules.


asukaminato0721 (approved):

asukaminato0721:
A small thought, also test pyrefly?

Just ignore the errors it shows, and I could fire && fix the issues in it.

PR #31794, opened 2026-01-31
fix: fix access token is still valid after logout

fatelei:
> [!IMPORTANT]
>

## Summary

fix #31793

save jti in jwt payload, when logout save jti in blacklist, when use access token visit api check the jti is in blacklist or not

it only delete the refresh_token, so the access token is still valid

here we set a access token blacklist, after logout, we add it to blacklist and set access token  valid ttl, when use it after logout, it will reject with error

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/configs/feature/__init__.py
+++ b/api/configs/feature/__init__.py
@@ -29,6 +29,14 @@ class SecurityConfig(BaseSettings):
         default="",
     )
 
+    SESSION_REVOCATION_STORAGE: str = Field(
+        description=(
+            "Session revocation storage backend for JWT jti revocation checks. "
+            "Options: 'null' (default, disabled) or 'redis' (use configured Redis)."
+        ),
+        default="null",
+    )
+
     RESET_PASSWORD_TOKEN_EXPIRY_MINUTES: PositiveInt = Field(
         description="Duration in minutes for which a password reset token remains valid",
         default=5,
--- a/api/controllers/console/auth/login.py
+++ b/api/controllers/console/auth/login.py
@@ -37,6 +37,7 @@ from libs.token import (
     clear_access_token_from_cookie,
     clear_csrf_token_from_cookie,
     clear_refresh_token_from_cookie,
+    extract_access_token,
     extract_refresh_token,
     set_access_token_to_cookie,
     set_csrf_token_to_cookie,
@@ -157,7 +158,9 @@ class LogoutApi(Resource):
         if isinstance(account, flask_login.AnonymousUserMixin):
             response = make_response({"result": "success"})
         else:
-            AccountService.logout(account=account)
+            # Extract access_token from request to revoke it
+            access_token = extract_access_token(request)
+            AccountService.logout(account=account, access_token=access_token)
             flask_login.logout_user()
             response = make_response({"result": "success"})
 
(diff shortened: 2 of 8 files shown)


douxc:
resolves EE-1525

laipz8200 (changes requested):
The stateless JWT design was chosen with performance and ease of use in mind; there is currently no reason to introduce additional state to change it.

fatelei:
> The stateless JWT design was chosen with performance and ease of use in mind; there is currently no reason to introduce additional state to change it.

no additional state, it just add a access token blacklist in redis to disable user visit api after logout, jwt has nothing change

QuantumGhost:
What about extract an interface named SessionRevocationStorage, then let deployers of Dify specify SessionRevocationStorage to use while deploying Dify?

I propose the follow API for SessionRevocationStorage:

```py

class SessionRevocationStorage:
    # revoke mark a token_id as revoked, store it in the SessionRevocationStorage.
    # Once a token_id is markdown as revoked, the `is_revoke` method below should return `True` for that token_id.
    # 
    # `token_id` is the unique identifier of the token, corresponds to the `jti` field in JWT.
    # `expiration_time `is the expiration time of the token, after which the token is considered invalid.
    # It corresponds to the `exp` field in JWT. 
    # 
    # The main purpose of `expiration_time` parameter is to ensure that the spaced used by SessionRevocationStorage
    # would not grow unbound.
    def revoke(self, token_id: str, expiration_time: datetime.datetime) -> None: ...
    
    # return whether a token identified by `token_id` is revoked.
    def is_revoked(self, token_id: str) -> bool: ...

    # expunge maintains the storage and removes expired token ids from the underlying storage.
    # There might be a better name for this method.
    def expunge(self) -> None: ...
```

The default would be `NullSessionRevocationStorage`, which does not store revoked jwt token id, and always returns false for `is_revoked`.

```py
class NullSessionRevocationStorage:
    def revoke(self, token_id: str, expiration_time: datetime.datetime) -> None: 
        pass
    
    def is_revoked(self, token_id: str) -> bool:
        return False

    def expunge(self) -> None:
        pass
```

cc @laipz8200 how about this approach? Given the low frequency of logout / session revocation, and the short valid time of JWT tokens, the storage size of `SessionRevocationStorage` should be minimal. However, it does put an extra check in every request to authenticated API, require a storage backend with high availability, otherwise invocations of `is_revoked` would fail and requests would be blocked.

There is one caveat with this approach: JWT tokens issued before this PR  do not contain the `jti` field, which means that there will be a small transition window for session revocation to be effective.

sorphwer:
I agree with @QuantumGhost since current 'design choice' has been continuously identified as a 'critical vulnerability' by enterprise clients via their pentests.

crazywoola:
Hi @fatelei, thank you for taking the time to contribute to Dify.

### Why this is being closed

This PR was opened before February 2026 (`2026-01-31`) and touches more than 3 files (`8` files). Older broad PRs are very likely to conflict with the current codebase and are difficult to review safely after the surrounding code has moved on. Per our triage rules, we are closing old/conflicting PRs of this size to keep the review queue manageable.

### Next steps

If this change is still needed, please open a fresh PR from the latest `main`, keep the scope small, and link the related issue using GitHub's closing syntax such as `Fixes #123`, `Closes #123`, or `Resolves #123`. Thanks again for your work and understanding.

PR #31861, opened 2026-02-03
fix: redis for api token

zyssyz123:
> [!IMPORTANT]
>

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


api/controllers/console/datasets/datasets.py
```diff
@@ -819,6 +820,10 @@ def delete(self, api_key_id):
 
         if key is None:
             console_ns.abort(404, message="API key not found")
+            return  # Type checker hint: abort() raises exception
```
hj24: abort() will raise exception, so can we remove the duplicate return here?

api/controllers/console/apikey.py
```diff
@@ -130,6 +131,10 @@ def delete(self, resource_id: str, api_key_id: str):
 
         if key is None:
             flask_restx.abort(HTTPStatus.NOT_FOUND, message="API key not found")
+            return  # Type checker hint: abort() raises exception
```
hj24: ditto

api/tasks/update_api_token_last_used_task.py
```diff
@@ -0,0 +1,63 @@
...
+from sqlalchemy.orm import Session
+
+from extensions.ext_database import db
+from libs.datetime_utils import naive_utc_now
+from models.model import ApiToken
+
+logger = logging.getLogger(__name__)
+
+
+@shared_task(queue="dataset", bind=True)
```
hj24: suggest to create a separate queue for the api token operation, it can help isolate tasks queue accumulation in production environment

api/tasks/update_api_token_last_used_task.py
```diff
@@ -0,0 +1,63 @@
...
+        scope: The token type/scope (e.g., 'app', 'dataset')
+        start_time_iso: ISO format timestamp of when the request started
+    """
+    try:
+        # Parse start_time from ISO format
+        start_time = datetime.fromisoformat(start_time_iso)
+        # Update database
+        current_time = naive_utc_now()
+
+        with Session(db.engine, expire_on_commit=False) as session:
```
hj24: duplicate with line 333 in api/controllers/service_api/wraps.py, suggest to create a unify method to handle api token update

api/tests/unit_tests/libs/test_api_token_cache.py
```diff
@@ -0,0 +1,256 @@
+"""
+Unit tests for API Token Cache module.
```
hj24, line 2: suggest to add an integration test to cover the redis and celery operation

api/libs/api_token_cache.py
```diff
@@ -0,0 +1,380 @@
+"""
+API Token Cache Module
+
+Provides Redis-based caching for API token validation to reduce database load.
+"""
+
+import json
```
hj24: consider using orjson for better performance in frequent api token scenario

api/libs/api_token_cache.py
```diff
@@ -0,0 +1,380 @@
...
+import logging
+from datetime import datetime
+from typing import Any
+
+from extensions.ext_redis import redis_client, redis_fallback
+
+logger = logging.getLogger(__name__)
+
+
+class CachedApiToken:
```
hj24: It's not a common lib, seems only used for api token scenario, maybe better to move to service layer?

api/libs/api_token_cache.py
```diff
@@ -0,0 +1,380 @@
...
+        self.created_at = created_at
+
+    def __repr__(self) -> str:
+        return f"<CachedApiToken id={self.id} type={self.type}>"
+
+
+# Cache configuration
+CACHE_KEY_PREFIX = "api_token"
+CACHE_TTL_SECONDS = 600  # 10 minutes
+CACHE_NULL_TTL_SECONDS = 60  # 1 minute for non-existent tokens (防穿透)
```
hj24: remove Chinese annotation

api/tasks/update_api_token_last_used_task.py
```diff
@@ -0,0 +1,63 @@
...
+    Args:
+        token: The API token string
+        scope: The token type/scope (e.g., 'app', 'dataset')
+        start_time_iso: ISO format timestamp of when the request started
+    """
+    try:
+        # Parse start_time from ISO format
+        start_time = datetime.fromisoformat(start_time_iso)
+        # Update database
+        current_time = naive_utc_now()
```
hj24: current_time maybe a false-positive value, it cannot reflect the real update time of last_used_at, especially in queue accumulation scenario

api/tasks/update_api_token_last_used_task.py
```diff
@@ -0,0 +1,63 @@
...
+
+from extensions.ext_database import db
+from libs.datetime_utils import naive_utc_now
+from models.model import ApiToken
+
+logger = logging.getLogger(__name__)
+
+
+@shared_task(queue="dataset", bind=True)
+def update_api_token_last_used_task(self, token: str, scope: str | None, start_time_iso: str):
```
hj24: start_time_iso -> update_time_iso

api/tasks/update_api_token_last_used_task.py
```diff
@@ -0,0 +1,63 @@
...
+        # Update database
+        current_time = naive_utc_now()
+
+        with Session(db.engine, expire_on_commit=False) as session:
+            update_stmt = (
+                update(ApiToken)
+                .where(
+                    ApiToken.token == token,
+                    ApiToken.type == scope,
+                    (ApiToken.last_used_at.is_(None) | (ApiToken.last_used_at < start_time)),
```
hj24: If we want to skip the task out-of-order situation, consider the following should be enough?

```python
ApiToken.last_used_at < update_time
values(last_used_at=update_time)
```

```suggestion
update(ApiToken)
    .where(
        ApiToken.token == token,
        ApiToken.type == scope,
        (ApiToken.last_used_at.is_(None) | (ApiToken.last_used_at < update_time)),
    )
    .values(last_used_at=update_time)
```

api/tasks/update_api_token_last_used_task.py
```diff
@@ -0,0 +1,63 @@
...
+            if rowcount > 0:
+                session.commit()
+                logger.info("Updated last_used_at for token (async): %s... (scope: %s)", token[:10], scope)
+                return {"status": "updated", "rowcount": rowcount, "start_time": start_time_iso}
+            else:
+                logger.debug("No update needed for token: %s... (already up-to-date)", token[:10])
+                return {"status": "no_update_needed", "reason": "last_used_at >= start_time"}
+
+    except Exception as e:
+        logger.warning("Failed to update last_used_at for token (async): %s", e)
```
hj24: I think we can raise exception in celery task to let flower monitor aware of the failure

api/controllers/service_api/wraps.py
```diff
@@ -308,8 +316,20 @@ def validate_and_get_api_token(scope: str | None = None):
...
+        # Asynchronously update last_used_at (non-blocking)
+        _async_update_token_last_used_at(auth_token, scope)
+        return cached_token
+
+    # Cache miss - query database
+    logger.debug("Token cache miss, querying database for scope: %s", scope)
     current_time = naive_utc_now()
     cutoff_time = current_time - timedelta(minutes=1)
+
     with Session(db.engine, expire_on_commit=False) as session:
```
hj24: seems duplicate with task logic, is it possible to extract this into a unify method

api/controllers/service_api/wraps.py
```diff
@@ -308,8 +316,20 @@ def validate_and_get_api_token(scope: str | None = None):
...
         update_stmt = (
             update(ApiToken)
```
hj24: If cache missed, this logic is still not concurrent safe, maybe we can add a redis lock to make a single flight mode, use auth_token as lock key to allow only one call to update db and cache.

And this lock should be added to all code block that will update a token record, not only the validate_and_get_api_token.

api/libs/api_token_cache.py
hj24: Is it possible to use a pydantic model?

api/libs/api_token_cache.py
```diff
@@ -0,0 +1,380 @@
...
+
+                logger.info("Deleted cache for key: %s", cache_key)
+                return True
+            except Exception as e:
+                logger.warning("Failed to delete token cache: %s", e)
+                return False
+
+    @staticmethod
+    @redis_fallback(default_return=False)
+    def invalidate_by_tenant(tenant_id: str) -> bool:
```
hj24: when will invalidate_by_tenant be called?

api/libs/api_token_cache.py
```diff
@@ -0,0 +1,380 @@
...
+
+                logger.info(
+                    "Invalidated %d token cache entries for tenant: %s (via index)",
+                    deleted_count,
+                    tenant_id,
+                )
+                return True
+
+            # Index doesn't exist - fallback to scanning (slow path)
+            logger.info("Tenant index not found, falling back to full scan for tenant: %s", tenant_id)
```
hj24: Is this a necessary fallback? Seems it will scan across all tenants, looks not like an efficient way.

If a tenant status changed from normal to archive, is it possible to leave it expired automatically?

hj24 (changes requested):

api/controllers/service_api/wraps.py
```diff
@@ -308,29 +316,103 @@ def validate_and_get_api_token(scope: str | None = None):
...
+
+            stmt = select(ApiToken).where(ApiToken.token == auth_token, ApiToken.type == scope)
+            api_token = session.scalar(stmt)
+
+            if not api_token:
+                ApiTokenCache.set(auth_token, scope, None)
+                raise Unauthorized("Access token is invalid")
+
+            ApiTokenCache.set(auth_token, scope, api_token)
+            return api_token
```
laipz8200: I think this piece of code for handling the cache should be a separate function.

api/tasks/update_api_token_last_used_task.py
```diff
@@ -0,0 +1,59 @@
...
+
+from celery import shared_task
+
+from libs.api_token_updater import update_token_last_used_at
+
+logger = logging.getLogger(__name__)
+
+
+@shared_task(queue="api_token_update", bind=True)
+def update_api_token_last_used_task(self, token: str, scope: str | None, update_time_iso: str):
```
laipz8200: To avoid frequent updates to the database, I suggest setting up this task as a scheduled job with a 30-minute interval.

api/libs/api_token_cache.py
```diff
@@ -0,0 +1,279 @@
...
+        Args:
+            api_token: ApiToken model instance or CachedApiToken
+
+        Returns:
+            JSON bytes representation
+        """
+        # If it's already a Pydantic model, use model_dump
+        if isinstance(api_token, CachedApiToken):
+            # Pydantic model -> dict -> orjson
+            return orjson.dumps(api_token.model_dump(mode="json"))
```
laipz8200: You can directly use `model_dump_json()` here instead of `orjson`. The same applies to the section below as well.

api/docker/entrypoint.sh
```diff
@@ -35,10 +35,10 @@ if [[ "${MODE}" == "worker" ]]; then
...
+      DEFAULT_QUEUES="api_token_update,dataset,priority_dataset,priority_pipeline,pipeline,mail,ops_trace,app_deletion,plugin,workflow_storage,conversation,workflow_professional,workflow_team,workflow_sandbox,schedule_poller,schedule_executor,triggered_workflow_dispatcher,trigger_refresh_executor,retention"
     else
       # Community edition (SELF_HOSTED): dataset, pipeline and workflow have separate queues
-      DEFAULT_QUEUES="dataset,priority_dataset,priority_pipeline,pipeline,mail,ops_trace,app_deletion,plugin,workflow_storage,conversation,workflow,schedule_poller,schedule_executor,triggered_workflow_dispatcher,trigger_refresh_executor,retention"
+      DEFAULT_QUEUES="api_token_update,dataset,priority_dataset,priority_pipeline,pipeline,mail,ops_trace,app_deletion,plugin,workflow_storage,conversation,workflow,schedule_poller,schedule_executor,triggered_workflow_dispatcher,trigger_refresh_executor,retention"
```
hj24: Maybe a queue named api_token for all api_token related operations, not only for update

api/tests/integration_tests/libs/test_api_token_cache_integration.py
```diff
@@ -0,0 +1,386 @@
...
+        mock_redis.setex.side_effect = RedisError("Connection failed")
+
+        result_get = ApiTokenCache.get("test-token", "app")
+        assert result_get is None
+
+        result_set = ApiTokenCache.set("test-token", "app", None)
+        assert result_set is False
+
+
+if __name__ == "__main__":
```
hj24: Is this necessary? If not, could we remove it?

api/tests/unit_tests/libs/test_api_token_cache.py
```diff
@@ -0,0 +1,256 @@
...
+        """Test that non-existent tokens are cached as null."""
+        # Set null token (cache miss)
+        ApiTokenCache.set("non-existent-token", "app", None)
+
+        args = mock_redis.setex.call_args[0]
+        assert args[2] == b"null"
+        assert args[1] == CACHE_NULL_TTL_SECONDS  # Shorter TTL for null values
+
+
+if __name__ == "__main__":
```
hj24: ditto

api/controllers/service_api/wraps.py
```diff
@@ -308,29 +315,91 @@ def validate_and_get_api_token(scope: str | None = None):
...
+        # Record usage in Redis for later batch update (no Celery task per request)
+        _record_token_usage(auth_token, scope)
+        return cast(ApiToken, cached_token)
+
+    # Cache miss - use Redis lock for single-flight mode
+    # This ensures only one request queries DB for the same token concurrently
+    return _fetch_token_with_single_flight(auth_token, scope)
+
+
+def _query_token_from_db(auth_token: str, scope: str | None) -> ApiToken:
```
hj24: suggest to move those db/redis operations into service or lib layer instead of the controller layer.

api/schedule/update_api_token_last_used_task.py
```diff
@@ -0,0 +1,102 @@
...
+        with Session(db.engine, expire_on_commit=False) as session:
+            for token, scope in token_scope_pairs:
+                stmt = (
+                    update(ApiToken)
+                    .where(
+                        ApiToken.token == token,
+                        ApiToken.type == scope,
+                        (ApiToken.last_used_at.is_(None) | (ApiToken.last_used_at < current_time)),
+                    )
+                    .values(last_used_at=current_time)
```
hj24: Should we update this field by reading the value from redis? Seems current_time is generated by tasks, is it a correct value?

api/controllers/service_api/wraps.py
```diff
@@ -308,29 +315,91 @@ def validate_and_get_api_token(scope: str | None = None):
...
+def _record_token_usage(auth_token: str, scope: str | None):
+    """
+    Record token usage in Redis for later batch update by a scheduled job.
+
+    Instead of dispatching a Celery task per request, we simply SET a key in Redis.
+    A Celery Beat scheduled task will periodically scan these keys and batch-update
+    last_used_at in the database.
+    """
+    try:
+        key = f"api_token_active:{scope}:{auth_token}"
```
hj24: From the perspective of maintainability, it's better to use const `ACTIVE_TOKEN_KEY_PREFIX` instead of use string directly.

Consider to add a method like `ApiTokenCache.make_active_key` for invocation.

api/controllers/service_api/wraps.py
```diff
@@ -308,29 +315,91 @@ def validate_and_get_api_token(scope: str | None = None):
...
+        redis_client.set(key, naive_utc_now().isoformat(), ex=3600)  # TTL 1 hour as safety net
```
hj24: `naive_utc_now().isoformat()` is aim to record the `last_updated_at` right? Is so, we should use it in the schedule task to update db

api/libs/api_token_cache.py
```diff
@@ -0,0 +1,277 @@
...
+
+        Args:
+            tenant_id: The tenant ID
+            cache_key: The cache key to add to the index
+        """
+        if not tenant_id:
+            return
+
+        try:
+            index_key = f"tenant_tokens:{tenant_id}"
```
hj24: `_make_token_index_key(self, tenant_id)`

hj24 (changes requested):

api/controllers/service_api/wraps.py
zyssyz123: Yes, naive_utc_now().isoformat() records the actual request time. The schedule task reads this value from Redis and uses it as last_used_at when updating the database (see schedule/update_api_token_last_used_task.py line usage_time = datetime.fromisoformat(value)).

api/schedule/update_api_token_last_used_task.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if not token_entries:
+            click.echo(click.style("batch_update_api_token_last_used: no active tokens found.", fg="yellow"))
+            # Still clean up any invalid keys
+            if keys_to_delete:
+                redis_client.delete(*keys_to_delete)
+            return
+
+        # Batch update in database using actual usage timestamps from Redis
+        with Session(db.engine, expire_on_commit=False) as session:
```
hj24:
```suggestion
for token, scope, usage_time in token_entries:
    with Session(db.engine, expire_on_commit=False) as session, session.begin():
```

It's better to change the order, move session block into the for loop to avoid the risk of long transaction.

api/libs/api_token_cache.py
```diff
@@ -0,0 +1,369 @@
...
+def query_token_from_db(auth_token: str, scope: str | None) -> Any:
+    """
+    Query API token from database and cache the result.
+
+    last_used_at is NOT updated here -- it is handled by the periodic batch
+    task via record_token_usage().
+
+    Raises Unauthorized if token is invalid.
+    """
+    from models.model import ApiToken
```
hj24: Suggestion: if lib depends on model, then it's better to move those logic into service layer. lib should be a pure layer that should not be biz-aware.

hj24 (changes requested):
Except the code structure and db transaction issue, others LGTM

api/libs/api_token_cache.py
```diff
@@ -0,0 +1,305 @@
...
+
+from pydantic import BaseModel
+
+from extensions.ext_redis import redis_client, redis_fallback
+from libs.datetime_utils import naive_utc_now
+
+logger = logging.getLogger(__name__)
+
+
+class CachedApiToken(BaseModel):
```
hj24: Seems other libs are all not biz aware common utils, but looks like the rest of the api token logic more focus on a specific biz (api token) right? Is is possible to move to service layer also?

api/extensions/ext_celery.py
```diff
@@ -184,6 +184,15 @@ def __call__(self, *args: object, **kwargs: object) -> object:
             "task": "schedule.trigger_provider_refresh_task.trigger_provider_refresh",
             "schedule": timedelta(minutes=dify_config.TRIGGER_PROVIDER_REFRESH_INTERVAL),
         }
+
+    if dify_config.ENABLE_API_TOKEN_LAST_USED_UPDATE_TASK:
+        imports.append("schedule.update_api_token_last_used_task")
+        beat_schedule["batch_update_api_token_last_used"] = {
+            "task": "schedule.update_api_token_last_used_task.batch_update_api_token_last_used",
+            # "schedule": timedelta(minutes=dify_config.API_TOKEN_LAST_USED_UPDATE_INTERVAL),
+            "schedule": timedelta(minutes=2),
```
hj24: Is it for local test? Should we revert it back to use `API_TOKEN_LAST_USED_UPDATE_INTERVAL`?

hj24 (changes requested):

hj24 (approved):
LGTM

PR #31877, opened 2026-02-03
refactor: replace line-clamp package with native CSS

niveshdandyan:
## Summary

Partially addresses #31875

Replaces the JavaScript-based `line-clamp` npm package with Tailwind's native `line-clamp-4` CSS utility class.

## Changes

- Removed `line-clamp` dependency from `web/package.json`
- Updated `edit-slice.tsx` to use Tailwind's `line-clamp-4` utility instead of the JS-based approach

## Why

- The `line-clamp` package uses JavaScript DOM manipulation to truncate text
- Modern CSS `line-clamp` (via `-webkit-line-clamp`) is well-supported across all browsers
- The codebase already uses Tailwind's `line-clamp-*` utilities in 50+ other places
- CSS approach is more performant - no JavaScript runtime manipulation needed

## Test Plan

1. Navigate to a dataset document that uses the formatted text component
2. Verify text is properly truncated to 4 lines
3. Verify the edit slice renders correctly with the floating element reference


hyoban (changes requested):
Please update the lock

niveshdandyan:
Thanks for the review @hyoban\! I've updated the lock file by running `pnpm install`. The CI should pass now.

hyoban (approved):

PR #31955, opened 2026-02-04
fix: dynamically update image URLs in agent_thoughts with valid signatures

jackjin1997:
Fixes #31907. 

### Description
Currently, image URLs in `agent_thoughts.thought` and `agent_thoughts.observation` have static signatures that expire after ~10 minutes. This PR adds dynamic re-signing logic to these fields, consistent with how the `answer` field is handled in the `Message` model.

### Changes
- Added `_re_sign_file_url` helper and related properties (`re_sign_file_url_thought`, `re_sign_file_url_observation`) to `MessageAgentThought` model.
- Updated `AgentThought` Pydantic schema to use these properties via `validation_alias`.

This is a more lightweight approach compared to other proposed fixes, focusing on the core issue of signature expiration in API responses.

Fixes #31907


crazywoola:
Thanks for your contribution, @jackjin1997 — really appreciate the effort here.

Closing this in favor of #31929, which was opened earlier and targets the same issue (#31907). To keep duplicate cleanup fair, we're going FIFO on dupes — the older PR gets the merge consideration first.

PR #31974, opened 2026-02-05
fix(docker): Use named volumes for PostgreSQL and Redis to resolve macOS permission issues

### LINKED ISSUE #36357: PostgreSQL bind mount causes silent data loss on Docker Desktop WSL2 due to 9p stale file handles
## Self Checks

- [x] I have read Dify's [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
- [x] I have searched to make sure this is not a [discussion](https://github.com/langgenius/dify/discussions) question
- [x] I have searched both [open](https://github.com/langgenius/dify/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/langgenius/dify/issues?q=is%3Aissue+is%3Aclosed) issues
- [x] I have confirmed this bug has not been submitted before
- [x] I am submitting this bug report in English
- [x] I have not modified this template

## Dify version

1.14.1 (also affects earlier versions using the same docker-compose template)

## Cloud or Self Hosted

Self Hosted (Docker)

## Steps to reproduce

1. Deploy Dify via `docker compose` on **Docker Desktop for Windows with WSL2 backend**
2. The default `docker-compose.yaml` mounts PostgreSQL data via bind mount:
   ```yaml
   volumes:
     - ./volumes/db/data:/var/lib/postgresql/data
   ```
3. Let the deployment run continuously for **2-3 days** with active usage (apps, datasets, conversations created)
4. After extended uptime, query the database:
   ```sql
   SELECT COUNT(*) FROM apps;  -- returns 0, even though data exists on disk
   ```
5. Check the host filesystem — `volumes/db/data/pgdata/base/16384/` still contains all data files on disk
6. Restart the PostgreSQL container — all data reappears immediately

## ✔️ Expected Behavior

PostgreSQL should consistently read and return data stored in `volumes/db/data/` regardless of container uptime. Bind mounts should provide reliable, persistent storage indistinguishable from native filesystem access.

## ❌ Actual Behavior

After ~2-3 days of uptime, PostgreSQL returns **empty query results** for all tables (0 apps, 0 datasets, 0 workflows, 0 tenants, 0 accounts). The Dify web UI shows an empty workspace with no apps, redirecting to `/install` as if it were a fresh deployment. All data files remain intact and uncorrupted on disk. A PostgreSQL restart fixes the issue immediately, confirming the data was never actually lost.

**PostgreSQL logs show no errors, crashes, or recovery events** during the affected period — the database believes it is operating normally.

## Root Cause

Docker Desktop on Windows accesses the WSL2 Linux filesystem via the **9p (Plan 9) network filesystem protocol**. Unlike a native Linux mount, 9p file handles held by long-running processes can become **stale** — the kernel dentry cache and the 9p server state drift apart over time. When PostgreSQL's shared buffer pool and file descriptors age past a certain threshold, reads from the 9p-backed bind mount return empty/inconsistent data without surfacing I/O errors. The process (PostgreSQL) sees no reason to invalidate its cache or reopen files, so the condition persists until restart.

This is a known class of issue with 9p-based filesystems — they were designed for low-throughput, short-lived access patterns, not for database workloads with long-held file descriptors and mmap'd pages.

## Proposed Fix

Change the `db_postgres` volume mount in `docker-compose-template.yaml` from a **bind mount** to a **Docker named volume**:

```yaml
# Before (bind mount — affected by 9p staleness)
volumes:
  - ./volumes/db/data:/var/lib/postgresql/data

# After (named volume — Docker manages the filesystem directly)
volumes:
  - dify_db_data:/var/lib/postgresql/data

# Top-level volumes section
volumes:
  dify_db_data:
```

**Why this fixes it:** Docker named volumes on Docker Desktop WSL2 use Docker's internal volume driver, which stores data inside the WSL2 VM's native ext4 filesystem. This bypasses the 9p translation layer entirely — PostgreSQL interacts with a real Linux filesystem, not a network filesystem proxy. File handles, fsync, and mmap all work correctly regardless of uptime.

**Migration path for existing users:** A one-time data copy from bind mount to named volume (via `docker compose run`) preserves all existing data. No downtime beyond a normal restart.

**Other affected services (for consideration):** The same 9p staleness risk applies to other bind-mounted stateful volumes in the template:
- `redis` → `./volumes/redis/data`
- `weaviate` → `./volumes/weaviate`
- `plugin_daemon` → `./volumes/plugin_daemon`

These may experience similar issues and could be migrated in a follow-up.

## Additional Context

Reproduced with Docker Desktop v5.1.1 on Windows 11 + WSL2 (Ubuntu), PostgreSQL 15-alpine. After migrating to a named volume, the problem has not recurred.

kemalelmizan:
> [!IMPORTANT]
>

## Summary

fix #36357

When running Dify on macOS (particularly with Rancher Desktop or Docker Desktop), the PostgreSQL and Redis containers fail to start with permission errors:

```
db_postgres-1 | chown: /var/lib/postgresql/data/pgdata: Permission denied
redis-1       | chown: .: Permission denied
```

This results in:
- PostgreSQL container marked as "optional dependency failed to start"
- Redis container in constant restart loop
- Cascading failures in api, worker, worker_beat, and plugin_daemon with: `redis.exceptions.ConnectionError: Error -2 connecting to redis:6379. Name or service not known.`

## Root Cause

The issue stems from UID/GID mismatch between the host macOS user and container users:

1. **macOS filesystem behavior**: macOS uses a different user namespace than Linux containers
2. **Bind mount limitations**: When using bind mounts (`./volumes/db/data:/var/lib/postgresql/data`), the directory is owned by the macOS user (typically UID 501)
3. **Container user mismatch**: PostgreSQL runs as user `postgres` (UID 70 or 999) and Redis runs as user `redis` (UID 999), neither of which have permissions to modify host-owned directories
4. **Rancher Desktop/Docker Desktop**: Both use VM-based solutions on macOS, which exacerbates the permission mapping issue

This is a known issue on macOS with both Docker Desktop and Rancher Desktop when using bind mounts with containers that run as non-root users.

## Changes Made

Changed storage from bind mounts to Docker-managed named volumes for:

1. **PostgreSQL (docker-compose-template.yaml)**
   - Old: `./volumes/db/data:/var/lib/postgresql/data`
   - New: `dify-postgres-data:/var/lib/postgresql/data`

2. **Redis (docker-compose-template.yaml)**
   - Old: `./volumes/redis/data:/data`
   - New: `dify-redis-data:/data`

3. **Volume declarations**
   - Added `dify-postgres-data:`
   - Added `dify-redis-data:`

## Why This Fixes the Problem

**Docker-managed named volumes** resolve the issue by:

1. **Proper permission management**: Docker automatically handles UID/GID mapping within the volume
2. **Container-native ownership**: The volume is created with ownership matching the container's user
3. **Cross-platform compatibility**: Named volumes work consistently across Linux, macOS, and Windows
4. **VM abstraction**: On macOS, volumes are created within the Docker VM where Linux permissions work natively, avoiding the host filesystem permission translation layer

## References

- Docker documentation: https://docs.docker.com/storage/volumes/
- PostgreSQL Docker Hub: https://hub.docker.com/_/postgres#arbitrary---user-notes (see "Arbitrary --user Notes")
- Related issue pattern: https://github.com/docker-library/postgres/issues/435

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


crazywoola:
Please link an issue in the description. :)

crazywoola:
Please link an existing issue or create a new one in the description so that we can proceed :)

PR #32087, opened 2026-02-07
fix: preserve spaces between words when splitting by space separator

veeceey:
## Summary
- Fixes a bug where spaces between words were removed in child chunks when using Knowledge Pipeline with Parent Chunk Mode.
- When the text splitter falls back to space as the separator with `keep_separator` enabled (the default for `RecursiveCharacterTextSplitter`), the merge separator was incorrectly set to an empty string (`""`), causing words to be concatenated without spaces.
- The space split path uses `re.split(r" +", text)` which removes all spaces from the text, but unlike other separators, the `keep_separator` logic does not re-append the space to each split. The merge step then joined splits with `""` instead of `" "`, losing all word spacing.

## Root Cause
In `FixedRecursiveCharacterTextSplitter.recursive_split_text()`:

```python
# Space split does NOT append separator to splits (unlike other separators)
if separator == " ":
    splits = re.split(r" +", text)
else:
    splits = text.split(separator)
    if self._keep_separator:
        splits = [s + separator for s in splits[:-1]] + splits[-1:]

# But _separator was still set to "" when keep_separator is True
_separator = "" if self._keep_separator else separator
```

Since `keep_separator` defaults to `True` in `RecursiveCharacterTextSplitter`, and the space case doesn't append the separator to each split, the merge step lost all spaces.

## Fix
When the separator is a space, always use `" "` as the merge separator regardless of `_keep_separator`, since the space split path does not append the separator to splits.

## Test plan
- [x] All 94 existing text splitter tests pass
- [ ] Verify with a document containing space-separated words that child chunks preserve word spacing

Closes #32066


veeceey:
## Manual Test Results

I've tested the space preservation fix with the following scenario:

**Test Setup:**
- Text: "This is a long sentence that needs to be split into smaller chunks"
- Chunk size: 20 characters
- Separator: space

**Results After Fix:**

```
Chunk 1: 'This is a long' (length: 14)
Chunk 2: 'sentence that needs' (length: 19)
Chunk 3: 'to be split into' (length: 16)
Chunk 4: 'smaller chunks' (length: 14)
```

**Reconstructed:** `This is a long sentence that needs to be split into smaller chunks`
**Original:** `This is a long sentence that needs to be split into smaller chunks`

✅ **Spaces are correctly preserved!**

**Without the fix (hypothetical):**
The chunks would be joined without spaces: `This is a longsentence that needsto be split intosmaller chunks`
❌ This is incorrect - all spaces between chunks would be removed.

**Conclusion:**
The fix successfully preserves spaces when splitting text by space separator, ensuring that child chunks can be properly reconstructed with spaces between words.

veeceey:
Hi team -- friendly ping! This PR has been open for a couple of days without review. Would love any feedback when you get a chance. Thanks!

veeceey:
Friendly bump -- this fixes the space-stripping issue when splitting text chunks. Manual test results are included above. Would appreciate a review when you have a moment!

veeceey:
hey, any updates here? the space preservation fix is pretty important for non-English text processing

PR #32118, opened 2026-02-09
fix: resolve file validation failure on file return for workflow-as-tool

lilongfei2000:
Fixes #32112

### Bug
When a workflow is released as a tool and used elsewhere, it always raise validation error if the return value contains files.

### Root Cause
The returned files are created via
https://github.com/langgenius/dify/blob/0428ac5f3aca02ebaeb3bedad866eb7957adf5a7/api/core/tools/__base/tool.py#L171-L176

While the class `FileMessage` has an attribute `file_marker` which would be validated **before** the model is instantiated and the default value assigned.
https://github.com/langgenius/dify/blob/0428ac5f3aca02ebaeb3bedad866eb7957adf5a7/api/core/tools/entities/tool_entities.py#L146-L154

### Solution
I fix it by simply pass in the default parameter value explicitly.

## Screenshots

| Before | After |
|--------|-------|
| bug2 | fixed |

## Checklist

### DIFF
--- a/api/core/tools/__base/tool.py
+++ b/api/core/tools/__base/tool.py
@@ -171,7 +171,7 @@ class Tool(ABC):
     def create_file_message(self, file: File) -> ToolInvokeMessage:
         return ToolInvokeMessage(
             type=ToolInvokeMessage.MessageType.FILE,
-            message=ToolInvokeMessage.FileMessage(),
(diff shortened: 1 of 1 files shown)


lilongfei2000:
I just saw #32114 working on this bug. I'll close this pr.

PR #32195, opened 2026-02-10
fix: add distributed lock and fix session persistence for OAuth token refresh

### LINKED ISSUE #32174: Bug: OAuth token refresh credentials never persist to DB + concurrent refresh race condition
### Self Checks

- [x] This is only a bug report, not a feature request.
- [x] I have searched for [existing issues](https://github.com/langgenius/dify/issues) to avoid creating duplicates.
- [x] I understand this is an open source project and will provide as much detail as possible.

### Dify version

v1.12.1 (Docker deployment)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Configure a built-in tool provider with OAuth2 credentials (e.g., QuickBooks)
2. Create a workflow with 2+ tool nodes using the same OAuth provider
3. Wait for the access token to expire (1 hour)
4. Execute the workflow
5. Check the database: `SELECT updated_at, created_at, expires_at FROM tool_builtin_providers WHERE provider LIKE '%quickbooks%';`
6. Observe `updated_at == created_at` — credentials were never updated despite successful refresh HTTP calls

### ✅ Expected Behavior

- After a successful OAuth token refresh, the new `encrypted_credentials` and `expires_at` should be persisted to the database via `db.session.commit()`
- Only ONE refresh should happen per token expiry cycle (~1 per hour)
- Concurrent requests detecting expired credentials should be serialized

### ❌ Actual Behavior

**Two critical bugs:**

#### Bug 1: `db.session.commit()` after refresh does not persist

The refresh code in `tool_manager.py` calls `db.session.commit()` after updating credentials, but the changes never persist to the database.

**Evidence from production database:**

```
| expires_at | updated_at          | created_at          |
|------------|---------------------|---------------------|
| 1770606647 | 2026-02-09 02:10:47 | 2026-02-09 02:10:47 |
```

- `created_at == updated_at`: the row has **never been updated** since creation
- `expires_at = created_at + 3600` (standard 1-hour token lifetime), never refreshed
- The `updated_at` column has `onupdate=func.current_timestamp()`, so any ORM commit would update it

**Yet 693 refresh HTTP calls were made in 24 hours** (528 from API container + 165 from Worker). All returned HTTP 200 OK from the plugin daemon. None persisted to the DB.

**Same-thread proof:** The same greenlet makes 4-6 sequential refresh calls ~1s apart. If the first commit worked, the second call would see fresh `expires_at` and skip:

```
10:37:17 [Dummy-1497] refresh_credentials "200 OK"
10:37:18 [Dummy-1497] refresh_credentials "200 OK"  ← should have skipped
10:37:19 [Dummy-1497] refresh_credentials "200 OK"  ← should have skipped
10:37:20 [Dummy-1497] refresh_credentials "200 OK"  ← should have skipped
```

#### Bug 2: No locking on concurrent refresh

Multiple `GraphWorker` threads in workflow execution concurrently detect expired credentials and call `refresh_credentials` simultaneously:

```
02:18:24 [GraphWorker-0] refresh_credentials → SUCCESS (consumes old refresh_token)
02:18:25 [GraphWorker-0] tool/invoke → SUCCESS
02:18:28 [GraphWorker-1] refresh_credentials → FAIL "refresh token expired or revoked"
02:18:28 Node ABORT → workflow fails
```

Both workers read the same expired credentials from DB. GraphWorker-0 refreshes first (rotates the token). GraphWorker-1 retries with the now-invalidated old token → permanent failure.

This pattern was observed throughout the day with near-simultaneous refreshes from two workers:

```
05:30:44.680 [GraphWorker-0] refresh  ← 3ms apart
05:30:44.683 [GraphWorker-1] refresh  ← same req_id
```

These succeeded earlier due to the OAuth provider's grace period, but eventually (02:18), the stale token was rejected.

#### Cascading permanent failure

After the first failure, ALL subsequent requests fail because:
- The DB still stores the original (never-updated) refresh token from credential creation
- That token has been invalidated by the OAuth provider after 693+ uses
- Every new request reads the invalid token → fails → requires manual re-authorization

### Impact

- **Performance**: 693 unnecessary OAuth refresh calls/day (expected: ~24)
- **Reliability**: Eventual permanent failure requiring manual re-authorization
- **User experience**: All workflows using the affected OAuth provider stop working permanently

### Affected code

- `api/core/tools/tool_manager.py` — `get_tool_runtime()`, OAuth refresh block (lines ~270-298)
- `api/services/datasource_provider_service.py` — `get_datasource_credentials()`, same pattern

### Proposed fix

1. **Add Redis distributed lock** around the refresh block (non-blocking acquire + double-check pattern)
2. **Add exponential backoff polling** for requests that don't acquire the lock
3. **Investigate commit persistence** — ensure `db.session.commit()` persists in all execution contexts (API server, Celery worker, gevent greenlets)

A PR with the distributed lock fix is being prepared.

Alai-sudo:
## Summary

Fixes two critical bugs in the OAuth token refresh logic that cause:

1. **Refreshed credentials never persist to the database** — `db.session.commit()` after a successful OAuth refresh does not persist in Celery worker and gevent greenlet contexts because Flask-SQLAlchemy's scoped `db.session` is not reliable in these contexts. Every subsequent request re-reads the stale `expires_at` and re-triggers the refresh, leading to hundreds of unnecessary OAuth calls per day.

2. **No distributed lock on concurrent refresh** — Multiple concurrent workers (e.g., `GraphWorker-0` and `GraphWorker-1` in workflow execution) can simultaneously detect expired credentials and call the OAuth provider to refresh. With providers that use refresh token rotation (e.g., QuickBooks), this leads to `invalid_grant` errors and permanent failure requiring manual re-authorization.

Together, these bugs create a cascading failure: the token is refreshed hundreds of times but never saved, until the original refresh token is eventually invalidated by the OAuth provider.

### `api/core/tools/tool_manager.py`
- Add non-blocking Redis distributed lock around the OAuth refresh block
- Replace `db.session.refresh()` / `db.session.commit()` with explicit `Session(db.engine)` and `session.get()` / `session.commit()` to ensure persistence in all execution contexts
- Add double-check after lock acquisition (another request may have already refreshed)
- Add exponential backoff polling for requests that don't acquire the lock

### `api/services/datasource_provider_service.py`
- Add the same non-blocking Redis distributed lock + double-check pattern
- This file already uses explicit `Session(db.engine)`, so only the lock was needed

### New test files
- `api/tests/unit_tests/core/tools/test_tool_manager_oauth_refresh_lock.py` — 6 tests covering lock acquired + refresh, double-check skip, lock not acquired + polling, lock release on exception, non-expired skip, never-expires skip
- `api/tests/unit_tests/services/test_datasource_provider_oauth_refresh_lock.py` — 6 matching tests for datasource provider

## Evidence from production

Before fix (24 hours):
- **693** OAuth refresh HTTP calls (expected: ~24)
- Database `updated_at == created_at` — credentials never persisted
- Concurrent refresh race condition caused permanent `invalid_grant` failure

After fix (5.5 hours):
- **3** OAuth refresh calls (1 manual test + 2 automatic hourly refreshes)
- Database `updated_at` correctly updated on each refresh
- Zero errors, zero concurrent conflicts
- All workflows executing successfully

## Related issue

Closes #32174

## Test plan

- [x] 6 unit tests for `tool_manager.py` OAuth refresh lock — all pass
- [x] 6 unit tests for `datasource_provider_service.py` OAuth refresh lock — all pass
- [x] 78 existing tools tests — zero regressions
- [x] Production deployment verified: token refresh persists, lock prevents concurrent refresh

### DIFF
--- a/api/core/tools/tool_manager.py
+++ b/api/core/tools/tool_manager.py
@@ -278,9 +278,7 @@ class ToolManager:
                         with Session(db.engine) as session:
                             # double-check after acquiring lock: re-read from DB
                             bp = session.get(BuiltinToolProvider, builtin_provider.id)
-                            if bp is not None and bp.expires_at != -1 and (
-                                bp.expires_at - 60
-                            ) < int(time.time()):
+                            if bp is not None and bp.expires_at != -1 and (bp.expires_at - 60) < int(time.time()):
                                 # TODO: circular import
                                 from core.plugin.impl.oauth import OAuthHandler
                                 from services.tools.builtin_tools_manage_service import BuiltinToolManageService
@@ -292,9 +290,7 @@ class ToolManager:
                                     f"{dify_config.CONSOLE_API_URL}/console/api/oauth/plugin/"
                                     f"{provider_id}/tool/callback"
                                 )
-                                system_credentials = BuiltinToolManageService.get_oauth_client(
-                                    tenant_id, provider_id
-                                )
+                                system_credentials = BuiltinToolManageService.get_oauth_client(tenant_id, provider_id)
 
                                 oauth_handler = OAuthHandler()
                                 refreshed_credentials = oauth_handler.refresh_credentials(
--- a/api/core/tools/tool_manager.py
+++ b/api/core/tools/tool_manager.py
@@ -272,58 +272,67 @@ class ToolManager:
                 lock = redis_client.lock(lock_key, timeout=30)
                 if lock.acquire(blocking=False):
                     try:
-                        # double-check after acquiring lock: another request may have already refreshed
-                        db.session.refresh(builtin_provider)
-                        if builtin_provider.expires_at != -1 and (
-                            builtin_provider.expires_at - 60
-                        ) < int(time.time()):
-                            # TODO: circular import
-                            from core.plugin.impl.oauth import OAuthHandler
-                            from services.tools.builtin_tools_manage_service import BuiltinToolManageService
+                        # Use an explicit session to ensure commit persistence in
+                        # Celery worker and gevent greenlet contexts where the
+                        # Flask-SQLAlchemy scoped db.session does not reliably persist.
+                        with Session(db.engine) as session:
+                            # double-check after acquiring lock: re-read from DB
+                            bp = session.get(BuiltinToolProvider, builtin_provider.id)
+                            if bp is not None and bp.expires_at != -1 and (
+                                bp.expires_at - 60
+                            ) < int(time.time()):
+                                # TODO: circular import
+                                from core.plugin.impl.oauth import OAuthHandler
+                                from services.tools.builtin_tools_manage_service import BuiltinToolManageService
 
-                            # refresh the credentials
-                            tool_provider = ToolProviderID(provider_id)
-                            provider_name = tool_provider.provider_name
-                            redirect_uri = (
-                                f"{dify_config.CONSOLE_API_URL}/console/api/oauth/plugin/"
-                                f"{provider_id}/tool/callback"
-                            )
-                            system_credentials = BuiltinToolManageService.get_oauth_client(tenant_id, provider_id)
+                                # refresh the credentials
+                                tool_provider = ToolProviderID(provider_id)
+                                provider_name = tool_provider.provider_name
+                                redirect_uri = (
+                                    f"{dify_config.CONSOLE_API_URL}/console/api/oauth/plugin/"
+                                    f"{provider_id}/tool/callback"
+                                )
+                                system_credentials = BuiltinToolManageService.get_oauth_client(
+                                    tenant_id, provider_id
+                                )
 
-                            oauth_handler = OAuthHandler()
-                            refreshed_credentials = oauth_handler.refresh_credentials(
-                                tenant_id=tenant_id,
-                                user_id=builtin_provider.user_id,
-                                plugin_id=tool_provider.plugin_id,
-                                provider=provider_name,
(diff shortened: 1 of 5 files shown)


Petrus-Han:
@crazywoola this PR has been manually check for 24 hours, works well.

api/core/tools/tool_manager.py
```diff
@@ -267,35 +268,71 @@ def get_tool_runtime(
...
-                builtin_provider.encrypted_credentials = json.dumps(
-                    encrypter.encrypt(refreshed_credentials.credentials)
-                )
-                builtin_provider.expires_at = refreshed_credentials.expires_at
-                db.session.commit()
-                decrypted_credentials = refreshed_credentials.credentials
-                cache.delete()
+                lock_key = f"oauth_token_refresh_lock:{tenant_id}_{provider_id}_{builtin_provider.id}"
+                lock = redis_client.lock(lock_key, timeout=30)
+                if lock.acquire(blocking=False):
```
QuantumGhost: It's better to use context manager to  `with redis_client.lock(...)` to acquire / release lock. This improves readability of the following code.

Alai-sudo: Thanks for the suggestion! This has been addressed in commit f556590 — we replaced the manual `lock.acquire()/lock.release()` with `with redis_client.lock(...)` context manager + `except LockError:` pattern.

Alai-sudo:
@laipz8200 Could you help review this PR when you have a chance? All previous review feedback has been addressed. Thanks!

PR #32209, opened 2026-02-10
fix: convert UUID to str in deleted_tools API provider check

### LINKED ISSUE #32205: bug: deleted_tools property marks all API tools as "removed" due to UUID vs string comparison
## Bug Description

In the `deleted_tools` property of the `App` model (`api/models/model.py`), all API tool providers are incorrectly reported as deleted/removed even when they exist in the `tool_api_providers` table. This causes the UI to show a warning triangle with "tool removed" on every API tool in agent configurations.

## Root Cause

The property at line ~228-234 builds `existing_api_providers` using a list comprehension:

```python
existing_api_providers = [
    api_provider.id  # Returns UUID object from SQLAlchemy Row
    for api_provider in session.execute(
        text("SELECT id FROM tool_api_providers WHERE id IN :provider_ids"),
        {"provider_ids": tuple(api_provider_ids)},
    ).fetchall()
]
```

Later, it checks:
```python
if provider_id not in existing_api_providers:  # str NOT IN [UUID, UUID, ...]
    deleted_tools.append(...)
```

The issue is that `api_provider.id` returns a `uuid.UUID` object (from SQLAlchemy's Row), while `provider_id` (from the JSON config) is a `str`. In Python, `"38eb56c1-..." != UUID("38eb56c1-...")`, so the `in` check always fails, marking ALL API tools as deleted.

## Steps to Reproduce

1. Create an Agent Chat app
2. Add an API tool provider with an OpenAPI schema
3. Add tools from that provider to the agent
4. View the agent configuration in the UI
5. All API tools show a warning triangle with "tool removed"

Despite the warning, the tools still function correctly at runtime because `ToolManager.get_tool_runtime()` resolves the provider separately.

## Expected Behavior

API tools that exist in `tool_api_providers` should NOT be marked as deleted.

## Suggested Fix

Change line ~230 in `api/models/model.py`:

```diff
existing_api_providers = [
-    api_provider.id
+    str(api_provider.id)
    for api_provider in session.execute(
        text("SELECT id FROM tool_api_providers WHERE id IN :provider_ids"),
        {"provider_ids": tuple(api_provider_ids)},
    ).fetchall()
]
```

## Environment

- Dify version: 1.12.1 (self-hosted, Kubernetes)
- Database: PostgreSQL 16
- Python: 3.12

## Additional Context

Verified by running the `deleted_tools` property through the ORM:
- Before fix: 44 tools marked as deleted (all API tools)
- After fix: 0 tools marked as deleted

The fix has been verified in production and resolves the issue completely.

agent-steven:
## Summary
Fix a type mismatch bug where all API tool providers are incorrectly marked as deleted in the `deleted_tools` property.

## Problem
`api_provider.id` returns a `uuid.UUID` object from SQLAlchemy, but `provider_id` from the app JSON config is a `str`. Since `"abc" != UUID("abc")` in Python, the `in` check always fails, causing every API tool to appear as "removed" in the UI.

## Fix
Cast `api_provider.id` to `str()` so the type comparison works correctly.

```diff
existing_api_providers = [
-    api_provider.id
+    str(api_provider.id)
    for api_provider in session.execute(...)
]
```

## Verification
As reported in the issue: before fix 44 tools marked deleted, after fix 0.

Closes #32205

**Disclosure**: AI-assisted identification of the fix pattern.

### DIFF
--- a/api/models/model.py
+++ b/api/models/model.py
@@ -227,7 +227,7 @@ class App(Base):
         with Session(db.engine) as session:
             if api_provider_ids:
                 existing_api_providers = [
-                    api_provider.id
+                    str(api_provider.id)
                     for api_provider in session.execute(
                         text("SELECT id FROM tool_api_providers WHERE id IN :provider_ids"),
                         {"provider_ids": tuple(api_provider_ids)},


crazywoola:
Sorry, but duplicated with this one https://github.com/langgenius/dify/pull/32207

PR #32299, opened 2026-02-13
fix(api): make DB migration Redis lock TTL configurable and prevent LockNotOwnedError from masking failures

### LINKED ISSUE #32297: DB migrations: Redis lock TTL (60s) can expire during long migrations causing concurrent upgrades and LockNotOwnedError masking failures (Kubernetes)
### Self Checks

### Dify version

v1.13 / 84d090db33c6cd0d7c6f3139b03eb3a5e4fb6797 / Latest

### Cloud or Self Hosted

Self Hosted (Docker), Cloud, Self Hosted (Source)

### Steps to reproduce

1. Deploy Dify with multiple components/pods that share the same API image entrypoint (e.g. Kubernetes deployments for `api`, `worker`, `worker_beat`, or docker-compose services).
2. Ensure migrations run on startup (default behavior):
   - `MIGRATION_ENABLED=true`
   - container entrypoint runs `flask upgrade-db` before starting the main process.
3. Create/trigger a migration that can take longer than 60 seconds in real environments (large tables/indexes, slow DB, constrained resources, busy DB).
4. While the migration is still running, trigger one of the following (common in Kubernetes):
   - rolling update
   - pod restart / crash-loop recovery
   - scale up/down
   - node reschedule
5. Observe logs and DB state when multiple instances attempt `flask upgrade-db` during the same deployment window.

### ✔️ Expected Behavior

- Database migrations (Alembic / Flask-Migrate) should run exactly once per deployment event and never run concurrently against the same database.
- If a migration fails, logs and exit status should clearly reflect the real root cause (Alembic/DB error), not be obscured by Redis lock cleanup errors.
- Deployments (especially Kubernetes) should have a clear, safe migration strategy (e.g. a dedicated migration job/init step), rather than having every long-running service attempt migrations at startup.

### ❌ Actual Behavior

There are multiple migration-related issues that can surface especially on Kubernetes:

### 1) Redis migration lock TTL is too short (60s), allowing concurrent migrations
- `flask upgrade-db` uses a Redis lock with a fixed `timeout=60` seconds.
- If a migration takes longer than 60 seconds, the lock key can expire while the migration process is still running.
- A second instance (pod/container) starting after the TTL expiry may acquire the lock and start a concurrent migration.
- This can cause migration failures, deadlocks, partial upgrades, or inconsistent schema states.

### 2) `LockNotOwnedError` (“owner” error) can mask the real migration failure
- After the migration attempt (success or failure), code unconditionally calls `lock.release()` in a `finally` block.
- When the lock has expired (or was re-acquired by another instance), `lock.release()` may raise `LockNotOwnedError` / owner errors.
- In practice this can become the most visible error in logs and make debugging confusing, while the real failure is the migration itself (or a concurrency conflict triggered by TTL expiry).

### 3) Default deployment behavior runs migrations from multiple services
- The same API image is used by multiple services (`api`, `worker`, `worker_beat`) and all of them run migrations on startup when `MIGRATION_ENABLED=true`.
- This “many services run migrations” pattern increases the probability of races during restarts/rollouts, and makes correctness depend entirely on the Redis lock behavior.

### Relevant code paths
- Startup migration trigger:
  - `api/docker/entrypoint.sh` runs `flask upgrade-db` when `MIGRATION_ENABLED=true` (before starting the main process).
- Migration command:
  - `api/commands.py` defines `upgrade-db` and runs `flask_migrate.upgrade()`.
- Redis lock (hard-coded TTL):
  - `redis_client.lock(name="db_upgrade_lock", timeout=60)`

### Why this is worse on Kubernetes
- Pod restarts/rollouts are common.
- Multiple replicas/components can start at different times.
- Migrations can exceed 60 seconds under constrained resources or busy databases.

### Suggested direction (for discussion)
- Make migration locking robust for long migrations (configurable TTL and safe release behavior).
- Provide an official Kubernetes-friendly pattern:
  - run migrations in a dedicated Job/init step
  - set `MIGRATION_ENABLED=false` for long-running services (`api/worker/beat`) to avoid redundant attempts

lin-snow:
> [!IMPORTANT]
>

Fixes #32297

This PR hardens startup database migrations (`flask upgrade-db`) in multi-replica environments (e.g. Kubernetes) to prevent concurrent migrations and ensure failures are observable.

### Background / Problems

- The migration command uses a Redis lock (`db_upgrade_lock`) to avoid concurrent upgrades, but long-running migrations can outlive a short lock TTL, allowing a second instance to acquire the lock and start a concurrent migration.
- When the lock expires or ownership changes, lock cleanup can raise `LockNotOwnedError`, which can distract from or mask the real Alembic/DB migration failure in logs.
- Simply increasing the TTL is not sufficient: if a pod is killed mid-migration and cannot release the lock, a long TTL can leave a stale lock and block future upgrades for too long.

### Changes

- Use a short base lock TTL (fixed in code) and keep it alive via heartbeat renewal while migrations are running:
  - Introduce a migration-specific helper `DbMigrationAutoRenewLock` that renews the redis-py lock TTL using `lock.reacquire()` from a daemon heartbeat thread.
  - The underlying lock is created with `thread_local=False` so the renewal thread can access the token.
  - If the process is killed, renewal stops and the lock naturally expires soon, avoiding long stale-lock windows.
- Make lock release best-effort and non-masking:
  - Lock renewal/release errors (`LockNotOwnedError` / `RedisError`) are logged at `warning` level with stack traces (`exc_info=True`) for debugging.
  - These errors never change the migration exit code and do not mask the real migration failure (Alembic/DB error still results in exit 1).
- Remove TTL env/config plumbing:
  - Drop the `MIGRATION_LOCK_TTL` environment/config option and any docker compose exposure, to keep behavior consistent and avoid stale-lock tuning pitfalls.
- Tests:
  - Unit tests cover skip-on-lock-busy, failure not masked by release errors, heartbeat renewal behavior, and renewal error tolerance.
  - Add a TestContainers integration test using real Redis to validate TTL renewal and release end-to-end.

No new dependencies.

## Screenshots

| Before | After |
|--------|-------|
| N/A (backend-only, no UI change) | N/A |

## Checklist

- [ ] I ran `make lint` and `make type-check` (backend) and `cd web && npx lint-staged` (frontend) to appease the lint gods


api/tests/unit_tests/commands/test_upgrade_db.py
```diff
@@ -0,0 +1,83 @@
+import sys
+import types
+from unittest.mock import MagicMock
+
+import commands
```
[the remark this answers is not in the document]
lin-snow: In this case the unit under test is the Click command implementation living in `commands.py`, and the behavior relies on module-level state (e.g. `redis_client`, `DB_UPGRADE_LOCK_TTL_SECONDS`). Importing the module directly keeps the patching explicit and avoids accidentally patching a stale reference.

If this command is refactored into a smaller injectable unit in the future, we can switch to importing specific functions/classes. For now, the direct module import is intentional to keep the test aligned with the actual CLI entrypoint behavior.

api/tests/unit_tests/commands/test_upgrade_db.py
```diff
@@ -0,0 +1,83 @@
...
+        return int(e.code or 0)
+    return 0
+
+
+def test_upgrade_db_skips_when_lock_not_acquired(monkeypatch, capsys):
+    monkeypatch.setattr(dify_config, "MIGRATION_LOCK_TTL", 1234)
+
+    lock = MagicMock()
+    lock.acquire.return_value = False
+    commands.redis_client.lock.return_value = lock
```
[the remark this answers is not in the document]
lin-snow: In our unit test setup, `api/tests/unit_tests/conftest.py` provides a shared `redis_mock` and resets it before each test (autouse fixture), including re-pointing any loaded modules’ `redis_client` references to the mock. That gives us test isolation and prevents cross-test pollution.

Here we only override `redis_client.lock.return_value` for the duration of each test, and the mock is reset automatically between tests. If we see flakiness later, we can tighten this further with `monkeypatch.setattr(commands.redis_client, "lock", ...)`, but current isolation should be sufficient.

api/commands.py
```diff
@@ -54,6 +56,35 @@
 
 logger = logging.getLogger(__name__)
 
+if TYPE_CHECKING:
+    from redis.lock import Lock
+
+DB_UPGRADE_LOCK_TTL_SECONDS = 60
+
+
+def _heartbeat_db_upgrade_lock(lock: "Lock", stop_event: threading.Event, ttl_seconds: float) -> None:
```
hj24: Seems it's a general logic that can be used for any redis lock, can we move it to api/libs?

api/commands.py
```diff
@@ -727,8 +758,22 @@ def create_tenant(email: str, language: str | None = None, name: str | None = No
...
-    lock = redis_client.lock(name="db_upgrade_lock", timeout=60)
+    # Use a short base TTL + heartbeat renewal, so a crashed process doesn't block migrations for long.
+    # thread_local=False is required because heartbeat runs in a separate thread.
+    lock = redis_client.lock(
+        name="db_upgrade_lock",
+        timeout=DB_UPGRADE_LOCK_TTL_SECONDS,
+        thread_local=False,
+    )
     if lock.acquire(blocking=False):
+        stop_event = threading.Event()
```
hj24: Consider packing it into a general clean API, so that invoker don't need to consider too much on the threading things, for example:

```suggestion
lock = AutoRenewRedisLock(name="xxx", ttl=xxx, ...)
lock.acquire(blocking=False)
```

api/commands.py
```diff
@@ -737,14 +782,31 @@ def upgrade_db():
...
         finally:
-            lock.release()
+            stop_event.set()
+            heartbeat_thread.join(timeout=5)
+            # Lock release errors should never mask the real migration failure.
+            try:
+                lock.release()
+            except LockNotOwnedError:
+                status = "successful" if migration_succeeded else "failed"
+                logger.warning(
```
hj24: logger.exception

api/commands.py
```diff
@@ -737,14 +782,31 @@ def upgrade_db():
...
+                    "DB migration lock not owned on release after %s migration (likely expired); ignoring.", status
+                )
+            except RedisError:
+                status = "successful" if migration_succeeded else "failed"
+                logger.warning(
```
hj24: ditto

api/tests/unit_tests/commands/test_upgrade_db.py
```diff
@@ -0,0 +1,145 @@
...
+
+def _invoke_upgrade_db() -> int:
+    try:
+        commands.upgrade_db.callback()
+    except SystemExit as e:
+        return int(e.code or 0)
+    return 0
+
+
+def test_upgrade_db_skips_when_lock_not_acquired(monkeypatch, capsys):
```
hj24, line 26: consider to add a containers_integration test case to use real redis to test the lib

api/commands.py
```diff
@@ -54,6 +56,35 @@
 
 logger = logging.getLogger(__name__)
```
QuantumGhost, line 57: IMO `if TYPE_CHECKING:` should only be used if absolutely necessary. (E.G. to avoid a circular import).

api/tests/unit_tests/commands/test_upgrade_db.py
QuantumGhost: I recommend **against** extracting this into a shared library.

Using a background thread to renew the lock is a workaround specific to the constraints of his migration framework:

1. The migration execution time is unbounded.
2. The main thread is blocked by long-running DDL/DML operations, preventing it from manually refreshing the lock.

However, this pattern does not apply to most application code, where we have better control over the execution flow.

Relying on background threads for lock renewal is risky and generally discouraged. For example, if the main task crashes or hangs but the background thread continues to renew the lock, it defeats the purpose of the TTL and creates a "zombie" lock (effectively a deadlock).

Since `redis-py` already provides `Lock.extend()`, we should encourage developers to manage lock lifecycles explicitly for better safety and predictability.

**Update:** I see the extraction is already part of this change. **If we do want to merge** `AutoRenewRedisLock` as a shared utility, please add a clear warning in the comments. We need to ensure developers understand the risks and use `Lock.extend` for standard scenarios.

lin-snow: Thanks for the thoughtful feedback — I agree that auto-renewing locks via a background thread is generally not a good default for typical application code.

In this specific case (startup DB migrations), we have two constraints:
1) migration execution time is effectively unbounded (long DDL/DML),
2) the main thread is blocked inside `flask_migrate.upgrade()` so we can’t reliably call `Lock.extend()` / `reacquire()` from the same thread.

Without renewal, the lock can expire mid-migration and allow concurrent upgrades, which is the failure mode we’re trying to prevent.

On the “zombie lock” concern: if the process is killed, the daemon thread stops and the lock expires quickly (we keep a short base TTL). If the migration hangs but the process is still alive, keeping the lock held is arguably the safer failure mode than allowing another instance to start a concurrent migration; operators can restart the stuck pod to stop renewal.

That said, I agree on scope: to avoid encouraging misuse, we can keep this helper migration-specific (naming/location/docstring) rather than presenting it as a general-purpose shared lock utility.

QuantumGhost: Yes, I see that for migration the deadlock is not a problem as the  migration happens inside a dedicated process. The background worker process stop when the migration process exits.

QuantumGhost: One more points: You should set the `daemon` argument of the background to `True` or the worker thread will block the main thread from exitting.

[ref: Python docs about the daemon argument.](https://docs.python.org/3/library/threading.html#threading.Thread)

api/libs/db_migration_lock.py
```diff
@@ -0,0 +1,196 @@
+"""
+DB migration Redis lock with heartbeat renewal.
+
+This is intentionally migration-specific. Background renewal is a trade-off that makes sense
+for unbounded, blocking operations like DB migrations (DDL/DML) where the main thread cannot
+periodically refresh the lock TTL.
+
+Do NOT use this as a general-purpose lock primitive for normal application code. Prefer explicit
+lock lifecycle management (e.g. redis-py Lock context manager + `extend()` / `reacquire()` from
```
QuantumGhost, line 9: 👍

api/tests/unit_tests/commands/test_upgrade_db.py
lin-snow: The heartbeat thread is created with `daemon=True` (see `DbMigrationAutoRenewLock._start_heartbeat()`), so it won’t block process exit if the main migration flow finishes or exits early.

api/libs/db_migration_lock.py
```diff
@@ -0,0 +1,195 @@
...
+
+        self._stop_heartbeat()
+
+        # Lock release errors should never mask the real error/exit code.
+        try:
+            lock.release()
+        except LockNotOwnedError:
+            self._logger.warning(
+                "DB migration lock not owned on release%s%s; ignoring.",
+                f" after {status} operation" if status else "",
```
QuantumGhost: avoiding use f-string for logging, use log string with [printf-style format](https://docs.python.org/3/library/stdtypes.html#old-string-formatting) instead.

QuantumGhost (approved):
LGTM

PR #32317, opened 2026-02-13
fix: json body corruption when string variables contain special characters

### LINKED ISSUE #31927: HTTP Request corrupts input
### Self Checks

### Dify version

1.11.2

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Create a Code block that returns string 
`### a \n\n#### a \n\n| a | a | a | a | a (a) | a | a | a / a | a |\n| --- | --- | --- | --- | --- | --- | --- | --- | --- |\n| a | a | a\\a\\a\\a | a (...) a (...) a .a .a : a — a — a ;a , a .a [a.a](a://a).*(a) a : "[a.a](a://a)" – a, a + "&a\\a=a-a-a". a "[a://a.a.a/a?a=a&a=a-a-a](a://a.a.a/a?a=a)"* | a/a | a /a/@a , a @a -a /a/@a != a(a)*a – a "a" a -a .* | a = a.a.a | a | a |\n| a | a | a\\a\\a\\a | a (...) a (...) a .a .a : a — a — a ;a , a .a [a.a](a://a).*(a) a : "[a.a](a://a)" – a, a + "&a\\a=a-a-a". a "[a://a.a.a/a?a=a&a=a-a-a](a://a.a.a/a?a=a)"* | a/a | a /a/@a , a @a | a = a.a.a | a | a |`  and a HTTP Request block that makes request with body (type JSON):

```markdown
{
    "model": "pro",
    "messages": [{
        "role": "user",
        "content": {{#1770191412056.result#}}
     }]
}
```

`{{#1770191412056.result#}}` is an output from Code block.

### ✔️ Expected Behavior

HTTP Request block does NOT parse outputs from previous blocks! It has to use it as it is.

### ❌ Actual Behavior

When you run this flow you can get an error in HTTP Request block because HTTP Request block somehow parses string from Code block and the resulting Processed Data becomes malformed:

```markdown
{
  "request": "POST /some_method HTTP/1.1\r\nHost: my-host\r\nContent-Type: application/json\r\nAuthorization: ********************************\r\nContent-Length: 307\r\n\r\n{\"model\": \"pro\", \"messages\": [{\"role\": \"user\", \"content\": \"\", \"a\": \"a \\u2014 a \\u2014 a ;a , a .a [a.a](a://a).*![(a)](/a/a/a/a/a/a/a/a/a.a) a :\", \"a.a](a\": \"\"}, {\"a\": \"\"}, {\"a\": \"\"}, {\"a\": \"a \\u2014 a \\u2014 a ;a , a .a [a.a](a://a).*![(a)](/a/a/a/a/a/a/a/a/a.a) a :\", \"a.a](a\": \"\"}, {\"a\": \"\"}]}"
}
```

So, basically the HTTP Request block or something between blocks corrupts the string.

I tried to put {{#1770191412056.result#}} in quotes or double quotes but it works only if there is no quotes or doubles quotes in string correspondingly.

Mr-Neutr0n:
When the HTTP Request node builds a JSON body from a template like `{"content": {{#node.result#}}}`, it was doing raw string substitution via `convert_template().text` and then running `repair_json` on the result. If the variable value contains quotes, backslashes, newlines, or pipe characters (common in markdown), the substitution breaks the JSON structure and `repair_json` corrupts the data rather than fixing it.

The root cause is that string values weren't being JSON-escaped before substitution. For example, a value like `hello "world"` would produce `{"content": hello "world"}` — broken JSON that `repair_json` can't reliably reconstruct.

Fixed by adding `_convert_template_for_json()` which uses `json.dumps()` to properly encode string variable values before they land in the JSON structure. Numbers, objects, and arrays continue to use their existing `.text` representation which already produces valid JSON tokens. Also handles the edge case where template authors already wrap variables in quotes (e.g. `"{{#var#}}"`).

Added two regression tests covering strings with quotes/backslashes/newlines and markdown table content.

Closes #31927

### DIFF
--- a/api/core/workflow/nodes/http_request/executor.py
+++ b/api/core/workflow/nodes/http_request/executor.py
@@ -14,8 +14,9 @@ from configs import dify_config
 from core.file.enums import FileTransferMethod
 from core.file.file_manager import file_manager as default_file_manager
 from core.helper.ssrf_proxy import ssrf_proxy
-from core.variables.segments import ArrayFileSegment, FileSegment
+from core.variables.segments import ArrayFileSegment, FileSegment, StringSegment
 from core.workflow.runtime import VariablePool
+from core.workflow.runtime.variable_pool import VARIABLE_PATTERN
 
 from ..protocols import FileManagerProtocol, HttpClientProtocol
 from .entities import (
@@ -189,12 +190,16 @@ class Executor:
                 case "json":
                     if len(data) != 1:
                         raise RequestBodyError("json body type should have exactly one item")
-                    json_string = self.variable_pool.convert_template(data[0].value).text
+                    json_string = self._convert_template_for_json(data[0].value)
                     try:
-                        repaired = repair_json(json_string)
-                        json_object = json.loads(repaired, strict=False)
-                    except json.JSONDecodeError as e:
-                        raise RequestBodyError(f"Failed to parse JSON: {json_string}") from e
+                        json_object = json.loads(json_string, strict=False)
+                    except json.JSONDecodeError:
+                        # Fall back to repair_json for malformed templates
+                        try:
+                            repaired = repair_json(json_string)
+                            json_object = json.loads(repaired, strict=False)
+                        except json.JSONDecodeError as e:
+                            raise RequestBodyError(f"Failed to parse JSON: {json_string}") from e
                     self.json = json_object
                     # self.json = self._parse_object_contains_variables(json_object)
                 case "binary":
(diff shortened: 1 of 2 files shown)


MezentsevIlya:
Why not just remove this JSON "repairing" and modification? It's non-obvious behaviour (hidden changes that the user cannot know about or turn off) and definitely an anti-pattern. If a user creates JSON, they want to send it to the endpoint as-is. If the JSON is invalid, it's the user's problem and the user fixes it. Period.

Or at least there should be an option to disable this behaviour.

Mr-Neutr0n:
@MezentsevIlya you're right in direction, and I owe you a proper answer — sorry it took this long.

But I think the disagreement dissolves once we're precise about *who* produces the invalid JSON, because it isn't the user.

## The user's JSON is valid. Dify breaks it.

The template from #31927 is well-formed:

```json
{"model": "pro", "messages": [{"role": "user", "content": {{#node.result#}}}]}
```

The node then pastes the raw variable value into a JSON string position with no encoding. Reproduced against the real code path:

```
substituted:                    {"model": "pro", ... "content": ### a\n\n| a | a\a\a | a : "[a.a](a://a)" ...
valid JSON after substitution?  NO — Expecting value
after repair_json:              content = ''        <-- the whole value, discarded
```

So `"If the JSON is invalid, it's the user's problem and the user fixes it"` doesn't have a target here. There is nothing for the author to fix — they wrote valid JSON and got a corrupted request. That's the platform's bug, and `repair_json` exists to paper over it.

And note what "repairing" actually did: not a hidden modification, a **silent total data loss**. The request went out with `"content": ""`. There's a second variant too, where it *injects* a character: `{"content": "prefix {{#var#}} suffix"}` with value `a"b` comes back as `prefix a"b suffix"`.

## Which makes your position and this patch the same position

Escaping at substitution time means well-formed templates never produce invalid JSON, so `repair_json` stops being load-bearing. **This is the change that makes removing it possible** — right now it can't be removed, because the platform depends on it to clean up its own mess.

I kept it as a fallback only for templates that were malformed *before* substitution, since some existing workflows likely lean on that and dropping it is a breaking change. If you'd rather it be removed outright or put behind a switch, I agree that's the better end state, and it's a clean follow-up once the cause is fixed. I'd just rather not bundle a breaking change with a data-loss fix.

## This PR can't merge, though — the code moved

While this sat, `api/core/workflow/nodes/http_request/` was extracted into [`langgenius/graphon`](https://github.com/langgenius/graphon) (`api/pyproject.toml` pins `graphon==0.6.0`). The file this PR edits no longer exists in this repo, which is why the branch shows as conflicting.

The bug is verbatim intact in graphon 0.7.0 (`src/graphon/nodes/http_request/executor.py:275`), so I've rewritten the fix against it and opened **langgenius/graphon#235**. It's a cleaner implementation than what's here: quote state is tracked across the literal parts instead of the fragile "is it already wrapped in quotes" adjacency check, the log-safe copy is encoded the same way, and there are six regression tests, three of which fail on unpatched `main`.

Closing this one in favour of that. Thanks for pushing on the design — the rewrite is better for it.

Issue #32362, opened 2026-02-16
Hydration warning: markdown inline image renderer nests <div> inside <p>


lyzno1:
## Summary
When rendering Markdown that includes an inline image inside a paragraph, the web app logs an invalid DOM nesting warning:

`In HTML, <div> cannot be a descendant of <p>. This will cause a hydration error.`

This can lead to hydration mismatch/noise in Next.js.

## Environment
- Dify web frontend
- Next.js `16.1.5` (Turbopack)

## Reproduction
1. Open a page that renders Markdown through `ReactMarkdownWrapper`.
2. Render content where a paragraph contains an inline image, for example:
   - `Text before ![image](https://example.com/a.png) text after`
   - or raw HTML image inside text: `Text before <img src="https://example.com/a.png" alt="image" /> text after`
3. Open browser console.

## Actual Result
Console warning/error appears:
- `In HTML, <div> cannot be a descendant of <p>. This will cause a hydration error.`

## Expected Result
Markdown renderer should output valid HTML nesting without `<div>` inside `<p>`.

## Suspected Root Cause
The `img` renderer returns block-level wrappers even when used in paragraph (inline) context:

- `web/app/components/base/markdown-blocks/img.tsx` returns `<div className="markdown-img-wrapper"><ImageGallery .../></div>`
- `web/app/components/base/markdown-blocks/plugin-img.tsx` has the same pattern
- `web/app/components/base/image-gallery/index.tsx` root is `<div ...>`

When `Paragraph` falls back to `<p>{children}</p>` (e.g. image is not the first child), the rendered tree becomes `<p> ... <div> ... </div> ... </p>`, which is invalid HTML and triggers hydration warnings.

## Affected Call Chain (from runtime stack)
- `ReactMarkdownWrapper` (`web/app/components/base/markdown/react-markdown-wrapper.tsx`)
- `Paragraph` (`web/app/components/base/markdown-blocks/paragraph.tsx`)
- `Img` (`web/app/components/base/markdown-blocks/img.tsx`)
- `ImageGallery` (`web/app/components/base/image-gallery/index.tsx`)

## Notes
This issue is isolated to report/triage only; no code changes are included in this report.

lyzno1:
https://github.com/user-attachments/assets/464cc2db-b1ce-4e85-8879-ae60749165f7

vibeyclaw:
[...]

PR #32366, opened 2026-02-16
fix(web): avoid invalid DOM nesting when inline images appear in paragraphs

vibeyclaw:
## Summary
When a Markdown paragraph contains an inline image that is **not** the first child (e.g. `Text before img text after`), the `Img` component renders a block-level `<div>` wrapper. This was placed inside a `<p>` tag, producing invalid HTML and triggering React/Next.js hydration warnings:

> `In HTML, <div> cannot be a descendant of <p>. This will cause a hydration error.`

## Changes
- Updated `Paragraph` component (`web/app/components/base/markdown-blocks/paragraph.tsx`) to detect when **any** child node (not just the first) is an image element.
- When a non-leading image is found, renders as `<div class="markdown-paragraph">` instead of `<p>`, keeping the DOM valid.
- The existing first-child-is-image logic (dedicated `ImageGallery` block) is unchanged.

## Testing
- Rendered Markdown with inline images in various positions within paragraphs
- Confirmed no hydration warning in browser console
- Confirmed existing image-as-first-child rendering is unaffected

Closes #32362

### DIFF
--- a/web/app/components/base/markdown-blocks/paragraph.tsx
+++ b/web/app/components/base/markdown-blocks/paragraph.tsx
@@ -1,15 +1,31 @@
 /**
  * @fileoverview Paragraph component for rendering <p> tags in Markdown.
  * Extracted from the main markdown renderer for modularity.
- * Handles special rendering for paragraphs that directly contain an image.
+ * Handles special rendering for paragraphs that contain images.
+ *
+ * When a paragraph's first child is an image, it renders as a dedicated
+ * image block with ImageGallery. When images appear elsewhere among the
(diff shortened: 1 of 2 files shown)


vibeyclaw:
Thanks for the thorough review! Great catch on `plugin-paragraph.tsx` — I've pushed a follow-up commit (eb0bc90) that applies the same `<div>` fallback for non-first-child images to `PluginParagraph` as well.

Regarding the style suggestions: updated accordingly. Let me know if anything else needs attention!

lyzno1:
Closing as duplicate of #32419 which addresses the same issue with a shared utility approach.

PR #32410, opened 2026-02-19
fix: avoid invalid DOM nesting when inline image is not first paragraph child

tysoncung:
## Problem

When rendering Markdown where an inline image appears inside a paragraph but is **not** the first child (e.g. `Text before img text after`), the `Paragraph` component falls through to rendering `<p>`, while the `Img` renderer wraps the image in a `<div>`. This creates invalid `<p><div>...</div></p>` nesting, which triggers Next.js hydration warnings:

> In HTML, `<div>` cannot be a descendant of `<p>`. This will cause a hydration error.

## Solution

Added a check for image tags among **all** children of the paragraph node, not just the first child. When any child is an image, the component renders a `<div>` wrapper instead of `<p>`, preventing invalid DOM nesting.

The existing first-child optimization (which extracts the image into an `ImageGallery`) is preserved for backwards compatibility.

## Changes

- `web/app/components/base/markdown-blocks/paragraph.tsx`: Added `hasImageChild` helper and fallback `<div>` rendering when images are found in non-first positions.

Fixes #32362

### DIFF
--- a/web/app/components/base/markdown-blocks/paragraph.tsx
+++ b/web/app/components/base/markdown-blocks/paragraph.tsx
@@ -6,13 +6,36 @@
 import * as React from 'react'
 import ImageGallery from '@/app/components/base/image-gallery'
 
-const Paragraph = (paragraph: any) => {
-  const { node }: any = paragraph
+type MdastNode = {
+  tagName?: string
+  children?: MdastNode[]
+  properties?: Record<string, unknown>
+}
+
+type ParagraphProps = {
+  node: MdastNode
+  children: React.ReactNode[]
+}
+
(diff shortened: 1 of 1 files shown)


tysoncung:
Thanks for the review feedback! Addressed both suggestions:

- Made `hasImageChild` recursive to handle nested images (e.g., linked images `[alt](link)` and formatted images `**alt**`)
- Added explicit `false` return for null/undefined children
- Changed styling from `mb-4` to `markdown-img-wrapper` for consistency with the first-child image rendering path

tysoncung:
Closing in favour of #32419 which covers the same fix. Thanks!

PR #32419, opened 2026-02-19
fix: prevent hydration warning from div nesting inside p for inline markdown images

tysoncung:
## Summary

Fixes #32362

When a markdown paragraph contains an inline image that is **not** the first child (e.g. `Text before img text after`), the paragraph renderer fell through to a `<p>` wrapper while the image component rendered block-level `<div>` elements inside it. This produced invalid HTML (`<div>` inside `<p>`) and triggered React/Next.js hydration warnings.

## Changes

- **`paragraph.tsx`**: Added `hasImageChild()` helper to detect any image element among paragraph children. When an image is found (but isn't the first child), renders a `<div class="markdown-paragraph">` instead of `<p>` to avoid invalid nesting.
- **`plugin-paragraph.tsx`**: Same logic applied to the plugin paragraph variant.

## How it works

The existing code already handled the case where an image is the **first** child by rendering a `<div>` wrapper. This fix extends that pattern to cover images appearing **anywhere** in the paragraph, ensuring valid HTML output in all cases.


tysoncung:
Thanks for the review! Addressed the code duplication feedback:

- Extracted `hasImageChild` to the shared `utils.ts` file as suggested
- Both `paragraph.tsx` and `plugin-paragraph.tsx` now import from the shared utility
- Made the implementation recursive to handle nested images (linked/formatted)
- Applied the same div-fallback pattern to `plugin-paragraph.tsx` for consistency

tysoncung:
Friendly bump - CI is green, tests passing. Ready for merge whenever convenient. Thanks!

lyzno1 (approved):

PR #32581, opened 2026-02-25
fix: replace 13 bare except clauses with except Exception

haosenwang1018:
## What
Replace 13 bare `except:` clauses with `except Exception:` across 12 files.

## Why
Bare `except:` catches `BaseException`, including `KeyboardInterrupt` and `SystemExit`, which can prevent clean process shutdown and mask critical errors.


haosenwang1018:
CI failures require manual investigation.

PR #32586, opened 2026-02-25
feat: support redis xstream

fatelei:
> [!IMPORTANT]
>

## Summary

fix #32518

use redis xstream to publish and consume topic, eliminating the subscribe-before-publish requirement entirely

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/configs/middleware/cache/redis_pubsub_config.py
+++ b/api/configs/middleware/cache/redis_pubsub_config.py
@@ -1,7 +1,7 @@
 from typing import Literal, Protocol
 from urllib.parse import quote_plus, urlunparse
 
-from pydantic import Field
+from pydantic import AliasChoices, Field
 from pydantic_settings import BaseSettings
 
 
@@ -23,39 +23,54 @@ class RedisConfigDefaultsMixin:
 
 class RedisPubSubConfig(BaseSettings, RedisConfigDefaultsMixin):
     """
-    Configuration settings for Redis pub/sub streaming.
+    Configuration settings for event transport between API and workers.
+
+    Supported transports:
+    - pubsub: Redis PUBLISH/SUBSCRIBE (at-most-once)
+    - sharded: Redis 7+ Sharded Pub/Sub (at-most-once, better scaling)
+    - streams: Redis Streams (at-least-once, supports late subscribers)
     """
 
     PUBSUB_REDIS_URL: str | None = Field(
-        alias="PUBSUB_REDIS_URL",
+        validation_alias=AliasChoices("EVENT_BUS_REDIS_URL", "PUBSUB_REDIS_URL"),
         description=(
-            "Redis connection URL for pub/sub streaming events between API "
-            "and celery worker, defaults to url constructed from "
-            "`REDIS_*` configurations"
+            "Redis connection URL for streaming events between API and celery worker; "
+            "defaults to URL constructed from `REDIS_*` configurations. Also accepts ENV: EVENT_BUS_REDIS_URL."
         ),
         default=None,
     )
 
     PUBSUB_REDIS_USE_CLUSTERS: bool = Field(
+        validation_alias=AliasChoices("EVENT_BUS_REDIS_CLUSTERS", "PUBSUB_REDIS_USE_CLUSTERS"),
         description=(
-            "Enable Redis Cluster mode for pub/sub streaming. It's highly "
-            "recommended to enable this for large deployments."
+            "Enable Redis Cluster mode for pub/sub or streams transport. Recommended for large deployments. "
+            "Also accepts ENV: EVENT_BUS_REDIS_CLUSTERS."
         ),
         default=False,
     )
 
-    PUBSUB_REDIS_CHANNEL_TYPE: Literal["pubsub", "sharded"] = Field(
+    PUBSUB_REDIS_CHANNEL_TYPE: Literal["pubsub", "sharded", "streams"] = Field(
+        validation_alias=AliasChoices("EVENT_BUS_REDIS_CHANNEL_TYPE", "PUBSUB_REDIS_CHANNEL_TYPE"),
         description=(
-            "Pub/sub channel type for streaming events. "
-            "Valid options are:\n"
-            "\n"
-            " - pubsub: for normal Pub/Sub\n"
-            " - sharded: for sharded Pub/Sub\n"
-            "\n"
-            "It's highly recommended to use sharded Pub/Sub AND redis cluster "
-            "for large deployments."
+            "Event transport type. Options are:\n\n"
+            " - pubsub: normal Pub/Sub (at-most-once)\n"
+            " - sharded: sharded Pub/Sub (at-most-once)\n"
+            " - streams: Redis Streams (at-least-once, recommended to avoid subscriber races)\n\n"
+            "Note: Before enabling 'streams' in production, estimate your expected event volume and retention needs.\n"
+            "Configure Redis memory limits and stream trimming appropriately (e.g., MAXLEN and key expiry) to reduce\n"
+            "the risk of data loss from Redis auto-eviction under memory pressure.\n"
+            "Also accepts ENV: EVENT_BUS_REDIS_CHANNEL_TYPE."
         ),
         default="pubsub",
+    )
+
+    PUBSUB_STREAMS_RETENTION_SECONDS: int = Field(
+        validation_alias=AliasChoices("EVENT_BUS_STREAMS_RETENTION_SECONDS", "PUBSUB_STREAMS_RETENTION_SECONDS"),
+        description=(
+            "When using 'streams', expire each stream key this many seconds after the last event is published. "
+            "Also accepts ENV: EVENT_BUS_STREAMS_RETENTION_SECONDS."
+        ),
+        default=600,
     )
 
     def _build_default_pubsub_url(self) -> str:
--- a/api/libs/broadcast_channel/redis/streams_channel.py
+++ b/api/libs/broadcast_channel/redis/streams_channel.py
@@ -1,0 +1,159 @@
+from __future__ import annotations
+
+import logging
+import queue
+import threading
+from collections.abc import Iterator
+from typing import Self
+
+from libs.broadcast_channel.channel import Producer, Subscriber, Subscription
+from libs.broadcast_channel.exc import SubscriptionClosedError
+from redis import Redis, RedisCluster
+
+logger = logging.getLogger(__name__)
+
+
+class StreamsBroadcastChannel:
+    """
+    Redis Streams based broadcast channel implementation.
+
+    Characteristics:
+    - At-least-once delivery for late subscribers within the stream retention window.
+    - Each topic is stored as a dedicated Redis Stream key.
+    - The stream key expires `retention_seconds` after the last event is published (to bound storage).
+    """
+
+    def __init__(self, redis_client: Redis | RedisCluster, *, retention_seconds: int = 600):
+        self._client = redis_client
+        self._retention_seconds = max(int(retention_seconds or 0), 0)
+
+    def topic(self, topic: str) -> StreamsTopic:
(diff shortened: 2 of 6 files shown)


api/libs/broadcast_channel/redis/streams_channel.py
```diff
@@ -0,0 +1,154 @@
...
+        self._queue: deque[bytes] = deque()
+        self._lock = threading.Lock()
+        self._listener: threading.Thread | None = None
+
+    def _listen(self) -> None:
+        # Use XREAD in a loop; block up to 1 second to allow timely shutdown.
+        while not self._closed.is_set():
+            try:
+                streams = self._client.xread({self._key: self._last_id}, block=1000, count=100)
+            except Exception:
```
QuantumGhost: Please please do not swallow exceptions. 

If the exception catching here is absolutely necessary, at least log the exception with `logger.exception` to signal the developer an error has occurred and the developer could fix the bug accordingly.

api/libs/broadcast_channel/redis/streams_channel.py
```diff
@@ -0,0 +1,154 @@
...
+
+            if not streams:
+                continue
+
+            # streams is a list of (key, [(id, {field: value}), ...])
+            for _key, entries in streams:
+                for entry_id, fields in entries:
+                    data = None
+                    if isinstance(fields, dict):
+                        # Support both new 'data' and legacy 'd' keys, in bytes or str forms.
```
QuantumGhost: If I understand this correctly, this feature is new and there is no  legacy `"d"` keys?

QuantumGhost (commented):
I'm wondering why this PR uses `deque` instead of relying on `queue.Queue` as the other two implementations do. Is there any special reason?

api/configs/middleware/cache/redis_pubsub_config.py
```diff
@@ -23,41 +23,45 @@ def _redis_defaults(self: RedisConfigDefaults) -> RedisConfigDefaults:
...
-            "\n"
-            " - pubsub: for normal Pub/Sub\n"
-            " - sharded: for sharded Pub/Sub\n"
-            "\n"
-            "It's highly recommended to use sharded Pub/Sub AND redis cluster "
-            "for large deployments."
+            "Event transport type. Options are:\n\n"
+            " - pubsub: normal Pub/Sub (at-most-once)\n"
+            " - sharded: sharded Pub/Sub (at-most-once)\n"
+            " - streams: Redis Streams (at-least-once, recommended to avoid subscriber races)"
```
QuantumGhost: I suggest adding a comment here to recommend performing volume computation prior to applying the `streams` transport type. This helps prevent data loss issues caused by Redis auto-eviction.

api/configs/middleware/cache/redis_pubsub_config.py
```diff
@@ -23,41 +23,45 @@ def _redis_defaults(self: RedisConfigDefaults) -> RedisConfigDefaults:
...
         ),
         default="pubsub",
     )
 
+    PUBSUB_STREAMS_RETENTION_SECONDS: int = Field(
```
QuantumGhost, line 67: Maybe we should update the configuration names too?

Since the stream does not use the PubSub mechanism, prefixing the configuration name with `PUBSUB_` is a little weird.

I would suggest use `EVENT_BUS_REDIS_URL` and rename other configuration items accordingly:

- PUBSUB_REDIS_URL -> EVENT_BUS_REDIS_URL
- PUBSUB_REDIS_USE_CLUSTERS -> EVENT_BUS_REDIS_CLUSTERS
- PUBSUB_STREAMS_RETENTION_SECONDS -> EVENT_BUS_STREAMS_RETENTION_SECONDS

That said, we may adjust this in another PR.

fatelei:
> I'm wondering why this PR uses `deque` instead of relying on `queue.Queue` as the other two implementations do. Is there any special reason?

streams aims for at-least-once delivery without local drops, so a lightweight FIFO (deque) with a small lock fits the semantics and keeps overhead down

api/libs/broadcast_channel/redis/streams_channel.py
```diff
@@ -0,0 +1,161 @@
...
+        self.max_length = 5000
+
+    def as_producer(self) -> Producer:
+        return self
+
+    def publish(self, payload: bytes) -> None:
+        try:
+            self._client.xadd(self._key, {b"data": payload}, maxlen=self.max_length)
+        except TypeError:
+            self._client.xadd(self._key, {b"data": payload})
```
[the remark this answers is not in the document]
QuantumGhost: The `MAXLEN` option is [supported since Redis 5.0](https://redis.io/docs/latest/commands/xadd/). The try catch here is not necessary.

QuantumGhost:
> > I'm wondering why this PR uses `deque` instead of relying on `queue.Queue` as the other two implementations do. Is there any special reason?
> 
> streams aims for at-least-once delivery without local drops, so a lightweight FIFO (deque) with a small lock fits the semantics and keeps overhead down

I understand the goal of keeping the implementation lightweight. However, `deque` is only "faster" because it omits the synchronization logic that we are now manually adding back via `self._lock` and `time.sleep()`. 

The current busy-loop (polling) approach in `receive` is a bit concerning. It introduces an artificial latency floor (up to 50ms) and may be less CPU-efficient than an event-driven approach. If we were to implement a custom queue properly using condition variables to avoid polling, we would essentially be re-inventing the `queue.Queue` class.

Most importantly, since this project relies on `gevent`, [`queue.Queue` is already monkey-patched](https://github.com/gevent/gevent/blob/master/src/gevent/monkey/__init__.py#L264-L294) to use `gevent.queue.Queue`. This implementation is specifically designed for our environment: it [uses `deque` as its underlying data structure and handles greenlet yielding/waiting properly without the need for `time.sleep()`](https://github.com/gevent/gevent/blob/master/src/gevent/queue.py). 

By switching to `queue.Queue`, we get:
1. **Zero-latency wake-ups**: No more waiting for the 50ms polling interval; consumers are notified immediately when data is available.
2. **Cleaner code**: Removing manual synchronization and polling logic makes the implementation more robust and easier to maintain.

I suggest we stick to the standard `queue.Queue` to keep the codebase idiomatic and take full advantage of gevent's optimizations. If you do not need watermark control and back pressure, you could use `queue.SimpleQueue`.

I have also vibed a benchmark script. The result shows that `queue.Queue` is about 3x faster than the manual implementation.

Result on my computer:

```
--- Environment: Native Thread ---
=== Concurrency Benchmark (Native Thread) ===

Starting test: Manual Deque (Lock + Sleep)...
[Manual Deque (Lock + Sleep)] Processed 10000000 messages.
[Manual Deque (Lock + Sleep)] Total Duration: 2.6738 seconds
[Manual Deque (Lock + Sleep)] Avg Throughput: 3739970.63 msg/sec

Starting test: Standard Queue (Native Thread)...
[Standard Queue (Native Thread)] Processed 10000000 messages.
[Standard Queue (Native Thread)] Total Duration: 7.4632 seconds
[Standard Queue (Native Thread)] Avg Throughput: 1339916.43 msg/sec

Starting test: Standard Simple Queue (Native Thread)...
[Standard Simple Queue (Native Thread)] Processed 10000000 messages.
[Standard Simple Queue (Native Thread)] Total Duration: 1.8146 seconds
[Standard Simple Queue (Native Thread)] Avg Throughput: 5510827.41 msg/sec


--- Environment: Gevent Patched ---
=== Concurrency Benchmark (Gevent Patched) ===

Starting test: Manual Deque (Lock + Sleep)...
[Manual Deque (Lock + Sleep)] Processed 10000000 messages.
[Manual Deque (Lock + Sleep)] Total Duration: 5.8528 seconds
[Manual Deque (Lock + Sleep)] Avg Throughput: 1708585.46 msg/sec

Starting test: Standard Queue (Gevent Patched)...
[Standard Queue (Gevent Patched)] Processed 10000000 messages.
[Standard Queue (Gevent Patched)] Total Duration: 1.9640 seconds
[Standard Queue (Gevent Patched)] Avg Throughput: 5091610.81 msg/sec

Starting test: Standard Simple Queue (Gevent Patched)...
[Standard Simple Queue (Gevent Patched)] Processed 10000000 messages.
[Standard Simple Queue (Gevent Patched)] Total Duration: 1.7282 seconds
[Standard Simple Queue (Gevent Patched)] Avg Throughput: 5786518.69 msg/sec
```

Benchmark script:

```py
import os

# 1. Environment Setup: Apply Gevent Monkey Patching if requested
# This must be done before importing threading/queue to ensure proper patching.
if os.environ.get('USE_GEVENT') == '1':
   from gevent import monkey
   monkey.patch_all()
   mode_label = "Gevent Patched"
   print(f"--- Environment: {mode_label} ---")
else:
    mode_label = "Native Thread"
    print(f"--- Environment: {mode_label} ---")

import sys
import time
import collections
import threading
import queue


# 2. Manual Implementation (The one being reviewed)
class ManualDequeQueue:
    def __init__(self):
        self._queue = collections.deque()
        self._lock = threading.Lock()
        self._stop_event = threading.Event()

    def enqueue(self, data):
        with self._lock:
            self._queue.append(data)

    def dequeue(self):
        while not self._stop_event.is_set():
            with self._lock:
                if self._queue:
                    return self._queue.popleft()
            # This sleep introduces artificial latency and polling overhead
            time.sleep(0.05)
        return None

    def stop(self):
        self._stop_event.set()

# 3. Standard Library Implementation (Recommended)
class StandardQueue:
    def __init__(self, queue_class=queue.Queue):
        self._queue = queue_class()
        self._stop_event = threading.Event()

    def enqueue(self, data):
        self._queue.put(data)

    def dequeue(self):
        while not self._stop_event.is_set():
            try:
                # Use a small timeout to allow periodic stop_event checks
                return self._queue.get(timeout=0.05)
            except queue.Empty:
                continue
        return None

    def stop(self):
        self._stop_event.set()

# 4. Concurrency Test Logic
def run_concurrency_test(q_obj, name, total_messages=10_000_000):
    print(f"Starting test: {name}...")
    start_time = time.perf_counter()

    consumed_count = [0]

    def producer():
        for i in range(total_messages):
            q_obj.enqueue(f"msg-{i}")
            # Simulate a small production interval every 100 messages
            # if i % 100 == 0:
            #     time.sleep(0.0001)

    def consumer():
        while consumed_count[0] < total_messages:
            item = q_obj.dequeue()
            if item:
                consumed_count[0] += 1
        q_obj.stop()

    # Standard threading.Thread will be patched to Greenlets if USE_GEVENT=1
    t_prod = threading.Thread(target=producer)
    t_cons = threading.Thread(target=consumer)

    t_prod.start()
    t_cons.start()
    t_prod.join()
    t_cons.join()

    end_time = time.perf_counter()
    duration = end_time - start_time

    print(f"[{name}] Processed {total_messages} messages.")
    print(f"[{name}] Total Duration: {duration:.4f} seconds")
    print(f"[{name}] Avg Throughput: {total_messages / duration:.2f} msg/sec\n")

if __name__ == "__main__":
    # High load test to observe throughput and synchronization overhead
    MSG_COUNT = 500000

    manual_q = ManualDequeQueue()
    standard_q = StandardQueue()
    standard_simple_q = StandardQueue(queue.SimpleQueue)

    print(f"=== Concurrency Benchmark ({mode_label}) ===\n")

    run_concurrency_test(manual_q, "Manual Deque (Lock + Sleep)")
    run_concurrency_test(standard_q, f"Standard Queue ({mode_label})")
    run_concurrency_test(standard_simple_q, f"Standard Simple Queue ({mode_label})")
```

benchmark command:

```bash
# Native thread
uv run bench-queue.py
# Gevent
USE_GEVENT=1 uv run bench-queue.py
```

QuantumGhost (approved):

BaHoang:
This fix is critical for our production deployment. We upgraded from 1.7.1 to 1.13.0 and the SSE race condition causes frequent message_end event loss. Looking forward to the next stable release that includes this fix. Is there an ETA for 1.14.0 stable?

PR #32602, opened 2026-02-25
Fix hydration warning for inline markdown images

mani-agah-esmaeilzad:
**What**
Render inline markdown images as  to avoid invalid HTML nesting (<div> inside <p>), preventing hydration warnings.

**Why**
Inline image renderer previously returned a wrapper <div> which can be placed inside <p> nodes by the markdown parser, producing invalid HTML and hydration errors.

**How to test**
Render markdown like: Text before image text after

Confirm no console warning: “In HTML, <div> cannot be a descendant of <p>”.

Fixes #32362

### DIFF
--- a/web/app/components/base/markdown-blocks/img.tsx
+++ b/web/app/components/base/markdown-blocks/img.tsx
@@ -4,10 +4,8 @@
  * Uses the ImageGallery component to display images.
  */
 import * as React from 'react'
-import ImageGallery from '@/app/components/base/image-gallery'
 
-const Img = ({ src }: any) => {
-  return <div className="markdown-img-wrapper"><ImageGallery srcs={[src]} /></div>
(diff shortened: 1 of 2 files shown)


mani-agah-esmaeilzad:
> # Comment
> 
> The hydration mismatch likely occurs because the markdown image rendering differs between server and client. Try ensuring consistent image URL resolution and lazy-loading attributes are identical on both sides, or consider moving image processing to a useEffect hook to defer client-side rendering until hydration is complete.



Thanks for the suggestion!
In this case, the hydration warning was caused by invalid HTML nesting (`<div> inside <p>`) rather than a server/client rendering mismatch.
Rendering inline markdown images as  ensures valid structure and resolves the hydration error without deferring rendering to useEffect.

lyzno1:

Issue #32379, opened 2026-02-17
Increase test coverage for core module


rajatagarwal-oss:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] Please do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing? Tell me about your story.

The core module currently has limited test coverage, would like to add missing tests to improve coverage.

### 2. Additional context or comments

Idea is to review the existing coverage and add missing unit tests for untested logic and edge cases in the core module.

### 3. Can you help us with this feature?

- [x] I am interested in contributing to this feature.

PR #32637, opened 2026-02-26
test: unit test cases for core.variables, core.plugin, core.prompt module

rajatagarwal-oss:
> [!IMPORTANT]
>
> 1. Use the correct syntax to link this PR: `Fixes #32379 `.

## Summary

Added pytest for core.plugin, core.prompt, core.variables module
Fix: #32379 

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


api/tests/unit_tests/core/plugin/test_plugin_entities.py
```diff
@@ -0,0 +1,344 @@
...
+        class _Unknown(StrEnum):
+            CUSTOM = "custom"
+
+        assert cast_parameter_value(_Unknown.CUSTOM, 12) == "12"
+
+        class _BadString:
+            def __str__(self):
+                raise RuntimeError("boom")
+
+        with pytest.raises(RuntimeError, match="boom"):
```
QuantumGhost: `cast_parameter_value` has a bug. When `str(value)` fails, the `except` block tries to construct a `ValueError` message by interpolating `value` again. That can fail for the same reason, which means the function may re-raise the original exception instead of raising `ValueError` as intended.

So this test is pinning the current defect in the implementation, not the expected contract. I think the correct fix is to make `cast_parameter_value` reliably raise `ValueError` in this case, and then update the test accordingly.

I suggest changing the final line of `cast_parameter_value` to:

```py
raise ValueError(
    f"The tool parameter value {repr(value)} is not in correct type of {as_normal_type(typ)}."
)
```

Using `repr(value)` is safer here than interpolating `value` directly, because it avoids invoking the normal string-formatting path again when constructing the error message.

QuantumGhost (changes requested):

QuantumGhost (approved):

PR #31723, opened 2026-01-29
refactor(typing): Fixup typing A2 - workflow engine & nodes

BeautyyuYanli:
## Context

Part of #31680 (**A2 stage: workflow engine & nodes**).

## What this PR does

Unblocks running `ty` on the workflow engine/nodes layer by tightening types, improving narrowing, and removing a few cross-layer import/typing issues.

Key changes:

- **Typed workflow node config**
  - Introduces `core.workflow.entities.graph_config` with `NodeConfigDict` + `NodeConfigDictAdapter`.
  - Updates workflow graph parsing (`core.workflow.graph.graph.Graph`) to validate node configs via Pydantic `TypeAdapter` and use typed indexing.
  - Updates `Workflow.get_node_config_by_id()` and workflow entry code to use the typed node config.

- **Protocol-friendly dependency injection for HTTP request node**
  - Adds thin adapters:
    - `core.helper.ssrf_proxy.SSRFProxy` implementing `HttpClientProtocol`
    - `core.file.file_manager.FileManager` implementing `FileManagerProtocol`
  - Updates `HttpRequestNode` / executor and `DifyNodeFactory` to accept optional injected deps with safe defaults (no behavioral change intended).

- **Type/narrowing cleanups inside workflow code**
  - Reduces `cast()` usage where types are already structurally guaranteed.
  - Improves runtime narrowing (`isinstance`) and a few small typing fixes in nodes (e.g. LLM prompt content handling, list operator file field extraction, datasource type conversion).

- **Tooling**
  - Bumps `ty` to `>=0.0.14` and updates `ty.toml` rules/excludes for this stage.

## Why

The workflow codebase has import-linter enforced layering, including that `core.workflow.nodes.*` must not import `core.workflow.graph.*`.

During the A2 work, `models/workflow.py` importing node-config typing from `core.workflow.graph.graph` created an indirect dependency chain that broke the rule once graph parsing started depending on node-typed configs.

Moving the shared node-config `TypedDict` into `core.workflow.entities` keeps the architecture boundary intact while still allowing `Graph`/`Workflow`/factories to share the same validated config shape.

## Testing

- `uv run --directory api --dev ty check`
- `uv run --directory api --dev lint-imports`

PR #32780, opened 2026-03-01
refactor: Unify NodeConfigDict.data and BaseNodeData

BeautyyuYanli:
> [!IMPORTANT]
>

## Summary

- Following #31723.
- Unified `NodeConfigDict.data` and `BaseNodeData`.
- Updated workflow/runtime/services and related tests to pass and validate typed node data consistently.

**This PR is drafted by `gpt-5.3-codex(xhigh)` and I'm responsible for all the changes. I have reviewed the code and varified the behavior, while minor breaks may still exist. Reach me to fix in this case.**

## Screenshots

| Before | After |
|--------|-------|
| N/A | N/A |

## Checklist


api/dify_graph/entities/base_node_data.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # Validate array element types
+        if validator["type"] == list and not self._validate_array(self.value, validator["element_type"]):
+            raise DefaultValueTypeError(f"All elements must be {validator['element_type'].__name__} for {self.value}")
+
+        return self
+
+
+class BaseNodeData(ABC, BaseModel):
+    model_config = ConfigDict(extra="allow")
```
laipz8200, line 129: `BaseNodeData` is now the shared DTO boundary for workflow node payloads, so `ConfigDict(extra="allow")` makes every node-data model permissive by default. That conflicts with the scoped backend rule in [`api/AGENTS.md`](https://github.com/langgenius/dify/blob/main/api/AGENTS.md#pydantic-usage): `Define DTOs with Pydantic v2 models and forbid extras by default.` If a legacy compatibility escape hatch is still needed, it should be opt-in on the specific model that needs it rather than on the base class.

api/dify_graph/entities/graph_config.py
```diff
@@ -4,21 +4,18 @@
...
-@with_config(extra="allow")
-class NodeConfigData(TypedDict):
-    type: str
-
-
 @with_config(extra="allow")
 class NodeConfigDict(TypedDict):
     id: str
-    data: NodeConfigData
+    data: BaseNodeData
```
laipz8200, line 20: Changing `NodeConfigDict.data` to `BaseNodeData` makes graph parsing validate every node's inner `data.type` as a `NodeType` before the existing `custom-note` filter runs in `Graph.init()`. That breaks real note payloads: the shipped templates in [`api/constants/pipeline_templates.json`](https://github.com/langgenius/dify/blob/main/api/constants/pipeline_templates.json) still contain nodes with top-level `type: custom-note` and inner `data.type: ""`, so workflows/templates containing notes now fail with a `ValidationError` instead of being filtered out.

api/dify_graph/nodes/trigger_webhook/entities.py
```diff
@@ -28,26 +43,31 @@ class WebhookParameter(BaseModel):
     """Parameter definition for headers, query params, or body."""
 
     name: str
+    type: SegmentType = SegmentType.STRING
```
laipz8200, line 62: `WebhookParameter.type` now accepts `object`, `array[...]`, and `file` for both query params and headers, but the runtime still extracts query params from `request.args` and routes them through `_convert_form_value()`, which only handles string/number/boolean; header types are never validated at all. So these new configs either fail every matching request (query params) or silently become a no-op (headers). Please either restrict `WebhookParameter` to the types `WebhookService` actually supports here, or update the service layer to handle the widened type set end to end.

laipz8200 (commented):
Here are some issues identified by the LLM.

laipz8200 (commented):
Here are some suggestions.

api/dify_graph/nodes/base/exc.py
laipz8200, line 1: I still believe that errors should be placed in a separate `exc.py` file.

api/dify_graph/nodes/base/entities.py
```diff
@@ -210,3 +85,18 @@ class MetaData(BaseModel):
         pass
 
     metadata: MetaData
+
+
+__all__ = [
```
laipz8200: I would like you to update the import paths later on, rather than using indirect imports.

api/dify_graph/entities/base_node_data.py
BeautyyuYanli: This is required since the derived NodeData types have extra fields relative to the BaseNodeData

api/dify_graph/entities/graph_config.py
BeautyyuYanli: fixed in [1ad18fa](https://github.com/langgenius/dify/pull/32780/commits/1ad18fa09f13afd51ab25e4f459e39f78fabb5f9)

api/core/workflow/node_factory.py
```diff
@@ -156,31 +161,29 @@ def _resolve_dify_context(run_context: Mapping[str, Any]) -> DifyRunContext:
...
 
         :param node_config: node configuration dictionary containing type and other data
         :return: initialized Node instance
-        :raises ValueError: if node type is unknown or configuration is invalid
+        :raises ValueError: if node type is unknown or no implementation exists for the resolved version
+        :raises ValidationError: if node_config fails NodeConfigDict/BaseNodeData validation
         """
-        # Get node_id from config
-        node_id = node_config["id"]
+        typed_node_config = NodeConfigDictAdapter.validate_python(node_config)
```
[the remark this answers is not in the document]
BeautyyuYanli: false positive. pydantic_core._pydantic_core.ValidationError is a subclass of ValueError

BeautyyuYanli:
@gemini-code-assist review

api/core/workflow/node_factory.py
```diff
@@ -157,178 +154,128 @@ def _resolve_dify_context(run_context: Mapping[str, Any]) -> DifyRunContext:
...
+        node_id = typed_node_config["id"]
+        node_data = typed_node_config["data"]
+        node_class = self._resolve_node_class(node_type=node_data.type, node_version=str(node_data.version))
+        node_type = node_data.type
+        node_init_kwargs_factories: Mapping[NodeType, Callable[[], dict[str, object]]] = {
+            NodeType.CODE: lambda: {
```
laipz8200, line 173: Is the Lambda necessary here?

laipz8200 (approved):

PR #32861, opened 2026-03-02
fix(web): resolve hydration warning caused by nesting div inside p in markdown image renderer

zqyhimself:
### Motivation
When rendering Markdown content that includes an inline image inside a paragraph, the web app logs an invalid DOM nesting warning and triggers a hydration mismatch in Next.js SSR:

> In HTML, `<div>` cannot be a descendant of `<p>`. This will cause a hydration error.

Fixes langgenius/dify#32362

### Approach
Replaced all block-level `<div>` wrappers used in the image rendering pipeline with `<span style={{ display: 'block' }}>`. A `<span>` is phrasing content and is valid inside `<p>`, while `display: block` preserves the existing visual layout. This change touches the full affected call chain identified in the issue:

- `img.tsx` — outer `markdown-img-wrapper`
- `plugin-img.tsx` — outer `markdown-img-wrapper`
- `paragraph.tsx` — image-first branch wrapper + remaining-children wrapper
- `plugin-paragraph.tsx` — image-first branch wrapper + remaining-children wrapper
- `image-gallery/index.tsx` — root container element (also replaced Tailwind `flex flex-wrap` utility with equivalent inline styles for consistency)

### Testing
- Verified that the DOM output no longer produces `<p><div>...</div></p>` nesting in any code path.
- Existing test suite (`plugin-img.spec.tsx`) remains green as the change is purely at the HTML element level.
- Checked that `display: block` on `<span>` visually matches the previous `<div>` rendering in all image gallery layouts (single, 2-up, 3-up grids).

### DIFF
--- a/web/app/components/base/image-gallery/index.tsx
+++ b/web/app/components/base/image-gallery/index.tsx
@@ -36,22 +36,22 @@ const ImageGallery: FC<Props> = ({
   const imgNum = srcs.length
   const imgStyle = getWidthStyle(imgNum)
   return (
-    <div className={cn(s[`img-${imgNum}`], 'flex flex-wrap')}>
+    <span className={cn(s[`img-${imgNum}`])} style={{ display: 'flex', flexWrap: 'wrap' }}>
       {srcs.map((src, index) => (
         !src
           ? null
           : (
-              <img
-                key={index}
-                className={s.item}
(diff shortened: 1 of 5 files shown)


zqyhimself:
Sorry, there was a slight error when I submitted the PR. I will modify it.

PR #32900, opened 2026-03-03
refactor: human input node decouple db

### LINKED ISSUE #32896: human input node decouple db
### Self Checks

human input node decouple db

_No response_

fatelei:
> [!IMPORTANT]
>

## Summary

fix #32896

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/core/repositories/human_input_repository.py
+++ b/api/core/repositories/human_input_repository.py
@@ -4,9 +4,10 @@ from collections.abc import Mapping, Sequence
 from datetime import datetime
 from typing import Any
 
-from sqlalchemy import Engine, select
-from sqlalchemy.orm import Session, selectinload, sessionmaker
+from sqlalchemy import select
+from sqlalchemy.orm import Session, selectinload
 
+from core.db.session_factory import session_factory
 from dify_graph.nodes.human_input.entities import (
     DeliveryChannelConfig,
     EmailDeliveryMethod,
@@ -198,12 +199,9 @@ class _InvalidTimeoutStatusError(ValueError):
 class HumanInputFormRepositoryImpl:
     def __init__(
         self,
-        session_factory: sessionmaker | Engine,
+        *,
         tenant_id: str,
     ):
-        if isinstance(session_factory, Engine):
-            session_factory = sessionmaker(bind=session_factory)
-        self._session_factory = session_factory
         self._tenant_id = tenant_id
 
     def _delivery_method_to_model(
@@ -217,7 +215,7 @@ class HumanInputFormRepositoryImpl:
             id=delivery_id,
             form_id=form_id,
             delivery_method_type=delivery_method.type,
-            delivery_config_id=delivery_method.id,
+            delivery_config_id=str(delivery_method.id),
             channel_payload=delivery_method.model_dump_json(),
         )
         recipients: list[HumanInputFormRecipient] = []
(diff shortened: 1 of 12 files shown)


api/core/repositories/human_input_repository.py
```diff
@@ -198,14 +200,26 @@ class _InvalidTimeoutStatusError(ValueError):
...
+        self._lock = threading.Lock()
         self._tenant_id = tenant_id
 
+    def _get_session_factory(self) -> sessionmaker:
+        if self._session_factory is not None:
+            return self._session_factory
+
+        with self._lock:
+            if self._session_factory is None:
+                self._session_factory = sessionmaker(bind=db.engine)
```
QuantumGhost: Is it better to use the global session factory here instead of relying on flask_sqlalchemy's db instance?

api/core/repositories/human_input_repository.py
```diff
@@ -343,7 +357,7 @@ def _query_workspace_members_by_ids(
     def create_form(self, params: FormCreateParams) -> HumanInputFormEntity:
         form_config: HumanInputNodeData = params.form_config
 
-        with self._session_factory(expire_on_commit=False) as session, session.begin():
+        with self._get_session_factory()(expire_on_commit=False) as session, session.begin():
```
QuantumGhost: This call syntax seems weird. How about add a `expire_on_commit: bool = False` default parameter to `_get_session_factory`?

api/dify_graph/nodes/human_input/human_input_node.py
```diff
@@ -38,6 +36,10 @@
 logger = logging.getLogger(__name__)
 
 
+def naive_utc_now() -> datetime.datetime:
+    return datetime.datetime.now(datetime.UTC).replace(tzinfo=None)
```
QuantumGhost: There is a `naive_utc_now` in api/libs directory. Is this duplication intentional?

If this is intentional, I suggest add a short comment above to explain its motivation.

QuantumGhost (commented):
Other changes LGTM

api/core/repositories/human_input_repository.py
```diff
@@ -198,12 +199,9 @@ class _InvalidTimeoutStatusError(ValueError):
 class HumanInputFormRepositoryImpl:
     def __init__(
         self,
-        session_factory: sessionmaker | Engine,
+        *,
```
QuantumGhost, line 202: I apologize for not expressing myself clearly in my previous response. Let me clarify my point about the session factory implementation.

What I want to express is that the type of `session_factory` injected into `HumanInputFormRepositoryImpl` should be the `SessionFactory` class in the `api/extensions/ext_session_factory.py` file, rather than directly using flask-sqlalchemy's db instance.

This approach decouples database access from the flask-sqlalchemy framework, which is tightly coupled with the web framework implementation. Consider this scenario: we might want to move away from flask in the future. If the injected database session is retrieved directly from flask-sqlalchemy's global db value, we would need to modify all those usage sites accordingly. 

By relying on the `SessionFactory` abstraction instead, we only need to change the initialization of the `SessionFactory` class when migrating frameworks, leaving all other call sites intact. This provides better maintainability and reduces the coupling between our domain logic and the web framework.

The `SessionFactory` in `ext_session_factory.py` already provides this abstraction layer, configuring itself with the engine from flask-sqlalchemy but exposing a cleaner interface that isn't tied to flask's implementation details.

fatelei: ext_session_factory, actually is session_factory = SessionFactory(), so we can use this global session_factory directly

QuantumGhost: I see that the global instance `session_factory` is of type `SessionFactory`. However, relying on global database factory makes the test of `HumanInputFormRepositoryImpl` harder. IMO  keeping the `session_factory` parameter for HumanInputFormRepositoryImpl is more clear and improves readability.

QuantumGhost (approved):

PR #33044, opened 2026-03-05
fix: enterprise API error handling and license enforcement

GareArc:
## Summary

- Handle enterprise API errors properly to prevent `KeyError` crashes when the enterprise sidecar returns error responses
- Add global license check middleware that blocks console and webapp API access when the enterprise license is inactive, expired, or lost
- Use `UnauthorizedAndForceLogout` to trigger frontend cookie clearing and reload on license expiry
- Extend license enforcement to webapp API endpoints
- Add Redis caching (10 min TTL) for enterprise license status to avoid hammering the enterprise API on every request
- Exempt console bootstrap APIs (`/system-features`, `/setup`, `/features`, `/account/profile`, `/workspaces/current`, `/version`, `/activate/check`) from the license check to prevent infinite reload loops on the login page
- Expose license status and expiry to unauthenticated `/system-features` callers so the login page can show the expiry UI after force-logout, while keeping workspace usage details auth-gated

## Checklist


api/services/enterprise/enterprise_service.py
```diff
@@ -223,3 +227,45 @@ def cleanup_webapp(cls, app_id: str):
...
+
+        # Try cache first — only valid statuses are cached
+        try:
+            cached_status = redis_client.get(LICENSE_STATUS_CACHE_KEY)
+            if cached_status:
+                if isinstance(cached_status, bytes):
+                    cached_status = cached_status.decode("utf-8")
+                return cached_status
+        except Exception:
+            logger.debug("Failed to get license status from cache, calling enterprise API")
```
QuantumGhost: Is it really necessary to use  `except Exception:` here? I would suggest avoid unnecessary `try ... catch` statements as it does not improve stability of the application, only hiding bugs and making debugging harder.

If it is really necessary, I suggest  log the exception with `exc_info=True` and level equal to or above `info`. Logging exception with `debug` level would cause the exception information disapper in production environment.

api/services/enterprise/enterprise_service.py
```diff
@@ -223,3 +227,45 @@ def cleanup_webapp(cls, app_id: str):
...
+
+        # Cache miss or failure — call enterprise API
+        try:
```
QuantumGhost, line 277: The same suggestion above about exception handling applies to here, too.

Also, the deeply nested `try ... except` statements makes this part of code hard to understand.

api/services/feature_service.py
```diff
@@ -379,11 +379,14 @@ def _fulfill_params_from_enterprise(cls, features: SystemFeatureModel, is_authen
             )
             features.webapp_auth.sso_config.protocol = enterprise_info.get("SSOEnforcedForWebProtocol", "")
 
-        if is_authenticated and (license_info := enterprise_info.get("License")):
+        # License status and expiry are always exposed so the login page can
+        # show the expiry UI after a force-logout (the user is unauthenticated
+        # at that point).  Workspace usage details remain auth-gated.
+        if license_info := enterprise_info.get("License"):
             features.license.status = LicenseStatus(license_info.get("status", LicenseStatus.INACTIVE))
             features.license.expired_at = license_info.get("expiredAt", "")
```
QuantumGhost: I see that this is necessary since all other endpoints are blocked when EE license expires. However I still want to remind that It may be considered as information leakage and we have received [a vulnerability reports] (https://nvd.nist.gov/vuln/detail/CVE-2025-63387) for this in the past.

QuantumGhost (commented):
I would suggest define an enumeration class  for license status and avoid using string literals here and there.

The enumeration class could be defined with `StrEnum` and can be compared to string values directly without type conversion.

api/services/feature_service.py
GareArc: The unauthed system-feature endpoint only returns the license status (e.g. active/expired) — no other license fields are populated. Frontend needs this to display the license-expired page.

api/app_factory.py
```diff
@@ -31,6 +35,59 @@ def before_request():
...
+                    # on unauthorized_and_force_logout, causing an infinite loop.
+                    console_exempt_prefixes = (
+                        "/console/api/system-features",
+                        "/console/api/setup",
+                        "/console/api/init",
+                        "/console/api/login",
+                        "/console/api/version",
+                        "/console/api/activate/check",
+                    )
+                    is_exempt = any(request.path.startswith(p) for p in console_exempt_prefixes)
```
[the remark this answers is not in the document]
GareArc: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/langgenius/dify/pull/33044#discussion_r2903533471)

zhangx1n (approved):
LGTM

BenjaminX:
Hi, I noticed a regression introduced by this PR. After logging in, the "License expiring in X days" banner does not appear on the first page load — it only shows after a manual browser refresh.
                                                                                                                                                                                                                             
  Root cause:                                               
                                                                                                                                                                                                                             
  1. The frontend GlobalPublicStoreProvider calls GET /console/api/system-features before login (unauthenticated). At this point, license.status is returned but expired_at is "" because it's gated behind is_authenticated 
  (line 389-390 in feature_service.py).
  2. React Query caches this response with a 5-minute staleTime (query-client-server.ts).                                                                                                                                    
  3. After successful login, router.replace('/apps') navigates to the dashboard. useSystemFeaturesQuery() hits the stale cache from step 1 — it does not re-fetch because the data is still "fresh" within the 5-minute      
  window.                                                                                                                                                                                                                    
  4. The LicenseNav component reads systemFeatures.license.expired_at which is "", so dayjs("").diff(...) produces no meaningful result and the expiring badge is not rendered.                                              
  5. After 5 minutes (or a page refresh), the query re-fetches with cookies → is_authenticated=True → expired_at is populated → badge appears.                                                                               
                                                                                                                                                                                                                             
  Before this PR, expired_at was always returned regardless of auth status (from the intermediate fix 099568f3da), so the cached pre-login response already contained the full data and the badge worked immediately.        
                                                                                                                                                                                                                             
  Suggested fix (pick one):                                                                                                                                                                                                  
                                                            
  - Option A (backend): Move expired_at out of the is_authenticated guard — it's not sensitive information (the status EXPIRING already implies it), and the login page may also need it for countdown display:              
  if license_info := enterprise_info.get("License"):
      features.license.status = LicenseStatus(license_info.get("status", LicenseStatus.INACTIVE))                                                                                                                            
      features.license.expired_at = license_info.get("expiredAt", "")  # always expose                                                                                                                                       
                                                                                                                                                                                                                             
      if is_authenticated:  # only gate workspace usage details                                                                                                                                                              
          if workspaces_info := license_info.get("workspaces"):                                                                                                                                                              
              ...                                                                                                                                                                                                            
  - Option B (frontend): Invalidate the systemFeatures query cache after login so it re-fetches with authentication.
                                                                                                                                                                                                                             
  Option A is simpler and restores the behavior from commit 099568f3da which was explicitly intended to fix this exact scenario.

@GareArc

PR #33150, opened 2026-03-09
fix(web): handle undefined type in FileItem component

Br1an67:
Fixes #33122
Fixes https://github.com/langgenius/dify/issues/33156

## Summary
The FileItem component crashed when rendering files with an undefined `type` property (e.g., images generated by Stability AI). The component was calling `.split('/')` on the `type` without checking if it was defined first, causing a "Cannot read properties of undefined (reading 'split')" error.

## Changes
- Added optional chaining (`type?.split('/')`) in FileItem component to safely handle undefined type values
- Added mocks for AudioPreview and VideoPreview components in tests to prevent timeout issues
- Updated tests to use data-testid attributes instead of document.querySelector for better reliability
- Added test case for undefined type scenario

## Testing
- All existing tests pass
- New test case verifies component renders without crashing when type is undefined


crazywoola:
Please resolve the lint errors.

PR #33174, opened 2026-03-09
fix(web): resolve hydration warning for inline images in markdown paragraphs

nil957:
## Summary
When a paragraph contains an inline image (not as the first child), the Img component renders a `<div>` wrapper which becomes nested inside the `<p>` tag. This is invalid HTML and causes Next.js hydration warnings:

```
In HTML, <div> cannot be a descendant of <p>. This will cause a hydration error.
```

## Changes
- Detect if ANY child (not just the first) is an image using `.some()`
- When images are present but not first, render paragraph as `<div className="markdown-paragraph">` instead of `<p>` to avoid invalid nesting
- Preserve existing behavior for image-first paragraphs

## Test
1. Render markdown with inline image: `Text before image text after`
2. Console should no longer show hydration warning

Fixes #32362

### DIFF
--- a/web/app/components/base/markdown-blocks/paragraph.tsx
+++ b/web/app/components/base/markdown-blocks/paragraph.tsx
@@ -5,37 +5,42 @@
  * nesting <div> inside <p> which causes hydration warnings in Next.js.
  */
 import * as React from 'react'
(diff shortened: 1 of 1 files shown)


lyzno1:

PR #33188, opened 2026-03-10
fix: ensure external knowledge API key updates are persisted

### LINKED ISSUE #33186: Edit External Knowledge API: Updated API key is not saved; backend always receives [__HIDDEN__]
### Self Checks

### Dify version

1.13

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Steps to reproduce
Go to Knowledge → External Knowledge API and create an API (or use an existing one) with a valid API Endpoint and API Key.
Save it, then use it in an external knowledge base and confirm retrieval works (e.g. Retrieval Test returns 200).
Edit the same External Knowledge API: change the API Key to a new token (e.g. after rotating the key) and click Save.
Re-open the edit modal, or trigger retrieval again (e.g. Retrieval Test for a knowledge base using this API).

Actual result:

In the edit modal, the field still shows the old API key (the one before step 3).
Retrieval calls still use the old token, so if the old key was revoked you get 401 Unauthorized from the external API; the new key is never used.
Root cause:
In the edit flow, the frontend always sends api_key: '[__HIDDEN__]' in the PATCH request when saving. The backend treats [__HIDDEN__] as “keep existing value”, so the new key is never persisted. Only when the user actually changes the API key should the new value be sent; [__HIDDEN__] should be sent only when the user did not change the key (e.g. when the field is still the placeholder/masked value).

### ✔️ Expected Behavior

After editing the External Knowledge API and entering a new API Key, then clicking Save, the new key is stored and used for all subsequent requests.
Re-opening the edit modal shows the new API key (or a masked placeholder if the UI masks it).
Retrieval (and any other use of this API) uses the new token, so rotation and updates work as intended.

### ❌ Actual Behavior

In the edit modal, the field still shows the old API key (the one before step 3).
Retrieval calls still use the old token, so if the old key was revoked you get 401 Unauthorized from the external API; the new key is never used.
Root cause:
In the edit flow, the frontend always sends api_key: '[__HIDDEN__]' in the PATCH request when saving. The backend treats [__HIDDEN__] as “keep existing value”, so the new key is never persisted. Only when the user actually changes the API key should the new value be sent; [__HIDDEN__] should be sent only when the user did not change the key (e.g. when the field is still the placeholder/masked value).

Saber-youMianJing:
Fixes #33186
This PR fixes an issue where editing an External Knowledge API and updating its API Key
does not actually persist the new key.

Previously, the frontend always sent `[__HIDDEN__]` in the `api_key` field when saving
an existing External Knowledge API. The backend interprets `[__HIDDEN__]` as “keep the
existing value”, so any new key entered by the user was never saved. As a result,
external retrieval calls continued using the old token, which caused 401 Unauthorized
responses after the key was rotated or corrected.

This change only sends `[__HIDDEN__]` when the user has not modified the key
(the field value is still `[__HIDDEN__]`), and otherwise sends the new `api_key`
from the form so that updates are persisted and used by subsequent requests.

N/A – behavior change is around API key persistence. Verified via manual testing by:
- Editing the External Knowledge API key
- Saving
- Reopening the modal and triggering retrieval to confirm the new key is used
-
image

- [x] I have created and linked issue #33186 for this bug.
- [x] I have manually verified that updating the External Knowledge API Key now persists the new value and is used by subsequent retrieval calls.


MontaEllis8 (approved):
✅ Code Review 通过

代码审查通过，建议合并。

crazywoola (approved):

PR #33216, opened 2026-03-10
fix: preserve timing metrics in parallel iteration

### LINKED ISSUE #33214: [Bug] Parallel iteration inflates node elapsed time
## Summary
Parallel iteration buffers child node result events until the iteration finishes, which makes persisted and streamed node elapsed time include delayed replay time instead of the node's actual finish time.

## Reproduction
Run an iteration in parallel mode with an LLM node followed by downstream work inside the iteration body.

## Expected
The LLM node elapsed time should reflect when the node actually finished.

## Actual
The elapsed time can be inflated compared with non-parallel iteration because finish time is derived when the buffered event is replayed.

BeautyyuYanli:
> [!IMPORTANT]
>

Fixes #33214

## Summary

This fixes two timing issues in parallel iteration runs.
First, it preserves each child node's actual `finished_at` timestamp so node-level elapsed time is no longer inflated when buffered iteration events are replayed later.
Second, it measures each iteration's `iteration_duration_map` entry inside the worker thread before buffered events are replayed, so per-iteration timing is no longer polluted by replay and consumer delay.
It propagates finish timestamps through graph events, queue events, and error handling, and adds regression tests for the response-converter, persistence, and parallel iteration duration paths.

**This PR is drafted by `gpt-5.4(xhigh)` and I'm responsible for all the changes. I have reviewed the code and varified the behavior, while breaks may still exist. Reach me to fix in this case.**

## Screenshots

| Before | After |
|--------|-------|
| N/A | N/A |

## Checklist

- [x] I ran `make lint` and `make type-check` (backend); frontend was not touched.


gambletan (commented):
This PR correctly identifies and fixes a real timing accuracy problem in parallel iterations — when nodes finish inside a thread pool, the elapsed time was being computed at event consumption time rather than at actual completion time, inflating metrics. The fix propagates `finished_at` through the entire event pipeline.

**What's done well:**
- The fix is applied consistently across all three finish event types (succeeded, failed, exception) in both graph events and queue entities.
- The `finished_at or naive_utc_now()` fallback pattern in persistence and response converter maintains backward compatibility — events without `finished_at` (from older code paths) still work.
- The iteration node change from passing `start_at` to passing pre-computed `iteration_duration` is elegant — it correctly captures duration inside the worker thread before events are buffered and replayed.
- The `_build_fallback_failure_event` helper in worker.py is a nice refactoring that consolidates error handling and ensures even unexpected failures get proper timestamps.
- Regression tests directly validate that the event's `finished_at` is preferred over the delayed processing time.

**Observations:**

1. **`_current_node_started_at` threading concern** — In `worker.py`, `_current_node_started_at` is set during event iteration inside `_execute_node` and read in the `except` block of `run()`. Since each `Worker` instance runs in its own thread, this is safe. However, if the worker design ever changes to process multiple nodes concurrently, this instance variable would be a race condition. A comment noting the single-node-at-a-time assumption would be helpful.

2. **`NodeRunStartedEvent` detection for `_current_node_started_at`** — The code captures `start_at` from `NodeRunStartedEvent` during event iteration. If a node's `run()` method raises before yielding `NodeRunStartedEvent`, `_current_node_started_at` remains `None` and the fallback in `_build_fallback_failure_event` uses `failure_time` for both `start_at` and `finished_at`, giving `elapsed_time = 0`. This is reasonable behavior for a node that failed before it truly started.

3. **The `finally: self._current_node_started_at = None` cleanup** — Good practice to avoid stale state leaking between node executions.

4. **Type change in `_execute_single_iteration_parallel`** — The return type changes from `datetime` to `float` for the first tuple element. Any callers destructuring this tuple would need to be updated. Since `_execute_parallel_iterations` is the only consumer and it's updated here, this is fine.

5. **Test in `test_worker.py` using `put_side_effect` to simulate queue failure** — This is a creative way to trigger the fallback path. The test validates that `start_at` is preserved from the `NodeRunStartedEvent` even when the worker hits an unexpected error later. Well constructed.

Solid fix that addresses a real production accuracy issue with a clean, consistent approach.

fatelei (approved):

PR #33224, opened 2026-03-10
fix: fix parse mailto / http link in table cell

fatelei:
> [!IMPORTANT]
>

## Summary

fix #33221

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

after

image

## Checklist

### DIFF
--- a/api/core/rag/extractor/word_extractor.py
+++ b/api/core/rag/extractor/word_extractor.py
@@ -204,26 +204,61 @@ class WordExtractor(BaseExtractor):
         return " ".join(unique_content)
 
     def _parse_cell_paragraph(self, paragraph, image_map):
-        paragraph_content = []
-        for run in paragraph.runs:
-            if run.element.xpath(".//a:blip"):
-                for blip in run.element.xpath(".//a:blip"):
-                    image_id = blip.get("{http://schemas.openxmlformats.org/officeDocument/2006/relationships}embed")
-                    if not image_id:
-                        continue
-                    rel = paragraph.part.rels.get(image_id)
-                    if rel is None:
-                        continue
-                    # For external images, use image_id as key; for internal, use target_part
-                    if rel.is_external:
-                        if image_id in image_map:
-                            paragraph_content.append(image_map[image_id])
-                    else:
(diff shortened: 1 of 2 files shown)


gambletan (commented):
Good fix — iterating over `paragraph._element` children instead of `paragraph.runs` is the right approach since `w:hyperlink` elements are not exposed through the `runs` property. A few observations:

**What's done well:**
- The fallback to plain `link_text` when there's no external relationship is sensible — anchor-only bookmarks degrade gracefully to text.
- The broad `except Exception` with `logger.exception` around relationship resolution is appropriate here since a single malformed hyperlink shouldn't abort the entire document extraction.
- Tests cover both HTTP and mailto scenarios end-to-end with real `.docx` round-trips, which is thorough.

**Suggestions:**

1. **Bare `except Exception` swallows silently for the caller** — While logging is fine, consider whether the caller (`_parse_cell`) should have any indication that a hyperlink was partially resolved. Currently if `rels.get()` or `target_ref` access fails, the link text is still appended (good), but the `except` block also appends nothing when `link_text` happens to be empty. This is probably fine in practice but worth a comment noting the intentional behavior.

2. **`urlparse` on `target_ref` for scheme check** — The condition `rel.is_external or parsed_target.scheme in ("http", "https", "mailto")` is good, but note that `urlparse("mailto:foo@bar.com").scheme` returns `"mailto"` while `urlparse("tel:+1234").scheme` returns `"tel"`. If future Word documents contain `tel:` links, they'd be excluded. This might be intentional — just calling it out since the comment says "HTTP/mailto, etc." suggesting there may be other schemes to support.

3. **Minor: `link_text_parts` could use a generator expression** — Instead of building a list and then joining:
   ```python
   link_text = "".join(
       run.text for run_elem in child.findall(qn("w:r"))
       if (run := Run(run_elem, paragraph)).text
   ).strip()
   ```
   This is purely stylistic and the current code is perfectly clear.

Overall this is a clean, well-tested fix. The approach of handling `w:hyperlink` and `w:r` separately at the XML level is correct.

crazywoola (approved):

PR #33226, opened 2026-03-10
feat: normalize keyword index storage

kenwoodjw:
> [!IMPORTANT]
>

## Summary
Fixed #33162 

This PR replaces the legacy dataset-level keyword index blob with normalized dataset_keyword_entries.

## Changes
- add dataset_keyword_entries for incremental keyword index storage
- add storage_version and migrated_at to dataset_keyword_tables
- refactor Jieba keyword indexing/search to use normalized entries
- keep legacy blob/file storage only as a compatibility layer
- backfill database-backed legacy data in migration
- lazily migrate file-backed legacy data on first keyword access

## Why
This removes the full deserialize/full rewrite behavior of the old keyword index design and reduces memory and I/O overhead for large datasets.

## Testing
```py
uv run --project api pytest \
api/tests/unit_tests/core/rag/datasource/test_jieba_keyword.py \
api/tests/unit_tests/models/test_dataset_models.py
```

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


gambletan (commented):
This is a significant refactoring that moves keyword index storage from a denormalized blob/file format to a proper normalized `dataset_keyword_entries` table. The approach is solid — let me highlight specific observations.

**What's done well:**
- The lazy migration pattern (`_ensure_keyword_index_migrated`) with `storage_version` tracking means existing deployments don't need downtime. The double-check pattern (check version, acquire lock, re-check) is correct for avoiding redundant migrations.
- Moving keyword search to SQL (`_retrieve_ids_by_query`) pushes the `GROUP BY` + `COUNT` aggregation to the database, which should scale better than the in-memory `defaultdict(int)` approach for large datasets.
- The migration script handles both online (`data_source_type == "database"`) and file-backed datasets, only marking database-backed ones as migrated — file-backed ones get migrated at runtime.
- Tests cover the key paths: add, search, and legacy migration.

**Potential issues:**

1. **`keyword` column is `String(255)` but jieba can produce multi-word segments** — While most CJK keywords extracted by jieba are short, the 255-char limit could silently truncate keywords or cause insertion failures for edge-case content. Consider whether a `Text` column type or at least a length check before insertion would be safer.

2. **`document_ids_filter` is now applied via JOIN in `_retrieve_ids_by_query` but was previously applied in the segment query afterward** — In the old code, `document_ids_filter` filtered the final segment query. Now it filters during the keyword match query via a JOIN on `DocumentSegment`. This is semantically correct and more efficient, but note that the `search()` method no longer applies `document_ids_filter` to the segment query (the old code had `if document_ids_filter: segment_query_stmt = segment_query_stmt.where(...)`). This is fine since the filtering already happened, but worth verifying that no caller depends on the segment query filtering.

3. **Missing `db.session.commit()` in `_replace_segment_keywords`** — Actually, looking closer, `_replace_segment_keywords` does call `db.session.commit()` at the end, which is correct. But `delete_by_ids` also calls `db.session.commit()` separately after `_delete_keyword_entries`. This means there are two commits for the delete path. Consider whether the commit in `delete_by_ids` is redundant given `_delete_keyword_entries` doesn't commit on its own (which is the right design for composability).

4. **Migration: `_build_entry_rows` in the Alembic migration skips non-list segment IDs** — The check `isinstance(segment_ids, list)` is good for defensive parsing, but the legacy format used `set` objects serialized via `orjson` with `set_orjson_default` converting to lists. If any legacy blob was written with a different serializer that preserved sets differently, those entries would be silently dropped. Probably fine in practice since `orjson` always produces lists.

5. **No index on `DatasetKeywordEntry.keyword` alone** — The composite index `(dataset_id, keyword)` covers queries filtered by dataset, but if there's ever a need to query keywords across datasets, a standalone index would be needed. Current usage doesn't require this, so this is fine for now.

6. **`_collect_segment_keywords` accesses `keywords_list[index]` without bounds checking** — If `keywords_list` is shorter than `texts`, this would raise an `IndexError`. The old code had the same pattern (`keywords_list[i]`), so this isn't a regression, but it might be worth adding a guard.

Clean refactoring overall. The normalized storage pattern is a clear improvement for both query performance and maintainability.

PR #33227, opened 2026-03-10
fix: set daemon=True on auxiliary threads to allow clean shutdown

gambletan:
## Summary
- Set `daemon=True` on moderation and TTS publisher threads
- Non-daemon threads prevent Python process from exiting when main thread finishes or receives SIGTERM
- These auxiliary threads should not block shutdown

## Test plan
- [x] Moderation and TTS still function correctly
- [x] Process shuts down cleanly on SIGTERM


gambletan:
Makes sense — I'll refactor to store thread references so we can `join()` on shutdown. This also addresses the existing FIXME on line 76.

PR #33228, opened 2026-03-10
fix: add raise_for_status() to GitHub OAuth email API response

gambletan:
## Summary
- Add missing `raise_for_status()` to GitHub email API response in OAuth flow
- Without this check, non-200 responses (rate limits, auth failures) silently produce empty email
- Also fix `== True` to `is True` for Pythonic comparison

## Test plan
- [x] OAuth flow handles email API errors properly
- [x] Successful flow unchanged


gambletan:
Good point — switching to `.get('primary')` is more Pythonic and handles missing keys gracefully. Will update.

PR #33229, opened 2026-03-10
fix: parameterize document_ids_filter to prevent SQL injection in vector DB adapters

gambletan:
## Summary
- Fix SQL injection vulnerability in pgvector, tidb_vector, and milvus adapters where `document_ids_filter` values were directly interpolated into SQL/filter expressions via f-strings
- **pgvector**: Replace f-string interpolation with `%s` parameterized query placeholders in both `search_by_vector` and `search_by_full_text`
- **tidb_vector**: Replace f-string interpolation with SQLAlchemy `:named` bind parameters in `delete_by_ids`, `get_ids_by_metadata_field`, and `search_by_vector`
- **milvus**: Add input sanitization for document IDs used in filter expressions in `search_by_vector` and `search_by_full_text`

## Security Impact
If `document_ids_filter` contains crafted values, arbitrary SQL can be executed against the vector database. This is reachable when a user-controlled document ID propagates to a vector search call.

Follows precedent of PR #31946 (MyScale fix) and #28951 (OceanBase fix).

## Test plan
- [x] Parameterized queries produce same results for valid document IDs (UUIDs)
- [x] No SQL injection possible through document_ids_filter
- [ ] Verified pgvector search_by_vector and search_by_full_text with document_ids_filter
- [ ] Verified tidb_vector search_by_vector with document_ids_filter
- [ ] Verified milvus search_by_vector and search_by_full_text with document_ids_filter

### DIFF
--- a/api/core/rag/datasource/vdb/milvus/milvus_vector.py
+++ b/api/core/rag/datasource/vdb/milvus/milvus_vector.py
@@ -213,7 +213,7 @@ class MilvusVector(BaseVector):
     @staticmethod
     def _sanitize_document_id(doc_id: str) -> str:
         """Sanitize a document ID to prevent expression injection in Milvus filters."""
-        sanitized = re.sub(r'[^a-zA-Z0-9\-_]', '', str(doc_id))
+        sanitized = re.sub(r"[^a-zA-Z0-9\-_]", "", str(doc_id))
         if sanitized != doc_id:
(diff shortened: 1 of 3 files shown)


alvinttang:
Thanks for the review\! Good catch — the metadata key in `get_ids_by_metadata_field` is still being interpolated directly into the SQL string. I'll fix that to use a whitelist/validation approach or parameterize it as well to fully close the injection surface.

PR #33246, opened 2026-03-11
fix: prevent temp file leak in document extractor on API failure

alvinttang:
## Summary
- Fix temporary file leak in document extractor node when `partition_via_api()` raises an exception
- Affects DOC, PPT, PPTX, and EPUB extraction via Unstructured API

## Root Cause

The `_extract_text_from_doc`, `_extract_text_from_ppt`, `_extract_text_from_pptx`, and `_extract_text_from_epub` functions in `api/dify_graph/nodes/document_extractor/node.py` create temporary files with `delete=False` but only call `os.unlink()` on the success path.

If `partition_via_api()` raises an exception (e.g., network timeout, API error, malformed response), execution jumps to the outer `except` block and the `os.unlink()` call is skipped. The temporary file is never cleaned up, leaking disk space.

In long-running Dify deployments that process many documents, this can accumulate significant orphaned temp files in the system's temp directory.

## Fix

Move `os.unlink()` into a `finally` block so the temporary file is always removed regardless of whether the API call succeeds or fails.

**Before:**
```python
with tempfile.NamedTemporaryFile(suffix=".doc", delete=False) as temp_file:
    temp_file.write(file_content)
    temp_file.flush()
    with open(temp_file.name, "rb") as file:
        elements = partition_via_api(...)
    os.unlink(temp_file.name)  # skipped if partition_via_api raises
```

**After:**
```python
with tempfile.NamedTemporaryFile(suffix=".doc", delete=False) as temp_file:
    temp_file.write(file_content)
    temp_file.flush()
try:
    with open(temp_file.name, "rb") as file:
        elements = partition_via_api(...)
finally:
    os.unlink(temp_file.name)  # always runs
```

## Test plan
- [ ] Verify DOC/PPT/PPTX/EPUB extraction still works normally when Unstructured API succeeds
- [ ] Verify temp files are cleaned up when Unstructured API returns an error
- [ ] Check no temp file accumulation in `/tmp` after processing documents with API failures

### DIFF
--- a/api/dify_graph/nodes/document_extractor/node.py
+++ b/api/dify_graph/nodes/document_extractor/node.py
@@ -377,6 +377,7 @@ def _extract_text_from_doc(file_content: bytes, *, unstructured_api_config: Unst
         with tempfile.NamedTemporaryFile(suffix=".doc", delete=False) as temp_file:
             temp_file.write(file_content)
             temp_file.flush()
+        try:
             with open(temp_file.name, "rb") as file:
                 elements = partition_via_api(
                     file=file,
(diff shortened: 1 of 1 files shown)


crazywoola:
Hi @alvinttang, thank you for taking the time to contribute to Dify.

### Why this is being closed

### Next steps

PR #33266, opened 2026-03-11
fix: scope datasource binding updates to current tenant

samzong:
## Summary
- scope datasource binding PATCH lookups to the current tenant
- return not found for cross-tenant binding ids
- add a unit test to verify tenant scoping is included in the query

## Testing
- cd api && .venv/bin/pytest tests/unit_tests/controllers/console/datasets/test_data_source.py -q
- cd api && .venv/bin/pytest tests/unit_tests/controllers/console/datasets/test_data_source.py -q -k scoped_to_current_tenant

fixes #31839


alvinttang (commented):
Solid security fix — scoping the DataSourceOauthBinding lookup to tenant_id closes a horizontal privilege escalation where a user in tenant A could enable/disable bindings belonging to tenant B by guessing or brute-forcing UUIDs. The test correctly inspects the compiled WHERE clause to assert both columns are present, which is more meaningful than mocking at the query level. One nit: the test relies on string-matching the SQLAlchemy-compiled SQL (assert "tenant_id" in compiled_where) which could theoretically pass even if the filter were on a joined table; asserting filter_by args or using a stricter compiled output check would be more robust. Also worth verifying that current_account_with_tenant() is already tested to return the correct tenant in multi-tenant scenarios — if it ever falls back to a default tenant, the security guarantee weakens.

samzong:
@alvinttang Thanks for looking at this carefully.

On the test: fair point, I've tightened the assertion to check `data_source_oauth_bindings.tenant_id` instead of just `tenant_id`, so it won't pass on a coincidental column match from a JOIN.

On `current_account_with_tenant`: I dug into `libs/login.py` — it hard-fails if `current_tenant_id` is None (via assert), and `set_tenant_id` validates against `TenantAccountJoin` before setting anything. So there's no silent fallback path. Happy to be corrected if I'm reading it wrong.

crazywoola:
Thanks for your contribution, @samzong — really appreciate the effort here.

Closing this duplicate in favor of #34403, which targets the same issue (#31839). Normally we'd go FIFO and keep the earlier PR, but this one has been inactive for 30+ days, so the more recently active duplicate is the better candidate to land.

PR #33280, opened 2026-03-11
fix: do not block upsert for baidu vdb

letterbeezps:
## Summary

do not rebuild vector index after upsert finished, make sure we can upsert data immediately after table created. and config auto_build_index_policy so that index will be rebuilt automatically at the baidu vector db.

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [] I've updated the documentation accordingly.


alvinttang (commented):
Replacing the blocking post-upsert `rebuild_index` poll with `auto_build` is a meaningful improvement — the old approach blocked the entire upsert call path for potentially up to an hour. The new `AutoBuildRowCountIncrement` policy is a reasonable trade-off, but the defaults (500 rows or 5% increment) may need tuning for small collections: at low document counts the ratio threshold may never trigger, leaving the index in a degraded state. It is worth documenting this expectation explicitly. The `_wait_for_index_ready` refactor is good, but it is now only called at table creation time — if the initial rebuild at creation fails or times out, subsequent upserts will write to a table with no ready index and queries may silently return poor results. Consider logging a warning when `auto_build` triggers are not met.

WayneCao (approved):

crazywoola:
Please resolve the errors in the CI.

crazywoola (approved):

PR #33330, opened 2026-03-11
fix: prevent silent debug email loss when user_id is None or empty

karesansui-u:
## Summary

Fixes #33329

When `user_id` is `None`, the `or ""` pattern at call sites converts it to an empty string. `with_debug_recipient()` then treats `""` as falsy via `if not user_id:` and sets recipients to an empty list, **silently dropping all debug emails** with no error or warning.

### Changes

- **`with_debug_recipient()`**: Raise `ValueError` on empty `user_id` instead of silently creating empty recipients. This makes the contract explicit — callers must provide a valid user_id.
- **`apply_debug_email_recipient()`**: Guard against empty `user_id` before calling `with_debug_recipient()`. When `user_id` is empty, skip the debug transformation and return the original delivery method unchanged. Removed the `or ""` conversion.
- **`workflow_service.py`**: Remove unnecessary `account.id or ""` fallback that masked the empty value.

### Root cause

The `or ""` pattern at call sites and the `if not user_id:` check in `with_debug_recipient()` interact to create a silent failure:

```
user_id=None → or "" → "" → not "" is True → recipients=[] → email silently lost
```

### Affected call sites

| File | Line | Before | After |
|------|------|--------|-------|
| `entities.py` | 154 | `with_debug_recipient(user_id or "")` | `with_debug_recipient(user_id)` |
| `workflow_service.py` | 955 | `user_id=account.id or ""` | `user_id=account.id` |
| `human_input_node.py` | 183 | `user_id=dify_ctx.user_id` | (unchanged, already correct) |

### DIFF
--- a/api/dify_graph/nodes/human_input/entities.py
+++ b/api/dify_graph/nodes/human_input/entities.py
@@ -2,6 +2,7 @@
 Human Input node entities.
 """
 
+import logging
 import re
 import uuid
 from collections.abc import Mapping, Sequence
@@ -17,6 +18,8 @@ from dify_graph.runtime import VariablePool
 from dify_graph.variables.consts import SELECTORS_LENGTH
 
 from .enums import ButtonStyle, DeliveryMethodType, EmailRecipientType, FormInputType, PlaceholderType, TimeoutUnit
+
+logger = logging.getLogger(__name__)
 
 _OUTPUT_VARIABLE_PATTERN = re.compile(r"\{\{#\$output\.(?P<field_name>[a-zA-Z_][a-zA-Z0-9_]{0,29})#\}\}")
 
@@ -74,7 +77,7 @@ class EmailDeliveryConfig(BaseModel):
 
     def with_debug_recipient(self, user_id: str) -> "EmailDeliveryConfig":
         if not user_id:
-            raise ValueError("user_id must not be empty for debug email recipient")
+            raise AssertionError("user_id must not be empty for debug email recipient")
         debug_recipients = EmailRecipients(whole_workspace=False, items=[MemberRecipient(user_id=user_id)])
         return self.model_copy(update={"recipients": debug_recipients})
 
@@ -149,6 +152,7 @@ def apply_debug_email_recipient(
     if not method.config.debug_mode:
         return method
     if not user_id:
+        logger.warning("Debug email recipient skipped: user_id is None or empty")
         return method
     debug_config = method.config.with_debug_recipient(user_id)
     return method.model_copy(update={"config": debug_config})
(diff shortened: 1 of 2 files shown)


api/dify_graph/nodes/human_input/entities.py
```diff
@@ -74,8 +74,7 @@ class EmailDeliveryConfig(BaseModel):
 
     def with_debug_recipient(self, user_id: str) -> "EmailDeliveryConfig":
         if not user_id:
-            debug_recipients = EmailRecipients(whole_workspace=False, items=[])
-            return self.model_copy(update={"recipients": debug_recipients})
+            raise ValueError("user_id must not be empty for debug email recipient")
```
QuantumGhost: Raising a `ValueError` does not signal developers and users that this is a program bug, as `ValueError` exceptions are caught by the Flask global error handler and converted to 400 response without logging tracebacks.

I suggest raising `AssertionError` or raising a custom exception for this.

api/dify_graph/nodes/human_input/entities.py
```diff
@@ -149,7 +148,9 @@ def apply_debug_email_recipient(
         return method
     if not method.config.debug_mode:
         return method
-    debug_config = method.config.with_debug_recipient(user_id or "")
+    if not user_id:
```
QuantumGhost, line 154: I would suggest raising a `AssertionError` or custom exception here to signal users and developers that the precondition `user_id` is not `None` nor `""`  is not met when debug email recipients is applied.

alvinttang (commented):
The fix is correct — raising ValueError on empty user_id in with_debug_recipient prevents the silent creation of an empty recipient list that would cause debug emails to be silently dropped. The early-return guard in apply_debug_email_recipient is the right place to handle the None case, keeping with_debug_recipient a pure validator. One thing worth considering: the ValueError in with_debug_recipient is now technically unreachable since callers always guard with "if not user_id" before calling it — either document this as a defensive invariant or remove the raise and rely solely on the guard. Also, consider whether a warning/log should be emitted in the early-return path so operators can tell when debug email was silently skipped due to a missing user identity.

api/dify_graph/nodes/human_input/entities.py
karesansui-u: Good point — updated to AssertionError in f20e27c. This way it bypasses Flask's 400 handler and surfaces as a proper program bug with traceback.

api/dify_graph/nodes/human_input/entities.py
karesansui-u: Updated — added logger.warning() in the early-return path (f20e27c), and changed with_debug_recipient() to raise AssertionError as a defensive invariant. The caller guard + AssertionError safety net gives two layers: graceful skip with logging at the call site, loud failure if someone calls with_debug_recipient() directly without checking.

karesansui-u:
Closing in favor of #33373 which was merged with a simpler approach (`if user_id is None:` instead of raising). Thanks for the review feedback @QuantumGhost — learned a lot about the AssertionError vs ValueError distinction in the Flask context.

Issue #27976, opened 2025-11-07
Non-Image Type Icons Incorrectly Replaced with Default Emoji in Workflow DSL Export


Xiu-Lan:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.9.2

### Cloud or Self Hosted

Cloud

### Steps to reproduce

1. Edit application information in a workflow app
2. Upload a custom avatar (image type) or select an emoji avatar
3. Save application settings
4. Export the workflow DSL file (via "Export" feature)
5. Check the `app.icon` field in the exported YAML file

### ✔️ Expected Behavior

The exported DSL file should contain the actual configured avatar information:
- If it's an image type (`icon_type: "image"`), should export the image URL
- If it's an emoji type (`icon_type: "emoji"`), should export the actual emoji character
- If it's another custom avatar type, should export the actual avatar value

### ❌ Actual Behavior

- When `icon_type` is `"image"`, the image URL is correctly exported
- When `icon_type` is `"emoji"` or any non-`"image"` type, the exported DSL file shows the default emoji `"🤖"` instead of the actually configured avatar

<img width="1079" height="400" alt="Image" src="https://github.com/user-attachments/assets/8af40b6b-420a-4acb-a599-c54ee702ad8c" />

Issue #26301, opened 2025-09-26
升级到1.9.0 后，创建的工作流 自定义图标失效


BannyLon:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.9.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

<img width="1834" height="1408" alt="Image" src="https://github.com/user-attachments/assets/3429e872-02ba-4231-b18b-7e312d380aa4" />

如图，我创建的工作流，自定义图标时，没办法显示。我新升级的最新版本 1.9.0，之前有一个版本也出现过这样的问题。

### ✔️ Expected Behavior

<img width="1834" height="1408" alt="Image" src="https://github.com/user-attachments/assets/3429e872-02ba-4231-b18b-7e312d380aa4" />

如图，我创建的工作流，自定义图标时，没办法显示。我新升级的最新版本 1.9.0，之前有一个版本也出现过这样的问题。

### ❌ Actual Behavior

<img width="1834" height="1408" alt="Image" src="https://github.com/user-attachments/assets/3429e872-02ba-4231-b18b-7e312d380aa4" />

如图，我创建的工作流，自定义图标时，没办法显示。我新升级的最新版本 1.9.0，之前有一个版本也出现过这样的问题。

Issue #12732, opened 2025-01-14
Using a custom image for the copied app icon does not work as expected


kyatai:
### Self Checks

- [X] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [X] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [X] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [X] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [X] Please do not modify this template :) and fill in all the required fields.

### Dify version

0.15.1

### Cloud or Self Hosted

Cloud, Self Hosted (Source)

### Steps to reproduce

- step 1
select an app in workroom and choice a pic as copy's icon
click copy button
![21736847628_ pic](https://github.com/user-attachments/assets/b626f673-cc33-4909-bcd6-53301bb3124d)

- step 2
save it.
![31736847628_ pic](https://github.com/user-attachments/assets/b12313f0-1c53-44aa-86b5-25f6ab67529b)

custom image for the copied app icon does not work as expected

### ✔️ Expected Behavior

custom icon

### ❌ Actual Behavior

default emoji

time1023:
Version 1.1.2 still has this problem

PR #33424, opened 2026-03-13
fix: preserve custom icons in exported DSL

majiayu000:
## Summary
Fixes #33409

This PR fixes a bug where custom emoji icons were being replaced with the default robot emoji (🤖) when exporting workflow DSL, and the `icon_type` field was not being included in the export.

## Root Cause
The `export_dsl` method in `api/services/app_dsl_service.py` had inverted conditional logic:
```python
"icon": app_model.icon if app_model.icon_type == "image" else "🤖",
```
This meant that only image icons were preserved, while emoji and link icons were replaced with 🤖.

## Changes
- Remove conditional logic that replaced non-image icons with 🤖
- Export actual icon value for all icon types (emoji, image, link)
- Add `icon_type` field to exported DSL for proper import
- Convert `icon_type` enum to string value for YAML serialization
- Add comprehensive test to verify icon and icon_type preservation

## Test Plan
- Added `test_export_dsl_preserves_icon_and_icon_type` to verify both emoji and image icons are preserved
- All existing 53 tests in `test_app_dsl_service.py` pass
- Tested export → import cycle to ensure icons are correctly restored

## Related Issues
- #27976 - Non-Image Type Icons Incorrectly Replaced with Default Emoji
- #26301 - Custom icons lost after upgrade to 1.9.0
- #12732 - Custom image icons not working in copied apps

### DIFF
--- a/api/services/app_dsl_service.py
+++ b/api/services/app_dsl_service.py
@@ -556,8 +556,11 @@ class AppDslService:
             "app": {
                 "name": app_model.name,
                 "mode": app_model.mode.value if isinstance(app_model.mode, AppMode) else app_model.mode,
-                "icon": app_model.icon if app_model.icon_type == "image" else "🤖",
-                "icon_background": "#FFEAD5" if app_model.icon_type == "image" else app_model.icon_background,
+                "icon": app_model.icon,
+                "icon_type": (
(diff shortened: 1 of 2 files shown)


fatelei:
icon_type = app_model.icon_type.value if isinstance(app_model.icon_type, IconType) else app_model.icon_type, and has conflict

fatelei (approved):

PR #33440, opened 2026-03-14
fix(web): render Stability AI agent images from normalized thought files

Haohao-end:
## Summary
- normalize agent thought file references into `FileEntity` before they reach chat rendering
- stop re-processing already-normalized agent files in `AgentContent`
- add regression coverage for history payload normalization, direct agent file rendering, and `VisionFile` compatibility

## Root cause
Agent tool image outputs were already converted to frontend `FileEntity` objects during streaming, but `AgentContent` treated `agent_thoughts[].message_files` as raw backend file payloads and converted them again. That second conversion dropped `type`, which later caused `file.type.split('/')` to throw in the chat attachment card.

In addition, `addFileInfos` also needed to accept the `VisionFile` path used by log/history rendering so that all agent file inputs are normalized consistently before rendering.

## Testing
- `pnpm run type-check:tsgo`
- `pnpm test app/components/base/chat/chat/answer/__tests__/agent-content.spec.tsx`
- `pnpm test app/components/tools/utils/__tests__/index.spec.ts`
- `pnpm test __tests__/tools/tool-data-processing.test.ts`
- `pnpm exec eslint app/components/tools/utils/index.ts app/components/base/chat/chat/answer/agent-content.tsx app/components/tools/utils/__tests__/index.spec.ts app/components/base/chat/chat/answer/__tests__/agent-content.spec.tsx __tests__/tools/tool-data-processing.test.ts --max-warnings=0 --pass-on-unpruned-suppressions`

fixes #33122

### DIFF
--- a/web/__tests__/tools/tool-data-processing.test.ts
+++ b/web/__tests__/tools/tool-data-processing.test.ts
@@ -19,6 +19,7 @@ import {
   toType,
   triggerEventParametersToFormSchemas,
 } from '@/app/components/tools/utils/to-form-schema'
+import { TransferMethod } from '@/types/app'
 
 describe('Tool Data Processing Pipeline Integration', () => {
   describe('End-to-end: API schema → form schema → form value', () => {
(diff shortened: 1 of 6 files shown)


Haohao-end:
Added focused coverage for the new VisionFile normalization helpers and the `upload_file_id` fallback path, and also simplified `getVisionFileSupportType` as suggested. This keeps the main fix unchanged while addressing the diff coverage failure.

crazywoola:
Thanks for your contribution, @Haohao-end — really appreciate the effort here.

Closing this in favor of #33150, which was opened earlier and targets the same issue (#33122). To keep duplicate cleanup fair, we're going FIFO on dupes — the older PR gets the merge consideration first.

Issue #27998, opened 2025-11-08
refactor(web): In TypeScript, the use of enums should be avoided.


coding-ice:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

##  Runtime Code Bloat
Enums generate additional runtime code, increasing bundle size:

```ts
enum Status {
  Active = 'active',
  Inactive = 'inactive'
}

// Compiled JavaScript
var Status;
(function (Status) {
    Status["Active"] = "active";
    Status["Inactive"] = "inactive";
})(Status || (Status = {}));
```

## Recommended Alternatives

```ts
const Status = {
  Active: 'active',
  Inactive: 'inactive',
} as const;

type Status = typeof Status[keyof typeof Status]; // 'active' | 'inactive'
```

hyoban:
I do not like enum either. But considering this is a big change,  I recommend enabling a lint rule for it first.

- https://www.typescriptlang.org/docs/handbook/enums.html
- https://www.totaltypescript.com/why-i-dont-like-typescript-enums
- https://github.com/antfu/eslint-config/pull/760
- https://www.totaltypescript.com/erasable-syntax-only

jubinsoni:
adding enum restrict rule is resulting in 3000+ warnings which is not letting the pr pass
@hyoban how do we approach this as we cannot raise a single pr with 3000+ changes ?

my opinion is that possible we can break this into sub-tasks and move module by module or directory by directory

Let me know your thoughts
 
<img width="2118" height="558" alt="Image" src="https://github.com/user-attachments/assets/24c5d433-40a4-45c9-b3c6-5484e24f75a1" />

mahmoodhamdi:
[...]

PR #33462, opened 2026-03-15
refactor(web): replace enums with as const in models and service files

mahmoodhamdi:
## Summary
- Converted 5 TypeScript enums to `as const` objects: `PromptMode`, `PromptRole`, `SubjectType`, `AccessMode`, `AppSourceType`
- Primary files changed: `models/debug.ts`, `models/access-control.ts`, `service/share.ts`
- Also converted `PromptRole` in `app/components/workflow/types.ts` to maintain cross-module type compatibility (both modules define `PromptRole` with the same values; keeping one as an enum and the other as a string union caused assignment errors)
- Added explicit `useState<PromptMode>` type parameter in `app/components/app/configuration/index.tsx` so TypeScript correctly widens the inferred state type beyond the literal `'simple'`
- No usage-site changes needed beyond the above — all existing references work identically

## Why
TypeScript enums generate runtime IIFE wrappers. The `as const` pattern provides the same API with zero runtime overhead.

This is the second incremental batch for #27998.

## Checklist
- [x] I understand that this PR may be closed if there was no previous discussion or issues (ref: #27998)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change
- [x] I've updated the documentation accordingly
- [x] I ran `cd web && npx lint-staged` to appease the lint gods
- [x] TypeScript type check passes

Part of #27998

### DIFF
--- a/web/app/components/app/configuration/index.tsx
+++ b/web/app/components/app/configuration/index.tsx
@@ -392,7 +392,7 @@ const Configuration: FC = () => {
     }
   }, [textGenerationModelList, hasFetchedDetail, modelModeType, currModel, modelConfig])
 
-  const [promptMode, doSetPromptMode] = useState(PromptMode.simple)
+  const [promptMode, doSetPromptMode] = useState<PromptMode>(PromptMode.simple)
   const isAdvancedMode = promptMode === PromptMode.advanced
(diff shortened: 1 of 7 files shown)


mahmoodhamdi:

PR #33463, opened 2026-03-15
refactor(api): remove unnecessary type: ignore comments

mahmoodhamdi:
## Summary

- Removed 8 `# type: ignore` comments across 6 files by fixing the underlying type issues
- Added proper type annotations where needed instead of suppressing type errors

### `api/models/dataset.py`
Fixed `DocumentSegmentSummary.error` field type from `Mapped[str]` to `Mapped[str | None]` to correctly reflect the `nullable=True` column constraint. This is the root fix that enables the `summary_index_service.py` changes below.

### `api/services/summary_index_service.py`
Removed 4 `# type: ignore[assignment]` comments on `error = None` assignments. These are now valid because the model field is correctly typed as `str | None`.

### `api/schedule/mail_clean_document_notify_task.py`
Added explicit type annotation `dict[str, list[str]]` to `dataset_auto_dataset_map` instead of relying on a bare `{}` with `# type: ignore`.

### `api/services/workflow/queue_dispatcher.py`
Annotated `PLAN_DISPATCHER_MAP` as `dict[str, type[BaseQueueDispatcher]]`. With this annotation, `PLAN_DISPATCHER_MAP.get(...)` returns `type[BaseQueueDispatcher]`, so calling `dispatcher_class()` correctly produces a `BaseQueueDispatcher` instance without needing `# type: ignore`.

### `api/services/trigger/trigger_request_service.py`
Removed `# type: ignore` from `TypeAdapter(...).dump_json(payload)` call. `dump_json` is typed to return `bytes` and `storage.save` accepts `bytes`, so no suppression is needed.

### `api/extensions/ext_logging.py`
Removed `# type: ignore[attr-defined]` from `handler.formatter.converter = time_converter`. `logging.Formatter.converter` is declared in the Python typeshed stubs, so the assignment is valid after the `if handler.formatter:` null check.

## Why
Reducing the number of `# type: ignore` comments improves type safety and makes the codebase easier to maintain. Each removed comment here represents a type issue that has been properly resolved rather than suppressed.

## Checklist
- [x] I understand that this PR may be closed if there was no previous discussion or issues (ref: #24494)
- [x] I've verified each change is safe and does not alter runtime behavior
- [x] I ran ruff linter (pre-commit hook passed)

Part of #24494


mahmoodhamdi:
Applied the `defaultdict(list)` suggestion — much cleaner. The `if key not in dict` check and manual `[]` initialization are no longer needed. Thanks!

mahmoodhamdi:

PR #33461, opened 2026-03-15
refactor(web): replace TypeScript enums with as const objects

mahmoodhamdi:
## Summary
- Replaced 6 TypeScript `enum` declarations with `as const` objects + companion type aliases
- Files changed: `types/common.ts`, `types/feature.ts`, `app/components/app/annotation/type.ts`
- Converted enums: `FlowType`, `SSOProtocol`, `LicenseStatus`, `InstallationScope`, `AnnotationEnableStatus`, `JobStatus`

## Why
TypeScript enums generate additional runtime code (IIFE wrappers), increasing bundle size unnecessarily. The `as const` pattern provides the same developer experience (dot-access for values, type-level usage for annotations) with zero runtime overhead — the values are inlined by the compiler.

```ts
// Before (generates runtime IIFE)
export enum FlowType {
  appFlow = 'appFlow',
  ragPipeline = 'ragPipeline',
}

// After (zero runtime overhead)
export const FlowType = {
  appFlow: 'appFlow',
  ragPipeline: 'ragPipeline',
} as const
export type FlowType = typeof FlowType[keyof typeof FlowType]
```

## Scope
This is an incremental step toward the full enum migration described in #27998. I intentionally kept this PR small and focused on type definition files with limited usage scope to make review straightforward. No usage-site changes were needed — all existing `FlowType.appFlow`, `SSOProtocol.SAML`, etc. references continue to work identically.

## Checklist
- [x] I understand that this PR may be closed if there was no previous discussion or issues (ref: #27998)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change
- [x] I've updated the documentation accordingly
- [x] I ran `cd web && npx lint-staged` to appease the lint gods
- [x] TypeScript type check passes (`pnpm tsc --noEmit`)
- [x] ESLint passes on all changed files

Closes partially #27998

PR #33465, opened 2026-03-15
refactor(web): replace enums with as const objects (batch 3)

mahmoodhamdi:
## Summary

Continues the enum-to-as-const refactoring tracked in issue #27998 (batch 3, following PRs #33461 and #33462).

Converts 6 enums across `models/` and `types/` to `as const` objects with companion type aliases:

| Enum | File |
|------|------|
| `DSLImportMode` | `web/models/app.ts` |
| `DSLImportStatus` | `web/models/app.ts` |
| `WorkflowRunTriggeredFrom` | `web/models/log.ts` |
| `ProviderName` | `web/models/common.ts` |
| `DataSourceCategory` | `web/models/common.ts` |
| `AgentStrategy` | `web/types/app.ts` |

### Conversion pattern applied

```typescript
// Before
export enum Foo {
  Bar = 'bar',
}

// After
export const Foo = {
  Bar: 'bar',
} as const

// eslint-disable-next-line ts/no-redeclare -- value-type pair
export type Foo = typeof Foo[keyof typeof Foo]
```

All call sites are unaffected — member access (`Foo.Bar`), type annotations (`Foo`), and mapped types (`[K in Foo]`) continue to work identically.

## Test plan

- [x] ESLint passes with zero warnings (`npx eslint` on changed files)
- [x] TypeScript type-check passes via `tsgo --noEmit` (pre-commit hook)
- [x] Unit tests pass (pre-commit hook)
- [x] `ProviderName` used in mapped types (`[Name in ProviderName]`) verified to work correctly with union type
- [x] `WorkflowRunTriggeredFrom` used as `Record<WorkflowRunTriggeredFrom, string>` key — verified compatible

### DIFF
--- a/web/eslint-suppressions.json
+++ b/web/eslint-suppressions.json
@@ -9897,14 +9897,9 @@
       "count": 2
     }
   },
-  "models/app.ts": {
-    "erasable-syntax-only/enums": {
-      "count": 2
-    }
-  },
   "models/common.ts": {
     "erasable-syntax-only/enums": {
-      "count": 3
+      "count": 1
     },
     "ts/no-explicit-any": {
       "count": 3
@@ -9927,9 +9922,6 @@
     }
   },
   "models/log.ts": {
-    "erasable-syntax-only/enums": {
-      "count": 1
-    },
     "ts/no-explicit-any": {
       "count": 7
     }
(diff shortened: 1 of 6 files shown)


mahmoodhamdi:

PR #33469, opened 2026-03-15
fix(web): display file icons in workflow response node output panel

mahmoodhamdi:
## Summary

Fixes #33419 — files returned by the response/answer/end node in the workflow result panel do not show file icons; only a raw UUID link is displayed.

### Root Cause

`output-panel.tsx` (used in the RESULT tab for historical workflow runs) had two bugs in its `fileList` memo:

1. **Early return on multi-key outputs** — `if (Object.keys(outputs).length > 1) return fileList` caused the file list to always be empty. The Answer/End node always emits at least two output keys (e.g. `"answer"` + `"files"`), so the guard fired unconditionally and no files were ever collected.

2. **Wrong utility + wrong data shape** — files were accumulated manually and passed to `getProcessedFilesFromResponse(FileResponse[])`, which returns a flat `FileEntity[]`. The `FileList` component in the rest of the run panel (e.g. `result-text.tsx`) expects a `{varName: string, list: FileEntity[]}[]` grouped structure from `getFilesInLogs`.

### Changes

**`web/app/components/workflow/run/output-panel.tsx`**
- Replace the manual file-collection loop and `getProcessedFilesFromResponse` with `getFilesInLogs(outputs)`, consistent with `result-text.tsx` and the rest of the run panel.
- Add `isFileOnlyOutput` memo to suppress the raw JSON code editor when every output value is a file object (previously raw JSON with UUIDs was always rendered alongside/instead of file icons).
- Render files as grouped rows with a variable-name label above each `FileList`, matching the established pattern.

**`web/app/components/workflow/panel/workflow-preview.tsx`**
- Add a branch in the `workflowRunningData` effect to auto-switch to the RESULT tab when a run finishes with file outputs but no text result (previously it fell through to DETAIL unconditionally).

## Test plan

- [ ] Run a workflow with an Answer node that returns file outputs; confirm files are shown with proper icons and filenames in the RESULT tab.
- [ ] Run a workflow with an End node that returns only files; confirm `isFileOnlyOutput` suppresses the JSON code editor and only file icons are shown.
- [ ] Run a workflow with a mix of text and file outputs; confirm text renders as Markdown and files render below it.
- [ ] Run a workflow with only text outputs; confirm existing text rendering is unchanged.
- [ ] Verify that after a file-only workflow run completes, the panel auto-switches to the RESULT tab.

### DIFF
--- a/web/app/components/workflow/panel/workflow-preview.tsx
+++ b/web/app/components/workflow/panel/workflow-preview.tsx
@@ -61,6 +61,8 @@ const WorkflowPreview = () => {
 
     if ((status === WorkflowRunningStatus.Succeeded || status === WorkflowRunningStatus.Failed) && !workflowRunningData.resultText && !workflowRunningData.result.files?.length)
       switchTab('DETAIL')
+    if ((status === WorkflowRunningStatus.Succeeded || status === WorkflowRunningStatus.Failed) && !workflowRunningData.resultText && !!workflowRunningData.result.files?.length)
+      switchTab('RESULT')
     if (status === WorkflowRunningStatus.Paused)
       switchTab('RESULT')
   }, [workflowRunningData])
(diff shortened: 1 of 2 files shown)


mahmoodhamdi:

PR #33545, opened 2026-03-16
fix: make AceDataCloud workflows visible on Explore page

Germey:
## Summary

- **Switch Explore page to DB mode**: Set `HOSTED_FETCH_APP_TEMPLATES_MODE=db` in production configmap so the Explore page reads from our `RecommendedApp` table instead of fetching from `tmpl.dify.ai` (which only shows Dify official templates)
- **Fix missing Site records**: The `DatabaseRecommendAppRetrieval` skips any app without a `Site` record (`if not site: continue`). Added `_ensure_site_exists()` that creates Site records alongside RecommendedApp records
- **Decouple Explore from user login**: Previously, Explore registration only happened when a new user logged in via OAuth (and was gated by a one-time Redis key). Now:
  - `_register_explore_apps()` is fully idempotent via DB checks (safe to call repeatedly)
  - New Celery beat task `sync_acedatacloud_explore_task` runs every 30 min to ensure all workflow YAMLs in `/workflows/*.yml` are registered in Explore, independent of user login
  - Controlled by `ENABLE_ACEDATACLOUD_EXPLORE_SYNC_TASK` flag (default: `false`, enabled in production configmap)

## Root Causes

| Problem | Fix |
|---------|-----|
| `HOSTED_FETCH_APP_TEMPLATES_MODE` defaulted to `"remote"` → DB records ignored | Set to `"db"` in configmap |
| `_register_explore_apps` created `RecommendedApp` but not `Site` → filtered out by DB retrieval | Added `_ensure_site_exists()` |
| Explore registration gated by Redis key + coupled to new user OAuth login | Removed Redis gate, added periodic Celery beat task |

## Test plan

- [ ] Deploy and verify `https://dify.acedata.cloud/explore/apps` shows all 13 AceDataCloud workflow templates
- [ ] Verify new users still get workflows imported into their workspace on first login
- [ ] Verify the Celery beat task runs every 30 min and is idempotent (no duplicate records)
- [ ] Verify adding a new `.yml` file to `/workflows/` directory causes it to appear on Explore within 30 min


Germey:
Opened against wrong repo — this should target AceDataCloud/Dify.

PR #33580, opened 2026-03-17
refactor(api): continue decoupling dify_graph from API concerns

### LINKED ISSUE #33433: [Refactor/Chore] Separate model runtime abstractions from plugin-specific execution concerns
## Self Checks

## Description

The current backend model and provider execution flow appears to mix two responsibilities that would be better kept separate:

1. the generic runtime-facing responsibilities that workflow, service, and graph code depend on; and
2. the plugin-specific responsibilities required to discover providers, resolve schemas, validate credentials, and execute model operations.

This creates a broad surface area where transport-specific assumptions influence otherwise generic model access paths. It also makes tenant-scoped model composition feel inconsistent across backend flows that need provider access, model invocation, or schema lookup.

This proposal is to introduce a clearer runtime boundary for model-related operations and align backend callers around a single tenant-aware access path. The intended outcome is a cleaner separation of concerns, more predictable composition across services and workflows, and easier long-term maintenance of model-capable features.

## Motivation

The current shape makes it harder to reason about where provider/runtime concerns should live, especially when the same model operations are used by multiple backend subsystems. That kind of coupling raises the maintenance cost of future changes and increases the chance that behavior diverges across workflows, provider management, retrieval, audio, moderation, and related model-enabled paths.

There is also a practical need for model execution paths that do not always have a meaningful end-user identity at the call site. Treating user context as optional where appropriate is a cleaner contract than forcing every caller to behave as if a user is always present.

From a maintainability perspective, a stable runtime-facing contract would improve testability, reduce duplicate composition logic, and keep plugin-specific routing from leaking into generic model abstractions.

## Additional Context

This request is primarily about architecture and boundary cleanup, not about changing the user-facing product behavior.

Useful success criteria:

- provider discovery, credential validation, schema access, and model execution can be reasoned about through one consistent backend boundary;
- tenant-aware access stays explicit and consistent across the major model-enabled backend flows;
- model operations continue to behave consistently across LLM, embeddings, rerank, speech, moderation, and related capabilities;
- incidental environment or compose-file adjustments, if any, are treated as separate follow-up work unless they are directly required by this refactor.

laipz8200:
## Summary

Fixes #33433

This branch advances the ongoing effort to decouple `dify_graph` from the rest of the API.

The goals of these improvements are to:
- strengthen `dify_graph` as an independent runtime boundary that the rest of the API can depend on without reverse coupling;
- move workflow, model runtime, and file-related responsibilities toward graph-owned abstractions instead of API-specific concepts;
- reduce implicit dependencies on user, conversation, and persistence concepts so the graph layer can evolve more cleanly as its own module.

## Screenshots

| Before | After |
|--------|-------|
| N/A | N/A |

## Checklist


api/services/workflow_service.py
```diff
@@ -1468,38 +1474,48 @@ def _setup_variable_pool(
...
     if is_start_node_type(node_type):
-        system_variable = SystemVariable(
-            user_id=user_id,
-            app_id=workflow.app_id,
-            timestamp=int(naive_utc_now().timestamp()),
-            workflow_id=workflow.id,
-            files=files or [],
-            workflow_execution_id=str(uuid.uuid4()),
-        )
+        system_variable_values: dict[str, Any] = {
```
fatelei, line 1532: why not use SystemVariable ?

laipz8200: To ensure the general versatility of the variable pool, we removed the system variables from it and moved that concept back into the dify.

As a result, we are now only left with system variable key.

41tair (approved):

PR #33642, opened 2026-03-18
fix(api): add try-except to get_tool_icon to prevent API error when plugin is uninstalled (#33635)

### LINKED ISSUE #33635: Missing try-except in WorkflowNodeExecution.extras causes API error when plugin is uninstalled
## Self Checks

- [x] This is only for bug reports, not feature requests.

## Dify version

1.x (after plugin architecture introduction)

## Cloud or Self-Hosted

Self-Hosted

## Steps to reproduce

1. Install a plugin and create a workflow that uses it as a tool node
2. Execute the workflow (execution history is saved to DB)
3. Uninstall the plugin
4. Open the workflow execution history in the console (GET `/console/api/apps/<app_id>/workflow-runs/<run_id>/node-executions`)
5. API returns an error, making the entire execution history inaccessible

This also reproduces when the Plugin Daemon is temporarily down.

## Expected Behavior

The node-executions API should return the execution history with a fallback icon (or no icon) instead of failing.

`workflow_response_converter.py` already handles this exact scenario correctly with try-except and an explicit comment:

```python
except Exception:
    # metadata fetch may fail, for example, the plugin daemon is down or plugin is uninstalled.
    logger.warning("failed to fetch icon for %s", event.provider_id)
```

## Actual Behavior

`WorkflowNodeExecution.extras` property in `models/workflow.py` calls `ToolManager.get_tool_icon()` **without try-except**. When the plugin is not found, `ToolProviderNotFoundError` propagates to the API response, making the entire execution history inaccessible.

## Root Cause

Two code paths call `ToolManager.get_tool_icon()` for the same purpose (fetching tool node icons), but only one has error handling:

| Call site | File | try-except | Behavior on plugin not found |
|---|---|---|---|
| SSE streaming (runtime) | `workflow_response_converter.py` L466-490 | Yes | Logs warning, continues |
| REST API (history) | `models/workflow.py` L916-942 | **No** | Error propagates, API fails |

Additionally, `agent_service.py` (L112, L120) calls `get_tool_icon()` without try-except, which could cause the same issue for agent execution history.

## Suggested Fix

Add try-except to `models/workflow.py` `extras` property, matching the pattern already used in `workflow_response_converter.py`:

```python
try:
    extras["icon"] = ToolManager.get_tool_icon(
        tenant_id=self.tenant_id,
        provider_type=tool_info["provider_type"],
        provider_id=tool_info["provider_id"],
    )
except Exception:
    logger.warning("failed to fetch icon for %s", tool_info.get("provider_id"))
```

The same fix should be applied to `agent_service.py`.

shanliuling:
## What is the problem?

When a plugin is uninstalled or the Plugin Daemon is temporarily unavailable, the ToolManager.get_tool_icon() method may fail and throw an exception. Currently, there is a lack of error handling in certain code paths, leading to API crashes (e.g., when accessing workflow execution history via GET /console/api/apps/<app_id>/workflow-runs/<run_id>/node-executions).

## How did I fix it?

I added 	ry-except blocks to catch potential exceptions when fetching tool icons in the following locations:
1.  **api/models/workflow.py**: Added error handling to the WorkflowNodeExecution.extras property.
2.  **api/services/agent_service.py**: Added error handling and logging to AgentService.get_agent_logs().

When an error occurs, the system now:
*   Logs a warning message with the relevant provider ID.
*   Falls back to an empty string for the icon, ensuring the API response remains valid and accessible.

This pattern aligns with the existing robust implementation found in workflow_response_converter.py.

## How was this tested?

- Code analysis confirms that the proposed changes address the missing error handling reported in the issue.
- Verified that the fallback logic correctly assigns an empty string to the icon.

Fixes #33635

### DIFF
--- a/api/models/workflow.py
+++ b/api/models/workflow.py
@@ -928,15 +928,11 @@ class WorkflowNodeExecutionModel(Base):  # This model is expected to have `offlo
         if execution_metadata:
             if self.node_type == BuiltinNodeTypes.TOOL and "tool_info" in execution_metadata:
                 tool_info: dict[str, Any] = execution_metadata["tool_info"]
-                try:
-                    extras["icon"] = ToolManager.get_tool_icon(
-                        tenant_id=self.tenant_id,
-                        provider_type=tool_info["provider_type"],
-                        provider_id=tool_info["provider_id"],
-                    )
-                except Exception:
-                    logger.warning("failed to fetch icon for %s", tool_info.get("provider_id"))
-                    extras["icon"] = ""
+                extras["icon"] = ToolManager.get_tool_icon(
+                    tenant_id=self.tenant_id,
+                    provider_type=tool_info["provider_type"],
+                    provider_id=tool_info["provider_id"],
+                )
             elif self.node_type == BuiltinNodeTypes.DATASOURCE and "datasource_info" in execution_metadata:
                 datasource_info = execution_metadata["datasource_info"]
                 extras["icon"] = datasource_info.get("icon")
--- a/api/models/workflow.py
+++ b/api/models/workflow.py
@@ -928,11 +928,15 @@ class WorkflowNodeExecutionModel(Base):  # This model is expected to have `offlo
         if execution_metadata:
             if self.node_type == BuiltinNodeTypes.TOOL and "tool_info" in execution_metadata:
                 tool_info: dict[str, Any] = execution_metadata["tool_info"]
-                extras["icon"] = ToolManager.get_tool_icon(
-                    tenant_id=self.tenant_id,
-                    provider_type=tool_info["provider_type"],
-                    provider_id=tool_info["provider_id"],
-                )
+                try:
+                    extras["icon"] = ToolManager.get_tool_icon(
+                        tenant_id=self.tenant_id,
+                        provider_type=tool_info["provider_type"],
+                        provider_id=tool_info["provider_id"],
+                    )
+                except Exception:
+                    logger.warning("failed to fetch icon for %s", tool_info.get("provider_id"))
+                    extras["icon"] = ""
             elif self.node_type == BuiltinNodeTypes.DATASOURCE and "datasource_info" in execution_metadata:
                 datasource_info = execution_metadata["datasource_info"]
                 extras["icon"] = datasource_info.get("icon")
--- a/api/services/agent_service.py
+++ b/api/services/agent_service.py
@@ -1,3 +1,4 @@
+import logging
 import threading
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


api/models/workflow.py
```diff
@@ -928,11 +928,15 @@ def extras(self) -> dict[str, Any]:
...
-                    provider_type=tool_info["provider_type"],
-                    provider_id=tool_info["provider_id"],
-                )
+                try:
+                    extras["icon"] = ToolManager.get_tool_icon(
+                        tenant_id=self.tenant_id,
+                        provider_type=tool_info["provider_type"],
+                        provider_id=tool_info["provider_id"],
+                    )
+                except Exception:
```
QuantumGhost: Is it possible to catch a more specific exception type here instead of catch the `Exception` type?

api/models/workflow.py
```diff
@@ -928,11 +928,15 @@ def extras(self) -> dict[str, Any]:
...
+                    logger.warning("failed to fetch icon for %s", tool_info.get("provider_id"))
```
QuantumGhost: If catching `Exception` is absolutely necessary, I suggest add `exc_info=True` to log the traceback of the exception above and reduce the difficulty of troubleshooting.

api/services/agent_service.py
```diff
@@ -109,19 +112,28 @@ def find_agent_tool(tool_name: str):
...
-                        provider_type=tool_config.get("tool_provider_type", ""),
-                        provider_id=tool_config.get("tool_provider", ""),
-                    )
+                    try:
+                        tool_icon = ToolManager.get_tool_icon(
+                            tenant_id=app_model.tenant_id,
+                            provider_type=tool_config.get("tool_provider_type", ""),
+                            provider_id=tool_config.get("tool_provider", ""),
+                        )
+                    except Exception:
```
QuantumGhost: ditto

api/services/agent_service.py
```diff
@@ -109,19 +112,28 @@ def find_agent_tool(tool_name: str):
...
                     if not tool_icon:
                         tool_entity = find_agent_tool(tool_name)
                         if tool_entity:
-                            tool_icon = ToolManager.get_tool_icon(
-                                tenant_id=app_model.tenant_id,
-                                provider_type=tool_entity.provider_type,
-                                provider_id=tool_entity.provider_id,
-                            )
+                            try:
+                                tool_icon = ToolManager.get_tool_icon(
```
QuantumGhost: ditto

QuantumGhost (changes requested):
Besides, the logic is the same for the three invocation of `ToolManager.get_tool_icon`.

I suggest handle it inside the `ToolManager.get_tool_icon`, or extract a private utility for the logic.

shanliuling:
Hi, I've moved the try-except inside the get_tool_icon function. Now it handles exceptions internally and returns an empty string on failure. The calling code is now clean. PTAL, thanks!

PR #33678, opened 2026-03-18
fix: pass default root to OpenDAL Operator for fs scheme

RickDamon:
When using OpenDAL with fs scheme and no OPENDAL_FS_ROOT env var set, the default root value 'storage' was not written back to kwargs, causing Operator initialization to fail with 'ConfigInvalid: root is not specified'.

This fix ensures the resolved root value is always passed to the Operator.

Fixes #33661

> [!IMPORTANT]
>

## Summary

When using `STORAGE_TYPE=opendal` with `OPENDAL_SCHEME=fs` and no `OPENDAL_FS_ROOT` env var configured, the `OpenDALStorage.__init__` resolves a default root value `"storage"` via `kwargs.get("root", "storage")`, but never writes it back to `kwargs`. This causes `Operator(scheme="fs", **kwargs)` to fail with:ConfigInvalid: root is not specified

**Root Cause**: In commit 99f40a9682 (PR #11754), the code was refactored from `Path(kwargs["root"])` to `root = kwargs.get("root", "storage")` to support a default value, but the resolved value was never assigned back to `kwargs`.

**Fix**: Added `kwargs["root"] = root` after resolving the default, ensuring the Operator always receives the root parameter.

**Tests Added**:
- Default root (`"storage"`) when no root is specified
- Explicit root passed as keyword argument
- Root from `OPENDAL_FS_ROOT` environment variable

## Screenshots

| Before | After |
|--------|-------|
| `ConfigInvalid: root is not specified` when `OPENDAL_FS_ROOT` is not set | OpenDAL fs storage initializes correctly with default root `"storage"` |

## Checklist


api/tests/unit_tests/oss/opendal/test_opendal.py
```diff
@@ -11,6 +11,59 @@
 )
 
 
+class TestOpenDALFsDefaultRoot:
```
QuantumGhost: This is actually an integration testing, as the OpenDALStorage does read and write files. 

IMO it's better to move this to the  test_containers_integration_tests folder.

api/extensions/storage/opendal_storage.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, scheme: str, **kwargs):
         if scheme == "fs":
             root = kwargs.get("root", "storage")
```
QuantumGhost: This line and the `kwargs["root"] = root` below could be merged into a single 
`root = kwargs.setdefault("root", "storage")` statment

QuantumGhost (commented):
Other changes LGTM.

api/extensions/storage/opendal_storage.py
RickDamon: done

api/tests/unit_tests/oss/opendal/test_opendal.py
RickDamon: done

QuantumGhost (approved):
LGTM.

PR #33719, opened 2026-03-19
fix: correct retrieval_model type in HitTestingPayload from dict to RetrievalModel

lyfuci:
## Problem

`HitTestingPayload.retrieval_model` was typed as `dict[str, Any] | None`, while the equivalent fields in `DatasetCreatePayload` and `DatasetUpdatePayload` use the structured `RetrievalModel` Pydantic model.

This caused Flask-RESTX's `register_schema_model` to generate incorrect Swagger documentation — the `weights` field (which is actually a `WeightModel` nested object) was shown as a primitive `float` instead of its correct nested structure.

## Fix

Changed `retrieval_model` type in `HitTestingPayload` from `dict[str, Any] | None` to `RetrievalModel | None`, consistent with `DatasetCreatePayload` and `DatasetUpdatePayload`.

The existing `parse_args` behavior is preserved since Pydantic's `model_dump(exclude_none=True)` correctly serializes nested models to dicts.

## Affected Files

- `api/controllers/console/datasets/hit_testing_base.py`


lyfuci:
Closing this PR — the `data_source_type` fix it contained has already been merged into upstream `main`. The original intent (fixing `HitTestingPayload.retrieval_model` type) is now correctly addressed in #33750.

PR #33752, opened 2026-03-19
fix(weaviate): proactive schema migration for pre-upgrade collections

RickDamon:
Collections created before `doc_type` was introduced cause `no such prop with name 'doc_type'` GRPC errors after upgrading.

- `_ensure_properties()` checks/adds missing properties, returns bool
- `_maybe_ensure_properties()` uses double-checked locking with a module-level set for once-per-process-per-collection verification
- All query/filter paths call `_maybe_ensure_properties()` before executing, transparently fixing old schemas on first access
- Failed migrations are not cached, allowing automatic retry

Closes #33716

> [!IMPORTANT]
>

## Summary

After upgrading Dify, Weaviate collections created before the `doc_type` property was introduced cause GRPC query failures:

```
Query call with protocol GRPC search failed with message no such prop with name 'doc_type'
found in class 'Vector_index_xxx_Node' in the schema.
```

**Root cause:** New code includes `doc_type` in `return_properties` when querying Weaviate, but old collections don't have this property in their schema.

**Fix:** Proactive schema migration with process-level caching — on first access to each collection, verify and add any missing properties *before* the query executes. This maybe a good way to solve it IMO, if you guys have better ways, then don't merge it.

### Key Changes

| Component | Change |
|-----------|--------|
| `_ensure_properties() → bool` | Checks collection schema and adds missing properties (`doc_type`, `document_id`, `doc_id`, `chunk_index`). Returns `True` if all confirmed present, `False` if any addition failed. |
| `_maybe_ensure_properties()` | Double-checked locking with module-level `_ensured_collections: set[str]`. Each collection is verified **once per process lifetime**. Failed migrations are **not** cached, allowing automatic retry on next request. |
| 4 protected entry points | `search_by_vector`, `search_by_full_text`, `text_exists`, `delete_by_metadata_field` all call `_maybe_ensure_properties()` before executing. |
| `_create_collection` | Also marks collections as ensured after successful schema verification to avoid redundant checks. |

**Cost:** First access to each old collection adds one schema-check RPC call (milliseconds). All subsequent queries have zero overhead (a single Python `set.__contains__` check).

### Why not catch-retry?

The [previous approach](https://github.com/langgenius/dify/pull/33747) used exception-triggered migration (catch `WeaviateQueryError` → fix schema → retry). It was abandoned because:
1. Only protected 2 of 5 code paths
2. Relied on fragile `"no such prop"` string matching
3. Could not handle same-type schema evolution (e.g. property type changes)

## Screenshots

N/A — backend-only change, no UI impact.

## Checklist

```

### DIFF
--- a/api/core/rag/datasource/vdb/weaviate/weaviate_vector.py
+++ b/api/core/rag/datasource/vdb/weaviate/weaviate_vector.py
@@ -37,6 +37,12 @@ logger = logging.getLogger(__name__)
 _weaviate_client: weaviate.WeaviateClient | None = None
 _weaviate_client_lock = threading.Lock()
 
+# Process-level cache: tracks collections whose required properties have been
+# verified in this process lifetime.  Guarded by _ensured_collections_lock so
+# concurrent Celery/thread workers can share the set safely.
+_ensured_collections: set[str] = set()
+_ensured_collections_lock = threading.Lock()
+
 
 def _shutdown_weaviate_client() -> None:
     """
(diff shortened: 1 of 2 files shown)


themavik (commented):
Double-checked locking looks correct. One nit: _get_uuids param renamed to texts but the type is list[Document] — could be confusing

jlfloresTeika:
Having this exact issue

PR #33766, opened 2026-03-19
fix: use constant-time comparison for API keys and secrets

gambletan:
## Summary

Fixes #33763

- Replace `==` and `!=` operators with `hmac.compare_digest()` for all API key, token, and secret comparisons to prevent timing side-channel attacks
- Added `import hmac` where needed
- Also fixed additional locations beyond those listed in the issue: CSRF token comparison, HMAC signature comparison, init password comparison, and OAuth client secret comparison

## Affected files

| File | Change |
|------|--------|
| `api/libs/token.py` | Admin API key + CSRF token comparisons |
| `api/extensions/ext_login.py` | Admin API key comparison |
| `api/controllers/inner_api/wraps.py` | Inner API key (3 locations) + HMAC signature comparison |
| `api/controllers/console/admin.py` | Admin API key comparison |
| `api/controllers/console/init_validate.py` | Init password comparison |
| `api/controllers/console/auth/oauth_server.py` | OAuth client secret comparison |

## Why this matters

String comparison with `==`/`!=` in Python short-circuits on the first differing character, leaking timing information proportional to the matching prefix length. `hmac.compare_digest()` always compares all bytes in constant time, preventing timing side-channel attacks against secret values.

## Test plan

- [ ] Verify existing tests pass (no behavioral change for correct inputs)
- [ ] Verify authentication still works with valid API keys
- [ ] Verify invalid API keys are still rejected


crazywoola:
Please fix the lint errors.

PR #33770, opened 2026-03-19
fix: resolve empty answer in blocking mode API responses

### LINKED ISSUE #33721: API blocking mode returns empty answer even though workflow execution succeeded in logs
### Self Checks

### Dify version

1.13.1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

When invoking a workflow via Dify API with response_mode: "blocking", the returned JSON contains an empty answer field. However, the same workflow runs successfully when tested directly in the Dify Workbench, and the execution logs for the API request also show a normal completion with the expected output.
Create a workflow in Dify (or use an existing one) that produces a non-empty output.

Obtain an API key and construct a blocking API request:

Endpoint: POST /v1/workflows/run

Headers: Authorization: Bearer <your-api-key>, Content-Type: application/json

Body (example):

json
{
  "inputs": { "query": "Hello" },
  "response_mode": "blocking",
  "user": "test-user"

Image
Image

}
Send the request using curl, Postman, or your application.

Observe that the answer field in the response is "" (empty).

Check the execution logs in Dify Console for that specific API request – it shows the workflow completed successfully with the correct output.

### ✔️ Expected Behavior

The answer field in the blocking API response should contain the final output of the workflow, identical to what is shown in the execution logs

### ❌ Actual Behavior

_No response_

gambletan:
Fixes #33721

## Summary
- **Race condition fix**: `AppQueueManager.stop_listen()` eagerly cleared `graph_runtime_state` for GC, but it is called from the worker thread immediately after enqueueing a terminal event. The consumer thread may not have cached the reference yet via `_resolve_graph_runtime_state()`, causing it to see `None` and either raise `ValueError` or produce an empty response. Removed the eager `None` assignment; the reference is released naturally when the queue manager is garbage-collected.

- **Chatflow blocking answer fallback**: The advanced-chat (Chatflow) blocking path built the `answer` exclusively from `QueueTextChunkEvent` streaming chunks accumulated in `_task_state.answer`. When the response coordinator did not emit text chunks for certain variable types (e.g. workflow-tool outputs fed directly into an Answer node without an intermediate Code/Template node), the accumulated answer stayed empty. Added a fallback that reads `answer` from `graph_runtime_state.outputs` when text chunks are missing, matching the value already persisted in execution logs.

## Test plan
- [ ] Deploy with a Chatflow that pipes a sub-workflow tool output directly into an Answer node. Call `/v1/chat-messages` with `response_mode: "blocking"` and verify the `answer` field is populated.
- [ ] Verify the same Chatflow works in streaming mode (no regression).
- [ ] Deploy a Workflow app and call `/v1/workflows/run` with `response_mode: "blocking"` — verify `data.outputs` is populated.
- [ ] Run a fast-executing workflow to exercise the race-condition fix path.


themavik (commented):
the race fix in stop_listen makes sense. quick question: where do we clear _graph_runtime_state now for long-lived workers, or is GC on manager teardown the only cleanup path?

gambletan:
It is cleared in `listen()`'s `finally` block, after the queue consumer finishes or closes the generator—not only on manager teardown. Keeping it until then preserves the terminal event's access to the state while still releasing the reference at the end of consumption.

PR #33788, opened 2026-03-20
refactor(workspace): optimize /workspaces plan resolution for SaaS and enterprise with resilient fallback

lin-snow:
> [!IMPORTANT]
>

Fixes #33718

## Summary

Optimizes the console **`GET /workspaces`** (`TenantListApi`) plan resolution by **deployment mode** (SaaS vs enterprise-only vs community/self-hosted). On the **SaaS happy path**, it replaces per-tenant `FeatureService.get_features` with a single `BillingService.get_plan_bulk` call. It adds **tiered fallbacks** when bulk billing data is **partially missing** or **empty**.

## Motivation

Previously, every joined workspace triggered `FeatureService.get_features(tenant_id)` mainly to derive `plan`, increasing fan-out into billing-related resolution. SaaS benefits from bulk plan fetch; enterprise-only deployments (`ENTERPRISE_ENABLED` + `!BILLING_ENABLED`) do not need billing-backed feature resolution for this field.

### 1. Explicit deployment-path branching

- `is_enterprise_only = ENTERPRISE_ENABLED && !BILLING_ENABLED`
- `is_saas = EDITION == "CLOUD" && BILLING_ENABLED`

Plan resolution is chosen deterministically by these flags for clarity and testability.

### 2. SaaS: bulk plan retrieval (happy path)

- In SaaS mode, collect all joined `tenant_id`s and call `BillingService.get_plan_bulk(tenant_ids)` **once**.
- For tenants present in the bulk map, `plan` comes from bulk (`plan` with `CloudPlan.SANDBOX` when empty).

### 3. SaaS: partial fallback (tenant missing from bulk map)

- If bulk returns a **non-empty** map but **omits** some tenant IDs, only those missing tenants call `FeatureService.get_features` via `_plan_from_billing_features`.
- Partial path semantics: **`billing.enabled` gates** plan display — disabled → `CloudPlan.SANDBOX`; enabled → `subscription.plan or CloudPlan.SANDBOX`.

### 4. SaaS: full fallback (empty bulk for non-empty input)

- `get_plan_bulk` may swallow errors internally and return `{}`. When `tenant_ids` is non-empty but the merged bulk result is **empty**, log a **warning** and use the **legacy per-tenant** path for **all** tenants.

### 5. Legacy / community (and SaaS full fallback)

- Uses `features.billing.subscription.plan or CloudPlan.SANDBOX` (**no** `billing.enabled` gate on this path).

### 6. Enterprise-only shortcut

- When `is_enterprise_only`, skip `FeatureService`; `plan` remains `CloudPlan.SANDBOX` (consistent with prior behavior for this endpoint).

## Files changed

- `api/controllers/console/workspace/workspace.py`
- `api/tests/unit_tests/controllers/console/workspace/test_workspace.py`

## Test coverage

- SaaS happy path: `get_plan_bulk` called once; `FeatureService` not used when all tenants are covered by bulk.
- SaaS partial omission: `FeatureService` called **only** for missing tenant(s); covers `billing.enabled` true/false.
- SaaS empty bulk: full legacy path; per-tenant `FeatureService`; warning logged.
- Community (`BILLING_ENABLED=false`, self-hosted): expectations unchanged.
- Enterprise-only: no `FeatureService`; default plan; empty tenant list covered.

## Risk / notes for reviewers

- Bulk `plan` is still assumed to match product expectations when billing would be “enabled”; partial fallback re-applies `billing.enabled` **only** for tenant IDs **omitted** from bulk.
- `get_plan_bulk` error handling lives inside `BillingService`; the controller treats an **empty aggregate result** as the signal for full fallback.

## Checklist

- [x] I've added tests for the paths introduced/changed.
- [ ] I ran `make lint` and `make type-check` (backend) and `cd web && npx lint-staged` (frontend) as required by the repo

### DIFF
--- a/api/controllers/console/workspace/workspace.py
+++ b/api/controllers/console/workspace/workspace.py
@@ -127,7 +127,9 @@ class TenantListApi(Resource):
         for tenant in tenants:
             plan: str = CloudPlan.SANDBOX.value
             if is_saas and not use_legacy_feature_path:
-                plan = tenant_plans.get(tenant.id, {}).get("plan", CloudPlan.SANDBOX.value)
+                tenant_plan = tenant_plans.get(tenant.id)
+                if tenant_plan:
+                    plan = tenant_plan["plan"] if tenant_plan["plan"] else CloudPlan.SANDBOX.value
             elif not is_enterprise_only:
                 features = FeatureService.get_features(tenant.id)
                 plan = features.billing.subscription.plan or CloudPlan.SANDBOX.value
--- a/api/controllers/console/workspace/workspace.py
+++ b/api/controllers/console/workspace/workspace.py
@@ -30,7 +30,7 @@ from libs.helper import TimestampField
 from libs.login import current_account_with_tenant, login_required
 from models.account import Tenant, TenantStatus
 from services.account_service import TenantService
-from services.billing_service import BillingService
+from services.billing_service import BillingService, SubscriptionPlan
 from services.enterprise.enterprise_service import EnterpriseService
 from services.feature_service import FeatureService
 from services.file_service import FileService
@@ -112,7 +112,7 @@ class TenantListApi(Resource):
         tenant_dicts = []
         is_enterprise_only = dify_config.ENTERPRISE_ENABLED and not dify_config.BILLING_ENABLED
         is_saas = dify_config.EDITION == "CLOUD" and dify_config.BILLING_ENABLED
-        tenant_plans: dict[str, dict] = {}
+        tenant_plans: dict[str, SubscriptionPlan] = {}
         use_legacy_feature_path = not is_enterprise_only and not is_saas
 
         if is_saas:
--- a/api/controllers/console/workspace/workspace.py
+++ b/api/controllers/console/workspace/workspace.py
@@ -40,14 +40,6 @@ logger = logging.getLogger(__name__)
 DEFAULT_REF_TEMPLATE_SWAGGER_2_0 = "#/definitions/{model}"
 
 
-def _plan_from_billing_features(tenant_id: str) -> str:
-    """Resolve plan from billing feature flags (SaaS partial-fallback path)."""
-    features = FeatureService.get_features(tenant_id)
-    if features.billing.enabled:
-        return features.billing.subscription.plan or CloudPlan.SANDBOX
-    return CloudPlan.SANDBOX
-
-
 class WorkspaceListQuery(BaseModel):
     page: int = Field(default=1, ge=1, le=99999)
     limit: int = Field(default=20, ge=1, le=100)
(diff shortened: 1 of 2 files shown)


api/controllers/console/workspace/workspace.py
```diff
@@ -108,17 +118,39 @@ def get(self):
...
+        use_legacy_feature_path = not is_enterprise_only and not is_saas
+
+        if is_saas:
+            tenant_ids = [tenant.id for tenant in tenants]
+            if tenant_ids:
+                tenant_plans = BillingService.get_plan_bulk(tenant_ids)
+                # If bulk fetch returned empty for non-empty input, fall back to legacy path
+                if not tenant_plans:
+                    logger.warning("get_plan_bulk returned empty result, falling back to legacy feature path")
+                    use_legacy_feature_path = True
```
wylswz: I would suggest removing this variable assignment of `use_legacy_feature_path ` and just leave an empty `tenant_plans` dict as it is, then in the loop below, call `_plan_from_billing_features` if the `tenant_id` is missing in the dict.

Will this improve the readability while keeping the semantics equivalent?

lin-snow: @wylswz 

Thanks for the suggestion — agreed on the readability angle.

We’ve **removed `use_legacy_feature_path`** and now keep `tenant_plans` as returned by `get_plan_bulk` (possibly empty). In the loop, SaaS resolves `plan` as: **use bulk when present**, otherwise **`FeatureService.get_features` + `subscription.plan or SANDBOX`** (inlined, no separate helper).

**Semantic equivalence note:** using a *single* missing-tenant path like `_plan_from_billing_features` for **all** misses is **not** equivalent to the older behavior if that helper applies a `billing.enabled` gate — when bulk returns **empty** for non-empty input, the legacy path was **`subscription.plan or SANDBOX` without that gate** (our unit test covered `enabled=False` still returning the mocked plan). We initially preserved that split; later we aligned with the SaaS contract that **`billing.enabled` is always true**, so the missing-tenant path converged to the same **`subscription.plan or SANDBOX`** resolution anyway.

api/controllers/console/workspace/workspace.py
```diff
@@ -108,17 +110,37 @@ def get(self):
...
+        is_enterprise_only = dify_config.ENTERPRISE_ENABLED and not dify_config.BILLING_ENABLED
+        is_saas = dify_config.EDITION == "CLOUD" and dify_config.BILLING_ENABLED
+        tenant_plans: dict[str, SubscriptionPlan] = {}
+
+        if is_saas:
+            tenant_ids = [tenant.id for tenant in tenants]
+            if tenant_ids:
+                tenant_plans = BillingService.get_plan_bulk(tenant_ids)
+                if not tenant_plans:
+                    logger.warning("get_plan_bulk returned empty result, falling back to legacy feature path")
```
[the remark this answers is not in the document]
lin-snow: Good point on cache vs consistency.

We’re intentionally using **`get_plan_bulk` here** (no Redis cache) so **`GET /workspaces` stays close to billing truth** for plan labels; the cached variant (`get_plan_bulk_with_cache`) trades **~TTL staleness** for fewer API calls, which we’re not taking on this hot path without an explicit product decision.

Also, the suggested snippet calls **both** `get_plan_bulk` and `get_plan_bulk_with_cache` in sequence — we’d only **replace** the call with one or the other, not both.

If we later want to optimize further, we can switch to `get_plan_bulk_with_cache` and document the staleness window in the PR/release notes.

zyssyz123 (approved):
LGTM

PR #33802, opened 2026-03-20
fix(api): replace bare except with scoped exception handling for issue #33425

Haohao-end:
## Summary
This PR removes unnecessary bare `except:` clauses in several API modules and narrows them to `except Exception:` or `except ImportError` where appropriate.

## Changes
- Preserve parser fallback behavior in CoT agent output parsing without swallowing interpreter-exit exceptions
- Keep CSRF verification semantics unchanged while allowing `KeyboardInterrupt` / `SystemExit` / `GeneratorExit` to propagate
- Narrow storage `exists()` and vector `text_exists()` probe handlers to ordinary runtime exceptions
- Restrict AnalyticDB optional dependency handling to `ImportError` only

## Validation
- Added focused unit tests to verify `KeyboardInterrupt` is no longer swallowed in parser and CSRF token paths
- Ran targeted pytest for the modified test files successfully

Closes #33425


Haohao-end:
The failing checks are now limited to the JavaScript code-executor integration tests.

The actual failure is:

`fork/exec /usr/local/bin/node: no such file or directory`

So the current blocker appears to be the test/sandbox environment missing Node at the expected path, rather than a regression caused by this PR’s scoped exception-handling changes.

crazywoola:
Thanks for your contribution, @Haohao-end — really appreciate the effort here.

Closing this in favor of #33426, which was opened earlier and targets the same issue (#33425). To keep duplicate cleanup fair, we're going FIFO on dupes — the older PR gets the merge consideration first.

PR #33827, opened 2026-03-20
feat(workflow): add MVP human-in-the-loop pause support for workflow agent nodes

Haohao-end:
## Summary
This PR is an implementation draft for the problem discussed in #34032.

When an agent strategy emits a dedicated clarification payload in its JSON output, the agent node now pauses through the existing workflow pause pipeline instead of forcing a succeeded state.

## What changed
- detect `human_required` / `clarification` payloads in workflow agent node JSON output
- emit a standard pause event instead of the normal success completion event
- preserve partial outputs and metadata when the node pauses
- create a standard human-input form for clarification requests
- keep the normal success path unchanged

## Scope
This is an MVP for workflow agent nodes only.
It does not implement provider-specific resume continuation yet.

## Validation
- `uv run --project api python -m py_compile ...`
- focused unit test added for:
  - clarification payload -> `PAUSED`
  - normal path -> `SUCCEEDED`


Haohao-end:
The type-check failure was caused by my patch and I will fix it.
The remaining JavaScript code executor failures look unrelated to this PR and point to a missing Node binary in the CI environment:
/usr/local/bin/node: no such file or directory

Haohao-end:
I pushed a follow-up fix for the type-check issue, and the PR-specific typing problem is now resolved.

The remaining failures appear unrelated to this change. They are isolated to JavaScript code executor integration tests and fail because the CI environment cannot find Node at:

`/usr/local/bin/node`

Error:
`fork/exec /usr/local/bin/node: no such file or directory`

This PR does not touch the JavaScript code executor path.

QuantumGhost:
Hi @Haohao-end,

Thank you for your contribution!

This appears to be a significant change to the current implementation. Could you please open an issue to provide more context regarding the background and your proposed approach?

In the meantime, I have converted this to a draft to prevent it from being merged prematurely.

Haohao-end:
Hi @QuantumGhost, thanks for the guidance.

You're right — this change is relatively significant, so aligning on the background and approach first makes sense.

I opened an issue to describe the problem, motivation, MVP scope, and the proposed direction for reusing the existing workflow pause pipeline for workflow Agent nodes:

#34032

I’ll use that thread to align on scope and approach first, then update this PR accordingly.

crazywoola:
Thank you for taking the time to contribute to Dify. This pull request currently has merge conflicts with the latest `main`, so we are closing it to keep the pull request queue current. If the change is still relevant, please rebase onto the latest `main`, resolve the conflicts, and open an updated pull request or let us know if you would like this one reopened. We appreciate your contribution and understanding.

PR #33854, opened 2026-03-21
fix: codes/code_executor/test_code_javascript.py failed

fatelei:
> [!IMPORTANT]
>

## Summary

fix #33850

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/tests/test_containers_integration_tests/conftest.py
+++ b/api/tests/test_containers_integration_tests/conftest.py
@@ -170,6 +170,9 @@ class DifyTestContainers:
         self.dify_sandbox.with_exposed_ports(8194)
         self.dify_sandbox.env = {
             "API_KEY": "test_api_key",
+            # Resolve Node via the container PATH so JavaScript execution tests
+            # do not depend on a specific absolute installation path.
+            "NODEJS_PATH": "node",
         }
         self.dify_sandbox.start()
         sandbox_host = self.dify_sandbox.get_container_host_ip()


hjlarry (approved):

wangji0923:
Hi, my PR #33853 is currently blocked by the same CI failure, so I read through this PR and its logs to understand the cause.

I may be missing some context, but I think this patch is very close while still addressing the symptom rather than the root cause.

From a local repo check:
- `api/tests/test_containers_integration_tests/conftest.py` currently starts `langgenius/dify-sandbox:latest`.
- The repo's compose files pin the sandbox image to `langgenius/dify-sandbox:0.2.12`.
- `docker/volumes/sandbox/conf/config.yaml.example` still uses `nodejs_path: /usr/local/bin/node`.

And from the CI logs:
- before this change, the failing JS tests were erroring with `fork/exec /usr/local/bin/node: no such file or directory`
- in this PR, the failure changes to `exec: "node": executable file not found in $PATH`

So the `NODEJS_PATH=node` change does change the failure mode, but it seems to move it from a missing absolute path to a missing PATH binary, rather than making the tests pass.

That makes me wonder if the bigger problem is that the testcontainers setup is pulling a drifting `latest` sandbox image, while the repo itself is pinned to `0.2.12` elsewhere.

If so, would it make sense to first pin the testcontainers sandbox image in `api/tests/test_containers_integration_tests/conftest.py` to the same version as the compose files, and then only keep a `NODEJS_PATH` override if it is still needed?

If helpful, I can try to prepare a small follow-up patch or help validate that direction.

PR #33884, opened 2026-03-23
feat: change xread to xreadgroup

fatelei:
> [!IMPORTANT]
>

## Summary

fix #33881

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/libs/broadcast_channel/redis/streams_channel.py
+++ b/api/libs/broadcast_channel/redis/streams_channel.py
@@ -91,9 +91,7 @@ class _StreamsSubscription(Subscription):
             if "BUSYGROUP" in str(e):
                 self._group_ready = True
             else:
-                logger.warning(
-                    "xgroup create failed for %s/%s: %s", self._key, self._group_name, e, exc_info=True
-                )
+                logger.warning("xgroup create failed for %s/%s: %s", self._key, self._group_name, e, exc_info=True)
         except Exception as e:  # pragma: no cover - safety net for different redis-py versions
             logger.warning(
                 "xgroup create unexpected error for %s/%s: %s",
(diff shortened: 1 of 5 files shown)


api/libs/broadcast_channel/redis/streams_channel.py
```diff
@@ -69,10 +73,55 @@ def __init__(self, client: Redis | RedisCluster, key: str):
...
+            # Use mkstream=True in case the stream key doesn't exist yet
+            self._client.xgroup_create(self._key, self._group_name, id="0", mkstream=True)
+            self._group_ready = True
+        except ResponseError as e:
+            # Group might already exist if recreated quickly; mark ready on BUSYGROUP, otherwise retry later
+            if "BUSYGROUP" in str(e):
+                self._group_ready = True
+            else:
+                logger.warning("xgroup create failed for %s/%s: %s", self._key, self._group_name, e, exc_info=True)
+        except Exception as e:  # pragma: no cover - safety net for different redis-py versions
```
QuantumGhost, line 95: This is not necessary IMO. Unnecessary `try ... catch` makes debugging more complicated.

QuantumGhost (commented):
It seems that this PR is functionally equivalent to the previous one, if I understand the semantics of `XREADGROUP` correctly?

The current implementation creates a distinct group for each `_Subscription` created in Dify, and set the start posiition of the group to `0`. Then the consumption part consumes the event that never delivered to other consumers.

However, since each subscription creates a distinct consumer group, there is only one consumer inside for each consumer group. The logic essentially is equivalent to consuming with XREAD and from `0-0`, which is the logic in PR in #32586.

Could you clarify if there are specific advantages to using XREADGROUP here that I might be overlooking?

QuantumGhost:
Superseded by #34030

PR #33927, opened 2026-03-23
fix(dev): load middleware env in start-docker-compose

dominciyue:
## Summary
- load docker/middleware.env in dev/start-docker-compose
- keep the helper script consistent with other middleware startup commands in the repo
- make custom middleware ports and compose variables generated during setup take effect when using the dev script

## Test plan
- [x] verify the script diff only adds --env-file middleware.env
- [x] docker compose --env-file middleware.env -f docker-compose.middleware.yaml -p dify config

### DIFF
--- a/dev/start-docker-compose
+++ b/dev/start-docker-compose
@@ -5,4 +5,4 @@ SCRIPT_DIR="$(dirname "$(realpath "$0")")"
 ROOT="$(dirname "$SCRIPT_DIR")"
 
 cd "$ROOT/docker"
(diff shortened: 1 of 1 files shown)


dominciyue:
Good catch. `docker/README.md` and `docker/middleware.env.example` already rely on `COMPOSE_PROFILES` from `middleware.env`, so I removed the explicit `--profile` flags to keep the script consistent with the existing setup. I also updated the test plan accordingly.

crazywoola (approved):

PR #33929, opened 2026-03-23
fix: trust Content-Type for known binary MIME types in HTTP Request node

### LINKED ISSUE #33897: HTTP Request node misidentifies PDF as text when binary marker is absent
https://github.com/langgenius/dify/blob/110b8c925e5265f48f8c74b937fa4204ca321c6e/api/dify_graph/nodes/http_request/entities.py#L152-L171

# HTTP Request node misidentifies PDF as text when binary marker is absent

## Self Checks
- [x] This is only for bug report
- [x] I have searched for existing issues, including closed ones
- [x] I confirm that I am using English to submit this report

## Dify version
Latest (main branch, `api/dify_graph/nodes/http_request/`)

## Cloud or Self Hosted
Self Hosted (Docker)

## Steps to reproduce

1. Upload a PDF generated by **Copper PDF** (a Java-based HTML-to-PDF engine) to S3
2. Create a workflow with an HTTP Request node that GETs the PDF via a pre-signed S3 URL
3. S3 returns the file with `Content-Type: application/pdf`
4. The HTTP Request node fails to recognize it as a file

## ✔️ Expected Behavior

The response should be recognized as a file (in the `files` output), since `Content-Type: application/pdf` clearly indicates binary file content.

## ❌ Actual Behavior

- `body` contains the raw PDF bytes decoded as text: `"%PDF-1.5\r\n1 0 obj\r\n<< /Type /Catalog ..."`
- `files` is empty — no file object is created
- The same PDF, after being rebuilt with `qpdf --linearize`, is correctly recognized as a file

## Root Cause

The bug is in `api/dify_graph/nodes/http_request/entities.py`, in the `is_file` property of the `Response` class.

The relevant code path:

```python
# For application types, try to detect if it's a text-based format
if content_type.startswith("application/"):
    # ...snip common text types check...

    # Try to detect if content is text-based by sampling first few bytes
    try:
        content_sample = self.response.content[:1024]
        content_sample.decode("utf-8")
        text_markers = (b"{", b"[", b"<", b"function", b"var ", b"const ", b"let ")
        if any(marker in content_sample for marker in text_markers):
            return False  # ← BUG: returns "not a file" for valid PDFs
    except UnicodeDecodeError:
        return True
```

**What happens step by step:**

1. `Content-Type` is `application/pdf` → passes through the `json/xml/javascript` check
2. The first 1024 bytes are sampled and decoded as UTF-8 → **succeeds** because the PDF lacks a binary marker comment on line 2
3. Text marker `<` is found in the PDF dictionary syntax `<< /Type /Catalog` → **matches**
4. `return False` → the response is treated as text, not a file

**Why some PDFs work and others don't:**

The PDF specification (ISO 32000) **recommends** (but does not require) a binary marker comment on the second line — a `%` followed by 4+ bytes with values > 128 (e.g., `%âãÏÓ`). Most PDF generators (Adobe, iText, qpdf, etc.) include this marker, which causes `content_sample.decode("utf-8")` to throw `UnicodeDecodeError`, correctly reaching `return True`.

**Copper PDF** does not output this binary marker. Its output starts with:

```
%PDF-1.5\r\n
1 0 obj\r\n
<< /Type /Catalog /Version /1.5 /Pages 2 0 R ...
```

The first 353 bytes are pure ASCII, so UTF-8 decoding succeeds, and the `<` in `<<` triggers the text marker check.

## Suggested Fix

For well-known binary MIME types like `application/pdf`, `application/zip`, `application/octet-stream`, etc., skip the heuristic byte-sampling entirely and trust the `Content-Type` header:

```python
if content_type.startswith("application/"):
    # Known binary application types — trust Content-Type
    known_binary_types = ("pdf", "zip", "gzip", "octet-stream", "msword",
                          "vnd.openxmlformats", "vnd.ms-excel", "vnd.ms-powerpoint")
    if any(bt in content_type for bt in known_binary_types):
        return True

    # Common text-based application types
    if any(text_type in content_type for text_type in ("json", "xml", "javascript", ...)):
        return False

    # Heuristic for unknown application/* types (existing logic)
    ...
```

Alternatively, a more general fix: if `Content-Type` is not in a known text type list, default to treating it as a file rather than applying heuristic byte detection.

## Environment

- **Dify**: Self-hosted, latest main branch
- **PDF generator**: Copper PDF 3.2.18 (used by some Japanese financial disclosure systems)
- **S3 response headers**: `Content-Type: application/pdf`, status 200

gambletan:
## Summary

Fixes #33897

- **Root cause**: PDFs generated by Copper PDF lack the optional binary marker comment (`%âãÏÓ`) on line 2. Without it, the first 1024 bytes are valid UTF-8, and the `<` character in PDF dictionary syntax (`<< /Type /Catalog`) matches the `text_markers` heuristic in `is_file`, causing `return False` — the response is treated as text instead of a file.
- **Fix**: For known binary application MIME subtypes (`pdf`, `zip`, `gzip`, `x-gzip`, `octet-stream`, `msword`, `vnd.openxmlformats`, `vnd.ms-excel`, `vnd.ms-powerpoint`, `x-tar`, `x-rar`, `x-7z-compressed`, `wasm`), skip the heuristic byte-sampling entirely and trust the `Content-Type` header. This check is inserted **before** the byte-sampling block so the heuristic is only used for genuinely unknown application types.
- **Test**: Added a parametrized regression test that verifies all known binary types are correctly identified as files even when the content is UTF-8 decodable and contains text markers like `<`.

## Test plan

- [ ] Existing `test_entities.py` tests continue to pass
- [ ] New `test_known_binary_types_skip_heuristic` test passes for all 13 binary MIME types
- [ ] Manual: HTTP Request node correctly returns Copper PDF (no binary marker) in `files` output

### DIFF
--- a/api/dify_graph/nodes/http_request/entities.py
+++ b/api/dify_graph/nodes/http_request/entities.py
@@ -148,6 +148,27 @@ class Response:
 
         # For application types, try to detect if it's a text-based format
         if content_type.startswith("application/"):
+            # Known binary application types — trust Content-Type, skip heuristic byte-sampling.
+            # Must be checked before the text-type substring check below, because types like
+            # "vnd.openxmlformats-officedocument.spreadsheetml.sheet" contain "xml" as a
+            # substring and would otherwise be misidentified as text.
+            known_binary_subtypes = (
+                "pdf",
+                "zip",
+                "gzip",
+                "x-gzip",
+                "msword",
+                "vnd.openxmlformats",
+                "vnd.ms-excel",
+                "vnd.ms-powerpoint",
+                "x-tar",
+                "x-rar",
+                "x-7z-compressed",
+                "wasm",
+            )
+            if any(subtype in content_type for subtype in known_binary_subtypes):
+                return True
+
             # Common text-based application types
             if any(
                 text_type in content_type
--- a/api/tests/unit_tests/core/workflow/nodes/http_request/test_entities.py
+++ b/api/tests/unit_tests/core/workflow/nodes/http_request/test_entities.py
@@ -231,3 +231,39 @@ def test_text_property_with_escaped_unicode(mock_response, json_content, descrip
     # The text should be valid JSON that can be parsed back to proper Unicode
     parsed = json.loads(response.text)
     assert isinstance(parsed, dict), f"Invalid JSON for {description}"
+
+
+@pytest.mark.parametrize(
+    "content_type",
+    [
+        "application/pdf",
+        "application/zip",
+        "application/gzip",
+        "application/x-gzip",
+        "application/msword",
+        "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+        "application/vnd.ms-excel",
+        "application/vnd.ms-powerpoint",
+        "application/x-tar",
+        "application/x-rar",
+        "application/x-7z-compressed",
+        "application/wasm",
+    ],
+)
+def test_known_binary_types_skip_heuristic(mock_response, content_type):
+    """Test that known binary MIME types are identified as files even when content is UTF-8 decodable.
+
+    Regression test for https://github.com/langgenius/dify/issues/33897
+    PDFs without a binary marker comment (e.g. Copper PDF) have UTF-8-decodable content
+    containing '<' in dictionary syntax (<< /Type /Catalog), which previously triggered
+    the text_markers heuristic and caused the response to be misidentified as text.
+    """
+    # Simulate a Copper PDF without binary marker — valid UTF-8, contains '<'
+    pdf_content = b"%PDF-1.5\r\n1 0 obj\r\n<< /Type /Catalog /Pages 2 0 R >>\r\nendobj\r\n"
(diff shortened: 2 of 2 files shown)


keisuke-na:
@gambletan Nice fix! Three suggestions:

1. `octet-stream` in `known_binary_subtypes` conflicts with 
   `test_content_based_detection` — JSON content with 
   `application/octet-stream` should probably still be treated 
   as text. Consider removing `octet-stream` from the list.

2. `vnd.openxmlformats` doesn't match the full MIME type 
   `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`. 
   Need to check if the substring match logic handles this correctly.

3. Per gemini-code-assist's comment, moving `known_binary_subtypes` 
   to a module-level constant would be cleaner.

fatelei:
fix the conflict

gambletan:
Closing — the http_request node files this modifies were moved in the dify_graph refactor, so it'd need to be redone on the new structure rather than merged.

PR #33980, opened 2026-03-23
fix: SQLAlchemy deprecation warnings for default parameter

qianchongyang:
## Problem

SQLAlchemy emits deprecation warnings when callable objects are passed to the `default` parameter in ORM-mapped dataclasses. The warnings indicate this will raise an error in future releases.

## Solution

Replaced `default` parameter with `default_factory` for callable defaults in ORM-mapped dataclasses:
- Changed `default=<callable>` to `default_factory=<callable>` for dataclass-level defaults
- Ensures compatibility with future SQLAlchemy versions

## Validation

```python
# No more deprecation warnings
from models import SomeModel
# ✅ SADeprecationWarning no longer appears
```

Fixes #33978

### DIFF
--- a/api/models/model.py
+++ b/api/models/model.py
@@ -589,7 +589,9 @@ class AppModelConfig(TypeBase):
     __tablename__ = "app_model_configs"
     __table_args__ = (sa.PrimaryKeyConstraint("id", name="app_model_config_pkey"), sa.Index("app_app_id_idx", "app_id"))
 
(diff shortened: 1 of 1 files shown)


asukaminato0721:
@qianchongyang revert the change to class App(Base)

asukaminato0721 (dismissed):

asukaminato0721 (approved):

PR #33982, opened 2026-03-23
fix: replace deprecated datetime.utcnow() with timezone-aware alternative

BillionClaw:
Replace deprecated datetime.datetime.utcnow() (deprecated in Python 3.12+) with datetime.datetime.now(datetime.timezone.utc) across all 70 occurrences in test files. Fixes #33977.


asukaminato0721:
Fix ci error

PR #33984, opened 2026-03-23
fix: Replace deprecated datetime.utcnow() with timezone-aware datetime.now(datetime.UTC)

qianchongyang:
## Problem

Python's `datetime.utcnow()` is deprecated and scheduled for removal in a future version. Using this method causes deprecation warnings and will break in future Python versions.

## Solution

Replaced all instances of `datetime.utcnow()` with `datetime.now(datetime.UTC)` to use timezone-aware datetime objects. This ensures compatibility with future Python versions.

## Validation

```python
# No more deprecation warnings
from datetime import datetime, UTC
datetime.now(UTC)  # ✅ Timezone-aware, no warnings
```

Fixes #33977

### DIFF
--- a/api/tests/test_containers_integration_tests/helpers/execution_extra_content.py
+++ b/api/tests/test_containers_integration_tests/helpers/execution_extra_content.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from datetime import datetime, timedelta
+from datetime import UTC, datetime, timedelta
 from decimal import Decimal
 from uuid import uuid4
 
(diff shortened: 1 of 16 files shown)


QuantumGhost:
Thanks for the PR!

I have a few concerns regarding the database: SQLAlchemy sometimes has issues with timezone-aware `datetime`, and our current schema uses `TIMESTAMP` rather than `TIMESTAMPTZ`. Also, mixing naive and aware datetimes will trigger `TypeError`, so we’d need a comprehensive refactor to ensure consistency.

I believe moving to tz-aware `datetime` is the right direction, but the scope of this change might be broader than expected.

asukaminato0721:
Fix ci error

PR #34041, opened 2026-03-24
fix: resolve SADeprecationWarning for Callable default in DefaultFieldsMixin

vijaygopalbalasa:
## Summary

Replace deprecated `default=callable` with `insert_default` + `default_factory` in `DefaultFieldsMixin` to resolve `SADeprecationWarning` when used in `MappedAsDataclass` (TypeBase) contexts.

This is a follow-up to #33980 which fixed the same warning in `AppModelConfig` (`model.py`). This PR addresses the shared `DefaultFieldsMixin` in `base.py`, which is inherited by 6 models across `human_input.py`, `workflow.py`, and `execution_extra_content.py`.

## Problem

SQLAlchemy raises `SADeprecationWarning` when a callable is passed to the `default` parameter in an ORM-mapped dataclass context:

```
SADeprecationWarning: Callable object passed to the ``default`` parameter
for attribute 'id' in a ORM-mapped Dataclasses context is ambiguous,
and this use will raise an error in a future release.
```

## Fix

For each affected field in `DefaultFieldsMixin`:
- `insert_default=callable` for database-level INSERT defaults
- `default_factory=callable` for dataclass instance-level defaults

This is the same pattern already used throughout the codebase in `web.py`, `tools.py`, `trigger.py`, and `model.py` (via #33980).

## Changes

- `api/models/base.py`: Updated 3 field declarations in `DefaultFieldsMixin` (`id`, `created_at`, `updated_at`)

Closes #33978

### DIFF
--- a/api/models/base.py
+++ b/api/models/base.py
@@ -27,24 +27,29 @@ class DefaultFieldsMixin:
     id: Mapped[str] = mapped_column(
         StringUUID,
         primary_key=True,
-        # NOTE: The default serve as fallback mechanisms.
+        # NOTE: The insert_default serves as a fallback mechanism for INSERT.
         # The application can generate the `id` before saving to optimize
         # the insertion process (especially for interdependent models)
         # and reduce database roundtrips.
-        default=lambda: str(uuidv7()),
+        # Use insert_default + default_factory to avoid SADeprecationWarning
+        # when used in MappedAsDataclass (TypeBase) contexts.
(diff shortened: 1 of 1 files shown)


vijaygopalbalasa:
Closing this PR — after deeper analysis, DefaultFieldsMixin is not currently used by any MappedAsDataclass (TypeBase) model, so the SADeprecationWarning does not fire for this code path. The real fixes needed are in the individual model files. Apologies for the noise.

PR #34043, opened 2026-03-24
fix: handle null email in GitHub OAuth sign-in

Krishnachaitanyakc:
## Summary
Fixes #34014

When a GitHub user's profile email is null (hidden/private), the OAuth callback fails with HTTP 400 because `GitHubRawUserInfo` validates `email` as a required non-null string. Even after the type was relaxed to `NotRequired[str | None]` in #33882, the flow still raises a `ValueError` when no email can be resolved, blocking sign-in entirely.

This PR improves the email resolution strategy so that users with private GitHub emails can still sign in.

## Changes
- Added `verified` field to `GitHubEmailRecord` TypedDict to support verified-email fallback
- Extracted `_get_email_from_emails_endpoint()` for clearer email resolution logic with a priority chain: primary email > any verified email > empty string
- Replaced the `ValueError` in `_transform_user_info()` with a noreply email fallback (`{id}+{login}@users.noreply.github.com`) so sign-in can still proceed when no email is available
- Updated unit tests to cover verified email fallback and noreply email scenarios

## Test plan
- [x] Tested with GitHub account that has public email (primary email used)
- [x] Tested with GitHub account that has private email but primary email available via /user/emails endpoint (primary email used)
- [x] Tested with GitHub account that has only verified (non-primary) email (verified email used as fallback)
- [x] Tested with GitHub account where /user/emails endpoint fails (noreply email used)
- [x] Tested with GitHub account with no usable emails (noreply email used)
- [x] Existing OAuth tests updated and passing

### DIFF
--- a/api/libs/oauth.py
+++ b/api/libs/oauth.py
@@ -131,10 +131,11 @@ class GitHubOAuth(OAuth):
         response.raise_for_status()
         user_info = GITHUB_RAW_USER_INFO_ADAPTER.validate_python(_json_object(response))
 
-        # The profile email may be null when the user has "Keep my email addresses private" enabled.
-        # Fall back to the /user/emails endpoint to find a usable email address.
-        email_from_api = self._get_email_from_emails_endpoint(headers)
-        resolved_email = email_from_api or user_info.get("email") or ""
+        # Only call the /user/emails endpoint when the profile email is absent,
+        # i.e. the user has "Keep my email addresses private" enabled.
+        resolved_email = user_info.get("email") or ""
+        if not resolved_email:
+            resolved_email = self._get_email_from_emails_endpoint(headers)
 
         return {**user_info, "email": resolved_email}
 
@@ -170,10 +171,11 @@ class GitHubOAuth(OAuth):
         if not email:
             # When no email is available from the profile or /user/emails endpoint,
             # fall back to GitHub's noreply address so sign-in can still proceed.
+            # Use only the numeric ID (not the login) so the address stays stable
+            # even if the user renames their GitHub account.
             github_id = payload["id"]
-            github_login = payload["login"]
-            email = f"{github_id}+{github_login}@users.noreply.github.com"
-            logger.info("GitHub user %s has no public email; using noreply address", github_login)
+            email = f"{github_id}@users.noreply.github.com"
+            logger.info("GitHub user %s has no public email; using noreply address", payload["login"])
         return OAuthUserInfo(id=str(payload["id"]), name=str(payload.get("name") or ""), email=email)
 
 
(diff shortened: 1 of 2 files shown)


api/libs/oauth.py
```diff
@@ -130,25 +131,49 @@ def get_raw_user_info(self, token: str) -> JsonObject:
         response.raise_for_status()
         user_info = GITHUB_RAW_USER_INFO_ADAPTER.validate_python(_json_object(response))
 
+        # The profile email may be null when the user has "Keep my email addresses private" enabled.
+        # Fall back to the /user/emails endpoint to find a usable email address.
+        email_from_api = self._get_email_from_emails_endpoint(headers)
+        resolved_email = email_from_api or user_info.get("email") or ""
```
QuantumGhost: I would suggest only invoking the `_get_email_from_emails_endpoint` method if the `email` key is absent in `user_info`. This reduces unnecessary requests to GitHub.

api/libs/oauth.py
```diff
@@ -130,25 +131,49 @@ def get_raw_user_info(self, token: str) -> JsonObject:
...
         if not email:
-            raise ValueError(
-                'Dify currently not supports the "Keep my email addresses private" feature,'
-                " please disable it and login again"
-            )
+            # When no email is available from the profile or /user/emails endpoint,
+            # fall back to GitHub's noreply address so sign-in can still proceed.
+            github_id = payload["id"]
+            github_login = payload["login"]
+            email = f"{github_id}+{github_login}@users.noreply.github.com"
```
QuantumGhost: It seems that this is problematic if the user updated his username.

cc @crazywoola how do you think about this?

QuantumGhost (commented):
Other changes LGTM.

api/libs/oauth.py
crazywoola: It should be a rare case, but this is good than my previous solution.

api/libs/oauth.py
Krishnachaitanyakc: Good call — updated to only invoke `_get_email_from_emails_endpoint` when `user_info["email"]` is absent or `None`. This avoids the extra API round-trip for users who already have a public profile email.

api/libs/oauth.py
Krishnachaitanyakc: Great catch @QuantumGhost. Updated the noreply fallback to use `{id}@users.noreply.github.com` (without the login), so the address stays stable even if the user renames their GitHub account. GitHub routes both `{id}+{login}@...` and `{id}@...` to the same account, so this is safe.

QuantumGhost (approved):

PR #34236, opened 2026-03-29
fix: QA index processor preview key error

cyphercodes:
## Summary

Fixed a KeyError that occurred when using QA blocks in knowledge base workflows.

## Root Cause

The `QAIndexProcessor.format_preview` method was returning a dictionary with the key `'qa_preview'`, but the parent `IndexProcessor.format_preview` method expected the key `'preview'`. This caused a `KeyError: 'preview'` when iterating through the preview items.

## Changes

- `api/core/rag/index_processor/processor/qa_index_processor.py`: Changed the return key from `'qa_preview'` to `'preview'` to match the expected interface.

## Testing

This fix ensures that QA blocks in knowledge base workflows can be previewed without errors.

Fixes #34144

### DIFF
--- a/api/core/rag/index_processor/processor/qa_index_processor.py
+++ b/api/core/rag/index_processor/processor/qa_index_processor.py
@@ -237,7 +237,7 @@ class QAIndexProcessor(BaseIndexProcessor):
             preview.append({"question": qa_chunk.question, "answer": qa_chunk.answer})
         return {
             "chunk_structure": IndexStructureType.QA_INDEX,
-            "qa_preview": preview,
+            "preview": preview,
             "total_segments": len(qa_chunks.qa_chunks),
         }
 
--- a/api/tests/unit_tests/core/rag/indexing/processor/test_qa_index_processor.py
+++ b/api/tests/unit_tests/core/rag/indexing/processor/test_qa_index_processor.py
@@ -314,7 +314,7 @@ class TestQAIndexProcessor:
             with pytest.raises(ValueError, match="must be high quality"):
                 processor.index(dataset, dataset_document, {"qa_chunks": []})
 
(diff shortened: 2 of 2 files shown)


mahdirajaee (commented):
Clean, minimal fix that addresses the root cause directly. The QA index processor was returning `qa_preview` while the base `IndexProcessor.format_preview` interface expects `preview`, so this KeyError would have occurred every time a user attempted to preview QA-indexed segments through the knowledge base workflow.

One concern worth considering: this is a public-facing API response key change. If any external consumers (plugins, third-party integrations, or the frontend) were already handling the `qa_preview` key as a fallback or special case, changing it to `preview` could break those consumers in a different way. It would be worth verifying that the frontend chunk preview rendering code only reads from `preview` and never checks for `qa_preview`. That said, since the bug report indicates this was simply broken (KeyError), it is likely nobody was successfully consuming the old key anyway. The fix looks correct.

cyphercodes:
Updated the PR with a focused test fix after re-checking the current branch:

- Updated `TestQAIndexProcessor::test_format_preview_*` to expect the processor's internal `preview` payload shape introduced by this fix.
- The public preview response still maps QA items into `qa_preview` through `IndexProcessor.format_preview`, so the frontend-facing response key remains covered by the wrapper behavior on current `main`.

Local verification:
- `git diff --check` ✅
- `uv run --project api --group dev ruff check api/tests/unit_tests/core/rag/indexing/processor/test_qa_index_processor.py` ✅
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --project api --group dev pytest -o addopts='' api/tests/unit_tests/core/rag/indexing/processor/test_qa_index_processor.py::TestQAIndexProcessor::test_format_preview_returns_preview_items -q --tb=short` ✅

Issue #33294, opened 2026-03-11
[Refactor/Chore] use sa.Enum in mapped_column , instead of sa.String


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

it's more precise type.

@dosu give me a list on where I should change.

tmimmanuel:
@asukaminato0721 Can I work on this issue?

asukaminato0721:
@dosu now is there any remaining one

asukaminato0721:
@dosu now is there any remaining one

asukaminato0721:
@dosu
Now is there any remaining one to close this?

tmimmanuel:
Hi, @asukaminato0721 
Now is there any remaining one to close this?
I think remaining ones are remaining strings intentionally left.
Please review again. Thank. 🧐

asukaminato0721:
@dosu
Now is there any remaining one to close this?

tmimmanuel:
@asukaminato0721 Thanks. I will continue. 👍👌

asukaminato0721:
@dosu
Now is there any remaining one to close this?

asukaminato0721:
@dosu how can I lint so in the future no more port needed?

asukaminato0721:
@dosu
Now is there any remaining one to close this?

PR #34300, opened 2026-03-30
refactor: use EnumText for model_type in provider models

tmimmanuel:
## Summary
- Convert 5 `model_type` columns in provider.py from `Mapped[str] + String` to `Mapped[ModelType] + EnumText(ModelType)`
- Replace `to_origin_model_type()` calls with direct enum at write sites (12 call sites)
- Remove `ModelType.value_of()` wrappers at read sites (2 call sites)
- New rows store enum values ("llm", "text-embedding") instead of origin values ("text-generation", "embeddings")
- `ModelType.value_of()` handles both formats, so legacy DB rows are still readable

Related to #33294.

## Test plan
- [x] basedpyright: 0 errors
- [x] mypy: 0 errors
- [x] 176 unit tests pass

### DIFF
--- a/api/controllers/console/workspace/models.py
+++ b/api/controllers/console/workspace/models.py
@@ -287,12 +287,10 @@ class ModelProviderModelCredentialApi(Resource):
                 provider=provider,
             )
         else:
-            # Normalize model_type to the origin value stored in DB (e.g., "text-generation" for LLM)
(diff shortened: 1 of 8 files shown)


asukaminato0721 (approved):

tmimmanuel:
@asukaminato0721 Thanks for merging this. Please review again. I think it's the end.

Issue #34321, opened 2026-03-31
Hardcoded default secrets in docker-compose.yaml pose security risk for production deployments


mahdirajaee:
## Description

The `docker-compose.yaml` (and its `.env.example` defaults) ships with multiple hardcoded secret values that are used as fallback defaults via the `${VAR:-default}` syntax. If operators deploy without explicitly setting these variables, the instance runs with publicly known credentials.

## Specific instances

| Variable | Default value | Risk |
|----------|--------------|------|
| `SECRET_KEY` | `sk-9f73s3ljTXVcMT3Blb3ljTqtsKiGHXVcMT3BlbkFJLK7U` | Flask session signing key; known value allows session forgery |
| `DB_PASSWORD` | `difyai123456` | PostgreSQL database password |
| `REDIS_PASSWORD` | `difyai123456` | Redis instance password |
| `INNER_API_KEY_FOR_PLUGIN` | `QaHbTe77CtuXmsfyhR7+vRjI/+XbV1AaFy691iy+kGDv2Jvy0/eAh8Y1` | Internal API key for plugin daemon communication |
| `MINIO_SECRET_KEY` | `minioadmin` | MinIO object storage credential |

## Current behavior

These defaults are embedded directly in the compose YAML via `${SECRET_KEY:-sk-9f73s3ljTXVcMT3Blb3ljTqtsKiGHXVcMT3BlbkFJLK7U}` syntax. Users who follow a quick-start guide or forget to set `.env` values will silently run production services with well-known secrets that are visible in the public repository.

## Suggested improvements

1. **Remove default values for security-critical variables.** Instead of `${SECRET_KEY:-sk-...}`, use `${SECRET_KEY:?SECRET_KEY must be set}` so Docker Compose fails fast with a clear error if the variable is unset.
2. **Add a startup validation script or init container** that checks for default/unset secrets and prints a warning or refuses to start in production (`DEPLOY_ENV=PRODUCTION`).
3. **Document the requirement** prominently in the deployment guide. The current `.env.example` file has comments but does not prevent the defaults from being silently used.
4. **Consider generating random defaults** during `docker compose up` via an init script, similar to how some projects use `openssl rand -base64 32` in their setup scripts.

## Impact

Any Dify instance deployed with default secrets is vulnerable to:
- Session token forgery (via known `SECRET_KEY`)
- Direct database access if the PostgreSQL port is exposed
- Redis data exfiltration or manipulation
- Unauthorized plugin daemon API calls

[...]

PR #34152, opened 2026-03-26
fix: handle basic auth and credential cleanup

EndlessLucky:
Part of #34105 

## Summary

- Ensure web/CLAUDE.md is always present/valid for CI runs (no broken symlink ENOENT).
- Add web/scripts/ensure-claude-md.mjs to repair missing/dangling CLAUDE.md by copying from web/AGENTS.md.
- Load the guard early for both next build and eslint by importing it from web/next.config.ts and web/eslint.config.mjs.
- Make e2e/scripts/setup.ts call the same guard before triggering pnpm run build during full-stack E2E.

## Why this change

CI was failing with ENOENT on web/CLAUDE.md in both:

- Web build / E2E (next build does a stat() and exits)
- Lint (ESLint worker reads the file and exits)

A git-tracked symlink can still produce ENOENT on Linux when the symlink target is invalid (e.g. CRLF in link text or missing target). Making the file reliable and adding a small repair step stabilizes both pipelines.

## Changes

- web/CLAUDE.md: Convert from symlink to a regular markdown file (avoid Linux symlink resolution issues).
- web/scripts/ensure-claude-md.mjs: New guard script to ensure web/CLAUDE.md exists (and replace broken symlink if needed).
- web/next.config.ts: Import the guard so it runs before Next config is evaluated.
- web/eslint.config.mjs: Import the guard so it runs before ESLint config is evaluated.
- e2e/scripts/setup.ts: Run the guard before building the web app in E2E.

## Test plan
vp run lint:ci passes in CI (no web/CLAUDE.md ENOENT; config import-order/style lint fixed)
vp run e2e:full passes in CI (Next build no longer fails on web/CLAUDE.md)

PR #34408, opened 2026-04-01
fix(docker): Remove insecure default secret fallbacks in compose

EndlessLucky:
Part of #34321.

## Summary

- Add a CI helper script to generate missing Docker secrets and keep middleware credentials in sync.
- Prevent docker/generate_docker_compose from crashing by defining REQUIRED_SECRET_VARS and hardening shared-env interpolation rules.
- Update CI workflows (web E2E + VDB tests) to copy example dotenvs and run the helper script before starting containers.

## Why this change

CI (and some self-hosted runs) copy `docker/.env.example` → `docker/.env` and then start Docker Compose. Compose interpolation and post-job cleanup can fail if required secrets are blank or missing. Separately, some services must use the same Postgres/Redis passwords as the Compose stack, otherwise they fail authentication (e.g. plugin daemon SASL auth against Postgres).

This change makes CI deterministic by:

- ensuring required secrets are non-empty in `docker/.env`, and

- synchronizing `DB_PASSWORD / REDIS_PASSWORD` into `docker/middleware.env.`

## Changes

- `.github/scripts/prepare_docker_ci_env.py`: New script — fills blank required secrets in `docker/.env`, then syncs `DB_PASSWORD` and `REDIS_PASSWORD` into `docker/middleware.env`.
- `docker/generate_docker_compose`: Define `REQUIRED_SECRET_VARS` and ensure shared env interpolation treats those as required (no weak defaults).
- `docker/.env.example`: Adjust defaults/required-secret handling to be compatible with Compose interpolation and CI setup.
- `.github/workflows/web-e2e.yml`: Copy `.env.example` files and run `prepare_docker_ci_env.py` before E2E.
- `.github/workflows/vdb-tests.yml`: Run the same dotenv setup + script before bringing up vector stores.
- `.github/workflows/vdb-tests-full.yml`: Same as above for the full VDB matrix.

## Test plan

- CI: `web-e2e` passes (API + middleware credentials match; no missing-secret failures).
- CI: `vdb-tests` and `vdb-tests-full` pass with generated/synced Docker secrets.


EndlessLucky:
@crazywoola Would you please review this PR? Thanks.

web/package.json
```diff
@@ -35,7 +35,7 @@
     "i18n:check": "tsx ./scripts/check-i18n.js",
     "knip": "knip",
     "lint": "eslint --cache --concurrency=auto",
-    "lint:ci": "eslint --cache --cache-strategy content --concurrency 2",
+    "lint:ci": "eslint --concurrency 2",
```
fatelei: why modify this ？

EndlessLucky:
@fatelei I reverted it. Would you please review? Thanks.

EndlessLucky:
@fatelei Would you please review? Thanks.

e2e/support/web-server.ts
```diff
@@ -32,6 +33,18 @@ export const startWebServer = async ({
   reuseExistingServer,
   timeoutMs,
 }: WebServerStartOptions) => {
+  const readWebServerLog = async () => {
```
fatelei, line 36: why add this ?

EndlessLucky: When the web server times out or dies early, the error alone doesn’t tell you what went wrong. 
This pulls the last bit of the log we already write to disk and sticks it on the error so CI output usually shows the real failure (build error, bad env, etc.) without digging through artifacts.

EndlessLucky:
@fatelei @crazywoola Would you please review this PR? Thanks.

kurokobo:
Thanks for working on this, I'm in favor of removing default passwords.

I noticed that there are other hardcoded default keys and passwords in `.env.example`. Was there a reason for limiting the changes to just these five: `SECRET_KEY`, `DB_PASSWORD`, `REDIS_PASSWORD`, `PLUGIN_DIFY_INNER_API_KEY`, and `MINIO_SECRET_KEY`?

Also, I feel that the user instructions could use some improvement.
Users are expected to update the `.env` file themselves, while for CI, you've added a password auto-generation script.
Wouldn't it be more straightforward to create such a script for users to use as well—so it could replace `cp .env.example .env` for them—and have CI run the same script too? What do you think?

EndlessLucky:
@asukaminato0721 Would you please review this PR?

EndlessLucky:
@asukaminato0721 I fixed feedbacks. Would you please review? Thanks.

asukaminato0721:
fix ci

EndlessLucky:
@asukaminato0721 I fixed the ci. Would you please review? Thanks.

EndlessLucky:
@asukaminato0721 I fixed the ci. Would you please review? Thanks.

EndlessLucky:
@lyzno1 Would you please review this PR? Thanks.

EndlessLucky:
@JohnJyong @laipz8200 @QuantumGhost @crazywoola @Yeuoly Would you please review these PRs? Thanks.

EndlessLucky:
@crazywoola Would you please review this PR?

EndlessLucky:
@asukaminato0721 Would you please review this PR? Thanks.

.github/actions/setup-web/action.yml
laipz8200, line 1: This change appears to be unrelated to this PR.

laipz8200 (changes requested):
Removing default values is reasonable, but we must ensure that deployment doesn't become more complex as a result. Also, we should not break any previously deployed instances.

I suggest modifying the program to load configurations according to the following priority:
1. Environment variables
2. .env file
3. Docker named volume
4. Randomly generated values written to a Docker named volume

EndlessLucky:
@laipz8200 Thank you for your revision. I'll fix feedback and push soon.

EndlessLucky:
@laipz8200 Meanwhile, would you please review these PRs?
https://github.com/langgenius/dify/pull/34152
https://github.com/langgenius/dify/pull/34448
https://github.com/langgenius/dify/pull/35085

EndlessLucky:
@laipz8200 I fixed the feedback. Would you please review?

PR #34422, opened 2026-04-01
refactor: replace dict params with BaseModel payloads in TagService

YB0y:
## Summary

Part of #31497 

**Before:** `TagService.save_tags`, `update_tags`, `save_tag_binding`, and `delete_tag_binding` accepted raw `dict` args, with callers either calling `.model_dump()` on already-validated Pydantic models or manually constructing dicts.

**After:** Four new typed payloads (`SaveTagPayload`, `UpdateTagPayload`, `TagBindingCreatePayload`, `TagBindingDeletePayload`) defined in the service layer. All callers in console and service API controllers now construct and pass these models directly, eliminating dict intermediaries and gaining compile-time type safety with Pydantic validation.

### Files changed
- `api/services/tag_service.py` — new payload models + updated method signatures
- `api/controllers/console/tag/tags.py` — pass models instead of `.model_dump()`
- `api/controllers/service_api/dataset/dataset.py` — pass models instead of hand-built dicts

### DIFF
--- a/api/controllers/console/tag/tags.py
+++ b/api/controllers/console/tag/tags.py
@@ -9,7 +9,14 @@ from controllers.common.schema import register_schema_models
 from controllers.console import console_ns
 from controllers.console.wraps import account_initialization_required, edit_permission_required, setup_required
 from libs.login import current_account_with_tenant, login_required
-from services.tag_service import TagService
+from models.enums import TagType
+from services.tag_service import (
+    SaveTagPayload,
+    TagBindingCreatePayload,
+    TagBindingDeletePayload,
+    TagService,
+    UpdateTagPayload,
+)
 
 dataset_tag_fields = {
     "id": fields.String,
@@ -25,19 +32,19 @@ def build_dataset_tag_fields(api_or_ns: Namespace):
 
 class TagBasePayload(BaseModel):
     name: str = Field(description="Tag name", min_length=1, max_length=50)
-    type: Literal["knowledge", "app"] | None = Field(default=None, description="Tag type")
+    type: TagType = Field(description="Tag type")
 
 
 class TagBindingPayload(BaseModel):
     tag_ids: list[str] = Field(description="Tag IDs to bind")
     target_id: str = Field(description="Target ID to bind tags to")
-    type: Literal["knowledge", "app"] | None = Field(default=None, description="Tag type")
+    type: TagType = Field(description="Tag type")
 
 
 class TagBindingRemovePayload(BaseModel):
     tag_id: str = Field(description="Tag ID to remove")
     target_id: str = Field(description="Target ID to unbind tag from")
-    type: Literal["knowledge", "app"] | None = Field(default=None, description="Tag type")
+    type: TagType = Field(description="Tag type")
 
 
 class TagListQueryParam(BaseModel):
@@ -82,7 +89,7 @@ class TagListApi(Resource):
             raise Forbidden()
 
         payload = TagBasePayload.model_validate(console_ns.payload or {})
-        tag = TagService.save_tags(payload.model_dump())
+        tag = TagService.save_tags(SaveTagPayload(name=payload.name, type=payload.type))
 
         response = {"id": tag.id, "name": tag.name, "type": tag.type, "binding_count": 0}
 
@@ -103,7 +110,7 @@ class TagUpdateDeleteApi(Resource):
             raise Forbidden()
 
         payload = TagBasePayload.model_validate(console_ns.payload or {})
-        tag = TagService.update_tags(payload.model_dump(), tag_id)
+        tag = TagService.update_tags(UpdateTagPayload(name=payload.name, type=payload.type), tag_id)
 
         binding_count = TagService.get_tag_binding_count(tag_id)
 
@@ -136,7 +143,9 @@ class TagBindingCreateApi(Resource):
             raise Forbidden()
 
         payload = TagBindingPayload.model_validate(console_ns.payload or {})
-        TagService.save_tag_binding(payload.model_dump())
+        TagService.save_tag_binding(
+            TagBindingCreatePayload(tag_ids=payload.tag_ids, target_id=payload.target_id, type=payload.type)
+        )
 
         return {"result": "success"}, 200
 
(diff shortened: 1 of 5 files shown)


bri-tong:
## Code Review

### Summary

Clean refactoring that replaces raw `dict` parameters with typed Pydantic `BaseModel` payloads in `TagService`. The direction is good — it improves type safety at the service boundary and catches invalid inputs earlier via Pydantic validation. However, **the existing test suites were not updated** and will break at runtime, and there is a **type-narrowing issue** when the console controller passes `None` into the new non-optional `type` field.

---

### Blocking Issues

#### 1. Tests not updated — will fail at runtime

All integration tests in `api/tests/test_containers_integration_tests/services/test_tag_service.py` still pass raw `dict` arguments to the refactored methods:

```python
# Line 691
tag_args = {"name": "test_tag_name", "type": "knowledge"}
result = TagService.save_tags(tag_args)  # Now expects SaveTagPayload, not dict

# Line 759
result = TagService.update_tags(update_args, tag.id)  # Now expects UpdateTagPayload

# Line 992
TagService.save_tag_binding(binding_args)  # Now expects TagBindingCreatePayload

# Line 1117
TagService.delete_tag_binding(delete_args)  # Now expects TagBindingDeletePayload
```

The unit test in `api/tests/unit_tests/controllers/service_api/dataset/test_dataset.py` (line 873) also passes a raw dict:
```python
result = TagService.save_tags({"name": "New Tag", "type": "knowledge"})
```

All of these will raise `TypeError` or `ValidationError` since the methods now expect `BaseModel` instances. **These tests must be updated to construct the new payload models.**

#### 2. `type: None` from console controller will cause Pydantic `ValidationError`

In `api/controllers/console/tag/tags.py`, the controller-level models allow `type` to be `None`:

```python
class TagBasePayload(BaseModel):
    type: Literal["knowledge", "app"] | None = Field(default=None, ...)
```

But the new service-level payloads require it:

```python
class SaveTagPayload(BaseModel):
    type: TagTypeStr  # Literal["knowledge", "app"] — no None allowed
```

When a console API request omits `type`, `payload.type` is `None`, and constructing `SaveTagPayload(name=payload.name, type=None)` will raise a Pydantic `ValidationError`. This applies to all four console controller call sites (`SaveTagPayload`, `UpdateTagPayload`, `TagBindingCreatePayload`, `TagBindingDeletePayload`).

Note: the *old* code would also fail later at `TagType(None)`, so this isn't a net-new failure mode. But since this PR is adding explicit validation, consider also fixing the controller-level models to make `type` required:
```python
class TagBasePayload(BaseModel):
    name: str = Field(..., min_length=1, max_length=50)
    type: Literal["knowledge", "app"]  # Remove | None
```

---

### Non-blocking Suggestions

#### 3. `SaveTagPayload` and `UpdateTagPayload` are structurally identical

Both have exactly `name: str` + `type: TagTypeStr`. Consider a shared base to reduce duplication:

```python
class _TagPayloadBase(BaseModel):
    name: str = Field(min_length=1, max_length=50)
    type: TagTypeStr

class SaveTagPayload(_TagPayloadBase): pass
class UpdateTagPayload(_TagPayloadBase): pass
```

This preserves separate types for independent evolution while eliminating the duplicated field definitions.

#### 4. Model placement is inconsistent with codebase convention

Other services (e.g., `services/entities/knowledge_entities/`) place payload/entity models in a separate `entities/` directory. Defining them inline in `tag_service.py` works but diverges from the existing pattern. Consider moving to `services/entities/tag_entities.py` for consistency, especially if more tag-related models are added later.

#### 5. Minor: variable rename in `delete_tag_binding` is a nice bonus cleanup

The rename from `tag_bindings` → `tag_binding` (singular) correctly reflects that `.first()` returns a single result. 👍

---

### Questions / Clarifications

- Is the `type: None` default in the console controller models (`TagBasePayload`, `TagBindingPayload`, `TagBindingRemovePayload`) intentionally allowed by the frontend, or is it effectively dead code? If it's dead code, cleaning it up as part of this PR would make the type contracts fully consistent across layers.

YB0y:
Hi @asukaminato0721, is this PR not ready to be merged? Could you please provide any feedback?

api/controllers/console/tag/tags.py
```diff
@@ -25,19 +32,19 @@ def build_dataset_tag_fields(api_or_ns: Namespace):
 
 class TagBasePayload(BaseModel):
     name: str = Field(description="Tag name", min_length=1, max_length=50)
-    type: Literal["knowledge", "app"] | None = Field(default=None, description="Tag type")
+    type: TagType = Field(description="Tag type")
```
asukaminato0721, line 35: @dosu now here requires a type field, can you tell me what is the ideal default value?

api/controllers/console/tag/tags.py
```diff
@@ -25,19 +32,19 @@ def build_dataset_tag_fields(api_or_ns: Namespace):
...
 
 
 class TagBindingPayload(BaseModel):
     tag_ids: list[str] = Field(description="Tag IDs to bind")
     target_id: str = Field(description="Target ID to bind tags to")
-    type: Literal["knowledge", "app"] | None = Field(default=None, description="Tag type")
+    type: TagType = Field(description="Tag type")
```
asukaminato0721, line 41: @dosu now here requires a type field, can you tell me what is the ideal default value?

api/controllers/console/tag/tags.py
```diff
@@ -25,19 +32,19 @@ def build_dataset_tag_fields(api_or_ns: Namespace):
...
 
 
 class TagBindingRemovePayload(BaseModel):
     tag_id: str = Field(description="Tag ID to remove")
     target_id: str = Field(description="Target ID to unbind tag from")
-    type: Literal["knowledge", "app"] | None = Field(default=None, description="Tag type")
+    type: TagType = Field(description="Tag type")
```
asukaminato0721, line 47: @dosu now here requires a type field, can you tell me what is the ideal default value?

YB0y:
Hi @asukaminato0721, sorry for the ping. @dosu mentioned that my approach looks good. could you share your thoughts as well?

asukaminato0721 (approved):

Issue #25193, opened 2025-09-05
[Chore/Refactor] 'use' is preferred over 'useContext' because it is more flexible


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

use use

### Motivation

'use' is preferred over 'useContext' because it is more flexible 

### Additional Context

https://react.dev/reference/react/use#reading-context-with-use

atif09:
Ill change it

asukaminato0721:
@dosu how to fix it

majiayu000:
I’d like to take a small non-overlapping batch for this issue.

Scope of this batch:
- replace React native  with  in files that are currently **not** covered by the open PR #25548
- keep behavior unchanged and avoid touching  usage
- run web lint/type-check and related unit tests before opening the PR

This keeps risk low and should help move the issue forward incrementally.

majiayu000:
Correction for formatting issue in my previous comment.

I’d like to take a small non-overlapping batch for this issue.

Scope of this batch:
- replace React native useContext with use in files that are currently not covered by the open PR #25548
- keep behavior unchanged and avoid touching use-context-selector usage
- run web lint/type-check and related unit tests before opening the PR

This keeps risk low and should help move the issue forward incrementally.

agenthaulk:
I'd like to work on this issue. I'll replace React native useContext with the use hook in the remaining uncovered files and submit a PR within 24 hours.

PR #34450, opened 2026-04-02
refactor: replace useContext with use in selected batch

agenthaulk:
## Summary
This PR updates a small non-overlapping batch of remaining React native `useContext` usages to React 19 `use(Context)` for #25193.

## Changes
- replace selected native `useContext` calls with `use`
- keep behavior unchanged
- leave `use-context-selector` usages untouched
- update related tests accordingly

## Validation
- passed 3 focused test files
- passed 10 tests

### DIFF
--- a/web/app/components/base/features/__tests__/context.spec.tsx
+++ b/web/app/components/base/features/__tests__/context.spec.tsx
@@ -1,10 +1,10 @@
 import { render, screen } from '@testing-library/react'
 import * as React from 'react'
-import { useContext } from 'react'
+import { use } from 'react'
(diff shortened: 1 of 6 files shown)


CodingOnStar:
The replacement itself looks fine and I don't see a blocking issue. But “remaining batch” feels misleading because native react useContext still exists in nearby wrapper hooks. Could we either rename this to a selected batch or migrate the adjacent wrappers in the same area as well?

agenthaulk:
Updated the PR title to `selected batch` so the scope matches the actual changes more accurately. The implementation itself stays the same.

CodingOnStar (approved):

Issue #24645, opened 2025-08-27
[Chore/Refactor] remove redundant-cast


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

Many are leaving redundant-cast.

how to do it:

go to a check log like https://github.com/langgenius/dify/actions/runs/17258598012/job/48975128610 (you can choose any related log)

search for `is the same type as ` in pyrefly check, and fix them.

example #24646 

### Motivation

Clean code.

### Additional Context

https://pyrefly.org/en/docs/error-kinds/#redundant-cast

Gnomeek:
I can do this.

asukaminato0721:
```
 WARN Redundant cast: `int` is the same type as `int` [redundant-cast]
  --> core/ops/tencent_trace/utils.py:26:20
   |
26 |         return cast(int, uuid_obj.int)
   |                    -------------------
   |
 WARN Redundant cast: `int` is the same type as `int` [redundant-cast]
  --> core/ops/tencent_trace/utils.py:55:82
   |
55 |             trace_id = int(trace_id_str, 16) if len(trace_id_str) == 32 else cast(int, uuid.UUID(trace_id_str).int)
   |                                                                                  ----------------------------------
```

asukaminato0721:
@dosu what do you think

majiayu000:
I’m picking up a small follow-up batch for this issue: remove remaining safe `redundant-cast` warnings (starting with `api/core/ops/tencent_trace/utils.py`) and run pyrefly + type checks before opening a PR.

majiayu000:
[...]

PR #34461, opened 2026-04-02
fix: remove redundant cast in MCP base session

majiayu000:
## Summary
- Remove two `redundant-cast` warnings in `api/core/mcp/session/base_session.py` flagged by pyrefly
- `model_validate()` already returns the correct generic type (`ReceiveRequestT` / `ReceiveNotificationT`), so the `cast()` calls are no-ops
- Removed the now-unused `cast` import from `typing`
- Added targeted `# type: ignore[arg-type]` for mypy, which cannot narrow constrained TypeVars from `model_validate()` return types (pyrefly and basedpyright handle this correctly)

Closes part of #24645

cc @asukaminato0721


asukaminato0721:
mypy disagree with pyright... Well

asukaminato0721 (approved):

Issue #34504, opened 2026-04-03
Plugin API key modal is hidden behind the settings panel


Innocent-children:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

latest

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

---

Title: Plugin API key modal is hidden behind the settings panel

Body:

```markdown
### Summary

When configuring plugin credentials (e.g. MinIO S3) from the settings page, the API key modal opens behind the settings panel and is not interactable.

### Steps to reproduce

1. Install a plugin that requires API key configuration (e.g. MinIO S3)
2. Navigate to Settings → Data Source (`/apps?action=showSettings&tab=data-source`)
3. Click "Configure" / "Add credentials" on the plugin

### Expected behavior

The API key modal appears on top of the settings panel.

### Actual behavior

The modal opens behind the settings panel. The user cannot see or interact with the credential form.

### Root cause

`ApiKeyModal` (`web/app/components/plugins/plugin-auth/authorize/api-key-modal.tsx`) uses `wrapperClassName="z-1002!"`, which overrides the legacy `Modal` component's default `z-9998` down to `z-1002`. The settings panel (`MenuDialog`) uses the new `Dialog` component whose backdrop and popup are also at `z-1002`. Since both share the same z-index, DOM order determines stacking, and the settings dialog covers the modal.

The same issue exists in `OAuthClientSettings` (`oauth-client-settings.tsx`).

### Fix

Remove the `z-1002!` override from both components so they use the `Modal` default `z-9998`:

```diff
- wrapperClassName="z-1002!"
+ wrapperClassName=""


### ✔️ Expected Behavior

all settings are run well

### ❌ Actual Behavior

_No response_

lyzno1:
[...]

PR #34505, opened 2026-04-03
fix: plugin credential modal hidden behind settings panel

Innocent-children:
---

Title: `fix: plugin credential modal hidden behind settings panel`

Body:

```markdown
## Summary

When configuring plugin credentials from the settings page (e.g. Data Source tab), the API key modal opens behind the settings panel and is not interactable.

## Root cause

`ApiKeyModal` and `OAuthClientSettings` both set `wrapperClassName="z-1002!"`, which overrides the legacy `Modal` component's default `z-9998` down to `z-1002` — the same z-index used by the settings panel's `Dialog`. DOM order then causes the settings panel to stack on top of the modal.

## Changes

Remove the `z-1002!` override from:
- `web/app/components/plugins/plugin-auth/authorize/api-key-modal.tsx`
- `web/app/components/plugins/plugin-auth/authorize/oauth-client-settings.tsx`

This restores the default `z-9998`, which correctly stacks above the settings dialog's `z-1002`.

## Test

1. Install any plugin requiring API key (e.g. MinIO S3)
2. Go to Settings → Data Source
3. Click configure / add credentials
4. Modal now appears on top of the settings panel

### DIFF
--- a/web/app/components/plugins/plugin-auth/authorize/api-key-modal.tsx
+++ b/web/app/components/plugins/plugin-auth/authorize/api-key-modal.tsx
@@ -138,7 +138,7 @@ const ApiKeyModal = ({
       onExtraButtonClick={onRemove}
       disabled={disabled || isLoading || doingAction}
       clickOutsideNotClose={true}
-      wrapperClassName="z-1002!"
+      wrapperClassName=""
     >
       {pluginPayload.detail && (
         <ReadmeEntrance pluginDetail={pluginPayload.detail} showType={ReadmeShowType.modal} />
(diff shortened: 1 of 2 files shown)


lyzno1:
This method cannot be used to modify components. Please follow the instructions in `web/overlay-migration.md`.
In addition, please create an issue first for easy tracking :)

Innocent-children:
> This method cannot be used to modify components. Please follow the instructions in `web/overlay-migration.md`. In addition, please create an issue first for easy tracking :)

In fact, there is an issue here https://github.com/langgenius/dify/issues/34504 I am confident that my changes have taken effect; I successfully built and tested the package locally on my MacBook.

PR #34524, opened 2026-04-03
Fixes #34520 [Bug] License expiring banner not shown after login until manual page refresh

Kilros0817:
### Summary

This PR fixes an issue where the license expiring banner did not appear immediately after login until the user manually refreshed the page.

The root cause was that the expired_at field in system features was only available for **authenticated requests**, while the frontend cached system features before login. This caused the dashboard to use stale data without expired_at, preventing the banner from showing.

**The fix (Option A)**: Move expired_at outside the is_authenticated guard so it is available even for unauthenticated calls. The status field remains gated because it contains sensitive information, but expired_at is safe to expose.

This ensures that the banner is displayed immediately after login without requiring a manual page refresh.

Fixes #34520

### DIFF
--- a/api/services/feature_service.py
+++ b/api/services/feature_service.py
@@ -379,15 +379,15 @@ class FeatureService:
             )
             features.webapp_auth.sso_config.protocol = enterprise_info.get("SSOEnforcedForWebProtocol", "")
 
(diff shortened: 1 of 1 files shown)


Kilros0817:
@GarfieldDai
Could you please review this PR? 
PR about issue #34520
Thanks!

Issue #32863, opened 2026-03-02
[Refactor/Chore] migrate to TypedDict


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

now codebase is full of `: dict,` `-> dict` `-> Mapping` `: Mapping`

we should use TypedDict whenever possible.

@dosu give me a list of file need to change.

and show how to change.

asukaminato0721:
@dosu give me a full list

asukaminato0721:
@dosu tell me how to update them one by one.

statxc:
Hi @asukaminato0721 
I want to contribute to this repo. Can I work on this issue?

statxc:
@asukaminato0721 thanks for assigning to me.
there will be huge code refactoring to solve this issue.
do you want me to raise one PR for all changes or several PRs for each small group of changes?

asukaminato0721:
you can do them one by one. And make many small prs at a time. Since this is a huge task.

atipre:
@asukaminato0721 i would also like to contribute, since there are so many can I also work on this one?

statxc:
sorry @atipre i want to work on the entire refactoring and already made progress across several sections.
pls find another issue.

statxc:
Hi @asukaminato0721 
I raised a PR for migrating models scope. Pls review.
The PRs for other scope were almost prepared.
I will raise them after you approve and merge this one.
Thanks for your time.

asukaminato0721:
check the type errors

So I suggest one file by one file, or even one Dict by one Dict

statxc:
@atipre I found this issue really need a lot changes.
Pls contribute if you still want to do. Let's do this together.
Sorry for inconvenience.

atipre:
@statxc Sure no problem, do you have an idea of how you wanna split it up?

[...]

PR #34531, opened 2026-04-03
refactor(api): type workflow run delete/count results with RunsWithRelatedCountsDict TypedDict

YB0y:
## Summary
- Add `RunsWithRelatedCountsDict` TypedDict with keys `runs`, `node_executions`, `offloads`, `app_logs`, `trigger_logs`, `pauses`, `pause_reasons`
- Annotate return types of `delete_runs_with_related` and `count_runs_with_related` in both protocol and implementation

## Why this change
Both methods return a 7-key integer dict representing counts of deleted/counted workflow runs and their related records. The dict shape is identical across the abstract protocol and its SQLAlchemy implementation, and callers access keys by name. A TypedDict makes the contract explicit at the protocol level.

## Changes
- `repositories/api_workflow_run_repository.py`: Define `RunsWithRelatedCountsDict`, update 2 protocol method return types
- `repositories/sqlalchemy_api_workflow_run_repository.py`: Import TypedDict, update 2 implementation method return types

## Test plan
- [x] `ruff check` passes

Part of #32863 (`repositories/api_workflow_run_repository.py`, `repositories/sqlalchemy_api_workflow_run_repository.py`)

### DIFF
--- a/api/services/retention/workflow_run/clear_free_plan_expired_workflow_run_logs.py
+++ b/api/services/retention/workflow_run/clear_free_plan_expired_workflow_run_logs.py
@@ -173,6 +173,9 @@ class WorkflowRunCleanupMetrics:
         self._record(self._job_duration_seconds, job_duration_seconds, attributes)
 
 
+_RELATED_RECORD_KEYS = ("node_executions", "offloads", "app_logs", "trigger_logs", "pauses", "pause_reasons")
+
+
 class WorkflowRunCleanup:
     def __init__(
         self,
@@ -312,8 +315,8 @@ class WorkflowRunCleanup:
                         int((time.monotonic() - count_start) * 1000),
                     )
                     if related_totals is not None:
-                        for key in related_totals:
-                            related_totals[key] += batch_counts.get(key, 0)
+                        for k in _RELATED_RECORD_KEYS:
+                            related_totals[k] += batch_counts.get(k, 0)  # type: ignore[literal-required,operator]
                     sample_ids = ", ".join(run.id for run in free_runs[:5])
                     click.echo(
                         click.style(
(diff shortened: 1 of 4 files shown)


api/services/retention/workflow_run/delete_archived_workflow_run.py
```diff
@@ -23,7 +23,7 @@ class DeleteResult:
     run_id: str
     tenant_id: str
     success: bool
-    deleted_counts: dict[str, int] = field(default_factory=dict)
+    deleted_counts: RunsWithRelatedCountsDict | dict[str, int] = field(default_factory=dict)
```
asukaminato0721: can we just use RunsWithRelatedCountsDict here?

api/services/retention/workflow_run/clear_free_plan_expired_workflow_run_logs.py
```diff
@@ -340,10 +343,13 @@ def run(self) -> None:
 
                 try:
                     delete_start = time.monotonic()
-                    counts = self.workflow_run_repo.delete_runs_with_related(
-                        free_runs,
-                        delete_node_executions=self._delete_node_executions,
-                        delete_trigger_logs=self._delete_trigger_logs,
+                    counts = cast(
```
asukaminato0721: can we remove cast here?

api/services/retention/workflow_run/clear_free_plan_expired_workflow_run_logs.py
```diff
@@ -301,10 +301,13 @@ def run(self) -> None:
 
                 if self.dry_run:
                     count_start = time.monotonic()
-                    batch_counts = self.workflow_run_repo.count_runs_with_related(
-                        free_runs,
-                        count_node_executions=self._count_node_executions,
-                        count_trigger_logs=self._count_trigger_logs,
+                    batch_counts = cast(
```
asukaminato0721: same

asukaminato0721 (changes requested):

asukaminato0721 (approved):

PR #34537, opened 2026-04-03
ci: detect icon package changes for web checks

suyua9:
## Summary
- include the packages/iconify-collections path in the main CI filters for web checks
- trigger both web and e2e jobs when the shared icon workspace package changes
- keep the scope narrow to the package currently consumed by the web workspace

## Testing
- git diff --check

Fixes #34533


suyua9:
Closing this from my side as superseded by the current main branch. The workflow path filter now includes packages/** for web/e2e checks, which covers packages/iconify-collections/** and is broader than this older targeted path-filter patch.

Issue #22668, opened 2025-07-20
use select style api in orm


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] Please do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing? Tell me about your story.

now the using of `db.session.query` gives very terrible typing result.

use new select api will make it better.

### 2. Additional context or comments

https://docs.sqlalchemy.org/en/20/orm/queryguide/select.html

https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-orm-usage

### 3. Can you help us with this feature?

- [ ] I am interested in contributing to this feature.

asukaminato0721:
@dosu how to do it, give me code example

ai-mountain:
@asukaminato0721 I am interested in this issue. May I contribute?

ai-mountain:
Hi, @asukaminato0721 I am going to open several PRs to close this issue. Is it OK with you?

asukaminato0721:
Yes, and keep them small. Your can open many prs.

asukaminato0721:
@dosu give me a file list

Helios531:
Hi. @asukaminato0721 
I am new here and I noticed this issue is reopened.
Can I work on this issue?

asukaminato0721:
> Hi. [@asukaminato0721](https://github.com/asukaminato0721) I am new here and I noticed this issue is reopened. Can I work on this issue?

Go ahead

asukaminato0721:
@dosu what is the estimate remaining time for this issue.

wdeveloper16:
@dosu Please list me remaining tasks for this issue.

asukaminato0721:
@dosu Please list me remaining tasks for this issue

asukaminato0721:
@dosu Please list me remaining tasks for this issue

asukaminato0721:
@dosu Please list me remaining tasks for this issue

PR #34548, opened 2026-04-04
refactor: select in datasource_provider_service

RenzoMXD:
## Summary
- Migrate all 6 `db.session.query()` calls to SQLAlchemy 2.x `select()` style in `datasource_provider_service.py`
- Replace `.filter_by().count()` with `db.session.scalar(select(func.count()).where())`
- Replace `.filter_by().first()` with `db.session.scalar(select().where().limit(1))`
- Replace `.where().all()` with `list(db.session.scalars(select().where()).all())`
- Add default `scalar`/`scalars` return values to test fixture for compatibility
- Update 10 test mocks for affected methods only; other tests unchanged

## Test plan
- [x] All 63 unit tests pass for datasource_provider_service
- [x] Basedpyright type check passes (0 errors)

Part of #22668

### DIFF
--- a/api/services/datasource_provider_service.py
+++ b/api/services/datasource_provider_service.py
@@ -4,6 +4,7 @@ from collections.abc import Mapping
 from typing import Any
 
 from graphon.model_runtime.entities.provider_entities import FormType
+from sqlalchemy import func, select
 from sqlalchemy.orm import Session
 
 from configs import dify_config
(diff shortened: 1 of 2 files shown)


asukaminato0721 (approved):

ai-mountain:
@asukaminato0721 
I think this is the end for the @22668.
Pleas let me know if there is another issue for help.
Thanks.

PR #34553, opened 2026-04-05
fix(api): add missing error handling in Notion table rows extractor

lawrence3699:
I noticed that `_read_table_rows` in `notion_extractor.py` is the only method in the Notion extractor that doesn't validate the HTTP response before processing. Its sibling methods `_get_notion_block_data` (lines 171-188) and `_read_block` (line 245) both check the status code and validate the response structure, but `_read_table_rows` goes straight to `data["results"][0]` without any checks.

This crashes in two situations:

1. **Notion API returns an error** (401 expired token, 429 rate limit, 500 server error): The response body has no `"results"` key, so `data["results"]` raises `KeyError`.

2. **Empty table results**: If the results array is empty, `data["results"][0]` raises `IndexError`.

Both are easy to reproduce:
```python
# Error response (expired token):
data = {"object": "error", "status": 401, "code": "unauthorized"}
data["results"][0]  # → KeyError: 'results'

# Empty results:
data = {"results": [], "next_cursor": None}
data["results"][0]  # → IndexError: list index out of range
```

The fix adds the same validation pattern already used by `_get_notion_block_data`:
- HTTP status code check with `ValueError` on non-200
- `httpx.HTTPError` catch for network failures
- `"results"` key and empty-list check before indexing

**Tests added:** Four test cases covering HTTP error, network error, empty results, and error-in-200-body scenarios.

### DIFF
--- a/api/core/rag/extractor/notion_extractor.py
+++ b/api/core/rag/extractor/notion_extractor.py
@@ -291,17 +291,24 @@ class NotionExtractor(BaseExtractor):
         while not done:
             query_dict: dict[str, Any] = {} if not start_cursor else {"start_cursor": start_cursor}
 
-            res = httpx.request(
-                "GET",
-                block_url,
-                headers={
-                    "Authorization": "Bearer " + self._notion_access_token,
-                    "Content-Type": "application/json",
-                    "Notion-Version": "2022-06-28",
-                },
-                params=query_dict,
-            )
-            data = res.json()
+            try:
+                res = httpx.request(
+                    "GET",
+                    block_url,
+                    headers={
+                        "Authorization": "Bearer " + self._notion_access_token,
+                        "Content-Type": "application/json",
+                        "Notion-Version": "2022-06-28",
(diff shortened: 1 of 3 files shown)


lawrence3699:
Pushed a fix for the CI failures in `test_notion_provider.py`. The three existing `_read_table_rows` tests used `Mock(json=lambda: data)` without setting `status_code`, so the new HTTP status check was triggering on the mock objects. Updated the mocks to include `status_code = 200`.

The other failing tests in the CI run (`test_pause_resume`, `test_knowledge_retrieval_node`, etc.) are unrelated to this PR — they appear to be pre-existing failures on the base branch.

lawrence3699:
Updated the test fixes — found one more mock in `TestNotionExtractorTableAdvanced::test_read_table_rows_with_many_columns` that was also missing `status_code = 200`. All 4 failing tests should now be resolved. The new CI runs are waiting for maintainer authorization to execute.

PR #33834, opened 2026-03-20
refactor(web): replace MediaType enum with const object

Daedalus-Icarus:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

- Replace the `MediaType` string `enum` in `web/hooks/use-breakpoints.ts` with an `as const` object plus an exported union type alias (`MediaTypeValue`), avoiding non-erasable enum emit and matching the direction in #27998 / the `erasable-syntax-only` setup from #31487.
- Remove the corresponding `erasable-syntax-only/enums` suppression for `hooks/use-breakpoints.ts` in `web/eslint-suppressions.json`.
- Call sites keep using `MediaType.mobile` / `.tablet` / `.pc`; no UI or behavior change intended.

**Relates to #27998** 

## Screenshots

| Before | After |
|--------|-------|

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [x] I ran `make lint` and `make type-check` (backend) and `cd web && npx lint-staged` (frontend) to appease the lint gods

PR #34565, opened 2026-04-05
refactor(web): convert file-local Step enum to as-const in website crawlers 🤖🤖🤖

jeanibarz:
## Summary

Converts the file-local `Step` enum to an `as const` object with a companion type alias in three website crawler components (`firecrawl`, `jina-reader`, `watercrawl`). All three files declared an identical 3-value `Step` enum (`init` / `running` / `finished`) used for local crawl-state tracking.

Partially addresses #27998.

## Why this is safe

- `Step` is **file-local** in each component — never exported, never imported cross-file.
- All call sites use value lookup (`Step.init`, `Step.running`, `Step.finished`) or type position (`useState<Step>`), so the `as const` object produces identical runtime values and compile-time types.
- No runtime reflection (`Object.values`/`Object.keys`/iteration), so none of the special-case patterns from dosubot's comment on #27998 apply.

## Pattern

Matches the convention from the merged batches (e.g. #33960, #33834):

```ts
const Step = {
  init: 'init',
  running: 'running',
  finished: 'finished',
} as const
type Step = typeof Step[keyof typeof Step]
```

## Checklist

- [x] Ran `npx eslint` on the three touched files — 0 errors
- [x] Ran `npx tsc --noEmit -p tsconfig.json` — 0 errors
- [x] Ran `npx knip` — 0 errors
- [x] Removed the three `erasable-syntax-only/enums` suppressions from `web/eslint-suppressions.json` (other suppressions for these files left intact)

### DIFF
--- a/eslint-suppressions.json
+++ b/eslint-suppressions.json
@@ -2449,9 +2449,6 @@
     }
   },
   "web/app/components/datasets/create/website/firecrawl/index.tsx": {
-    "erasable-syntax-only/enums": {
-      "count": 1
-    },
     "no-console": {
       "count": 1
     },
@@ -2468,9 +2465,6 @@
     }
   },
   "web/app/components/datasets/create/website/jina-reader/index.tsx": {
-    "erasable-syntax-only/enums": {
-      "count": 1
-    },
     "no-console": {
       "count": 1
     },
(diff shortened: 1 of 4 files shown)


WTW0313 (approved):
LGTM

PR #34566, opened 2026-04-05
feat: redis add retry logic

### LINKED ISSUE #34557: Enhance Redis HA with client retry and connection health settings
### Self Checks

I am looking at Redis high availability behavior in the backend and found that the main Redis clients do not appear to be configured to retry failed calls during transient network issues or Redis failover windows.

The current implementation in `api/extensions/ext_redis.py` builds the main clients through:

- `redis.ConnectionPool(**redis_params)` for standalone Redis
- `sentinel.master_for(...)` for Sentinel
- `redis.Redis.from_url(...)` / `RedisCluster.from_url(...)` for pub/sub

However, the shared Redis parameters currently only include auth/db/encoding/cache settings and do not pass retry-oriented options such as:

- `retry`
- `retry_on_timeout`
- `retry_on_error`
- `socket_timeout`
- `socket_connect_timeout`
- `health_check_interval`

Since Dify is using `redis-py 7.3.0`. I verified that:

- clients created from `redis.ConnectionPool(...)` use connections with `retry._retries == 0`
- Sentinel-managed connections also end up with `retry._retries == 0`
- `Redis.from_url(...)` also results in connections with `retry._retries == 0`

So even though Sentinel/Cluster may help with topology discovery, individual failed commands are still surfaced immediately instead of being retried by the Redis client. This weakens HA behavior in practice, especially during:

- master failover
- brief network blips
- half-open or stale socket reuse

There is already one local example in `api/schedule/queue_monitor_task.py` that sets `socket_timeout`, `socket_connect_timeout`, and `health_check_interval`, which suggests this problem is already recognized in one isolated path but not applied consistently to the main backend Redis clients.

I would like to propose enhancing Redis HA by adding a shared retry/backoff and connection health policy for all backend Redis client construction paths in `api/extensions/ext_redis.py`.

Suggested direction:

- define one shared retry policy in `api/extensions/ext_redis.py`, for example with `redis.retry.Retry` plus exponential backoff
- pass that policy into standalone, Sentinel, Cluster, and pub/sub client creation
- add conservative `socket_timeout`, `socket_connect_timeout`, and `health_check_interval` defaults
- optionally expose the retry and timeout knobs in config so operators can tune them for different deployments

This would make Redis behavior more resilient without forcing every call site to implement its own retry logic.

It would also complement the existing `redis_fallback()` decorator, which is useful for best-effort paths but does not replace transport-level retry for transient failures.

fatelei:
> [!IMPORTANT]
>

## Summary

fix #34557

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


api/configs/middleware/cache/redis_config.py
```diff
@@ -117,6 +117,37 @@ class RedisConfig(BaseSettings):
         default=None,
     )
 
+    REDIS_RETRY_RETRIES: NonNegativeInt = Field(
+        description="Maximum number of retries per Redis command on "
+        "transient failures (ConnectionError, TimeoutError, socket.timeout)",
+        default=3,
+    )
```
QuantumGhost, line 124: I have examined redis-py's documentation and found that the retry logic does not distinguish between readonly commands and commands with side effects. This means that a non-idempotent command  (E.G. `INCR`)  could be executed multiple times when the command is processed but the response is not sent back to API.

I would suggest to disable retry by default and let the deployer to decided whether enable the retry feature.

api/.env.example
```diff
@@ -71,6 +71,12 @@ REDIS_USE_CLUSTERS=false
 REDIS_CLUSTERS=
 REDIS_CLUSTERS_PASSWORD=
 
+REDIS_RETRY_BACKOFF_BASE=1.0
```
QuantumGhost, line 75: REDIS_RETRY_RETRIES is missing here.

docker/.env.example
```diff
@@ -373,6 +373,18 @@ REDIS_USE_CLUSTERS=false
 REDIS_CLUSTERS=
 REDIS_CLUSTERS_PASSWORD=
 
+# Redis connection and retry configuration
```
QuantumGhost, line 376: ditto

api/configs/middleware/cache/redis_config.py
QuantumGhost: It seems that redis-py already set a default retry policy with a max 3 retries. IMO there's no need to adjust the default.

QuantumGhost (dismissed):

QuantumGhost (approved):

PR #34610, opened 2026-04-06
refactor(api): type document summary status detail with TypedDict

YB0y:
## Summary
- Add `SummaryEntryDict` TypedDict with keys `segment_id`, `segment_position`, `status`, `summary_preview`, `error`, `created_at`, `updated_at`
- Add `DocumentSummaryStatusDetailDict` TypedDict with keys `total_segments`, `summary_status`, `summaries`
- Annotate return type of `get_document_summary_status_detail` (was `dict[str, Any]`)
- Remove unused `Any` import

## Why this change
`get_document_summary_status_detail` constructs a dict with 3 fixed top-level keys and a list of summary entries with 7 fixed keys each, but was typed as `dict[str, Any]`. The TypedDicts make both the response shape and the per-entry shape explicit.

## Changes
- `services/summary_index_service.py`: Define 2 TypedDicts, update method return type

## Test plan
- [x] `ruff check` passes

Part of #32863 (`services/summary_index_service.py`)

### DIFF
--- a/api/services/summary_index_service.py
+++ b/api/services/summary_index_service.py
@@ -4,7 +4,7 @@ import logging
 import time
 import uuid
 from datetime import UTC, datetime
-from typing import Any
+from typing import TypedDict, cast
 
 from graphon.model_runtime.entities.llm_entities import LLMUsage
 from graphon.model_runtime.entities.model_entities import ModelType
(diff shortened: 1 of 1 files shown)


api/services/summary_index_service.py
```diff
@@ -1438,8 +1454,8 @@ def get_document_summary_status_detail(
...
                 )
 
-        return {
-            "total_segments": total_segments,
-            "summary_status": status_counts,
-            "summaries": summary_list,
-        }
+        return DocumentSummaryStatusDetailDict(
+            total_segments=total_segments,
+            summary_status=cast(dict[str, int], status_counts),
```
asukaminato0721, line 1459: this cast can be removed in following pr

asukaminato0721 (approved):

PR #31879, opened 2026-02-03
fix: refresh tool lists after custom and workflow provider changes

jiurenm:
## Summary

Fixes [#31372](https://github.com/langgenius/dify/issues/31372)

refresh tool lists after custom and workflow provider changes

## Screenshots

N/A

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `make lint` and `make type-check` (backend) and `cd web && npx lint-staged` (frontend) to appease the lint gods

PR #34629, opened 2026-04-07
fix: invalidate MCP tool cache on provider update to refresh descriptions

BrianWang1990:
## Summary

Fixes #31372

When updating an MCP provider configuration through the Studio UI (e.g., changing the server URL via MCPModal), the per-provider tool list cache (`useMCPTools(providerID)`) was not invalidated. This caused tool descriptions to remain stale even after the provider config was successfully updated.

## Root Cause

In `web/app/components/tools/mcp/detail/content.tsx`, the `handleUpdate` callback calls `onUpdate()` after a successful `updateMCP`, which refreshes the provider list via `useAllToolProviders`. However, it did **not** invalidate the individual provider's tool cache (`useMCPTools`), so the cached tool names and descriptions persisted.

The sibling `handleUpdateTools` callback already correctly calls `invalidateMCPTools(detail.id)` after updating tools — `handleUpdate` simply needed the same invalidation.

## Fix

Added `invalidateMCPTools(detail.id)` in the `handleUpdate` callback (and its dependency array) to force tool descriptions to refresh when the provider configuration changes.

This is a minimal one-line change with no side effects — `invalidateMCPTools` was already imported and used elsewhere in the same component.

### DIFF
--- a/web/app/components/tools/mcp/detail/content.tsx
+++ b/web/app/components/tools/mcp/detail/content.tsx
@@ -130,9 +130,10 @@ const MCPDetailContent: FC<Props> = ({
     if ((res as any)?.result === 'success') {
       hideUpdateModal()
       onUpdate()
+      invalidateMCPTools(detail.id)
       handleAuthorize()
     }
-  }, [detail, updateMCP, hideUpdateModal, onUpdate, handleAuthorize])
+  }, [detail, updateMCP, hideUpdateModal, onUpdate, invalidateMCPTools, handleAuthorize])
 
   const handleDelete = useCallback(async () => {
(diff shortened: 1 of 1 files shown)


crazywoola:
Hi @BrianWang1990, thanks for opening this pull request.

### Why this is being closed

Issue #31372 has multiple open pull requests. Under the duplicate-PR rule, pull requests from the issue author are ranked first, then creation time is used as the FIFO tie-breaker. PR #31879 is the retained pull request, so this one is being closed.

### Next steps

Please coordinate on PR #31879 if you would like to contribute further. Thank you for understanding.

Issue #34636, opened 2026-04-07
ERROR Argument `SimpleNamespace` is not assignable to parameter


asukaminato0721:
```diff
+   --> tests/test_containers_integration_tests/services/test_app_dsl_service.py:138:42
+ERROR Argument `SimpleNamespace` is not assignable to parameter `app_model` with type `App` in function `services.app_dsl_service.AppDslService.check_dependencies` [bad-argument-type]
+   --> tests/test_containers_integration_tests/services/test_app_dsl_service.py:579:55
+ERROR Argument `SimpleNamespace` is not assignable to parameter `app_model` with type `App` in function `services.app_dsl_service.AppDslService.check_dependencies` [bad-argument-type]
+   --> tests/test_containers_integration_tests/services/test_app_dsl_service.py:607:55
+ERROR Argument `SimpleNamespace` is not assignable to parameter `app` with type `App | None` in function `services.app_dsl_service.AppDslService._create_or_update_app` [bad-argument-type]
+   --> tests/test_containers_integration_tests/services/test_app_dsl_service.py:664:17
+ERROR Argument `SimpleNamespace` is not assignable to parameter `app` with type `App | None` in function `services.app_dsl_service.AppDslService._create_or_update_app` [bad-argument-type]
+   --> tests/test_containers_integration_tests/services/test_app_dsl_service.py:738:21
+ERROR Argument `SimpleNamespace` is not assignable to parameter `app` with type `App | None` in function `services.app_dsl_service.AppDslService._create_or_update_app` [bad-argument-type]
+   --> tests/test_containers_integration_tests/services/test_app_dsl_service.py:755:21
+ERROR Attribute `app_model_config` of class `App` is a read-only property and cannot be set [read-only]
+   --> tests/test_containers_integration_tests/services/test_app_dsl_service.py:772:9
+ERROR Argument `SimpleNamespace` is not assignable to parameter `app` with type `App | None` in function `services.app_dsl_service.AppDslService._create_or_update_app` [bad-argument-type]
+   --> tests/test_containers_integration_tests/services/test_app_dsl_service.py:793:21
+ERROR Argument `SimpleNamespace` is not assignable to parameter `app_model` with type `App` in function `services.app_dsl_service.AppDslService.export_dsl` [bad-argument-type]
+   --> tests/test_containers_integration_tests/services/test_app_dsl_service.py:833:34
[...]

PR #34659, opened 2026-04-07
fix(#34636): replace SimpleNamespace with MagicMock(spec=App) in test_app_dsl_service

iamPulakesh:
### Summary
Replace all `SimpleNamespace` test doubles with `MagicMock(spec=App)` in `test_app_dsl_service.py` to resolve 12 `bad-argument-type` errors and 1 `read-only` property error flagged by the type checker.

### Why this change
`SimpleNamespace` is not assignable to parameters typed as `App`. Using `MagicMock(spec=App)` satisfies the type checker while preserving the same lightweight test behavior. This also naturally resolves the `app_model_config` read-only property error since `MagicMock` allows freely setting any attribute.

### Changes
**`api/tests/unit_tests/services/test_app_dsl_service.py`**:
- Added `App` to model imports.
- Introduced `_app_mock(**kwargs)` helper (mirrors the existing `_account_mock()` pattern already in the file).
- Replaced 20 `SimpleNamespace(...)` instances with `_app_mock(...)` wherever the object is passed to a parameter typed as `App` or `App | None`.

### Test plan
- `ruff check` and `basedpyright` - passed.

### DIFF
--- a/api/tests/unit_tests/services/test_app_dsl_service.py
+++ b/api/tests/unit_tests/services/test_app_dsl_service.py
@@ -11,7 +11,7 @@ from core.trigger.constants import (
     TRIGGER_SCHEDULE_NODE_TYPE,
     TRIGGER_WEBHOOK_NODE_TYPE,
 )
-from models import Account, AppMode
+from models import Account, App, AppMode
 from models.model import IconType
 from services import app_dsl_service
 from services.app_dsl_service import (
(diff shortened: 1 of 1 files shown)


asukaminato0721:
@dosu how about directly use App, instead of MagicMock?

asukaminato0721:
since this is unittest, this well, is enough, I hope to use container though.

iamPulakesh:
Using real App() instances would require `enable_site`, `enable_api`, and 
other required fields everywhere, it will add noise to tests that don't care about 
those fields at all.

Happy to switch to real `App()` if the you prefer that approach. @asukaminato0721

asukaminato0721 (approved):

PR #34660, opened 2026-04-07
fix(web): handle IME composition in DelimiterInput

plind-junior:
Fixes #34364

### Problem

Using an IME (e.g. Microsoft Pinyin) in the knowledge base delimiter input field causes duplicated characters. Typing `wuliu` and pressing space to confirm produces `www五六` instead of `五六`.

This happens because `DelimiterInput` fires `onChange` on every keystroke during IME composition, before the user has committed their input. The intermediate keystrokes leak into the field value.

### Solution

Added `onCompositionStart` / `onCompositionEnd` handlers to `DelimiterInput`, following the same pattern already used by `SearchInput` in the codebase. During composition, changes are buffered in local state and only flushed to the parent `onChange` once the IME input is committed.

### Testing

- Added a test case that simulates IME composition and verifies `onChange` is suppressed during composition and only fires on `compositionEnd`
- All 17 existing tests in `inputs.spec.tsx` continue to pass


plind-junior:
@asukaminato0721 kindly check this PR

plind-junior:
@asukaminato0721 Would you check my PR?

plind-junior:
@hyoban Would you check it again?

plind-junior:
@lyzno1 Would you check this too? Ty

plind-junior:
@fatelei @asukaminato0721 Review my PR, ty

plind-junior:
@asukaminato0721 Sorry for the interrupt during the weekend but this PR has been opened for a week.
Do you have a chance to review this PR? ty

WTW0313:
hello, @plind-dm. The handler flushes `compositionValue` from React state, not the input’s actual DOM value at `compositionEnd`. That is weaker than the existing SearchInput pattern, which reads `e.currentTarget.value`.

plind-junior:
@WTW0313, I've fixed the issues following your comments. I attempted to update fetch.ts to resolve a pre-existing type error, but decided to leave it as is since it's unrelated to the original issue.

WTW0313 (approved):

PR #34661, opened 2026-04-07
fix(web): assign in-progress tracing items to latest loop/iteration record

plind-junior:
Fixes #34559

### Problem

During streaming execution, the loop/iteration tracing panel shows in-progress child nodes under the **first** record instead of the **latest** one.

In `addChildrenToLoopNode`, `loop_index` defaults to `0` via destructuring, so `(parallel_mode_run_id || loop_index)` evaluates to `0` when both are falsy — the `||` operator can't distinguish between "missing" and "actually index 0".

Same issue in `addChildrenToIterationNode`: `iteration_index` defaults to `0` in destructuring, making the subsequent `!== undefined` check always true.

### Fix

- Removed the `= 0` destructuring defaults so missing indices stay `undefined`
- Switched to nullish coalescing (`??`) which correctly preserves `0` as a valid index
- When the index is genuinely missing (streaming, metadata not yet populated), falls back to the latest known record via `Math.max(0, details.length - 1)`

### Testing

- Added test cases for both loop and iteration that verify items with missing index land on the latest record
- All 17 existing format-log tests pass


plind-junior:
@asukaminato0721 kindly check this PR

plind-junior:
@asukaminato0721 Would you check my PR?

lyzno1 (approved):

Issue #23579, opened 2025-08-07
[Chore/Refactor] Native Support for Dataclasses Mapped as ORM Models


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

continue of https://github.com/langgenius/dify/issues/22652

### Motivation

https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#native-support-for-dataclasses-mapped-as-orm-models

asukaminato0721:
progress

```
grep --perl-regexp "class .*?\(Base\):" -r --exclude-dir=".venv" --exclude-dir="tests". | xargs -I {} printf "- [ ] %s\n" {} |  wl-copy
```

- [ ] models/dataset.py:class Dataset(Base):
- [ ] models/dataset.py:class DatasetProcessRule(Base):  # bug
- [ ] models/dataset.py:class Document(Base):
- [ ] models/dataset.py:class DocumentSegment(Base):
- [ ] models/dataset.py:class ChildChunk(Base):
- [ ] models/dataset.py:class TidbAuthBinding(Base):
- [ ] models/dataset.py:class SegmentAttachmentBinding(Base):
- [ ] models/model.py:class App(Base):
- [ ] models/model.py:class AppModelConfig(Base):
- [ ] models/model.py:class Conversation(Base):
- [ ] models/model.py:class Message(Base):
- [ ] models/model.py:class MessageAnnotation(Base):
- [ ] models/model.py:class AppAnnotationHitHistory(Base):
- [ ] models/model.py:class Site(Base):
- [ ] models/model.py:class ApiToken(Base):  # bug: this uses setattr so idk the field.
- [ ] models/model.py:class UploadFile(Base):
- [ ] models/model.py:class TenantCreditPool(Base):
- [ ] models/workflow.py:class Workflow(Base):  # bug
[...]

PR #34686, opened 2026-04-07
refactor: make DefaultFieldsMixin compatible with TypeBase (MappedAsDataclass)

sxxtony:
## Summary

Part of #23579 — migrating ORM models from `Base` to `TypeBase`.

`DefaultFieldsMixin` is used by several models that still inherit from
`Base`. Before those models can be converted to `TypeBase`
(`MappedAsDataclass`), the mixin's field declarations must support the
dataclass protocol — otherwise the generated `__init__` will not handle
`id`, `created_at`, and `updated_at` correctly.

## Changes

**`api/models/base.py`**

For each of the three shared fields (`id`, `created_at`, `updated_at`):

| Parameter | Before | After |
|---|---|---|
| INSERT default | `default=callable` | `insert_default=callable` |
| Dataclass default | — | `default_factory=callable` |
| Dataclass init | — | `init=False` |

- `insert_default` replaces `default` for the database INSERT operation —
  identical behavior for all existing `Base` classes
- `default_factory` provides the `__init__` default for `TypeBase`
  (`MappedAsDataclass`) classes, where `default=callable` is not valid
- `init=False` excludes these fields from the generated `__init__` —
  they are auto-populated and should not be passed by the caller

## Backward compatibility

`Base` (`DeclarativeBase` without `MappedAsDataclass`) ignores
`default_factory` and `init` entirely — no behavior change for any
existing model.

## What this unblocks

Subsequent PRs can now convert `DefaultFieldsMixin, Base` →
`DefaultFieldsMixin, TypeBase` with a one-line change per class:

- `workflow.py`: `WorkflowPause`, `WorkflowPauseReason`
- `human_input.py`: `HumanInputForm`, `HumanInputDelivery`,
  `HumanInputFormRecipient`
- `execution_extra_content.py`: `ExecutionExtraContent`

### DIFF
--- a/api/models/base.py
+++ b/api/models/base.py
@@ -24,6 +24,8 @@ class TypeBase(MappedAsDataclass, DeclarativeBase):
 
 
 class DefaultFieldsMixin:
+    """Mixin for models that inherit from Base (non-dataclass)."""
+
     id: Mapped[str] = mapped_column(
         StringUUID,
         primary_key=True,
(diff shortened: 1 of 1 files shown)


asukaminato0721 (dismissed):

asukaminato0721:
fix ci

Issue #30001, opened 2025-12-22
[Refactor/Chore] use match when doing if on enum


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

now in codebase, we use if x == a.b ，but it's easy to forget add new branch.

### Motivation

Match case can check it *automatically.

### Additional Context

https://github.com/facebook/pyrefly/pull/1833 I add related check in this pr.

asukaminato0721:
@dosu give me a list of file and tell me how to fix

asukaminato0721:
@dosu not, don't add _

if missing it, pyrefly / pyright will err in CI, to remind the user.

agenthaulk:
Hi @asukaminato0721, I'd like to pick up this issue.

I'll start with `api/services/app_model_config_service.py` (lines 10-17), which only handles 3 of 7 `AppMode` values. The plan is to convert the if/elif chain to an exhaustive match/case — no `case _:` wildcard, so pyrefly/pyright will catch any missing branches in CI.

I can submit a PR within 24 hours. Let me know if this file is available or if you'd prefer I start with a different one. Thanks!

agenthaulk:
Hi @asukaminato0721, I've submitted a PR that addresses this issue: #34412

The PR refactors `if/elif` chains to `match/case` across **7 files**, covering both `AppMode` (7 values) and `ProviderQuotaType` (3 values). All enum values are explicitly listed — no `case _:` wildcard, so pyrefly/pyright will catch any missing branches when new values are added.

Includes 40 unit tests validating exhaustive coverage. Looking forward to your review!

asukaminato0721:
@dosu give me a list of file and tell me how to fix

asukaminato0721:
@dosu what is the estimate remaining time for this issue.

dataCenter430:
[...]

PR #34768, opened 2026-04-08
refactor: convert ToolProviderType if/elif to match/case (#30001)

dataCenter430:
## What
Convert `if/elif` chains on `ToolProviderType` enum to exhaustive `match/case` statements in `tool_manager.py`.

Part of #30001. 

## Files changed
- `api/core/tools/tool_manager.py`

All tests passed.

PR #34777, opened 2026-04-08
refactor: convert ToolProviderType if/elif to exhaustive match/case (#30001) 🤖🤖🤖

jeanibarz:
## Summary

Converts two `if/elif` chains on `ToolProviderType` to exhaustive `match/case` in `ToolManager`:

- **`get_tool_runtime`** (6 branches → 7 explicit cases)
- **`get_tool_icon`** (5 branches + else → 7 explicit cases)

All seven `ToolProviderType` enum values are listed explicitly with no wildcard `case _:`, so pyrefly/pyright will flag missing branches when new variants are added.

Also:
- Removes dead code in `get_tool_icon` (unreachable `raise` after `try/except` in the `PLUGIN` branch)
- Removes unnecessary self-assignments (`provider_type = provider_type`)
- Updates test to use actual unsupported enum values (`APP`, `DATASET_RETRIEVAL`) instead of a raw string that bypassed the type system

Part of #30001

## Test plan

- [x] `pytest api/tests/unit_tests/core/tools/test_tool_manager.py` — 41/41 pass
- [x] `ruff check` — no lint issues
- [x] `ruff format --check` — already formatted

### DIFF
--- a/api/core/tools/tool_manager.py
+++ b/api/core/tools/tool_manager.py
@@ -205,16 +205,170 @@ class ToolManager:
 
         :return: the tool
         """
-        if provider_type == ToolProviderType.BUILT_IN:
-            # check if the builtin tool need credentials
-            provider_controller = cls.get_builtin_provider(provider_id, tenant_id)
+        match provider_type:
+            case ToolProviderType.BUILT_IN:
+                # check if the builtin tool need credentials
+                provider_controller = cls.get_builtin_provider(provider_id, tenant_id)
 
-            builtin_tool = provider_controller.get_tool(tool_name)
-            if not builtin_tool:
-                raise ToolProviderNotFoundError(f"builtin tool {tool_name} not found")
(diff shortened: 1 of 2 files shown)


dataCenter430:
Hey, @jeanibarz  thanks for your enthusiasm.
But this PR is duplicated my PR that already opened.
I think there are some unnecessary changes.
Please check the pr: #34768
You can open another follow-up PR to complete the issue.
Thank you.

jeanibarz:
You're right, #34768 was opened first and covers the same `tool_manager.py` changes. Closing this in favor of yours.

One heads-up: the existing test `test_get_tool_icon_invalid_provider_type_raises` passes `"invalid"` (a raw string) which currently hits the `else` clause. After the match/case conversion, no case will match that string and the function will silently return `None` instead of raising `ValueError`, so the test will fail. You may want to update it to use actual enum values like `ToolProviderType.APP` or `ToolProviderType.DATASET_RETRIEVAL`.

Issue #26412, opened 2025-09-29
[Chore/Refactor] remove ignore in pyright config


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

now pyright ignore some folders, fix the typing errors in them and let pyright check them.

### Motivation

improve type check coverage.

arpan-artisan:
Hi is this issue still open? Would like to solve this issue.

asukaminato0721:
> Hi is this issue still open? Would like to solve this issue.

yes, feel free to do it

arpan-artisan:
Could you fill in some more details on what needs to be done and the success criteria(what done means in this case)?

asukaminato0721:
> Could you fill in some more details on what needs to be done and the success criteria(what done means in this case)?

for example, https://github.com/langgenius/dify/blob/365f749ed5794da584154a63eafaab9587f31343/api/pyrightconfig.json#L40

remove this line, run type check, fix all the type err.

arpan-artisan:
Ok understood

Hypertension42:
Is this the problem solved?

asukaminato0721:
> Is this the problem solved?

no

tmimmanuel:
Hi, @asukaminato0721 
Can I work on this issue? If it's available, please assign. Thanks.

tmimmanuel:
Hi, @asukaminato0721 
Please review my open PRs.  Thanks.

PR #34796, opened 2026-04-09
refactor(api): add null safety to extract_processor and firecrawl retry methods

tmimmanuel:
## Summary
- Fix `upload_file` possibly-unbound variable in `extract_processor.py` by
  binding it before the `file_path` check and adding assert guards at PDF/DOCX
  extractor call sites
- Fix `response` possibly-unbound variable in `firecrawl_app.py` by initializing
  before retry loops with assert after loop completion

## Test plan
- [x] `make type-check-core` passes with 0 errors
- [x] All 194 extractor unit tests pass (no test changes needed)
- [x] No runtime behavior changes — null safety assertions only

Part of #26412


tmimmanuel:
@asukaminato0721 could you please review and approve?

Issue #34878, opened 2026-04-10
[Refactor/Chore] use [T] instead of Any


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

now dify use >= 3.12, which means we can use `f[T1, T2](x: T1, y: T2)` to type functions.

### Motivation

easier to remove Any

asukaminato0721:
and also `[**P](**kwargs: P.args):` so no more Any.

@dosu give me more detailed explanation.

corevibe555:
@asukaminato0721 Would you assign to me please? I'd love to contribute.

asukaminato0721:
@dosu is this issue solved?

PR #34890, opened 2026-04-10
refactor(mcp): remove unused AnyFunction alias, tighten callback type

corevibe555:
Part of: https://github.com/langgenius/dify/issues/34878Part of: https://github.com/langgenius/dify/issues/34878

## Summary
- Remove unused `type AnyFunction = Callable[..., Any]` alias from `types.py`
- Narrow `_on_complete` callback return type from `Any` to `object` in `RequestResponder`

## Why this change
`AnyFunction` was defined but never imported anywhere in the codebase — dead code.

`_on_complete` callback's return value is always discarded (called as a statement on line 90, never captured). `Any` disables type checking on the return; `object` correctly says "returns something we don't use."

## Changes
- `core/mcp/types.py`: Remove dead `AnyFunction` type alias
- `core/mcp/session/base_session.py`: `Callable[..., Any]` → `Callable[..., object]` on `_on_complete`

## Test plan
- [x] `basedpyright` passes
- [x] `ruff check` passes
- [x] 383 MCP unit tests pass

### DIFF
--- a/api/core/mcp/session/base_session.py
+++ b/api/core/mcp/session/base_session.py
@@ -55,7 +55,7 @@ class RequestResponder[ReceiveRequestT: ClientRequest | ServerRequest, SendResul
 
     request: ReceiveRequestT
     _session: "BaseSession[Any, Any, SendResultT, ReceiveRequestT, Any]"
-    _on_complete: Callable[["RequestResponder[ReceiveRequestT, SendResultT]"], Any]
+    _on_complete: Callable[["RequestResponder[ReceiveRequestT, SendResultT]"], object]
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


api/core/mcp/types.py
```diff
@@ -31,7 +31,6 @@
 Cursor = str
 Role = Literal["user", "assistant"]
 RequestId = Annotated[int | str, Field(union_mode="left_to_right")]
-type AnyFunction = Callable[..., Any]
```
corevibe555, line 34: Not used anywhere

api/core/mcp/session/base_session.py
```diff
@@ -55,15 +55,15 @@ class RequestResponder[ReceiveRequestT: ClientRequest | ServerRequest, SendResul
 
     request: ReceiveRequestT
     _session: "BaseSession[Any, Any, SendResultT, ReceiveRequestT, Any]"
-    _on_complete: Callable[["RequestResponder[ReceiveRequestT, SendResultT]"], Any]
+    _on_complete: Callable[["RequestResponder[ReceiveRequestT, SendResultT]"], object]
```
asukaminato0721, line 58: use T here

corevibe555: To use T here, we need to add a third type parameter to RequestResponder with a default (OnCompleteT = object), which requires PEP 696 (Python 3.13+). And now project currently use 3.12.0.
So if you would open to bump to Python 3.13, that's what we should do.
Otherwise, I think we should stay with object.

asukaminato0721: then I guess add a comment to future is ok.

asukaminato0721 (approved):

PR #34891, opened 2026-04-10
refactor(api): replace Any with precise types in db_migration_lock

corevibe555:
Part of https://github.com/langgenius/dify/issues/34878

## Summary
- Type `redis_client` as `redis.Redis | RedisCluster | RedisClientWrapper` instead of `Any`
- Type `_lock` as `Lock | None` instead of `Any`
- Replace `acquire(*args: Any, **kwargs: Any)` with explicit `Lock.acquire` parameters (`sleep`, `blocking`, `blocking_timeout`, `token`)
- Type `_heartbeat_loop` lock param as `Lock` instead of `Any`
- Remove unused `type AnyFunction = Callable[..., Any]` from `mcp/types.py`
- Narrow `_on_complete` callback return from `Any` to `object` in `base_session.py`

## Why this change
`DbMigrationAutoRenewLock` typed its redis client and lock as `Any`, hiding the actual types. The `acquire` method forwarded `*args: Any, **kwargs: Any` to `Lock.acquire()` — now uses the explicit parameter signature for full type safety.

`AnyFunction` in `mcp/types.py` was defined but never imported — dead code.

`_on_complete` callback's return is always discarded; `object` correctly expresses "returns something we don't use" without disabling type checking like `Any` does.

## Changes
- `libs/db_migration_lock.py`: Replace all `Any` with `redis.Redis | RedisCluster | RedisClientWrapper`, `Lock`, and explicit params
- `core/mcp/types.py`: Remove dead `AnyFunction` alias
- `core/mcp/session/base_session.py`: `Callable[..., Any]` → `Callable[..., object]`

## Test plan
- [x] `basedpyright` passes
- [x] `mypy` passes (1387 source files)
- [x] `ruff check` passes
- [x] 383 MCP unit tests pass

### DIFF
--- a/api/libs/db_migration_lock.py
+++ b/api/libs/db_migration_lock.py
@@ -86,11 +86,11 @@ class DbMigrationAutoRenewLock:
     def name(self) -> str:
         return self._name
 
-    def acquire(self, **kwargs: Any) -> bool:
+    def acquire(self, *args: Any, **kwargs: Any) -> bool:
         """
         Acquire the lock and start heartbeat renewal on success.
 
-        Accepts the same keyword arguments as redis-py ``Lock.acquire()``.
+        Accepts the same args/kwargs as redis-py `Lock.acquire()`.
         """
         # Prevent accidental double-acquire which could leave the previous heartbeat thread running.
         if self._acquired:
@@ -103,7 +103,7 @@ class DbMigrationAutoRenewLock:
                 timeout=self._ttl_seconds,
                 thread_local=False,
             )
-        acquired = bool(self._lock.acquire(**kwargs))
+        acquired = bool(self._lock.acquire(*args, **kwargs))
         self._acquired = acquired
         if acquired:
             self._start_heartbeat()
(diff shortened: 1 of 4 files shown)


api/libs/db_migration_lock.py
```diff
@@ -80,7 +86,13 @@ def __init__(
...
         return self._name
 
-    def acquire(self, *args: Any, **kwargs: Any) -> bool:
+    def acquire(
+        self,
+        sleep: float | None = None,
+        blocking: bool | None = None,
+        blocking_timeout: float | None = None,
+        token: str | None = None,
+    ) -> bool:
```
asukaminato0721: @dosu is changing the signature of acquire safe here?

corevibe555: I scanned the entire api codebase and two places call acquire, confirmed they are safe.

commands/system.py:134	
**lock.acquire(blocking=False)**	Safety: Yes, keyword arg
tests/.../test_auto_renew_redis_lock_integration.py:27	
**lock.acquire(blocking=True, blocking_timeout=5)**	Safety: Yes, keyword args

corevibe555:
@asukaminato0721 I resolved the issues. Would you review this PR?

corevibe555:
@asukaminato0721 I think I understood concern, be safe and no behavioral change.
I reverted risky part - it's purely no risk part left for this PR.

asukaminato0721 (approved):

PR #34923, opened 2026-04-10
fix: sticky toolbar overlap with nav bar on scroll

xingxing21:
> [!IMPORTANT]
>

Fixes #31270

## Summary

The sticky toolbar on the Knowledge Base list page (containing the tag filter, search input, and controls) was overlapping the navigation bar when scrolling down the page.

Root cause: The outer scroll container was missing h-0 in its Tailwind classes. Without it, the container had no bounded height, so the browser treated the document as the scroll context instead of the div. This caused the inner sticky top-0 toolbar to pin to the viewport top — overlapping with the nav bar — rather than the top of the scroll container.

Fix: Added h-0 to the scroll container, consistent with the same pattern used in the Apps list (web/app/components/apps/list.tsx) and Tools list (web/app/components/tools/provider-list.tsx). With h-0 + grow, the container is properly bounded and becomes the true scroll context, so the sticky toolbar correctly sticks just below the nav bar.

```
- <div className="scroll-container relative flex grow flex-col overflow-y-auto bg-background-body">
+ <div className="scroll-container relative flex h-0 grow flex-col overflow-y-auto bg-background-body">
```

## Checklist

- [x] I ran `make lint` and `make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods


xingxing21:
@dosu review this PR

xingxing21:
Hi @crazywoola 

I hope you’re doing well.
I wanted to kindly ask if you might have time to take a look at this PR when convenient. I’d really appreciate any feedback or suggestions you may have.

Thank you for your time and for maintaining the project!

WTW0313:
Hello, @xingxing21. I tested this locally, and the layout is still incorrect.

I think `z-index` is the real issue here. The sticky toolbar in `index.tsx` and the corner labels in `corner-labels.tsx` are both at `z-10`. Since the cards are rendered later in the DOM, their `z-10` elements are painted above the toolbar, so the overlap still happens.

Because of that, I don’t think this PR fully fixes the issue yet

xingxing21:
Hi @WTW0313 
I will improve it more
Thank you for your feedback

xingxing21:
@WTW0313 

this is fixed. please check it again

https://github.com/user-attachments/assets/5ff125f9-ae75-4261-926d-78f836572fcc

WTW0313:
> @WTW0313
> 
> this is fixed. please check it again
> 
>  Recording.2026-04-14.223842.mp4

This recording doesn’t show the key issue described in #31270.

The problem in that issue is specifically about the dataset card with the knowledge pipeline label, but the datasets shown in this list don’t have that label, so the overlap case is not actually being verified here.

Could you create a new dataset from Knowledge Pipeline and check again whether the issue still occurs in that scenario? That would make the verification match the original report more closely.

PR #34944, opened 2026-04-10
fix(web): invalidate systemFeatures cache after login to show license banner

HamzaSwitch:
Fixes #34520.

The "license expiring in N days" banner doesn't show up on `/apps` right after login. You have to refresh the browser for it to appear. @BenjaminX's trace in the issue is correct, so I'll just summarize:

1. `GlobalPublicStoreProvider` hits `GET /console/api/system-features` before the user is logged in.
2. The backend deliberately only returns `license.expired_at` when the caller is authenticated (see the security comment in `api/services/feature_service.py` around line 389, from #33044). Pre-login response is `expired_at = ""`.
3. React Query caches that empty response.
4. After `router.replace('/apps')`, nothing invalidates the cache, so `LicenseNav` keeps reading `""` and hides the badge.
5. A manual refresh triggers a new fetch with the auth cookie and the badge appears.

## The fix

Went with Option B from the issue (frontend cache invalidation). Option A would mean moving `expired_at` out of the `is_authenticated` guard, but that guard has an explicit security note saying the expiry date is intentionally auth-gated. Didn't want to touch that.

Three small changes:

- `web/context/global-public-context.tsx`: exported `systemFeaturesQueryKey` so login handlers can reference it.
- `web/app/signin/components/mail-and-password-auth.tsx`: call `queryClient.invalidateQueries({ queryKey: systemFeaturesQueryKey })` after a successful login, before `router.replace('/apps')`.
- `web/app/signin/check-code/page.tsx`: same thing for the email-code login path.

Left the invite and `one-more-step` flows alone since they don't reproduce the symptom and I wanted to keep the diff minimal. Can extend to those too if you'd rather.

## Testing

Local, with `docker compose -p dify up -d`:

1. Logged out, went to `/signin`
2. Logged in as a user in a workspace with a license that's expiring soon
3. Banner shows up immediately on `/apps`, no refresh needed
4. Same thing with the email-code login path

Before the change the banner stayed blank until a manual refresh. After, it renders on first paint. Unauthenticated callers still get `expired_at = ""` from `/system-features`, so the backend boundary is unchanged.

### DIFF
--- a/web/app/signin/check-code/page.tsx
+++ b/web/app/signin/check-code/page.tsx
@@ -1,6 +1,7 @@
 'use client'
 import type { FormEvent } from 'react'
 import { RiArrowLeftLine, RiMailSendFill } from '@remixicon/react'
+import { useQueryClient } from '@tanstack/react-query'
 import { useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { trackEvent } from '@/app/components/base/amplitude'
@@ -8,6 +9,7 @@ import Button from '@/app/components/base/button'
 import Input from '@/app/components/base/input'
 import { toast } from '@/app/components/base/ui/toast'
 import Countdown from '@/app/components/signin/countdown'
+import { systemFeaturesQueryKey } from '@/context/global-public-context'
 import { useLocale } from '@/context/i18n'
 
 import { useRouter, useSearchParams } from '@/next/navigation'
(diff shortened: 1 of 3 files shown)


HamzaSwitch:
Closing as duplicate — an earlier PR already targets #34520. Thanks to the maintainers for the triage; deferring to the existing work.

Issue #33362, opened 2026-03-12
Chatflow conversation reset button does not clear conversationIdInfo, preventing new conversations


tukaelu:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.13.0

### Cloud or Self Hosted

Cloud

### Steps to reproduce

1. Open a Chatflow app on Dify Cloud.
2. Send one or more messages to start a conversation.
3. Click the **reset (new conversation)** button in the top-right corner of the chat UI.

No error messages are displayed in the UI; the bug is silent and only observable by inspecting the conversation session behavior.

### ✔️ Expected Behavior

Clicking the reset button should properly clear `conversationIdInfo` and initiate a brand-new conversation with a new session ID.

### ❌ Actual Behavior

Although the chat UI visually appears to reset, the `conversationIdInfo` is not properly cleared internally.
The session ID remains the same as before the reset, so any subsequent messages are sent within the old conversation session rather than starting a new one.

This issue is reproducible 100% of the time on both Google Chrome and Safari.

PR #34945, opened 2026-04-10
fix(web): clear stale conversation_id from localStorage on 404

### LINKED ISSUE #34731: Stale conversation_id in localStorage causes infinite 404 "Conversation Not Exists" loop in Web App chatbot
### Self Checks

### Dify version

v1.13.3

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Steps to reproduce
1. Deploy Dify with a Chatflow app and share it as a Web App (e.g. via /chatbot/<app_code> embed URL)
2. Open the chatbot URL in a browser, send some messages to create a conversation
3. Observe in browser DevTools → Application → Local Storage that conversationIdInfo now stores the conversation_id
4. Delete the conversation from Dify (e.g. admin deletes it via console, or the conversation is removed from the database)
5. Reload the chatbot page in the browser
Nginx access log confirms the loop — the same stale conversation_id is retried every 2-3 seconds, all returning 404:
GET /api/messages?conversation_id=xxxxxx-xxxx-xxxx-xxxx-xxxxxxxx06060&limit=20&last_id= HTTP/1.1" 404 201
GET /api/messages?conversation_id=xxxxxx-xxxx-xxxx-xxxx-xxxxxxxx06060&limit=20&last_id= HTTP/1.1" 404 201
GET /api/messages?conversation_id=xxxxxx-xxxx-xxxx-xxxx-xxxxxxxx06060&limit=20&last_id= HTTP/1.1" 404 201
... (repeated 28+ times)
6. Clicking "Reset conversation" in the UI does not resolve the issue.

Root Cause Analysis
The issue spans two areas:
1. useShareChatList in web/service/use-share.ts has no 404 error handling — TanStack Query retries 3 times by default, and refetches on window focus, creating an infinite retry loop for this unrecoverable
error.
2. Both chat-with-history/hooks.tsx and embedded-chatbot/hooks.tsx use useLocalStorageState to persist conversation_id. When GET /api/messages returns 404, neither hook clears the stale conversation_id from
localStorage or resets to a new conversation state. The "Reset conversation" button only resets the visual state but does not clear the localStorage entry.

### ✔️ Expected Behavior

When a conversation no longer exists (API returns 404), the frontend should automatically detect the stale conversation_id, clear it from localStorage, and reset to a new conversation state — providing a seamless user experience without requiring manual browser data clearing.

### ❌ Actual Behavior

When a conversation is deleted, the Web App chatbot enters an infinite 404 error loop. The user sees the error message "Conversation Not Exists. You have requested this URI [/api/messages] but did you mean /api/messages or /api/chat-messages or /api/saved-messages?" every 2-3 seconds. Clicking "Reset conversation" does not fix it. The only workaround is for the user to manually clear browser localStorage/site data for the domain.

HamzaSwitch:
Fixes #34731.

When a conversation has been deleted on the server but the Web App chatbot still has its `conversation_id` in localStorage, the page enters an infinite loop: it refetches `GET /api/messages?conversation_id=...`, gets `404`, shows a toast, and retries again every couple of seconds. The "Reset conversation" button only clears the visual state, not the localStorage entry, so the only real workaround is for the user to manually clear site data.

The reporter and @dosu already traced the root cause pretty well. Summarizing:

1. `useShareChatList` in `web/service/use-share.ts` doesn't configure `retry`, so TanStack Query uses its default (3 retries) and also refetches on window focus.
2. `fetchChatList` doesn't pass `silent: true`, so every failed attempt fires a toast via `afterResponseErrorCode` in `web/service/fetch.ts`.
3. Both `embedded-chatbot/hooks.tsx` and `chat-with-history/hooks.tsx` destructure only `data` and `isLoading` from `useShareChatList` — the `error` state is ignored, so nothing clears the stale `conversation_id` from localStorage when the 404 comes back.

## The fix

Three small changes, all frontend:

**`web/service/share.ts`** — `fetchChatList` now passes `silent: true` so a 404 response doesn't spam toasts. The 404 is a legitimate "conversation no longer exists" signal that the hook layer handles; it shouldn't look like a server error to the user.

**`web/service/use-share.ts`** — `useShareChatList` now disables retries for 404s:

```ts
retry: (failureCount, error: unknown) => {
  if (error instanceof Response && error.status === 404)
    return false
  return failureCount < 3
}
```

Other HTTP failures still get the default 3 retries.

**`web/app/components/base/chat/embedded-chatbot/hooks.tsx`** and **`web/app/components/base/chat/chat-with-history/hooks.tsx`** — both now read `error` from `useShareChatList` and run a `useEffect` that calls `removeConversationIdInfo(appId)` when the error is a 404 `Response`. Clearing the stale entry causes `currentConversationId` to fall back to `''`, which in turn disables the query (since `isEnabled` checks `!!params.conversationId`), and the chatbot starts a fresh conversation.

`chat-with-history/hooks.tsx` didn't previously have a `removeConversationIdInfo` helper, so I added one matching the shape of the existing `embedded-chatbot` implementation.

## Why this shape instead of alternatives

- **Backend fix (return 200 with an empty array on deleted conversations)** — would mask real 404s and lose the useful "this id is gone" signal. The frontend already has everything it needs to handle this cleanly.
- **Reset conversation button also clears localStorage** — the reporter mentioned that the button only resets the visual state. I looked at #33362 which tracks that separately. Fixing the infinite-loop symptom doesn't require touching the button, and wiring both into one PR would widen scope.
- **Suppressing the toast only, without retry/clear logic** — would stop the visible spam but leave the underlying retry storm running in the network tab. The retry stop + localStorage clear is what actually makes the page recover.

## Testing

Manually reproduced on a self-hosted stack (`docker compose -p dify up -d`) with a Chatflow app exposed as a Web App:

1. Opened the chatbot URL, sent a message, confirmed `conversationIdInfo` is populated in `Application → Local Storage → conversationIdInfo`.
2. Deleted the conversation from the console as admin.
3. Reloaded the chatbot page.

**Before the fix:** Network tab shows `GET /api/messages?...` returning 404 every 2–3 seconds, toast error pops up repeatedly, page never recovers. Clicking "Reset conversation" does nothing useful.

**After the fix:** Exactly one `GET /api/messages?...` 404 in the network tab, no toast, `conversationIdInfo` is cleared from localStorage on the same tick, the chatbot falls back to the new-conversation state automatically.

Also re-tested the happy path (send messages in a valid conversation, switch conversations, refresh) to confirm I didn't break the normal chat-list loading flow.

### DIFF
--- a/web/app/components/base/chat/chat-with-history/hooks.tsx
+++ b/web/app/components/base/chat/chat-with-history/hooks.tsx
@@ -144,6 +144,13 @@ export const useChatWithHistory = (installedAppInfo?: InstalledApp) => {
   const [conversationIdInfo, setConversationIdInfo] = useLocalStorageState<Record<string, Record<string, string>>>(CONVERSATION_ID_INFO, {
     defaultValue: {},
   })
+  const removeConversationIdInfo = useCallback((targetAppId: string) => {
+    setConversationIdInfo((prev) => {
+      const newInfo = { ...prev }
+      delete newInfo[targetAppId]
+      return newInfo
+    })
+  }, [setConversationIdInfo])
   const currentConversationId = useMemo(() => conversationIdInfo?.[appId || '']?.[userId || 'DEFAULT'] || '', [appId, conversationIdInfo, userId])
   const handleConversationIdInfoChange = useCallback((changeConversationId: string) => {
     if (appId) {
@@ -185,7 +192,7 @@ export const useChatWithHistory = (installedAppInfo?: InstalledApp) => {
     refetchOnWindowFocus: false,
     refetchOnReconnect: false,
   })
-  const { data: appChatListData, isLoading: appChatListDataLoading } = useShareChatList({
+  const { data: appChatListData, isLoading: appChatListDataLoading, error: appChatListDataError } = useShareChatList({
     conversationId: chatShouldReloadKey,
     appSourceType,
     appId,
@@ -194,6 +201,14 @@ export const useChatWithHistory = (installedAppInfo?: InstalledApp) => {
     refetchOnWindowFocus: false,
     refetchOnReconnect: false,
   })
+  // When the backend reports the conversation no longer exists (404), clear
+  // the stale conversation_id from localStorage so the chatbot falls back to
+  // starting a new conversation. Without this, the user would be stuck in an
+  // infinite retry loop (GitHub issue #34731).
+  useEffect(() => {
+    if (appChatListDataError instanceof Response && appChatListDataError.status === 404 && appId)
+      removeConversationIdInfo(appId)
+  }, [appChatListDataError, appId, removeConversationIdInfo])
   const invalidateShareConversations = useInvalidateShareConversations()
   const [clearChatList, setClearChatList] = useState(false)
   const [isResponding, setIsResponding] = useState(false)
--- a/web/app/components/base/chat/embedded-chatbot/hooks.tsx
+++ b/web/app/components/base/chat/embedded-chatbot/hooks.tsx
@@ -146,11 +146,19 @@ export const useEmbeddedChatbot = (appSourceType: AppSourceType, tryAppId?: stri
     pinned: false,
     limit: 100,
   })
-  const { data: appChatListData, isLoading: appChatListDataLoading } = useShareChatList({
+  const { data: appChatListData, isLoading: appChatListDataLoading, error: appChatListDataError } = useShareChatList({
     conversationId: chatShouldReloadKey,
     appSourceType,
     appId,
   })
+  // When the backend reports the conversation no longer exists (404), clear
+  // the stale conversation_id from localStorage so the chatbot falls back to
+  // starting a new conversation. Without this, the user would be stuck in an
+  // infinite retry loop (GitHub issue #34731).
+  useEffect(() => {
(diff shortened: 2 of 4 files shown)


6mvp6:
Hi @HamzaSwitch @iamjoel @zxhlyh @crazywoola 👋

Wanted to share that this issue is still reproducible on v1.15.0, and from checking the v1.16.0 source it looks like the fix here still hasn't landed in any release. On a self-hosted
deployment with an embedded Chatflow chatbot (/chatbot/<token>), once conversationIdInfo in localStorage holds a stale conversation_id (conversation deleted, or end-user identity
changed), the page hits the exact infinite 404 loop on GET /api/messages described in #34731 — the "Conversation Not Exists ... did you mean ..." toast every few seconds, and the only
workaround is clearing site data.

I notice this PR has been open for a while, has conflicts with main, and CI is red — it may be stuck on an outdated base. Could the maintainers confirm whether this PR is still being
actively pursued?

Either way, I'd be glad to help move it forward:
1. If you'd like to keep this PR, I can help rebase it onto the latest main and get CI green.
2. If a fresh start is easier, I can pull the latest code, re-apply the same fix (404-aware retry in useShareChatList, silent: true in fetchChatList, and clearing the stale id on 404
in both the embedded-chatbot and chat-with-history hooks), and open a new PR.

This has real user impact (users get stuck and have to clear browser data), so I'd like to help get it resolved. 🙏

PR #34948, opened 2026-04-10
fix(web): release responding lock on annotation-reply message_end

### LINKED ISSUE #34885: Chatflow API requests cannot be terminated when citation markers are matched.
### Self Checks

### Dify version

1.13.3

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Chatflow API requests cannot be terminated when citation markers are matched.

annotations
Image

Issues with annotations，does not terminate
Image

Issues outside the scope of annotations complete as expected.
Image

Image

### ✔️ Expected Behavior

Can complete successfully

### ❌ Actual Behavior

Does not finish as expected

HamzaSwitch:
Fixes #34885. Same root cause as the prematurely-closed #33103.

When a Chatflow message matches an annotation reply, the backend fires a single `message_end` event with `metadata.annotation_reply` and closes the stream without going through the normal terminal event that triggers `onCompleted`. So `handleResponding(false)`, which normally runs inside `onCompleted`, never fires, and the chat UI stays stuck in "generating" until the SSE connection physically closes.

The file already has a working version of this path. The regenerate-flow `onMessageEnd` handler around line 881 correctly does `handleResponding(false); return` in its annotation branch. The initial-send handler around line 359 just wasn't doing the same thing. I restructured it to match: check for `annotation_reply` at the callback level, apply the annotation via `updateChatTreeNode`, release the responding lock, return. Non-annotation branch is untouched.

686 tests in `app/components/base/chat/chat/` pass, lint and types clean. I don't have an annotation configured on a live Dify to reproduce visually, but since I'm mirroring a sibling function in the same file I'm pretty confident in the shape of the fix.

### DIFF
--- a/web/app/components/base/chat/chat/hooks.ts
+++ b/web/app/components/base/chat/chat/hooks.ts
@@ -357,14 +357,24 @@ export const useChat = (
         })
       },
       onMessageEnd: (messageEnd) => {
-        updateChatTreeNode(messageId, (responseItem) => {
-          if (messageEnd.metadata?.annotation_reply) {
+        const annotationReply = messageEnd.metadata?.annotation_reply
+        if (annotationReply) {
+          updateChatTreeNode(messageId, (responseItem) => {
             responseItem.annotation = ({
-              id: messageEnd.metadata.annotation_reply.id,
-              authorName: messageEnd.metadata.annotation_reply.account.name,
+              id: annotationReply.id,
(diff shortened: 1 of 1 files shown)


XavierMoon:
Hi @zxhlyh , just checking in on this PR.
Is there anything blocking the merge? If not, could you please help approve and merge it? Thanks.

PR #34949, opened 2026-04-11
fix(mcp): omit None fields from OAuth Dynamic Client Registration payload

HamzaSwitch:
Fixes #34857. @uskaidev filed this with the full root-cause analysis and the exact one-line fix already identified, so I just applied it and added a regression test.

Short version: `OAuthClientMetadata` has a bunch of optional fields defaulting to `None` (`scope`, `client_uri`, `grant_types`, etc.). `register_client()` was serializing with plain `model_dump()`, so unset fields went out as JSON `null`. Per RFC 7591 §2 they're supposed to just be absent from the payload, and strict MCP registration servers (the reporter hit this against GitLab's `gitlab-mcp`) respond with `invalid_client_metadata` when they see the nulls.

Fix is `model_dump(exclude_none=True)`. I left a comment above the call referencing the issue and the RFC so the next refactor of this file doesn't quietly undo it.

The existing `test_register_client_success` was asserting the serialized payload shape, so I updated that to match. I also added `test_register_client_omits_none_fields`: it constructs an `OAuthClientMetadata` with only the two required fields, calls `register_client`, and checks that `mock_post` received exactly `{"client_name", "redirect_uris"}` with none of the optional fields leaking through as null.

49 tests in `test_auth_flow.py` pass, ruff and format clean. I don't have a GitLab MCP instance to verify the end-to-end OAuth flow against the real server, but the regression test pins the exact payload shape that was being rejected.

### DIFF
--- a/api/core/mcp/auth/auth_flow.py
+++ b/api/core/mcp/auth/auth_flow.py
@@ -525,9 +525,16 @@ def register_client(
     else:
         registration_url = urljoin(server_url, "/register")
 
+    # Per RFC 7591 §2, optional client metadata fields that are not provided
+    # should be omitted from the registration request rather than sent as
+    # explicit JSON `null`. `OAuthClientMetadata` has several optional fields
+    # (`scope`, `client_uri`, `grant_types`, etc.) that default to `None`;
+    # without `exclude_none=True`, Pydantic serializes them as `null` and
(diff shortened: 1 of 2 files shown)


HamzaSwitch:
Closing as duplicate — an earlier PR already targets #34857. Thanks to the maintainers for the triage; deferring to the existing work.

PR #26419, opened 2025-09-29
use no-root user in docker image by default

41tair:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

fix #26418 
## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods

PR #34953, opened 2026-04-11
fix(docker): create /home/dify so Gunicorn 25 control socket can start

### LINKED ISSUE #34463: Bug: Gunicorn 25 control socket fails in api Docker image because /home/dify is missing
### Self Checks

### Dify version

v1.13.3

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

  1. Build and run the API container from the `api` directory, for example with a script that does:
     - `docker build ...`
     - `docker run ... -e MODE=api ...`
  2. Start the container as the default non-root `dify` user.
  3. Check the API container logs.

  Observed logs:

  ```text
  /bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
  /entrypoint.sh: line 8: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
  [2026-04-02 08:43:40 +0000] [1] [INFO] Starting gunicorn 25.3.0
  [2026-04-02 08:43:40 +0000] [1] [INFO] Listening at: http://0.0.0.0:5001 (1)
  [2026-04-02 08:43:40 +0000] [1] [INFO] Using worker: gevent
  [2026-04-02 08:43:40 +0000] [13] [INFO] Booting worker with pid: 13
  [2026-04-02 08:43:40 +0000] [1] [ERROR] Control server error: [Errno 13] Permission denied: '/home/dify'

  Additional runtime inspection inside the container shows:

  getent passwd dify
  dify:x:1001:1001::/home/dify:/bin/bash

  but /home/dify does not exist in the image.

  Root cause appears to be:

  - Gunicorn 25.x now enables a control socket by default.
  - Its default path falls back to $XDG_RUNTIME_DIR/gunicorn.ctl or $HOME/.gunicorn/gunicorn.ctl.
  - The image sets the runtime user home to /home/dify, but does not create that directory.
  - The image also sets LANG=en_US.UTF-8 and LC_ALL=en_US.UTF-8, but that locale is not available in python:3.12-slim-bookworm.

  A local workaround is to create /home/dify before startup, or set a writable XDG_RUNTIME_DIR. After creating /home/dify, Gunicorn starts cleanly and logs:

  [INFO] Control socket listening at /home/dify/.gunicorn/gunicorn.ctl

 
  ```

### ✔️ Expected Behavior

  The API Docker image should start without runtime warnings or permission errors.

  Specifically:

  - Gunicorn should initialize successfully without Permission denied: '/home/dify'
  - the image should provide a valid writable home/runtime directory for the dify user
  - locale configuration should not emit setlocale warnings on container startup

### ❌ Actual Behavior

  The container starts, but Gunicorn logs a control server error because it cannot initialize under /home/dify, and the shell also emits locale warnings because en_US.UTF-8 is not available in
  the base image.

  This makes the image noisy at startup and leaves the Gunicorn control socket misconfigured by default.

HamzaSwitch:
Fixes #34463.

When the api container starts as the non-root `dify` user, Gunicorn 25 crashes immediately with `[Errno 13] Permission denied: '/home/dify'`. The passwd entry correctly points at `/home/dify` as the user's home, but the directory doesn't exist on disk — `api/Dockerfile` creates the user with `useradd -r` (system user), and the `-r` flag suppresses home-directory creation.

This used to work. #26419 originally did an explicit `mkdir -p /home/dify`, #28756 removed that line while parameterizing the UID, and then #33298 updated Gunicorn to 25.x which introduced a control socket that writes into the user's HOME. The missing home directory that was previously a latent wart is now a startup crash.

Fix is the minimal recreation of the #26419 lines, right after `useradd`:

```dockerfile
mkdir -p /home/dify && \
chown dify:dify /home/dify && \
```

I went with explicit `mkdir` + `chown` rather than `useradd -m` because that's the shape that was working before, and because a standalone block with a comment explaining the *why* is harder to silently drop in a future refactor than a single flag buried in a long useradd invocation — which is exactly how the regression got reintroduced last time. Left a comment referencing #34463 and the Gunicorn reason above the new lines.

Sanity-checked the shell comment syntax I used in the RUN block (`cmd && \` → `# comment` → `cmd2 && \`) against both `bash` and `/bin/sh` — both handle comment lines inside a line-continued block correctly, the logical command stays `cmd && cmd2`. Also grep'd the other Dockerfiles in the repo; only `api/Dockerfile` has this `useradd -r` pattern, so the fix is scoped.

I don't have the api image rebuilt and running locally to visually confirm gunicorn comes up clean, so if a reviewer is already in a build env a quick sanity check that `/home/dify` exists as `dify:dify` in the final image would be appreciated.

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -55,6 +55,14 @@ ARG NODE_PACKAGE_VERSION=22.21.0-1nodesource1
 ARG NODESOURCE_KEY_FPR=6F71F525282841EEDAF851B42F59B5F99B1BE0B4
 RUN groupadd -r -g ${dify_uid} dify && \
     useradd -r -u ${dify_uid} -g ${dify_uid} -s /bin/bash dify && \
+    # Create /home/dify explicitly: `useradd -r` skips home-directory creation
+    # even though the passwd entry still points at /home/dify, which breaks
+    # tools that expect the user's HOME to exist on disk. Notably, Gunicorn
+    # 25's control socket writes into HOME and fails with
+    # `[Errno 13] Permission denied: '/home/dify'` if this directory is
+    # missing. See #34463.
+    mkdir -p /home/dify && \
+    chown dify:dify /home/dify && \
     chown -R dify:dify /app
 
 RUN \


HamzaSwitch:
Closing as duplicate — an earlier PR already targets #34463. Thanks to the maintainers for the triage; deferring to the existing work.

PR #35085, opened 2026-04-13
fix: Accept scalar user IDs for partial dataset member updates

EndlessLucky:
Part of #34392 .

## Summary

Accept partial_member_list entries as plain strings (account IDs) when PATCHing dataset settings for partial_members, instead of requiring every item to be an object with user_id. Aligns the API with callers that send ["uuid-a", "uuid-b"] and removes the TypeError that led to 500 and left datasets stuck as partial_members without usable dataset_permissions rows.

## Why this change

Console PATCH /console/api/datasets/<id> could receive partial_member_list as a list of strings. The service layer always did user["user_id"], so each string was treated like a dict and raised TypeError: string indices must be integers, not 'str'. After update_dataset had already persisted permission = partial_members, a failed update_partial_member_list could leave the dataset hidden from the list UI even though the row still showed partial_members in the database.

## Changes

- api/controllers/console/datasets/datasets.py — Add a Pydantic field_validator on DatasetUpdatePayload.partial_member_list (mode="before") so string items are normalized to {"user_id": ...} and dict items still support user_id / id (and optional role).
- api/services/dataset_service.py — In DatasetPermissionService.update_partial_member_list and check_permission (operator partial-members path), resolve account_id from either a string or a dict (user_id or id), and raise a clear ValueError if the shape is invalid.
- api/tests/unit_tests/services/test_dataset_service_dataset.py — Unit coverage for scalar-string partial member lists and operator permission checks with scalar IDs.

## Test plan

- uv run --project api pytest api/tests/unit_tests/services/test_dataset_service_dataset.py::TestDatasetPermissionService -q passes in CI (scalar partial_member_list + operator checks; no TypeError from string entries).
- make lint passes in CI (console datasets.py + dataset_service.py ruff/format clean).


EndlessLucky:
@crazywoola I closed the old PR and reponed a new PR for issue #34392 . Would you please review? Thanks.

EndlessLucky:
@crazywoola Would you please review this PR?

EndlessLucky:
@asukaminato0721 Would you please review this PR?

PR #29305, opened 2025-12-09
perf: optimize get_configurations

fatelei:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

fix #29304 
fix #27299

 | Before                 | After                  | Improvement|
 |------------------------|------------------------|--------------------|
 | 6 separate sessions    | 1 shared session       | 84% reduction|
 | 6x connection overhead | 1x connection overhead | 84% less overhead  |
 | 6 transaction setups   | 1 transaction setup    | 84% faster startup |

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `dev/reformat`(backend) and `cd web && npx lint-staged`(frontend) to appease the lint gods

Issue #27299, opened 2025-10-22
[Chore/Refactor] Workflow database performance optimization


qiqizjl:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

We found that a large amount of SQL statements are generated during workflow execution, including many duplicate statements. For example, when querying app and workflow information, can this be cached in Redis to reduce the pressure on the database?

qiqizjl:
In addition, regarding the connection pool properties of sqlalchemy, have you considered changing all db.session to with?

uply23333:
Same issue here. Within an **Iteration node**’s **Retrieve node,** each RAG step triggers `ProviderManager.get_configurations` ( core/provider_manager.py ), which issues multiple SQL queries and consistently takes ~1s per call in my environment.

PR #35088, opened 2026-04-13
perf(api): memoize ProviderManager.get_configurations within request/task scope

YgorLeal:
## Summary

Incremental step for #27299 — assembling `ProviderConfigurations` runs six independent DB queries plus a provider-factory load, costing ~1s per call in production. A single workflow step (e.g. a Retrieval node inside an Iteration) invokes this path many times through independently-created `ProviderManager` instances, so per-instance caching is not enough.

This PR memoizes `ProviderManager.get_configurations` within the current request/task scope, keyed by `tenant_id`. For a workflow with N retrieval/LLM calls that share the same tenant in a single run, the call path collapses from `N * ~1s` of DB work to `1 * ~1s` + `(N-1) * dict-lookup`.

Reported impact (from #27299 and related comments): within an Iteration node's Retrieve step, each RAG call triggers `ProviderManager.get_configurations`, which issues multiple SQL queries and consistently takes ~1s per call — causing multi-second stalls per iteration.

## Why request/task-scoped (and not cross-request)

A previous attempt (#29305) tried to share a `db.session` across helpers inside `get_configurations`. It was rejected for good reasons (`db.session` is Flask-SQLAlchemy scoped — breaks in Celery/CLI contexts; long-lived session holds transactions open across dozens of queries). This PR takes a different angle: **don't touch the queries themselves, but avoid re-running them inside the same request/task**.

- No Redis serialization of `ProviderConfigurations` (which would leak decrypted credentials across the trust boundary).
- No SQLAlchemy event listeners or invalidation at the ~30 provider write sites spread across `core/entities/provider_configuration.py`, `events/event_handlers/update_provider_when_message_created.py`, `commands/system.py`, `core/app/llm/quota.py`, etc.
- Cache dies at the end of the request/task — **zero staleness risk** for quota, credential, or load-balancing updates observed across requests.

## Changes

- **`core/provider_manager.py`**: add a module-level `RecyclableContextVar[dict[str, ProviderConfigurations]]` and memoize the result in `get_configurations` for the current scope, keyed by `tenant_id`. The cache is shared across `ProviderManager` instances because callers routinely spawn fresh ones via `create_plugin_provider_manager` (see `core/app/llm/model_access.py`, `services/workflow_service.py`, etc.).
- **`extensions/ext_celery.py`**: call `RecyclableContextVar.increment_thread_recycles()` inside `FlaskTask.__call__`, matching the Flask `before_request` hook in `app_factory.py`. This ensures per-task caches reset cleanly between Celery tasks that land on recycled worker threads. Strictly speaking this also benefits any existing `RecyclableContextVar` used from Celery paths (e.g. `plugin_tool_providers`).
- **`tests/unit_tests/core/test_provider_manager.py`**: cover the memoization behaviors behind an autouse fixture that bumps the recycle counter for each test.

## Verification

- `uv run --group dev basedpyright` → 0 errors, 0 warnings (project-wide).
- `uv run --group dev pytest tests/unit_tests/core/test_provider_manager.py tests/unit_tests/extensions/otel/test_celery_sqlcommenter.py tests/unit_tests/core/test_model_manager.py` → 40 passed.
- `uv run --group dev ruff check` / `ruff format --check` → clean.

## Test plan

- [x] Memoization returns the same object for repeated calls with the same `tenant_id` in one scope
- [x] Different `tenant_id`s are cached independently
- [x] Cache is shared across distinct `ProviderManager` instances in the same scope (the common caller pattern)
- [x] A new scope (`increment_thread_recycles`) starts empty
- [x] No cross-request leakage (cache is request/task-scoped only)
- [x] Existing `provider_manager` / `celery_sqlcommenter` unit tests still pass


youaodu:
Thanks! This saves me a lot of time.

PR #35110, opened 2026-04-13
refactor: replace bare dict with dict[str, Any] in VDB providers

wdeveloper16:
## Summary
Tighten bare `dict` / `dict | None` annotations across 12 vector database provider packages under `api/providers/vdb/`:
- Relyt, OpenSearch, Milvus, Lindorm, Huawei Cloud, Elasticsearch —  `validate_config.values`, `create_collection.metadatas`/`index_params`,  `similarity_search_with_score_by_vector.filter`
- PgVector, TiDB, Tablestore, Upstash, Vastbase, Weaviate —  `validate_config.values`

All VDB config values are provider-specific, so `dict[str, Any]` is the correct shape. `Any` is already imported in each file.

No behavior change — types only.

Part of #22651.

## Test plan
- [x] `make lint` passes
- [x] `make type-check-core` passes (1391 files)


crazywoola (approved):
Required checks are green and these type annotation tighten-ups look good to merge.

PR #35120, opened 2026-04-13
refactor(api): remove Any from file factory common helpers

carlos4s:
Part of #34878

## Summary
- replace `Mapping[str, Any]` with `Mapping[str, object]` in `api/factories/file_factory/common.py`
- add `get_mapping_str()` to centralize safe string extraction from persisted file mappings
- reuse the helper inside `resolve_mapping_file_id()`


asukaminato0721:
@dosu I think use object dont solve the problem here.

asukaminato0721 (changes requested):
see dosu comment

PR #35123, opened 2026-04-14
refactor: replace bare dict with dict[str, Any] in VDB providers and libs

wdeveloper16:
Replace bare `dict` with `dict[str, Any]` in `validate_config(cls, values:)` for 12 VDB providers: baidu, oracle, opengauss, oceanbase, alibabacloud-mysql, analyticdb (sql + openapi), couchbase, pgvecto-rs, clickzetta, matrixone, hologres. 
Also update `libs/sendgrid.py`, `libs/smtp.py`, `libs/exception.py`, `libs/helper.py`, and redis broadcast channel files. 

Add `from typing import Any` where missing.

Part of langgenius/dify#22651.

### DIFF
--- a/api/libs/broadcast_channel/redis/_subscription.py
+++ b/api/libs/broadcast_channel/redis/_subscription.py
@@ -3,7 +3,7 @@ import queue
 import threading
 import types
 from collections.abc import Generator, Iterator
-from typing import Self
+from typing import Any, Self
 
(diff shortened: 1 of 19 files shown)


crazywoola (approved):
Required checks are green and these type annotation tighten-ups look good to merge.

Issue #24138, opened 2025-08-18
[Chore/Refactor] use sqlalchemy engine


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

continue of #24115 

after it, all will be

```py
with Session(db.engine) as session:
    ...
```

then only db is from

```py
from flask_sqlalchemy import SQLAlchemy

db = SQLAlchemy(metadata=metadata)
```

then can use engine from SQLAlchemy itself.

### Motivation

One more step to remove flask_sqlalchemy

### Additional Context

part of #23647

asukaminato0721:
@dosu what is current progress

dataCenter430:
Hi, @dosu
are there still remaning tasks for this issue?
what is the next step to complete this?

dataCenter430:
@dosu what is the estimate remaining time for this issue.

PR #35228, opened 2026-04-14
refactor: replace db.session with sessionmaker in console controllers

wdeveloper16:
## Summary

- Migrates `controllers/console/admin.py` and  `controllers/console/explore/installed_app.py` to explicit `sessionmaker` sessions.
- `admin.py`: consolidates the previously mixed `session_factory.create_session()` +  `db.session` pattern in `InsertExploreAppListApi.post` into a single  `expire_on_commit=False` session so the lazy-loaded `app.site` relationship is
  accessible; `InsertExploreAppApi.delete` similarly consolidated from three separate  sessions into one; banner insert/delete and the chunked email-lookup loop use  targeted write sessions.
- `installed_app.py`: the `get` list-build is moved inside the session so the  lazy-loaded `installed_app.app` relationship is resolved before detach;  `post` consolidates all reads and the conditional insert into one session;  `delete` and `patch` use `session.merge()` to re-attach the passed-in  `installed_app` object before mutation.

## Test plan

- [ ] Explore apps — insert a new app into the explore list; update an existing one;  remove it via the admin delete endpoint.
- [ ] Trial apps — insert an explore app with `can_trial=true`; verify `TrialApp`  row is created.
- [ ] Banners — create and delete an explore banner.
- [ ] Installed apps — list, install, uninstall, and pin/unpin an app from the  explore page.
- [ ] Notification accounts — call `batch_add_notification_accounts` with a CSV file  and with a JSON body; verify matched account IDs are forwarded correctly.

Part of #24138.


wdeveloper16:
Hi, @asukaminato0721 
I'd be happy if you review this when  you have a chance. 😉 
Thanks,

asukaminato0721:
resolve conflict

asukaminato0721 (changes requested):
^

PR #35234, opened 2026-04-15
chore(api): migrate event handlers to use Session(db.engine)

jerryzai:
## Summary

- Replace `db.session` with `with Session(db.engine) as session:` in all linker event-handler files under `api/events/event_handlers/`
- Session scope is kept broad enough to cover every ORM object access in   each handler (e.g. the `IndexingRunner.run()` call in `create_document_index.py` runs inside the same `with` block so loaded `Document` instances remain accessible)

## Files changed

- `api/events/event_handlers/create_document_index.py`
- `api/events/event_handlers/create_installed_app_when_app_created.py`
- `api/events/event_handlers/create_site_record_when_app_created.py`
- `api/events/event_handlers/update_app_dataset_join_when_app_model_config_updated.py`
- `api/events/event_handlers/update_app_dataset_join_when_app_published_workflow_updated.py`

## Motivation

Part of #24138. Event handlers fire in Celery worker processes and therefore share the same concern as schedule tasks: they must not rely on the Flask-request-scoped `db.session`.

## Test plan

- [ ] Unit tests pass
- [ ] App creation, document indexing, and dataset-join update flows work  end-to-end in a local dev environment


jerryzai:
@asukaminato0721 please check this PR.

asukaminato0721 (approved):

PR #35235, opened 2026-04-15
chore(api): migrate mail task and OAuth data source to use Session(db…

jerryzai:
## Summary

- `mail_clean_document_notify_task.py`: wrap all DB operations in a single `with Session(db.engine) as session:` block so that ORM objects fetched early in the task (logs, tenants, accounts) remain live when their attributes are accessed later in the loop
- `oauth_data_source.py`: migrate each method (`get_access_token`, `save_internal_access_token`, `sync_data_source`) to use an explicit session; the `DataSourceOauthBinding` record is read, mutated, and committed within the same session block

## Files changed

- `api/schedule/mail_clean_document_notify_task.py`
- `api/libs/oauth_data_source.py`

## Motivation

Part of #24138. Both files perform multi-step read-then-write patterns that are a natural fit for pure SQLAlchemy unit-of-work sessions.

## Test plan

- [ ] Unit tests pass
- [ ] Document-clean notification emails are sent correctly in a sandboxed environment
- [ ] Notion OAuth token exchange and sync flows work correctly


asukaminato0721 (dismissed):

asukaminato0721:
for future: all this comment comes from origin code.

asukaminato0721 (approved):

PR #35263, opened 2026-04-15
fix: apply score threshold after reranking in hybrid search

### LINKED ISSUE #35233: Score threshold not applied to reranked score in hybrid search
### Self Checks

### Dify version

1.13.3

### Cloud or Self Hosted

Cloud

### Environment
- **Platform:** Dify Cloud
- **Retrieval mode:** Hybrid search
- **Embedding model:** `text-embedding-3-large` (OpenAI)
- **Rerank model:** `jina-reranker-v1-base-en` (Jina)

---

### Describe the bug

When using hybrid search with a rerank model configured, setting a score threshold of `0.60` returns empty results. However, lowering the threshold to `0.45` returns chunks that display scores of `0.84` – `0.96`.

If the returned chunks are displaying scores of `0.84` – `0.96`, they should have easily passed a threshold of `0.60`. This is contradictory and confusing behavior.

This indicates that the threshold is being applied against a **different score** than what is displayed in the result. The user sets a threshold expecting it to filter against the displayed score — but it does not.

---

### Steps to reproduce

1. Create a knowledge base with hybrid search enabled
2. Configure Jina reranker
3. Set score threshold to `0.60` on retrieval node
4. Send a query
5. Retrieval node returns **empty result**
Image

6. Lower threshold to `0.45` on a second retrieval node with the same query
7. Returns 10 chunks with **displayed scores of `0.84` – `0.96`**

---

### Proof

| Node | Threshold | Chunks returned | Displayed score range |
|---|---|---|---|
| Node 1 | `0.60` | 0 | — |
| Node 2 | `0.45` | 10 | `0.84` – `0.96` |

The same chunks that were blocked at threshold `0.60` display scores of `0.84` – `0.96` when retrieved at threshold `0.45`. This means the threshold and the displayed score are **not measuring the same thing**.

---

### Expected behavior

The score threshold should filter against the **reranked score** — which is the score displayed to the user in the result.

If a chunk displays a score of `0.84`, it must pass a threshold of `0.60`. The threshold value and the displayed score must be **consistent** so that users can set a meaningful threshold based on what they see.

---

### Actual behavior

The threshold filters against the **raw hybrid score** before reranking. The displayed score is the **reranked score**. These are two completely different scoring systems with different value distributions — but the same threshold value is silently applied to both without any indication to the user.

This means:
- A user sets threshold `0.60` expecting to filter out chunks scoring below `0.60`
- But chunks scoring `0.84` – `0.96` (displayed) are blocked
- The user has no visibility into the raw hybrid score that is actually used for filtering
- The displayed score gives a **false impression** of what passed the threshold

---

### Impact

- Valid queries return **empty results** at reasonable threshold values
- The score threshold setting is **misleading** — it does not behave as the UI implies
- Users are forced to use **unnecessary workarounds** such as multiple retrieval nodes with different thresholds to compensate for this behavior

---

### Current workaround

Two retrieval nodes are required:

| Node | Threshold | Purpose |
|---|---|---|
| Node 1 | `0.60` | Primary retrieval |
| Node 2 | `0.45` | Fallback when Node 1 returns empty |
| Code node | `0.65` | Manual filter on displayed reranked score |

This workaround should not be necessary.

---

### Suggested fix

When a rerank model is configured, the score threshold should be applied against the **reranked score** — the same score that is displayed to the user. The threshold and the displayed score must always refer to the same value.

---

### Related issue
#3146

aayushbaluni:
## Summary

Fixes #35233

In hybrid search with reranking enabled, the score threshold is applied to the pre-rerank/fusion score instead of the post-rerank score. This causes documents with high reranked scores (0.84-0.96) to be filtered out because their pre-rerank scores were below the threshold.

## Root Cause

`RetrievalService.embedding_search` passes `score_threshold` to the vector DB query for `HYBRID_SEARCH`, which filters on pre-rerank similarity scores. After reranking produces new scores, the threshold-filtered results are already gone.

## Fix

- For `HYBRID_SEARCH`, pass `score_threshold=None` to the vector DB search step so no pre-filtering occurs
- Let the reranking pipeline (`RerankModelRunner`, `WeightRerankRunner`) apply the threshold to the final reranked/fused scores
- When no rerank runner is configured, fall back to filtering by vector score threshold after retrieval

### DIFF
--- a/api/core/rag/datasource/retrieval_service.py
+++ b/api/core/rag/datasource/retrieval_service.py
@@ -196,6 +196,23 @@ class RetrievalService:
         return all_documents
 
     @classmethod
+    def _filter_documents_by_vector_score_threshold(
+        cls, documents: list[Document], score_threshold: float | None
+    ) -> list[Document]:
+        """Keep documents whose stored retrieval score meets the threshold.
+
+        Used when hybrid search skips early vector thresholding but no rerank
+        runner applies a threshold afterward (same rule as ``calculate_vector_score``).
+        """
+        if score_threshold is None:
+            return documents
+        return [
+            document
+            for document in documents
+            if document.metadata and document.metadata.get("score", 0) >= score_threshold
+        ]
+
+    @classmethod
     def _deduplicate_documents(cls, documents: list[Document]) -> list[Document]:
         """Deduplicate documents in O(n) while preserving first-seen order.
 
@@ -294,13 +311,20 @@ class RetrievalService:
 
                 vector = Vector(dataset=dataset)
                 documents = []
+                # Hybrid search merges keyword / full-text / vector hits and then reranks
+                # (weighted fusion or reranking model). Applying the user score threshold at
+                # vector retrieval time uses embedding similarity, which is not comparable to
+                # reranked or fused scores and incorrectly drops high-quality chunks (#35233).
+                embedding_score_threshold = (
+                    None if retrieval_method == RetrievalMethod.HYBRID_SEARCH else score_threshold
+                )
                 if query_type == QueryType.TEXT_QUERY:
                     documents.extend(
                         vector.search_by_vector(
                             query,
                             search_type="similarity_score_threshold",
                             top_k=top_k,
-                            score_threshold=score_threshold,
+                            score_threshold=embedding_score_threshold,
                             filter={"group_id": [dataset.id]},
                             document_ids_filter=document_ids_filter,
                         )
@@ -312,7 +336,7 @@ class RetrievalService:
                         vector.search_by_file(
                             file_id=query,
                             top_k=top_k,
-                            score_threshold=score_threshold,
+                            score_threshold=embedding_score_threshold,
                             filter={"group_id": [dataset.id]},
                             document_ids_filter=document_ids_filter,
                         )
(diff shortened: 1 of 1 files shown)


api/core/rag/datasource/retrieval_service.py
```diff
@@ -294,13 +311,20 @@ def embedding_search(
 
                 vector = Vector(dataset=dataset)
                 documents = []
+                # Hybrid search merges keyword / full-text / vector hits and then reranks
+                # (weighted fusion or reranking model). Applying the user score threshold at
+                # vector retrieval time uses embedding similarity, which is not comparable to
+                # reranked or fused scores and incorrectly drops high-quality chunks (#35233).
+                embedding_score_threshold = (
```
fatelei, line 318: embedding_score_threshold should be 0.0

fatelei (approved):

d5devgodai-blip:
Hi, I am using Dify Cloud and tested this after the PR was merged but the behavior appears unchanged — score threshold is still being applied before reranking in hybrid search. Could you confirm whether this fix has been deployed to Dify Cloud, or if I need to wait for a specific release? Thanks.

fatelei:
> Hi, I am using Dify Cloud and tested this after the PR was merged but the behavior appears unchanged — score threshold is still being applied before reranking in hybrid search. Could you confirm whether this fix has been deployed to Dify Cloud, or if I need to wait for a specific release? Thanks.

yes, in next release

PR #35325, opened 2026-04-16
fix: scope plugin inner API end-user lookup by tenant

### LINKED ISSUE #35323: plugin inner API can resolve EndUser across tenants when user_id is provided
### Self Checks

### Dify version

main, also reproducible from code shipped in 1.13.3

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

1. Call the plugin inner API path that goes through `controllers/inner_api/plugin/wraps.py`.
2. Provide a valid `tenant_id` for tenant A.
3. Provide a non-anonymous `user_id` that belongs to an `EndUser` row in tenant B.
4. Observe how `get_user()` resolves the foreign row before the request context is updated.

Minimal code-path reproduction:
- anonymous users are looked up with `session_id + tenant_id`
- non-anonymous users are currently looked up with `session.get(EndUser, user_id)`
- `get_user_tenant()` then injects that resolved user into the request context

### ✔️ Expected Behavior

A supplied `user_id` should stay tenant-scoped. If the user does not exist in the current tenant, Dify should not resolve an `EndUser` row from another tenant.

### ❌ Actual Behavior

`get_user()` can resolve an `EndUser` from another tenant when a concrete `user_id` is provided, because the non-anonymous lookup path does not include `tenant_id`.

That makes the tenant-scoping invariant inconsistent with the anonymous path and with other recent ownership hardening fixes in the codebase.

shaun0927:
## Summary

Restore tenant-scoped lookup semantics in `controllers/inner_api/plugin/wraps.py` for explicit end-user IDs.

Today the plugin inner API already documents that `user_id` is untrusted input, but the non-anonymous path uses `session.get(EndUser, user_id)`, which can resolve a row from another tenant before the request context is updated.

This PR keeps the existing anonymous behavior unchanged and only narrows the non-anonymous lookup to:
- `EndUser.id == user_id`
- `EndUser.tenant_id == tenant_id`

If no matching row exists in the current tenant, the existing fallback path still creates a new end user for the request.

## Test plan

- Added a regression test in `api/tests/unit_tests/controllers/inner_api/plugin/test_plugin_wraps.py`
- Reproduced the pre-fix behavior locally with a lightweight harness that returned a foreign-tenant `EndUser`
- Verified the same harness no longer resolves the foreign row after the code change

Fixes #35323

### DIFF
--- a/api/controllers/inner_api/plugin/wraps.py
+++ b/api/controllers/inner_api/plugin/wraps.py
@@ -20,10 +20,13 @@ class TenantUserPayload(BaseModel):
 
 def get_user(tenant_id: str, user_id: str | None) -> EndUser:
     """
-    Get current user
+    Get current user.
 
     NOTE: user_id is not trusted, it could be maliciously set to any value.
-    As a result, it could only be considered as an end user id.
+    As a result, it could only be considered as an end user id. Even when a
+    concrete end-user ID is supplied, lookups must stay tenant-scoped so one
+    tenant cannot bind another tenant's user record into the plugin request
+    context.
     """
     if not user_id:
         user_id = DefaultEndUserSessionID.DEFAULT_SESSION_ID
@@ -42,7 +45,14 @@ def get_user(tenant_id: str, user_id: str | None) -> EndUser:
                     .limit(1)
                 )
             else:
-                user_model = session.get(EndUser, user_id)
+                user_model = session.scalar(
+                    select(EndUser)
+                    .where(
+                        EndUser.id == user_id,
+                        EndUser.tenant_id == tenant_id,
+                    )
+                    .limit(1)
+                )
 
             if not user_model:
                 user_model = EndUser(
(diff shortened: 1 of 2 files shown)


shaun0927:
Follow-up pushed in `e6c5791`. The product fix is unchanged; this only aligns the plugin-wrap unit mocks with the new tenant-scoped `session.scalar(select(...))` path so the focused backend test now passes locally.

Verification rerun:
- `cd api && source .venv/bin/activate && python -m pytest -o addopts='' tests/unit_tests/controllers/inner_api/plugin/test_plugin_wraps.py -q`
- result: `17 passed`

shaun0927:
Re-checked this in a fresh local worktree from the current PR head (`3034abd`).

- The fix is still tightly scoped to the tenant filter in `api/controllers/inner_api/plugin/wraps.py` plus the focused regression coverage in `api/tests/unit_tests/controllers/inner_api/plugin/test_plugin_wraps.py`.
- Local targeted verification passed: `uv run --project api pytest api/tests/unit_tests/controllers/inner_api/plugin/test_plugin_wraps.py -q` → `17 passed`.
- The existing bot feedback on this PR is informational only (`pyrefly-diff` summary); I don’t see any unresolved review comments requesting follow-up changes.

From my side this looks ready for maintainer review once the remaining CI jobs finish.

fatelei (approved):

PR #35326, opened 2026-04-16
fix(web): stop Splash after useUserProfile errors

shaun0927:
## Summary

Keep the first-screen optimization from #35313, but stop `Splash` from blocking the page after the shared profile query has already failed.

After the migration to `useUserProfile()`, the component still renders the loader whenever `!data?.profile`. In the error case, that means the overlay can remain visible even though the request is no longer pending.

This PR narrows the blocking condition to the actual pending state:
- render the loader while the profile query is pending
- fall through when the query errors or returns no profile payload

That preserves the deduplicated `/account/profile` request while avoiding an indefinite loading state.

## Test plan

- Added `web/app/components/__tests__/splash.spec.tsx`
- Verified the splash renders during the pending state
- Verified the splash disappears when the query succeeds
- Verified the splash also disappears when the query errors
- Ran: `PATH=/tmp/node-v22.22.1-darwin-arm64/bin:$PATH corepack pnpm test app/components/__tests__/splash.spec.tsx`

Fixes #35324

### DIFF
--- a/web/app/components/__tests__/splash.spec.tsx
+++ b/web/app/components/__tests__/splash.spec.tsx
@@ -1,0 +1,59 @@
+import type { MockedFunction } from 'vitest'
+import { render, screen } from '@testing-library/react'
+import { useUserProfile } from '@/service/use-common'
+import Splash from '../splash'
+
+vi.mock('@/service/use-common', () => ({
+  useUserProfile: vi.fn(),
+}))
+
+const mockUseUserProfile = useUserProfile as MockedFunction<typeof useUserProfile>
+
+describe('Splash', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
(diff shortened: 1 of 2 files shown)


lyzno1:
Thanks for the contribution! The fix can be simplified further — since every path after `isPending` returns `null`, you only need to remove `|| !data?.profile` from the original condition and drop the extra `isError` / `data` destructuring entirely.

shaun0927:
Applied the simplification requested in review in `260915a`: `Splash` now only reads `isPending` from `useUserProfile()`, returns the loading overlay while pending, and falls through to `null` for every settled path.

Verification rerun with Node 22:
- `cd web && PATH=/tmp/node-v22.22.1-darwin-arm64/bin:$PATH corepack pnpm test -- --run app/components/__tests__/splash.spec.tsx`
- result: `3 passed`

lyzno1 (approved):

PR #35335, opened 2026-04-16
fix: splash keeps rendering when useUserProfile fails

Protocol-zero-0:
Fixes #35324.

The `Splash` component was previously rendering the loading state whenever `isPending || !data?.profile` was true. However, if the `useUserProfile` query failed, `data` would remain undefined, causing the splash screen to stay visible indefinitely.

This PR adds a check for `isError` to ensure the splash screen is dismissed if the profile query fails, allowing the normal error or redirect flow to proceed.


crazywoola:
Duplicated https://github.com/langgenius/dify/pull/35326

PR #35359, opened 2026-04-17
fix: avoid app detail refetch on tab switches

### LINKED ISSUE #40791: bug(web): app detail is refetched on in-app tab switches
-

hyl64:
## Summary
This PR stops app detail from being cleared and refetched on in-app tab switches by only refetching when the appId changes.

Fixes #40791

## Verification
- targeted frontend test passed locally
- eslint, tsgo --noEmit, and knip commit checks passed

### DIFF
--- a/web/__tests__/app/app-detail-layout-flow.test.tsx
+++ b/web/__tests__/app/app-detail-layout-flow.test.tsx
@@ -1,0 +1,126 @@
+import { act, render, screen, waitFor } from '@testing-library/react'
+import { useEffect } from 'react'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import LayoutMain from '@/app/(commonLayout)/app/(appDetailLayout)/[appId]/layout-main'
+import { useStore as useAppStore } from '@/app/components/app/store'
+import { AppModeEnum } from '@/types/app'
+
+const mockFetchAppDetailDirect = vi.fn()
+const mockReplace = vi.fn()
+
+let mockPathname = '/app/app-1/configuration'
+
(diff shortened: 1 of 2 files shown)


crazywoola:
Hi @hyl64! Thanks so much for taking the time to contribute to Dify — we really appreciate it.

hyl64:
Thanks for reopening this.

While rebasing onto current `main` to resolve the reported conflict, I noticed the source-line change this PR makes is already effectively present upstream:

- On the base this PR targeted, the fetch effect dependency was `[appId, pathname]`.
- On current `main` (after the refactor in #37433), it is `[appId, router, setAppDetail]` — `pathname` is gone, and there's also an early return that skips fetch when the store already has the same `appId`'s detail. So tab switches no longer trigger a refetch.

That means `web/app/(commonLayout)/app/(appDetailLayout)/[appId]/layout-main.tsx` no longer needs the one-line change, and rebasing cleanly produces a no-op for that file (diff/mostly conflicts coming from the larger #37433 refactor, not from this PR's intent).

The remaining value I can see in keeping this PR is the test added here, `web/__tests__/app/app-detail-layout-flow.test.tsx`, re-adapted to the current store/atom/i18n structure.

Two questions before I push the rebase:

1. Do you want me to keep this PR as "test only" (re-adapt the test to current main, drop the now-redundant layout-main.tsx change), since the fix itself is already in main?
2. Or would you prefer to close this PR and track the (optional) test coverage separately?

Happy to proceed either way — just want to confirm the direction so I don't push something you don't want.

PR #35389, opened 2026-04-19
fix: handle numpy scalar types in safe_json_value

avasis-ai:
## Summary
Add explicit handling for `np.integer` and `np.floating` scalar types in `safe_json_value()`.

## Problem
`safe_json_value()` handles `np.ndarray` (via `.tolist()`) but misses numpy scalar types like `np.int64`, `np.float64`, etc. These fall through to the `else` clause and are returned as-is. Since `json.dumps()` cannot serialize numpy scalar types, this causes a `TypeError` at runtime when tool results contain numpy scalars.

In numpy >= 2.0, scalar types no longer inherit from Python builtins, making this even more likely to fail.

## Change
Added two checks before the existing `np.ndarray` handling:
- `np.integer` → `int(v)`
- `np.floating` → `float(v)`

Related to #34405 (parameter precision loss with numeric types).

### DIFF
--- a/api/core/tools/utils/message_transformer.py
+++ b/api/core/tools/utils/message_transformer.py
@@ -41,6 +41,10 @@ def safe_json_value(v):
             return v.hex()
     elif isinstance(v, memoryview):
         return v.tobytes().hex()
+    elif isinstance(v, np.integer):
+        return int(v)
+    elif isinstance(v, np.floating):
+        return float(v)
     elif isinstance(v, np.ndarray):
         return v.tolist()
     elif isinstance(v, dict):


fatelei (approved):

fatelei:
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================ tests coverage ================================
_______________ coverage: platform linux, python 3.12.3-final-0 ________________

Coverage XML written to file coverage.xml
Coverage JSON written to file coverage.json
=========================== short test summary info ============================
FAILED api/tests/unit_tests/core/mcp/session/test_base_session.py::test_receive_loop_cancel_notification - AttributeError: 'NoneType' object has no attribute 'message'
===== 1 failed, 10659 passed, 4 skipped, 297 warnings in 311.62s (0:05:11) =====

PR #35451, opened 2026-04-21
feat(human-input): expose selected action value

### LINKED ISSUE #34597: Expose selected action value as an output in HITL Human Input(and extend its length as well)
### Self Checks

Image

I am using the HITL (`Human Input`) node to pause a workflow and let an operator choose one of several actions. In the current UI, each user action has two editable fields. The left field is the action identifier, which is exposed downstream as `__action_id`. The right field is another action field, but it is not currently exposed as an output.

That right-hand field is sometimes exactly the value downstream nodes need. In practice, I would like to treat it as an `__action_value` and use it in later workflow steps. For example, after the user clicks a HITL action, downstream nodes may need to receive a card number, serial ID, internal code, or some other business value associated with that selection.

Right now, only `__action_id` is exposed, so the workflow can tell which action was clicked, but it cannot directly access the corresponding value from the right-hand field. This makes HITL less useful in workflows where the human choice needs to carry real business data into downstream nodes.

The attached illustration highlights the current situation:

- the left action field maps to `__action_id`
- the right action field is not exposed in `Output Variables`
- the HITL node currently exposes `__action_id` and `__rendered_content`, but not the value from the right-hand action field

The requested enhancement is:

- expose the right-hand action field as an output, for example `__action_value`
- keep `__action_id` as the existing output for the left-hand field
- make the selected action's value available to downstream nodes

This also connects to the current 20-character limit. Once that right-hand field becomes a real downstream value, the current length restriction gets in the way, because action values may need to hold business-meaningful data such as:

- card numbers
- serial IDs
- internal codes
- other longer identifiers

At the moment, this limit is hard-coded as `20`, which makes the feature much less practical for real workflows.

So if the right-hand field is intended to serve as the actual action value, it is important not only to expose it, but also to make its allowed length configurable, or at least raise the limit significantly instead of keeping it fixed at 20 characters.

If the intended design is different, and the right-hand field is only meant to be button display text, then an alternative enhancement would be:

- keep the current right-hand field for display text only
- add a third field for each user action
- expose that third field as the real downstream value, for example `__action_value`
- make the business value field length configurable instead of hard-coded to `20`

Example desired behavior:

1. Configure a HITL node with an action where the left field is `approve`.
2. Set the right field, or a dedicated third field, to a business value such as `card_visa_enterprise_001`.
3. The user clicks that action in the Human Input form.
4. The node outputs:
   - `__action_id`: the selected action ID from the left field
   - `__action_value` (or similar): the business value associated with that action
   - `__rendered_content`: rendered form content
5. Downstream nodes can use the selected business value directly.

This would make HITL much more practical for workflows where manual choices need to pass real business values into later steps. Exposing the value alone is not enough if the business value is still constrained by a hard-coded 20-character limit.

Blackoutta:
## Summary

Implements the Dify-side companion for [langgenius/dify#34597](https://github.com/langgenius/dify/issues/34597).

This PR exposes the selected Human Input action value in preview/editor flows and updates the Human Input UI so the right-hand field is treated as the downstream action value.

## Changes

- add `__action_value` to Human Input preview submission outputs
- expose `__action_value` in Human Input output variable metadata
- update Human Input editor copy so the right-hand field is presented as `Action Value`
- raise the frontend action value limit from 20 to 100 while keeping the action id limit at 20
- avoid duplicate built-in variable injection by keeping Human Input built-ins in the shared workflow constants
- update backend and frontend tests

Closes #34597.

image

## Dependency

This PR depends on the companion Graphon work:

- Graphon issue: [langgenius/graphon#77](https://github.com/langgenius/graphon/issues/77)
- Graphon PR: [langgenius/graphon#78](https://github.com/langgenius/graphon/pull/78)

## Verification

```bash
uv run --project api --dev pytest ./api/tests/unit_tests/services/test_workflow_service.py -k test_submit_human_input_form_preview_success -q
pnpm exec vp test app/components/workflow/nodes/human-input/components/__tests__/user-action.spec.tsx app/components/workflow/nodes/human-input/__tests__/panel.spec.tsx app/components/workflow/nodes/human-input/__tests__/human-input.spec.tsx app/components/workflow/hooks/__tests__/use-workflow-variables.spec.ts
```


Blackoutta:
Release-order note for reviewers: this PR depends on the companion Graphon changes in [langgenius/graphon#78](https://github.com/langgenius/graphon/pull/78) / [langgenius/graphon#77](https://github.com/langgenius/graphon/issues/77).

Those Graphon changes need to be merged first, and a Graphon release containing them needs to be published first. Only after Dify can consume that released Graphon version should this PR be merged and included in a Dify release.

In short: merge/release Graphon first, then merge/release this Dify PR.

QuantumGhost (approved):

PR #35464, opened 2026-04-21
fix(plugin): persist tenant plugin auto-upgrade strategy changes

zyssyz123:
## Summary

`PluginAutoUpgradeService.change_strategy` and `exclude_plugin` opened a SQLAlchemy session via `with session_factory.create_session() as session:` but **never started a transaction or called `session.commit()`**. The session factory is a plain `sessionmaker(bind=engine)` without autocommit, so the session context manager only closes the session on exit and silently discards any pending INSERT/UPDATE.

As a result, any write to `tenant_plugin_auto_upgrade_strategies` from these two methods was lost, while callers still saw a successful `{\"success\": true}` response.

## Symptom

Reproduced on the plugin preferences page:

1. Admin opens the plugin reference-setting modal and adds a plugin to the exclude list.
2. Saving the form calls `POST /console/api/workspaces/current/plugin/preferences/change`, which returns 200 with `{\"success\": true}`.
3. Reopening the modal (`GET /preferences/fetch`) returns an empty `exclude_plugins` array.
4. Inspecting the DB row confirms the update never happened — `updated_at` still equals `created_at` down to the microsecond, and `exclude_plugins` remains `[]`.

The same code path is also used by `POST /preferences/autoupgrade/exclude`, which suffered the same silent data loss.

## Fix

Wrap both methods in `session.begin()`, matching the pattern already used in `PluginPermissionService.change_permission`. SQLAlchemy auto-commits the transaction when the `session.begin()` context exits normally.

```python
with session_factory.create_session() as session, session.begin():
    ...
```

Scope kept intentionally minimal — behaviour of the success / error paths is unchanged, only the commit semantics are fixed.

## Test plan

- [ ] Save plugin preferences with a non-empty exclude list → refresh → the list is preserved.
- [ ] After saving, verify in DB that `tenant_plugin_auto_upgrade_strategies.updated_at` advances past `created_at` and `exclude_plugins` contains the expected plugin ids.
- [ ] Trigger `/preferences/autoupgrade/exclude` for a single plugin on both a tenant with an existing strategy row and a tenant without one → verify the row is created / updated.
- [ ] Verify existing happy-path tests for `/preferences/change` and `/preferences/fetch` still pass.

## Related

- `api/services/plugin/plugin_auto_upgrade_service.py`
- `api/services/plugin/plugin_permission_service.py` (pattern reference)
- `api/controllers/console/workspace/plugin.py` — `PluginChangePreferencesApi`, `PluginFetchPreferencesApi`, `PluginAutoUpgradeExcludePluginApi`


CodingOnStar (approved):
LGTM

PR #35574, opened 2026-04-26
fix(web): preserve completed thought duration

### LINKED ISSUE #35424: Reasoning duration resets to 0.0s after switching browser tab/window in Chatflow
### Self Checks

### Dify version

1.13.3

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Open a Chatflow conversation page that uses a reasoning-capable model.
2. Send a message and wait until the reasoning duration is displayed normally and the response is fully completed.
3. Switch to another browser tab or another application window.
4. Switch back to the Chatflow page.
5. Check the displayed reasoning duration.

### ✔️ Expected Behavior

The final reasoning duration of a completed message should remain unchanged after switching tabs/windows or after the page regains focus.

It should continue showing the actual completed duration instead of being reset to 0.0s.

### ❌ Actual Behavior

The reasoning duration, which was previously displayed correctly, is reset to 0.0s.

princepal9120:
## Summary
- preserve the final elapsed time for completed reasoning/thought blocks across remounts
- reuse the cached completed duration when a finished thought block is rendered again
- add regression coverage for remounting a completed thought block

## Tests
- `git diff --check`

Could not run the targeted web test locally because this sparse checkout does not have `web/node_modules` installed, so `pnpm test app/components/base/markdown-blocks/__tests__/think-block.spec.tsx --runInBand` fails before running tests with `command not found: vp`.

Fixes #35424


Qodo-Free-For-OSS:
Hi, On remount of an in-progress think block, elapsedTime initializes from cachedElapsedTime but subsequent ticks recompute from a fresh startTime, so the displayed time can decrease and the cache is overwritten with smaller values.

**Severity:** action required | **Category:** correctness

**How to fix:** Anchor startTime to cached time

**Agent prompt to fix** - you can give this to your LLM of choice:

> ### Issue description
> `useThinkTimer` uses a cached elapsed time but resets `startTime` on remount. When the component remounts while still "thinking" (not complete), the next interval tick computes elapsed time from the new `startTime` and can reduce the displayed time, also overwriting the cache with a smaller value.
> 
> ### Issue Context
> The PR introduces `completedThinkDurations` caching to preserve durations across remounts. This is correct for completed blocks, but the current implementation also writes to the cache during active ticking, which makes in-progress remounts problematic.
> 
> ### Fix Focus Areas
> - web/app/components/base/markdown-blocks/think-block.tsx[55-94]
> 
> ### Suggested approach
> - Either **only cache on completion** (don’t write in the interval), or
> - If caching during ticking is required, adjust `startTime` so that `Date.now() - startTime` starts at `cachedElapsedTime` on mount (e.g., `const [startTime] = useState(() => Date.now() - cachedElapsedTime * 1000)`), and ensure cache updates are monotonic (`Math.max(previous, current)`), so time never goes backward.

We noticed a couple of other issues in this PR as well - happy to share if helpful.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review. FYI, Qodo is free for open-source.*

princepal9120:
Addressed the Qodo review finding: both cache write sites now use `Math.max` to ensure elapsed time is always monotonically increasing, preventing any possibility of time going backward on remount.

The `startTime` anchoring (`Date.now() - cachedElapsedTime * 1000`) was already in place from the original PR — this commit adds the monotonic guard as the additional defensive layer Qodo suggested.

PR #35640, opened 2026-04-28
fix: hit-testing response failed because of Pydantic check.

FFXN:
> [!IMPORTANT]
>

## Summary

This PR fixes a 400 invalid_param regression in the dataset hit-testing API introduced after the response layer was tightened with Pydantic validation. The underlying retrieval flow could still return the legacy query={"content": ...} shape, and some optional collection fields such as segment.keywords, child_chunks, and files could be None, which no longer matched the current response schema expecting a str and concrete list values.

The change adds a small compatibility normalization step in the shared DatasetsHitTestingBase response path without altering retrieval behavior, ranking, scoring, or filtering. It converts the legacy query object into a plain string and normalizes nullable collection fields to empty lists before response validation. Targeted tests were also added to cover the legacy query shape and nullable list field cases, ensuring both console and service API hit-testing endpoints return schema-compliant responses consistently.

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


crazywoola (approved):

PHclaw:
For workflow execution timeout issues:

The default timeout for workflow execution is often too short. Increase it in the configuration:

```yaml
# docker-compose.yaml
services:
  api:
    environment:
      WORKFLOW_EXECUTION_TIMEOUT: 3600  # 1 hour instead of default 5 minutes
```

Also implement async timeout handling:
```python
import asyncio

async def run_workflow(workflow_id, inputs, timeout=3600):
    try:
        result = await asyncio.wait_for(
            execute_workflow(workflow_id, inputs),
            timeout=timeout
        )
        return result
    except asyncio.TimeoutError:
        # Save partial state and return error
        partial_state = await save_partial_state(workflow_id)
        return {"error": "timeout", "partial_state": partial_state}
```

PHclaw:
For the workflow node timeout:

```python
class WorkflowNode:
    timeout: int = 300

    async def execute(self, inputs):
        try:
            return await asyncio.wait_for(
                self._run(inputs), timeout=self.timeout
            )
        except asyncio.TimeoutError:
            return {'status': 'timeout', 'node': self.name}
```

PR #35656, opened 2026-04-28
fix(api): sanitize CSV values starting with line feed

princepal9120:
## Summary

- treat leading line feed (`\n`) as a CSV formula injection prefix
- add unit coverage for values that start with a line feed
- update sanitizer docs to include line feed alongside tab and carriage return

## Test

- `PYTHONPATH=. python3 - <<'PY' ...` smoke checks for `CSVSanitizer.sanitize_value`
- `PYTHONPATH=. python3 - <<'PY' ...` executed all `TestCSVSanitizer` test methods directly

`uv run pytest tests/unit_tests/core/helper/test_csv_sanitizer.py -q` could not complete in this workspace because dependency setup hit `No space left on device` while populating the uv cache.

### DIFF
--- a/api/core/helper/csv_sanitizer.py
+++ b/api/core/helper/csv_sanitizer.py
@@ -12,12 +12,12 @@ class CSVSanitizer:
     (Excel, LibreOffice, Google Sheets).
 
     Formula injection occurs when user-controlled data starting with special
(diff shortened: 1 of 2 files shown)


princepal9120:
Closing this PR because it was not tied to an existing open issue. I'll follow the issue-first workflow for future fixes.

PR #35710, opened 2026-04-29
fix: preserve single-run input variable types

Jingyi-Dify:
> [!IMPORTANT]
>

## Summary

Fixes single-step run inputs so referenced variables keep their expected input types instead of falling back to text inputs.

## Change List

- Use `value_type` as a fallback when a single-run variable cannot be resolved from workflow output metadata.
- Resolve `sys.*` variables by their full variable name across available workflow variable groups.
- Preserve number input behavior for system variables such as `sys.timestamp` and `sys.dialogue_count`.

## Testing

- Verified the diff is limited to `web/app/components/workflow/nodes/_base/hooks/use-one-step-run.ts`.
- Manually checked the affected single-run input paths for normal variables and system variables.

## Checklist


lyzno1 (approved):
No blocking issues found.

Verified locally:
- `pnpm test app/components/workflow/nodes/_base/hooks/__tests__/use-one-step-run.spec.ts`
- `pnpm type-check`
- `pnpm exec eslint app/components/workflow/nodes/_base/hooks/__tests__/use-one-step-run.spec.ts`

Small non-blocking suggestion: consider adding one more fallback test for an unresolved variable with `options`, asserting it becomes `InputVarType.select` and preserves `options`. The implementation already handles this path, but the current spec only covers the number fallback.

PR #35762, opened 2026-05-01
fix: batch workflow online users requests

princepal9120:
## Summary
- split workflow online users lookups into batches of 50 app IDs
- preserve support for both object-map and array response shapes
- add coverage for empty input and batched requests

## Tests
- git diff --check
- cd web && corepack pnpm test service/__tests__/apps.spec.ts
- cd web && corepack pnpm type-check

### DIFF
--- a/web/service/__tests__/apps.spec.ts
+++ b/web/service/__tests__/apps.spec.ts
@@ -1,0 +1,63 @@
+import { fetchWorkflowOnlineUsers } from '../apps'
+import { consoleClient } from '../client'
+
+vi.mock('../client', () => ({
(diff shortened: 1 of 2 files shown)


lyzno1:
Thanks for the fix.

I’m going to close this PR for now because frontend-side batching feels like a workaround for an API contract issue. This endpoint already has a backend batch limit, but the frontend would need to hardcode that limit and multiply a periodically-polled request as the app list grows.

Could you open an issue first so we can discuss the expected API contract and fallback behavior? We can revisit the frontend change after that is clarified.

PR #35786, opened 2026-05-04
fix: workflow online users polling for large app lists

### LINKED ISSUE #35761: After loading more than 50 apps in a workspace, a bad request error occurred when get workflow online users
### Self Checks

### Dify version

1.14.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

After loading more than **50 apps** in a workspace, a bad request error occurred when getting workflow online users (/apps/workflows/online-users)

Image

### ✔️ Expected Behavior

No error is reported and online users are displayed normally

### ❌ Actual Behavior

The `/apps/workflows/online-users` API has a hardcoded limit `MAX_WORKFLOW_ONLINE_USERS_QUERY_IDS = 50`
However, the front-end logic did not perform batch processing for situations exceeding this limit

hjlarry:
> [!IMPORTANT]
>

## Summary

fix https://github.com/langgenius/dify/issues/35761

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/controllers/console/app/workflow.py
+++ b/api/controllers/console/app/workflow.py
@@ -60,7 +60,8 @@ _file_access_controller = DatabaseFileAccessController()
 LISTENING_RETRY_IN = 2000
 DEFAULT_REF_TEMPLATE_SWAGGER_2_0 = "#/definitions/{model}"
 RESTORE_SOURCE_WORKFLOW_MUST_BE_PUBLISHED_MESSAGE = "source workflow must be published"
-MAX_WORKFLOW_ONLINE_USERS_QUERY_IDS = 50
+MAX_WORKFLOW_ONLINE_USERS_REQUEST_IDS = 1000
+WORKFLOW_ONLINE_USERS_REDIS_BATCH_SIZE = 50
 
 # Register models for flask_restx to avoid dict type issues in Swagger
 # Register in dependency order: base models first, then dependent models
@@ -158,8 +159,13 @@ class WorkflowFeaturesPayload(BaseModel):
     features: dict[str, Any] = Field(..., description="Workflow feature configuration")
 
 
-class WorkflowOnlineUsersQuery(BaseModel):
-    app_ids: str = Field(..., description="Comma-separated app IDs")
+class WorkflowOnlineUsersPayload(BaseModel):
+    app_ids: list[str] = Field(default_factory=list, description="App IDs")
+
+    @field_validator("app_ids")
+    @classmethod
+    def normalize_app_ids(cls, app_ids: list[str]) -> list[str]:
+        return list(dict.fromkeys(app_id.strip() for app_id in app_ids if app_id.strip()))
 
 
 class DraftWorkflowTriggerRunPayload(BaseModel):
@@ -186,7 +192,7 @@ reg(ConvertToWorkflowPayload)
 reg(WorkflowListQuery)
 reg(WorkflowUpdatePayload)
 reg(WorkflowFeaturesPayload)
-reg(WorkflowOnlineUsersQuery)
+reg(WorkflowOnlineUsersPayload)
 reg(DraftWorkflowTriggerRunPayload)
 reg(DraftWorkflowTriggerRunAllPayload)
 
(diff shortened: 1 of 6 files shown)


lyzno1:
I think this fix is addressing the symptom through DOM visibility rather than fixing the API behavior for the app-list use case.

The reported failure happens because the app list can naturally contain more than 50 loaded apps, while `/apps/workflows/online-users` rejects requests above `MAX_WORKFLOW_ONLINE_USERS_QUERY_IDS`. Querying `[data-app-id]` from the rendered DOM and filtering by `getBoundingClientRect()` reduces the number of IDs in common cases, but it does not guarantee the request will stay within the backend limit. On large screens, high zoom levels, different grid density, or future layout changes, the visible range can still contain more than 50 cards.

It also makes the data flow harder to reason about: the parent list now depends on a DOM attribute inside `AppCard`, newly loaded pages do not trigger an immediate online-user refresh because `pages` is no longer part of the refresh dependency, and scrolling to another area can show stale or empty online-user state until the next polling interval.

I suggest fixing this at the API boundary instead. The app-list use case is effectively: "return workflow online users for these loaded apps." The backend should accept that business request, deduplicate app IDs, apply tenant/access filtering, and internally batch or pipeline the Redis reads as needed. The frontend contract can then stay focused on the business shape of the request, instead of exposing the backend implementation limit to the component layer.

The frontend should keep the list data flow explicit, for example deriving IDs from the loaded app data and filtering to app modes that can actually have workflow collaboration presence. It should not need to infer business data from DOM layout or know how many Redis lookups the backend wants to perform per internal batch.

hjlarry:
> I think this fix is addressing the symptom through DOM visibility rather than fixing the API behavior for the app-list use case.
> 
> The reported failure happens because the app list can naturally contain more than 50 loaded apps, while `/apps/workflows/online-users` rejects requests above `MAX_WORKFLOW_ONLINE_USERS_QUERY_IDS`. Querying `[data-app-id]` from the rendered DOM and filtering by `getBoundingClientRect()` reduces the number of IDs in common cases, but it does not guarantee the request will stay within the backend limit. On large screens, high zoom levels, different grid density, or future layout changes, the visible range can still contain more than 50 cards.
> 
> It also makes the data flow harder to reason about: the parent list now depends on a DOM attribute inside `AppCard`, newly loaded pages do not trigger an immediate online-user refresh because `pages` is no longer part of the refresh dependency, and scrolling to another area can show stale or empty online-user state until the next polling interval.
> 
> I suggest fixing this at the API boundary instead. The app-list use case is effectively: "return workflow online users for these loaded apps." The backend should accept that business request, deduplicate app IDs, apply tenant/access filtering, and internally batch or pipeline the Redis reads as needed. The frontend contract can then stay focused on the business shape of the request, instead of exposing the backend implementation limit to the component layer.
> 
> The frontend should keep the list data flow explicit, for example deriving IDs from the loaded app data and filtering to app modes that can actually have workflow collaboration presence. It should not need to infer business data from DOM layout or know how many Redis lookups the backend wants to perform per internal batch.

sounds reasonable. However, since this is a GET request, there's a risk of the URL becoming too long, which could hit limits imposed by some browsers, proxies, or gateways. what do you think about this? change to a `post` request?

lyzno1:
> sounds reasonable. However, since this is a GET request, there's a risk of the URL becoming too long, which could hit limits imposed by some browsers, proxies, or gateways. what do you think about this? change to a `post` request?


Yes, I agree URL length is a valid concern. For this use case, I think a POST endpoint with a JSON body would be cleaner:

POST /apps/workflows/online-users
{ "app_ids": [...] }

Then the backend can deduplicate, apply access filtering, enforce a reasonable total limit, and internally batch/pipeline Redis reads. The frontend contract stays focused on the business request and does not need to know the internal per-batch limit.

Are there any compatibility or deployment constraints we should consider, such as existing callers, gateway rules, or API compatibility requirements?

hjlarry:
> 
> Are there any compatibility or deployment constraints we should consider, such as existing callers, gateway rules, or API compatibility requirements?

if change to a post request, seems there is not any compatibility constriants.

lyzno1 (approved):

fatelei (approved):

Issue #35699, opened 2026-04-29
Follow Up On Huntr Reports


zafido:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

latest

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Hi maintainers,

I have opened several reports on Huntr which have now became public. I'm reaching out here to find out what is the current status, and how we can proceed - of course, I'm here for anything needed. 
Note that I also disclosed through GitHub - those are private at the moment, but are clones of the public Huntr submissions.

| Name | Huntr | GitHub | Notes |
| :--- | :--- | :--- | :--- |
| Cross-Tenant App Trace Configuration | https://huntr.com/bounties/a43076b2-fbc8-4750-9647-89a036b52f52 | https://github.com/langgenius/dify/security/advisories/GHSA-48xc-wmw8-3jr3 | Fix available in GitHub report |
| Cross-User File Read | https://huntr.com/bounties/181136ec-d957-4b75-8ea7-6fa7b8abd01d | https://github.com/langgenius/dify/security/advisories/GHSA-3wpp-8x73-fm48 | Fix available in GitHub report |
| File Preview Authorization Bypass | https://huntr.com/bounties/d50a0240-7951-4939-b989-9bded66c7682 | https://github.com/langgenius/dify/security/advisories/GHSA-2qwc-c2cc-2xwv | |
| Path Traversal in Plugin Daemon Inner API | https://huntr.com/bounties/35b7ad59-e35d-443f-bf77-387bfb932ec0 | https://github.com/langgenius/dify/security/advisories/GHSA-gvc6-fh3x-89xh | Fix available in GitHub report |

Thank you!!!

### ✔️ Expected Behavior

N/A

xr843:
[...]

PR #35797, opened 2026-05-05
fix(security): tenant-scope FilePreviewApi text-extract endpoint (GHSA-2qwc-c2cc-2xwv)

xr843:
## Summary

`api/controllers/console/files.py FilePreviewApi.get` (the console-authenticated `/files/<file_id>/preview` text-extract endpoint) only checked authentication; the underlying `FileService.get_file_preview(file_id)` looked the file up by `id` alone, with no tenant filter. An authenticated user from tenant A could read the extracted-text preview of any `UploadFile` whose UUID they could obtain (or guess), including files belonging to other tenants.

## Fix

Two surgical changes:

1. **Service layer** — `FileService.get_file_preview` now takes `tenant_id` and adds `UploadFile.tenant_id == tenant_id` to the lookup query.
2. **Controller** — `FilePreviewApi.get` resolves the caller's tenant via `current_account_with_tenant()` and passes it through.

5 existing test sites (3 integration, 2 unit) updated for the new signature.

## Scope of audit

I greppped for other callers of `get_file_preview`. Only the console controller and the 5 test sites — no other production callers to update.

The HMAC-signed file endpoints (`api/controllers/files/image_preview.py` `ImagePreviewApi` / `FilePreviewApi`) are a **different code path**: they verify a per-file HMAC signature instead of relying on session auth, so cross-tenant access there requires already possessing a valid signature for the target file. Not affected by this advisory; not changed in this PR.

## Refs

- GHSA-2qwc-c2cc-2xwv (private advisory, reporter: Ido Shani / @zafran-io via Huntr)
- Public follow-up: #35699

## Credit / authorship

This PR applies **Ido Shani's** original patch verbatim (`git am`). Both commits authored by him. PR opened by me on his behalf since he didn't have permission to add me to the GHSA — see #35699. Marking as draft until reporter / a maintainer signs off.


laipz8200 (approved):
Thanks for your contribution.

PR #35799, opened 2026-05-05
refactor: migrate workflow queries to contracts

lyzno1:
### Workflow comments and collaboration contracts
- add contract-backed oRPC coverage for workflow comments and collaboration-related API calls
- migrate workflow comment read/write call sites from ad hoc service requests to contract-derived query and mutation helpers
- replace manual collaboration polling with TanStack Query polling so the polling interval and cache lifecycle are owned by query options

### App list infinite queries on the frontend
- add the console app list contract and register it under `consoleQuery.apps.list`
- remove the thin `useAppList` and `useInfiniteAppList` passthrough hooks from the app service layer
- update app list, app nav, account page, and plugin app selector call sites to consume `consoleQuery.apps.list` directly
- use oRPC `infiniteOptions` with TanStack Query for paginated app list reads, including explicit page params and next-page calculation at the call site
- update related frontend tests to mock TanStack Query directly now that app list queries are no longer hidden behind service hooks

### Backend app tag filter parsing
- update the console app list controller to decode bracket-style `tag_ids[n]` query parameters into `AppListQuery.tag_ids`
- keep the service-layer app filtering contract unchanged: app services still receive validated `tag_ids` as a list of UUID strings
- add backend coverage for the new tag filter query boundary, including rejecting the old flat `tag_ids` shape


hjlarry (approved):
LGTM

PR #35809, opened 2026-05-05
feat: warn when agent app uses a subset of provider's available tool operations

### LINKED ISSUE #35807: Cloned agent app doesn't see Custom Tool operations added after the original was attached
I'm running a self-hosted Dify 1.14 stack (docker-compose) and hit a confusing failure mode while cloning an agent-chat app. Putting it here in case it's useful or worth fixing.

## What happens

I have a Custom Tool that was originally registered with one operation (`op_a`). I attached it to an agent-chat app, so `model_config.agent_mode.tools` for that app has `op_a` listed.

Later I updated the Custom Tool's OpenAPI schema to add a second operation (`op_b`) and re-imported it. The workspace tool list now shows the Custom Tool exposes both operations. Good.

But — the original app I attached the tool to still has only `op_a` in its tools list. New operations don't auto-attach to apps that already use the tool. Fine on its own, that's defensible behavior.

The problem is what happens when I clone that app. The clone gets `agent_mode.tools` copied verbatim, so the clone also only has `op_a`. The UI shows the Custom Tool attached and the model sees one operation. If I then prompt the agent to use `op_b` by name, it narrates "I will use op_b first..." and then actually invokes `op_a` with op_b's parameters, which errors out.

So functionally `op_b` is missing from the clone, even though the provider exposes it.

## Repro

1. Create a Custom Tool from an OpenAPI spec with one operation `op_a`.
2. Create an agent-chat app, attach the tool. `agent_mode.tools = [op_a]`.
3. Edit the Custom Tool's OpenAPI to add `op_b`, re-import. Workspace tool now shows both operations available.
4. `POST /console/api/apps/{src_id}/copy` to clone the app.
5. `GET /console/api/apps/{new_id}` and look at `model_config.agent_mode.tools`.
6. Clone has only `op_a`. `op_b` is not there, even though it exists on the provider.

## Why it's confusing

There's no visible indicator that the clone has a stale tool attachment. The tool icon shows up in the tool list, the model gets one operation, the UI looks normal. But the model can't reach `op_b`.

The user-facing failure looks like a *model* bug. The agent says "I will use op_b first" and then doesn't, or calls op_a with op_b's args. I spent a few hours debugging this as if it was a model orchestration issue — token budget pressure, agent-loop iteration limit, sampling temperature — before I realised the tool just wasn't attached. The narration was the model literally telling me it wanted to call a tool it didn't have, which is how an agent should behave when instructed to use a tool that isn't in it's tool list.

## Workaround

After cloning, read the source app's `agent_mode.tools`, deep-copy the existing entry shape, mutate `tool_name` + `tool_parameters` to match the new operation, append, then POST back via `/console/api/apps/{new_id}/model-config`. ~5 lines of patching. Doable but it's a footgun.

## Suggested fixes

A few options, increasing in scope:

- **Lowest:** surface a warning on the app's tool-config UI when the attached operations are a subset of what the provider exposes. Something like "Tool X has 3 operations in the workspace, this app uses 1. Manage attachments?"
- **Medium:** when cloning, re-resolve every attached Custom Tool against its current provider state and pull in all operations available now. (Would change clone semantics — might be controversial since it makes clones non-faithful copies.)
- **Most upstream:** when re-importing a Custom Tool with new operations, prompt "auto-attach `op_b` to N existing apps that already use this tool?"

I'd be happy with any of them. Right now the lack of any signal is what burns time.

Happy to put together a PR if there's a preferred approach — let me know.

ayushh0110:
fixes #35807

## Problem

When a Custom Tool's OpenAPI schema is updated to add new operations, or when an agent-chat app is cloned, the app's `agent_mode.tools` list may contain only a subset of what the provider currently exposes. There is no visual signal — users discover this only when the LLM hallucinates about calling tools it doesn't have.

As confirmed by @dosu, agent-chat apps have no clone-time or edit-time mechanism to detect that the attached tool operations are a subset of what the provider currently exposes.

## Solution

Surface a warning icon in the agent tool-config UI when the app's attached operations for a given provider are a **proper subset** of that provider's currently available operations. This is the "lowest" (least disruptive) fix suggested in the issue.

### How it works

The `AgentTools` component already has both data sources:
- `agentConfig.tools` — the app's attached tool operations
- `collectionList` — all available tool collections with their operations

A new `useMemo` cross-references these to build a `Map<provider_id, {attached, available}>`. For each tool row where `available > attached`, an amber warning icon with a descriptive tooltip is rendered.

### What this covers

- Cloned apps with stale tool attachments
- Apps where the tool schema evolved after initial configuration
- Non-blocking — warning icon + hover tooltip only
- Actionable — tooltip tells the user to use the "+" button

### What this does NOT change

- No backend changes
- No clone behavior changes
- No auto-attach of missing operations

## Changes

| File | Change |
|---|---|
| `web/.../agent-tools/index.tsx` | Added `outOfSyncProviders` useMemo + amber warning icon per tool row |
| `web/i18n/en-US/tools.json` | Added `toolOperationsOutOfSync` translation key |
| `web/.../agent-tools/__tests__/index.spec.tsx` | Added 3 tests for subset, full-match, and single-op cases |

## Tests

All 13 tests pass (10 existing + 3 new):

### DIFF
--- a/web/app/components/app/configuration/config/agent/agent-tools/__tests__/index.spec.tsx
+++ b/web/app/components/app/configuration/config/agent/agent-tools/__tests__/index.spec.tsx
@@ -465,4 +465,90 @@ describe('AgentTools', () => {
       expect((getModelConfig().agentConfig.tools[0] as { isDeleted: boolean }).isDeleted).toBe(false)
     })
   })
+
+  it('should show out-of-sync warning when attached operations are a subset of available', () => {
+    // Provider has 2 tools available (search + translate)
+    builtInTools = [
+      createCollection({
+        id: 'multi-op-provider',
+        name: 'vendor/multi-op-provider',
+        tools: [
+          createToolDefinition({ name: 'op_a' }),
+          createToolDefinition({ name: 'op_b' }),
+        ],
+      }),
+    ]
+
+    // But the app only attaches op_a
+    renderAgentTools([
+      createAgentTool({
+        provider_id: 'multi-op-provider',
+        provider_name: 'vendor/multi-op-provider',
+        tool_name: 'op_a',
+        tool_label: 'Operation A',
+      }),
+    ])
+
+    const warningIcon = screen.getByTestId('tool-ops-out-of-sync')
+    expect(warningIcon).toBeInTheDocument()
+  })
+
+  it('should not show out-of-sync warning when all operations are attached', () => {
+    // Provider has 2 tools available
+    builtInTools = [
+      createCollection({
+        id: 'full-provider',
+        name: 'vendor/full-provider',
+        tools: [
+          createToolDefinition({ name: 'op_a' }),
+          createToolDefinition({ name: 'op_b' }),
+        ],
+      }),
+    ]
+
+    // App attaches both operations
+    renderAgentTools([
+      createAgentTool({
+        provider_id: 'full-provider',
+        provider_name: 'vendor/full-provider',
+        tool_name: 'op_a',
+        tool_label: 'Operation A',
+      }),
+      createAgentTool({
+        provider_id: 'full-provider',
+        provider_name: 'vendor/full-provider',
+        tool_name: 'op_b',
+        tool_label: 'Operation B',
+      }),
+    ])
+
+    const warningIcon = screen.queryByTestId('tool-ops-out-of-sync')
+    expect(warningIcon).not.toBeInTheDocument()
+  })
+
+  it('should not show out-of-sync warning when provider has only one operation', () => {
+    // Provider has exactly 1 tool
+    builtInTools = [
+      createCollection({
+        id: 'single-op-provider',
+        name: 'vendor/single-op-provider',
+        tools: [
+          createToolDefinition({ name: 'only_op' }),
+        ],
+      }),
+    ]
+
+    // App attaches that single operation
+    renderAgentTools([
+      createAgentTool({
+        provider_id: 'single-op-provider',
+        provider_name: 'vendor/single-op-provider',
+        tool_name: 'only_op',
+        tool_label: 'Only Operation',
+      }),
+    ])
+
+    const warningIcon = screen.queryByTestId('tool-ops-out-of-sync')
+    expect(warningIcon).not.toBeInTheDocument()
+  })
 })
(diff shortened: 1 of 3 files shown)


ayushh0110:
(Re-posting — accidentally deleted earlier)
Thanks for the review @dosu — addressing each point:
1. Deleted operations: This is handled separately by the existing isDeleted flow, which already shows an AlertTriangle with a "Tool removed" tooltip. Our warning explicitly guards against this with !item.isDeleted — it only fires when the provider has more operations than the app uses, never when an operation was removed. These are two distinct states:
Provider removed an op → existing isDeleted warning (yellow triangle + "Tool removed")
Provider added new ops → new out-of-sync warning (amber icon + count mismatch)

2. Tooltip actionability: The "+" button opens the existing ToolPicker, which lets users select individual operations to add. Adding new operations does not touch the tool_parameters of already-attached ones — handleSelectTool appends to the tools array via produce() without mutating existing entries. So existing configurations are preserved.

3. i18n: Only en-US is updated in this PR. Other locales typically follow in a separate pass via the community translation process — this matches the pattern used by other recent PRs in the repo.

camerono:
@ayushh0110 hey, this looks ready from the dosu pass — anything blocking review? mergeable_state shows "blocked" but i can't tell if that's required-reviews or a base sync needed. if it'd help i can pull and test against my repro from #35807.

ayushh0110:
hi @camerono ! nothing blocking from my side, just waiting on a review from the code owners. would definitely appreciate it if you tested against your repro though, that'd help.

camerono:
Tested against our running Dify 1.14 stack. Our only agent-chat app with a custom tool attached is currently in-sync — uses both of the 2 operations its provider exposes — so no naturally-occurring subset case to flag, but that's by accident: the moment a third operation gets added to the provider's OpenAPI schema, that existing attachment becomes stale until someone re-attaches in the UI, and a clone of the app would propagate the staleness.

Walked the PR diff against the original repro. Frontend-only as expected — the backend cloning behavior in `app_dsl_service.py` stays the same, this fix just surfaces the situation so users see it instead of debugging as a model bug.

Comparison logic in `outOfSyncProviders` does the right thing: groups `agent_mode.tools` by `provider_id`, looks up the matching collection, fires the warning when `attachedNames.size < collection.tools.length`. That covers the op_a/op_b case from the issue regardless of which specific operation is missing.

The `!item.isDeleted` guard keeps the new warning distinct from the existing provider-removed-op state — matches dosu's earlier point, good separation.

Test coverage hits both cases (warning shows on subset, doesn't show on full match). i18n string is accurate and points the user at the `+` button to fix. No backend touch, no risk to existing configs.

LGTM from this side.

PR #35840, opened 2026-05-06
refactor: verticalize tag management and batch bindings

lyzno1:
## Summary
- move tag selection/management UI into `web/features/tag-management` with direct module imports, named exports, and dedicated component/hook coverage
- migrate tag list/create/update/delete/bind/unbind flows to oRPC console contracts and TanStack Query mutation helpers
- add user feedback for tag binding apply flows, including deduped toast ids when selectors close
- replace per-tag console unbind requests with a batch remove API using `POST /tag-bindings/remove` and `tag_ids`
- keep tag-management refresh ownership decoupled from app/dataset internals by letting host surfaces decide their own list refresh

## Backend/API
- canonicalize console tag binding writes on `POST /tag-bindings` and `POST /tag-bindings/remove`
- remove legacy console `/tag-bindings/create` and item-delete compatibility paths
- batch-delete matching tag bindings in one service call, scoped by tenant and target
- update dataset Service API tag unbinding to normalize batch `tag_ids` internally while preserving legacy single `tag_id` payload compatibility

## Frontend
- introduce `AppCardTags`, `DatasetCardTags`, `TagSelector`, `TagPanel`, `TagFilter`, and modal/editor pieces under `features/tag-management`
- remove old base tag-management service/state usage and stale local tag mirrors in app/dataset card flows
- keep card tag display driven by server list data and refresh callbacks instead of feature-level app/dataset query coupling
- keep dataset list query-key ownership outside tag-management; tag-management only invalidates tag list cache directly


api/services/tag_service.py
```diff
@@ -178,13 +178,17 @@ def save_tag_binding(payload: TagBindingCreatePayload):
     @staticmethod
     def delete_tag_binding(payload: TagBindingDeletePayload):
         TagService.check_target_exists(payload.type, payload.target_id)
-        tag_binding = db.session.scalar(
-            select(TagBinding)
-            .where(TagBinding.target_id == payload.target_id, TagBinding.tag_id == payload.tag_id)
-            .limit(1)
-        )
-        if tag_binding:
+        tag_bindings = db.session.scalars(
```
hjlarry: directly use bulk delete here seems better:
```
    result = db.session.execute(
        delete(TagBinding).where(
            TagBinding.target_id == payload.target_id,
            TagBinding.tag_id.in_(payload.tag_ids),
            TagBinding.tenant_id == current_user.current_tenant_id,
        )
    )
    if result.rowcount:
        db.session.commit()
```

api/controllers/console/tag/tags.py
hjlarry: we should also rename this method to `_remove_tag_bindings`, because already has a `_create_tag_bindings`

hjlarry (changes requested):

hjlarry (dismissed):
LGTM

hjlarry (approved):

PR #35843, opened 2026-05-06
fix(commands): purge tenant tool credentials on reset-encrypt-key-pair (#35396)

### LINKED ISSUE #35396: reset-encrypt-key-pair leaves stale MCP tool data and breaks tool-providers API
### Self Checks

### Dify version

1.13.3

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Connect an MCP server in Dify. For instance: https://demo-day.mcp.cloudflare.com/sse

Image

3. Run `flask reset-encrypt-key-pair` in dify-api terminal.
4. Delete the tenant private key cache in Redis, this mocks key expiration.

Image

5. Visit the Tool page in the UI, which requests `/console/api/workspaces/current/tool-providers`.

Image

No stack trace was shown, only 400 http request log.
Image

Additional context:

- After the key reset, old tenant-encrypted MCP data remains in the database.
- In my case, deleting the rows in `tool_mcp_providers` immediately made the Tool page load again. This looks like a workaround, not a real fix.
- From reading the current code path, the MCP list response is still decrypting `server_url` during list rendering even when the request is using the list-view `include_sensitive=false` path.

### ✔️ Expected Behavior

- `reset-encrypt-key-pair` should leave the system in a recoverable state for the Tool page.
- `/console/api/workspaces/current/tool-providers` should not return `400` just because one stored MCP provider contains ciphertext encrypted with the old tenant key.
- After key rotation, invalidated MCP/tool credentials should either be cleaned up automatically or shown in a reconnect/re-enter state so users can recover from the UI.

### ❌ Actual Behavior

- `/console/api/workspaces/current/tool-providers` returns `400`.
- The Tool page becomes empty / unusable, so I cannot re-enter credentials from the UI.
- The immediate trigger appears to be stale encrypted MCP provider data left behind after `reset-encrypt-key-pair`.

### Suggested fix direction:

- Expand `reset-encrypt-key-pair` so it also clears or invalidates tenant-encrypted tool data that cannot survive key rotation, instead of only resetting legacy model-provider data.
- Make MCP list serialization defensive so broken encrypted rows do not fail the entire `/tool-providers` response.

xr843:
## Summary

Closes #35396.

`flask reset-encrypt-key-pair` rotates each tenant's asymmetric key and purges `Provider` / `ProviderModel` rows that hold LLM credentials encrypted under the old key. It does **not** touch the tool provider tables that encrypt credentials under the same key:

| Table | Encrypted columns |
|---|---|
| `tool_builtin_providers` | `encrypted_credentials` |
| `tool_api_providers` | `credentials_str` |
| `tool_mcp_providers` | `encrypted_credentials`, `encrypted_headers`, `server_url` |

Stale ciphertext in these tables surfaces as a 500 from `/console/api/workspaces/current/tool-providers` because the new key cannot decrypt it. The reproducer in #35396 walks through it for an MCP tool provider; the same hazard exists for the builtin and API tool tables.

## Change

Extend the existing per-tenant purge in `api/commands/system.py` to also delete rows from `tool_builtin_providers`, `tool_api_providers`, and `tool_mcp_providers`. Update the help text and confirmation prompt so operators know tool provider records are wiped along with model credentials.

The existing UX guarantee already says *all LLM credentials will become invalid, requiring re-entry*, so this is a behavior fix rather than a contract change. I considered putting the new scope behind a `--purge-tools` flag, but defaulting to the buggy behavior felt wrong since the reproducer in #35396 leaves the tool-providers API broken until an operator manually deletes the rows.

## Tests

`api/tests/unit_tests/commands/test_reset_encrypt_key_pair.py` (3 tests, follow `test_upgrade_db.py` mocking pattern — `sessionmaker` and `db` are patched, no DB required):

- `test_reset_aborts_when_not_self_hosted` — guard rail unchanged.
- `test_reset_purges_provider_and_tool_tables_for_each_tenant` — verifies delete is issued against all 5 tables (`providers`, `provider_models`, `tool_builtin_providers`, `tool_api_providers`, `tool_mcp_providers`) for the active tenant.
- `test_reset_iterates_all_tenants` — multi-tenant deployments purge every tenant (regression guard against returning early in the loop).

```
$ uv run pytest tests/unit_tests/commands/test_reset_encrypt_key_pair.py -v
========== 3 passed in 0.07s ==========
```

`ruff check` and `ruff format --check` both pass on the touched files.

## Notes for reviewers

- I followed the issue plan I posted on 2026-05-02 ("purge with --force flag") but settled on making the broader purge the **default** rather than a flag, because the existing help text already promises wiping all credentials and the buggy default leaves the system in a broken state. Happy to flip this to a flag if you'd prefer to preserve strict backward compat.
- I did not touch `tool_workflow_providers` — that table stores `parameter_configuration` JSON without credentials encrypted under the tenant key, so it does not have the same hazard.
- I did not extend the purge to other potentially-encrypted tables (e.g. data source OAuth bindings) to keep this PR scoped to the reported bug.

cc @crazywoola @laipz8200


laipz8200 (approved):
LGTM, thanks for your contribution.

PR #34149, opened 2026-03-26
fix(workflow): use correct field ID in KB metadata filter selection

shawny011717:
> [!IMPORTANT]
>
> 1. Make sure you have read our contribution guidelines
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

Fixes #33577.
fix https://github.com/langgenius/dify/issues/35844

The `onClick` handler in the metadata field dropdown (`AddCondition` component) was attached to the inner text `<div>` instead of the outer row `<div>`. This meant clicks landing on the icon, type label, or row padding did not register the intended field — the resulting mis-hit caused the UI to appear to always select the first field. Moving the handler to the row element ensures the full clickable area maps to the correct metadata item.

## Screenshots

N/A (logic-only fix, no visual change)

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. [Contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [x] I ran `cd web && npx lint-staged` to appease the lint gods

PR #35857, opened 2026-05-06
fix(workflow): move onClick to row div in KB metadata filter field selector

### LINKED ISSUE #35844: Bug: Metadata Filter Conditions always returns document_name regardless of the operator selected
### Self Checks

### Dify version

1.13.3

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Open a Knowledge Retrieval node.
2. Go to Search Settings → Conditions.
3. In the Metadata Filter Conditions panel, set the field to permission with type string.
4. Try selecting different operators from the dropdown.
5. Observe that the filtered output always returns results keyed by document_name.

### ✔️ Expected Behavior

The filter should apply the selected operator against the permission metadata field and return results accordingly.

### ❌ Actual Behavior

Regardless of the operator chosen, the result always appears to use document_name as the filter key.

Image

Beandon13:
## Summary

Fixes #35844.

The `onClick` handler in the `AddCondition` component (Knowledge Retrieval node → Metadata Filter Conditions) was attached to the inner text `<div>` rather than the outer row `<div>`. This meant that clicks landing on the field icon, the type label (`string`, `number`, etc.), or the row padding were silently dropped — the event never reached the handler. As a result, the UI always fell back to the first field in the list (`document_name`), regardless of which field the user intended to select.

**Root cause (one-line diff):**

```diff
-<div key={metadata.name} className="flex h-6 cursor-pointer ...">
+<div key={metadata.name} className="flex h-6 cursor-pointer ..." onClick={() => handleAddConditionWrapped(metadata)}>
   <div className="mr-1 p-px"><MetadataIcon type={metadata.type} /></div>
-  <div className="grow truncate" title={metadata.name} onClick={() => handleAddConditionWrapped(metadata)}>
+  <div className="grow truncate" title={metadata.name}>
     {metadata.name}
   </div>
   <div className="shrink-0 ...">string</div>
</div>
```

Moving the handler to the outer row element ensures the entire clickable area (icon, name, type, padding) correctly maps to the intended metadata item.

## Test plan

- Added a regression test in `integration.spec.tsx` that renders `AddCondition` with two metadata items, opens the dropdown, and clicks the **type label** span (the area that was previously uncovered). Asserts that `handleAddCondition` is called exactly once with the correct `id` and `name` of the intended field.
- Manually verified in the running app: selecting any built-in or custom field by clicking anywhere on the row now correctly adds the condition for that field.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/web/app/components/workflow/nodes/knowledge-retrieval/__tests__/integration.spec.tsx
+++ b/web/app/components/workflow/nodes/knowledge-retrieval/__tests__/integration.spec.tsx
@@ -16,6 +16,7 @@ import { RETRIEVE_METHOD, RETRIEVE_TYPE } from '@/types/app'
 import { DatasetsDetailContext } from '../../../datasets-detail-store/provider'
 import { createDatasetsDetailStore } from '../../../datasets-detail-store/store'
 import { BlockEnum, VarType } from '../../../types'
+import AddCondition from '../components/metadata/add-condition'
 import AddDataset from '../components/add-dataset'
 import DatasetItem from '../components/dataset-item'
 import DatasetList from '../components/dataset-list'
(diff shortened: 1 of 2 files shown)


hjlarry:
sorry, already has a PR to fix this https://github.com/langgenius/dify/pull/34149

PR #35859, opened 2026-05-06
feat(api): pass app_id to model plugins for provider-side cost attribution

### LINKED ISSUE #35772: Pass app_id to model plugins for provider-side cost attribution
### Self Checks

We are building a per-app cost tracking dashboard (Grafana) for a Dify + Amazon Bedrock deployment. Bedrock's Converse API supports a `requestMetadata` parameter that lets you tag each request with custom key-value pairs, which are then recorded in CloudWatch Logs. We built a custom Bedrock model plugin to use this, but discovered that **model plugins do not receive `app_id`** — only `user_id` is available. This makes it impossible to tag LLM calls with the originating app.

This is not a Bedrock-specific limitation. All major LLM providers offer similar request-level metadata for cost attribution:

- **Amazon Bedrock**: `requestMetadata` (Converse API) → CloudWatch Logs
- **OpenAI / Azure OpenAI**: `metadata` (Chat Completions API) → Usage Dashboard
- **Google Vertex AI**: `labels` (generateContent API) → directly reflected in Cloud Billing
- **Anthropic**: `metadata` (Messages API)

None of these can be utilized from a Dify model plugin today because `app_id` is not passed.

Meanwhile, **tool plugins already receive `app_id`** (along with `conversation_id` and `message_id`) via the session. The model plugin dispatch path simply does not include it:

```python
# Tool plugin dispatch (core/plugin/impl/tool.py) — app_id is included
data={
    "user_id": user_id,
    "conversation_id": conversation_id,
    "app_id": app_id,
    "message_id": message_id,
    "data": { ... },
}

# Model plugin dispatch (core/plugin/impl/model.py) — app_id is missing
data=jsonable_encoder({
    "user_id": user_id,
    "data": { ... },
})
```

**Proposal:** Add `app_id` to the model plugin dispatch payload, as an optional field (`null` when called outside app context, e.g., RAG routing, title generation, suggested questions).

We are only proposing `app_id` — not `conversation_id`, `message_id`, or `workflow_run_id`. These IDs vary by app type (e.g., Workflow apps have `workflow_run_id` instead of `conversation_id`), which would make the change more complex. With `app_id` on the provider side, finer-grained drill-down (per-conversation, per-node, per-iteration) can be done by joining provider logs with Dify's existing database tables (`messages`, `workflow_node_executions`) using `app_id` as the key.

This change spans 2 repositories (langgenius/dify and langgenius/dify-plugin-sdks). The plugin daemon (dify-plugin-daemon) already uses a shared InvokePluginRequest struct with an AppID field for both tool and model dispatch, so no daemon changes are needed — once Dify includes app_id in the model dispatch payload, the daemon will pass it through to the SDK session automatically. On the SDK side, plugin_executor.py needs to forward the session to model plugin instances (it already does this for tool plugins). Fully backward compatible — existing model plugins require no changes.

This asymmetry was introduced in PR #13836 (Introduce Plugins, 2025-02-17). We understand the model plugin path was kept minimal because models can be invoked outside of app context (RAG indexing, utility LLM calls, etc.). Passing `app_id` as a nullable field preserves this design — it is simply `null` for those calls.

**Known limitation:** For Chat/Completion apps using multiple knowledge bases with "single retrieval" (LLM-based routing), the routing LLM call does not have `app_id` because it is not threaded through the retrieval chain. This is a minor cost (a few hundred tokens per call) and can be addressed separately.

We have prepared PRs for both repositories and can submit them once the design direction is agreed upon.

ryuta-kobayashi-ug:
Fixes #35772

## Summary

Pass `app_id` to model plugins so they can tag LLM requests with the originating Dify app for provider-side cost attribution (e.g., Bedrock `requestMetadata`, OpenAI `metadata`, Vertex AI `labels`).

## Changes

- Add `app_context.py` with `ContextVar`-based `get_current_app_id()` / `set_current_app_id()`
- Set `app_id` in all app runners (chat, completion, agent_chat, advanced_chat, workflow) before LLM invocation
- `PluginModelRuntime.invoke_llm()` reads `get_current_app_id()` and passes it to `PluginModelClient`
- `PluginModelClient` includes `app_id` in the HTTP payload to plugin daemon
- `app_id` is `None` when called outside app context (RAG routing, title generation, etc.)

## Design

Uses `contextvars.ContextVar` because `PluginModelRuntime` is created at tenant scope (before app execution), and graphon sits between the app runner and `PluginModelRuntime` with a fixed interface that cannot pass `app_id` as an argument.

## Testing

- Unit tests for `_dispatch_payload` with/without `app_id`
- E2E tested all 5 app types (chatbot, chatflow, workflow, text generator, agent) — all pass `app_id` correctly
- Regression tested: official plugin, RAG, blocking API mode — no degradation

## Checklist

- [x] This change requires a documentation update
- [x] I've added a test for each change that was introduced
- [x] I ran dev/reformat(backend) to appease the lint gods

### DIFF
--- a/api/core/agent/cot_agent_runner.py
+++ b/api/core/agent/cot_agent_runner.py
@@ -133,6 +133,7 @@ class CotAgentRunner(BaseAgentRunner, ABC):
                 stop=app_generate_entity.model_conf.stop,
                 stream=True,
                 callbacks=[],
+                request_metadata={"app_id": self.app_config.app_id},
             )
 
             usage_dict: dict[str, LLMUsage | None] = {}
--- a/api/core/agent/fc_agent_runner.py
+++ b/api/core/agent/fc_agent_runner.py
@@ -101,6 +101,7 @@ class FunctionCallAgentRunner(BaseAgentRunner):
                 stop=app_generate_entity.model_conf.stop,
                 stream=self.stream_tool_call,
                 callbacks=[],
+                request_metadata={"app_id": self.app_config.app_id},
             )
 
             tool_calls: list[tuple[str, str, dict[str, Any]]] = []
--- a/api/core/app/apps/chat/app_runner.py
+++ b/api/core/app/apps/chat/app_runner.py
@@ -232,6 +232,7 @@ class ChatAppRunner(AppRunner):
             model_parameters=application_generate_entity.model_conf.parameters,
             stop=stop,
             stream=application_generate_entity.stream,
+            request_metadata={"app_id": app_config.app_id},
         )
 
         # handle invoke result
--- a/api/core/app/apps/completion/app_runner.py
+++ b/api/core/app/apps/completion/app_runner.py
@@ -193,6 +193,7 @@ class CompletionAppRunner(AppRunner):
             model_parameters=application_generate_entity.model_conf.parameters,
             stop=stop,
             stream=application_generate_entity.stream,
+            request_metadata={"app_id": app_config.app_id},
         )
 
         # handle invoke result
--- a/api/core/model_manager.py
+++ b/api/core/model_manager.py
@@ -124,6 +124,7 @@ class ModelInstance:
         stop: list[str] | None = None,
         stream: Literal[True] = True,
         callbacks: list[Callback] | None = None,
+        request_metadata: Mapping[str, object] | None = None,
     ) -> Generator: ...
 
     @overload
@@ -135,6 +136,7 @@ class ModelInstance:
         stop: list[str] | None = None,
         stream: Literal[False] = False,
         callbacks: list[Callback] | None = None,
+        request_metadata: Mapping[str, object] | None = None,
     ) -> LLMResult: ...
 
     @overload
@@ -146,6 +148,7 @@ class ModelInstance:
         stop: list[str] | None = None,
         stream: bool = True,
         callbacks: list[Callback] | None = None,
+        request_metadata: Mapping[str, object] | None = None,
     ) -> Union[LLMResult, Generator]: ...
 
     def invoke_llm(
@@ -156,6 +159,7 @@ class ModelInstance:
         stop: Sequence[str] | None = None,
         stream: bool = True,
         callbacks: list[Callback] | None = None,
+        request_metadata: Mapping[str, object] | None = None,
     ) -> Union[LLMResult, Generator]:
         """
         Invoke large language model
@@ -166,6 +170,7 @@ class ModelInstance:
         :param stop: stop words
         :param stream: is stream response
         :param callbacks: callbacks
+        :param request_metadata: optional request metadata
         :return: full response or stream response chunk generator result
         """
         if not isinstance(self.model_type_instance, LargeLanguageModel):
@@ -182,6 +187,7 @@ class ModelInstance:
                 stop=list(stop) if stop else None,
                 stream=stream,
                 callbacks=callbacks,
+                request_metadata=request_metadata,
             ),
         )
 
--- a/api/core/plugin/impl/model.py
+++ b/api/core/plugin/impl/model.py
@@ -27,10 +27,12 @@ _POLLING_UNSUPPORTED_ERROR_MESSAGE = "does not support polling"
 
 class PluginModelClient(BasePluginClient):
     @staticmethod
-    def _dispatch_payload(*, user_id: str | None, data: dict[str, Any]) -> dict[str, Any]:
+    def _dispatch_payload(*, user_id: str | None, data: dict[str, Any], app_id: str | None = None) -> dict[str, Any]:
         payload: dict[str, Any] = {"data": data}
         if user_id is not None:
             payload["user_id"] = user_id
+        if app_id is not None:
+            payload["app_id"] = app_id
         return payload
 
     def fetch_model_providers(self, tenant_id: str) -> Sequence[PluginModelProviderEntity]:
@@ -166,6 +168,7 @@ class PluginModelClient(BasePluginClient):
         tools: list[PromptMessageTool] | None = None,
         stop: list[str] | None = None,
         stream: bool = True,
+        app_id: str | None = None,
     ) -> Generator[LLMResultChunk, None, None]:
         """
         Invoke llm
@@ -188,6 +191,7 @@ class PluginModelClient(BasePluginClient):
                         "stop": stop,
                         "stream": stream,
                     },
+                    app_id=app_id,
                 )
             ),
             headers={
--- a/api/core/plugin/impl/model_runtime.py
+++ b/api/core/plugin/impl/model_runtime.py
@@ -321,20 +321,22 @@ class PluginModelRuntime(ModelRuntime):
         if not isinstance(app_id, str):
             app_id = None
         plugin_id, provider_name = self._split_provider(provider)
-        result = self.client.invoke_llm(
-            tenant_id=self.tenant_id,
-            user_id=self.user_id,
-            plugin_id=plugin_id,
-            provider=provider_name,
-            model=model,
-            credentials=credentials,
-            model_parameters=model_parameters,
-            prompt_messages=list(prompt_messages),
-            tools=tools,
-            stop=list(stop) if stop else None,
-            stream=stream,
-            app_id=app_id,
-        )
+        invoke_kwargs = {
+            "tenant_id": self.tenant_id,
+            "user_id": self.user_id,
+            "plugin_id": plugin_id,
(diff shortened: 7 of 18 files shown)


WH-2099 (changes requested):
Thanks for the contribution. However, I think using Context for implicit prop drilling here might impose a certain burden on future state management and maintainability.

I would prefer to use an explicit prop-passing (explicit parameter) approach instead.

ghost:
Thanks @WH-2099 — agreed, and I'd like to move toward explicit passing on the host side rather than defend the ContextVar approach. Let me confirm the plan and one design choice with you.

**Where app_id has to travel (host side):**
`AppRunner → ModelInstance.invoke_llm() → graphon LargeLanguageModel.invoke() → _invoke_llm_via_runtime() → PluginModelRuntime.invoke_llm()`.

The two middle hops live in the external `langgenius/graphon` package, which is why I originally reached for a ContextVar to keep the change inside the Dify host. But I agree the implicit flow hurts readability and is a latent footgun around streaming/async, so I'm happy to do the explicit, cross-package version.

**Proposed approach (my recommendation):**
Thread the value explicitly through the call stack down to `PluginModelRuntime`, but instead of a bare `app_id` argument I'd pass a small `invocation_context` object (e.g. `{ "app_id": ... }`). Rationale: it keeps graphon's generic `invoke()` signature from accruing one Dify-specific provider-metadata field, and it leaves room for `conversation_id` / `workflow_run_id` later without another signature change. Concretely this touches:

- the 6 `ModelInstance.invoke_*` methods (Dify host),
- `graphon`'s `invoke()` + the `*ModelRuntime` protocols (separate PR in langgenius/graphon),
- `PluginModelRuntime` reads it from the context object and forwards it.

**Why not constructor injection (the way `user_id` is handled)?**
A `PluginModelRuntime` is composed only from `(tenant_id, user_id)` and has no notion of an app, so `user_id` is part of the runtime identity. `app_id` is a property of the individual invocation — present for app execution, `None` for RAG indexing, title generation, suggested questions, etc. So it has to ride with the call, not the runtime.

**SDK side (dify-plugin-sdks#313) is the one place I'd keep ContextVar.** There `_invoke()` is an `@abstractmethod` implemented by every published plugin and `__init__` is `@final`, so adding a parameter would raise `TypeError` across the ecosystem. `get_current_session()` keeps it opt-in and backward compatible.

**One thing I need your call on:** are you OK with a companion PR against `langgenius/graphon` to extend `invoke()` / the runtime protocols? If yes, I'll rework #35859 along these lines and open the graphon PR. If you'd rather not touch graphon's signature right now, I can keep the ContextVar bridge tightly scoped and documented as an interim step.

WH-2099:
@ryuta-kobayashi-ug Thanks for the detailed breakdown. I completely agree with your rationale. Threading an context object is indeed the cleanest, most extensible design here, and keeping the SDK side backward-compatible via ContextVar makes perfect sense.

Let's go with your proposed approach. Modifying graphon to achieve the ideal implementation is definitely the right move. Please go ahead and open the companion PR against langgenius/graphon alongside your reworked PR here. I'll follow up and keep an eye on both.

Looking forward to your PRs!

WH-2099:
@ryuta-kobayashi-ug Hi, I haven't seen any new progress on this task yet. To avoid duplicating our efforts, I wanted to check if you're still working on it.

If you haven't started yet, I'm happy to take over from here and finish the rest.

ghost:
@WH-2099 Thanks for the go-ahead, and sorry for the quiet stretch on my end — yes, I'm still on it and starting today.

Plan for the next push:
1. Open the companion PR against `langgenius/graphon` first: introduce a small `invocation_context` object and thread it through `invoke()` and the `*ModelRuntime` protocols (LLM, text-embedding, rerank, tts, speech2text, moderation).
2. Rework this PR on top of it: drop the ContextVar bridge, pass `invocation_context` explicitly from the 6 `ModelInstance.invoke_*` methods down to `PluginModelRuntime`, and populate `app_id` from the app runners.
3. Leave dify-plugin-sdks#313 on the ContextVar approach as discussed, since an explicit `_invoke` parameter would break existing plugins.

I'll link the graphon PR here as soon as it's up so you can track both together. Thanks for your patience, and for offering to take it on.

ghost:
@WH-2099 Quick update: the graphon companion PR is up — langgenius/graphon#191.

It threads the existing `invocation_context` (`Mapping[str, object] | None`) from the public `invoke()` surface down to `model_runtime.invoke_*` for all six model types, and adds it to the runtime protocols. It's keyword-only with a `None` default, so it's fully backward compatible. (`invocation_context` already existed in graphon but was only wired to the LLM callbacks, never forwarded to the runtime — so this is a small, additive change.) The structured-output entry point is intentionally left out of scope for now.

Next I'll rework this PR on top of it: drop the ContextVar bridge and pass `invocation_context={"app_id": ...}` explicitly from the app runners down through `ModelInstance` → `PluginModelRuntime` → the plugin client. I've already validated the seam locally — with graphon#191 installed, `app_id` flows end-to-end from `LargeLanguageModel.invoke(...)` to the client, and Dify's existing unit suites stay green.

I'll push the reworked branch here once I've extended coverage to the agent and workflow call sites (not just chat/completion), and will bump the graphon pin once #191 is released. Thanks!

ghost:
@WH-2099 Thanks for merging langgenius/graphon#191 (and for renaming the parameter to `request_metadata` — agreed, that's clearer).

The Dify-side rework is ready on my end: ContextVar bridge removed, and `request_metadata={"app_id": ...}` is threaded explicitly from the app runners (and the workflow LLM nodes) through `ModelInstance` → `PluginModelRuntime` → the plugin client, covering all five app types (chat, completion, agent, advanced-chat/chatflow, workflow). Unit tests are green locally.

One sequencing question before I push, though: `request_metadata` only landed in graphon **v0.6.0**, but `main` here still pins `graphon==0.5.3`, and v0.6.0 isn't import-compatible with Dify yet — e.g. v0.6.0 dropped `graphon.nodes.human_input.enums` (`HumanInputFormKind` / `HumanInputFormStatus`), which `api/models/human_input.py` imports and uses in a DB column. So bumping the pin to 0.6.0 breaks Dify at import time, for reasons unrelated to this PR.

What's the plan for moving Dify to graphon 0.6.0? Is a migration already in progress (so I can rebase this PR on top once it lands), or would you prefer I hold until then? Happy to follow whichever path you prefer — I just don't want to bundle an unrelated 0.6.0 migration into this PR.

WH-2099:
Thanks for the work so far. I will take over the remaining work on this PR and rework it according to the agreed direction. I will keep the change focused and follow up with updates once it is ready for review.

WH-2099 (dismissed):

WH-2099 (approved):

ghost:
Thanks @WH-2099 for taking this over the finish line and for the reworks — great to see it land on `main` and already ride along in 1.16.0-rc1. Appreciated the collaboration throughout the design discussion. 🙏

WH-2099:
@ryuta-kobayashi-ug I apologize that the implementation of this feature may have taken a little longer than expected, but I believe the final implementation here is a more robust version for future maintenance and long-term evolution.

Thank you for your patience and contribution.

There's a minor follow-up issue: the new support has been merged into the dify-plugin-sdk mainline (https://github.com/langgenius/dify-plugin-sdks/pull/313). 
We now need to plan the release schedule for dify-plugin-sdks. 
Do you need to release it as soon as possible to ensure that the relevant changes in dify-official-plugins are reimplemented and implemented?

ghost:
@WH-2099 Thank you again, and no worries at all about the timeline — I fully agree the final design is more robust and well worth the extra iterations.

On the release question: if it's feasible on your side, I'd gently advocate for releasing `dify-plugin-sdks` relatively soon. The host-side change now in Dify 1.16.0-rc1 forwards `request_metadata` (including `app_id`) all the way down to the plugin, but until a release containing #313 is available, the SDK still discards the session before it reaches the model implementation — so the end-to-end path stays inert and the host-side work can't actually surface to plugins yet. An earlier release would unblock `dify-official-plugins` to adopt `get_current_session()` and make provider-side cost attribution work end-to-end.

That said, I don't want to rush anything — if it's cleaner to fold this into your normal release cadence, that's completely fine by me. And I'm glad to help with the `dify-official-plugins` side once the SDK is out. Thanks again for taking this all the way to the finish line!

WH-2099:
@ryuta-kobayashi-ug Thanks for understanding. I'll make sure the new version of the `dify-plugin-sdks` is released before the final 1.16 launch.

PR #35887, opened 2026-05-07
fix(tools): scope builtin tool default-credential clear to tenant

GareArc:
## Summary

- Fixes a bug where two workspace members each setting their own credential as default for the same builtin tool provider left **both** rows with `is_default=True` for the same `(tenant_id, provider)`.
- The `clear-default` UPDATE in `BuiltinToolManageService.set_default_provider` was filtered by `user_id`, so it only reset the caller's own defaults. The consumer (`get_builtin_provider`) is tenant-scoped and breaks ties via `created_at`, producing user-visible inconsistencies (UI shows credential B as default while runtime uses credential A).
- Drops `user_id` from the clear filter, mirroring `DatasourceProviderService.set_default_datasource_provider`. The `is_default` flag is now correctly tenant-scoped (one default per provider per workspace).

## Reproduction

1. Member A in workspace W creates `cred1` (provider P), sets it as default
2. Member B in workspace W creates `cred2` (provider P), sets it as default
3. Both rows now have `is_default=True` — DB has two defaults for the same `(tenant_id, provider)`

Reproduced on production and on `console-platform.dify.dev`.

## Test plan

- [x] Unit test added: `TestSetDefaultProvider::test_clear_default_is_tenant_scoped_not_user_scoped` — asserts the compiled UPDATE WHERE clause contains `tenant_id`/`provider` but not `user_id`. Verified failing on pre-fix code, passing on fixed code.
- [x] Existing `TestSetDefaultProvider` tests still pass.
- [ ] Manual reproduction on staging once merged.

## Notes

- Pre-existing TOCTOU race between concurrent transactions remains (two callers for the same tenant could still produce two `is_default=True` rows in a narrow window). A partial unique index `UNIQUE (tenant_id, provider) WHERE is_default` would close this — out of scope here, flagging for follow-up.
- Equivalent fix backported to `lts/1.13.x` in a separate PR.
- `user_id` parameter retained on `set_default_provider` signature: still passed by the controller via kwargs and may be used by EE callers; removing it would be a breaking change for kwarg callers.


api/services/tools/builtin_tools_manage_service.py
wylswz: `user_id` is no longer referenced.

GareArc: Fixed in the latest commit.

wylswz (approved):

PR #35891, opened 2026-05-07
fix(api): "File validation failed" on Chatflow follow-up with custom file type + memory

lin-snow:
## Summary

A Chatflow whose LLM node has memory enabled and File Upload set to `Other file types` (CUSTOM) fails on the second turn with `File validation failed for file: <name>`, even when no new file is uploaded.

**Root cause.** A file uploaded into the CUSTOM slot is coerced to its detected type by `_resolve_file_type` (PNG → `IMAGE`), and `MessageFile.type` persists the *resolved* type. On history replay, `build_from_message_file` rebuilds `mapping["type"]` from `MessageFile.type`, so a file that passed round 1 (`mapping["type"]=="custom"`) is rejected on round 2 (`mapping["type"]=="image"`). The validator only bypassed the type gate for literal `"custom"`, not for "config has CUSTOM as a fallback bucket".

A parallel mismatch was reachable on round 1: the extension whitelist used a raw `in` check, so a user-typed list like `[".PNG", "png", "JPG", ...]` failed to match the upload-side `".png"` (always lowercased with leading dot).

**Fix.**

- Refactor `is_file_valid_with_config` to bucket semantics. `CUSTOM` is a fallback bucket gated by `allowed_file_extensions`, compared case- and dot-insensitively. Empty whitelist while in the CUSTOM bucket continues to deny (defensive against DSL/API paths that bypass the UI).
- Skip re-validation when rehydrating files from conversation history in `TokenBufferMemory` and `BaseAgentRunner`, mirroring the `build_file_from_stored_mapping` pattern. Validation belongs at upload time, not on replay.

## Validator behavior matrix

| `allowed_file_types` | `input_file_type` | `allowed_file_extensions` | `file_extension` | before | after |
|---|---|---|---|:-:|:-:|
| `[CUSTOM]` | `custom` (round 1) | `[".png"]` | `.png` | ✓ | ✓ |
| `[CUSTOM]` | `image` (replay) | `[".png"]` | `.png` | ✗ | **✓** |
| `[CUSTOM]` | `custom` | `[".PNG", "png", "JPG"]` | `.png` | ✗ | **✓** |
| `[IMAGE, CUSTOM]` | `document` | `[".pdf"]` | `.pdf` | ✗ | **✓** |
| `[CUSTOM]` | `custom` | `[]` | `.png` | ✗ | ✗ (defensive) |
| `[IMAGE]` | `video` | — | `.mp4` | ✗ | ✗ |
| any | any | any | any (`TOOL_FILE`) | ✓ | ✓ |

Bold rows are the user-visible fixes. The empty-whitelist row preserves the original deny-on-empty posture for paths that bypass the UI.

## Screenshots

N/A — backend-only change.

## Checklist


api/core/memory/token_buffer_memory.py
```diff
@@ -86,12 +86,13 @@ def _build_prompt_message_with_files(
 
         detail = ImagePromptMessageContent.DETAIL.HIGH
         if file_extra_config and app_record:
-            # Build files directly without filtering by belongs_to
+            # History files were validated on upload; replaying does not re-validate.
+            # See models/utils/file_input_compat.py for the same pattern.
```
wylswz: It looks like the `config` is always `None` regardless of the invoker.

lin-snow: Thanks for the review!

Intentional — replay rehydrates already-validated state, mirroring `build_file_from_stored_mapping` at `api/models/utils/file_input_compat.py:200` (same pattern as `Message.inputs`). Re-validating here was actually the original bug: when workflow `file_upload` config drifts between rounds, older history files get rejected even though they were valid at upload time.

`file_extra_config` is still load-bearing — it gates the `if file_extra_config and app_record:` block and supplies `image_config.detail`. Only the `config=` argument is unconditionally `None`.

wylswz: If the `config` must be `None` for this function to behave correctly for all known use cases, maybe consider removing the parameter from its signature?

lin-snow: Good call — done in `ee8b19b539`. Both `build_from_message_file` and `build_from_message_files` only have replay-path callers, so I dropped `config` from their signatures and moved the design intent into the module docstring.

wylswz (approved):

PR #30598, opened 2026-01-06
refactor: generate_url to support scenario to build url

fatelei:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

fix #30589 #25168

for plugin for_external is false, use internal file url
for render use file url

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [x] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `make lint` and `make type-check` (backend) and `cd web && npx lint-staged` (frontend) to appease the lint gods

Issue #30589, opened 2026-01-05
Workflow node file uploading use FILES_URL instead of INTERNAL_FILES_URL


Limbo-NKU:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.11.2

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

`.env`:
FILES_URL:http://localhost:8088/ (which is my dify app's frontend url)
INTERNAL_FILES_URL:http://api:5001/

upload a file at the User Input node and pass it to next node.

For more, see Discussion #30588 

### ✔️ Expected Behavior

the `url` in JSON which describes the uploaded file and passed to next node should started with `INTERNAL_FILES_URL` instead of `FILES_URL` in order to  be accessible for other nodes(especially plugins from plugin_daemon).

### ❌ Actual Behavior

Here's my actual JSON which describes the uploaded file
```json
{
  "input": {
    "dify_model_identity": "__dify__file__",
    "id": null,
    "tenant_id": "9b91cd8a-cdae-48d0-bd6b-2a203a687440",
    "type": "document",
    "transfer_method": "local_file",
    "remote_url": "http://localhost:8088/files/70360c66-bce1-4766-b34b-cf702f1d6b1e/file-preview?timestamp=1767637200&nonce=f5119f7f541707654370b27afae41068&sign=g8edWMIe8jKtCjJwb_kvac9zL34ZJ00xocDLS-lFIE8%3D",
    "related_id": "70360c66-bce1-4766-b34b-cf702f1d6b1e",
    "filename": "test.txt",
    "extension": ".txt",
    "mime_type": "text/plain",
    "size": 718,
[...]

PR #21382, opened 2025-06-23
fix: resolve Docker file URL networking issue for plugins (#21334)

krikera:
Fixes #21334

### Problem
Using `STORAGE_TYPE=local` and `FILES_URL=http://localhost` no longer works in Docker environments. Plugin daemon cannot access files uploaded by users.

### Solution  
- Added `INTERNAL_FILES_URL` config for Docker network communication
- Plugin file access now uses internal Docker service URLs (e.g., `http://api:5001`)
- User downloads continue using external URLs (e.g., `http://localhost:5001`)
- Simple fallback: uses `FILES_URL` if `INTERNAL_FILES_URL` not set

### Configuration
Add to Docker environment:
```bash
INTERNAL_FILES_URL=http://api:5001
```

### Testing
- ✅ Plugin file access works in Docker
- ✅ User file downloads work from browser  
- ✅ Backward compatible with existing setups

Issue #21334, opened 2025-06-22
Using STORAGE_TYPE=local and settings FILES_URL=http://localhost no longer works locally (docker)


Korayem:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:)
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.4.3

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Since v.1.4.x, plugins stopped accessing files uploaded by the user in a workflow or chatflow

In previous versions, I had set in .env
```
STORAGE_TYPE=local
FILES_URL=http://localhost
```
Plugins were able to access user uploaded files and also files generated by workflow were downloadable by the user

But as of v1.4.x (dont know which minor version exactly), network access between plugin_daemon and local storage is no longer working

# Quick hacky fix
Changing file_url to the following enables plugins access to uploaded files by the user in a workflow
```
FILES_URL=http://host.docker.internal
```

However, if the workflow generates a file for the user to download, the url becomes `http://host.docker.internal/files/tools/{filename}.md?timestamp={value}&nonce={value}&sign={value}=&as_attachment=true` which doesnt work outside docker local network. 

I don't know what happened that broke this. it was working fine before :)

### ✔️ Expected Behavior

1. User should upload file
2. Plugin should be able to access the file
3. If a workflow generates a file, user should be able to download it

### ❌ Actual Behavior

1. User should upload file
2. Plugin CANT access the file
3. When a workflow generates a file, user CANT download it

PR #35914, opened 2026-05-08
fix: Image rendering in the knowledge base failed.

### LINKED ISSUE #35910: Image rendering in the knowledge base failed
### Self Checks

### Dify version

1.14.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. In the environment variable configuration INTERNAL_FILES_URL, the value is the API service address in the container, such as http://dify-api-svc:5001.
2. Upload images for retrieval during the knowledge base retrieval test.
3. The image preview failed on the chunks details page of the knowledge base.

### ✔️ Expected Behavior

Pictures render and display normally.

### ❌ Actual Behavior

Image rendering failed。

Image

FFXN:
> [!IMPORTANT]
>

## Summary

fixes #35910 

In the environment variable configuration INTERNAL_FILES_URL, the value is the API service address in the container, such as http://dify-api-svc:5001/.
Upload images for retrieval during the knowledge base retrieval test.
Then the browser will not request the image asset.
## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


api/core/tools/signature.py
```diff
@@ -26,12 +26,15 @@ def sign_tool_file(tool_file_id: str, extension: str, for_external: bool = True)
     return f"{file_preview_url}?timestamp={timestamp}&nonce={nonce}&sign={encoded_sign}"
 
 
-def sign_upload_file(upload_file_id: str, extension: str) -> str:
+def sign_upload_file(upload_file_id: str, extension: str, for_external: bool = True) -> str:
```
QuantumGhost: Could we avoid adding `for_external: bool = False` here?

I checked the current call sites, and none of them pass `for_external=False`. They all use the default behavior, and the generated URL is used for user/browser-facing knowledge base or retrieval previews.

Given that, would it be clearer to rename this helper to something like `sign_upload_file_preview_url`, and document in the docstring that it is intended for user/browser-facing preview URLs, not for plugin daemon internal communication?

Some related context:

- #21334 / #21382 introduced `INTERNAL_FILES_URL` to fix plugin daemon file access in Docker. The key distinction there was: plugin daemon/internal access should use `INTERNAL_FILES_URL`, while user downloads should continue using external URLs.
- #30589 / #30598 later made this distinction explicit with `for_external`: plugin/internal consumers should pass `for_external=False`, while render/user-facing paths should use `FILES_URL`.
- #35910 is the opposite failure mode: browser-facing knowledge base image previews received an internal URL such as `http://dify-api-svc:5001/...`, which the browser cannot access.

So for this helper specifically, the current usage seems to be preview/rendering rather than plugin daemon communication. The plugin daemon already has separate internal paths, such as `File.to_plugin_parameter() -> generate_url(for_external=False)`.

(This comment is generated and polished by Codex with GPT 5.5 High)

api/models/dataset.py
```diff
@@ -1020,7 +1020,7 @@ def attachments(self) -> list[AttachmentItem]:
             encoded_sign = base64.urlsafe_b64encode(sign).decode()
 
             params = f"timestamp={timestamp}&nonce={nonce}&sign={encoded_sign}"
-            reference_url = dify_config.CONSOLE_API_URL or ""
+            reference_url = dify_config.FILES_URL or dify_config.CONSOLE_API_URL or ""
```
QuantumGhost, line 1023: Codex suggests replacing this logic with `sign_upload_file` unless we need to keep the compatibility logic for falling back to `CONSOLE_API_URL`.

FFXN: This change aims to fix :

Image

When using CONSOLE_API_URL to generate a preview image, the browser cannot request resources and reports an error 404.

api/core/tools/signature.py
QuantumGhost: What I want to clarify here is the semantics of `sign_upload_file`.

This helper looks like it is used only to generate preview links. If that is the intended contract, then it should not use `INTERNAL_FILES_URL` at all, because `INTERNAL_FILES_URL` is specifically for plugin/internal communication rather than browser-facing preview URLs.

From the current call sites, it seems to be used as a preview URL in all cases.

For the LLM plugin path, I suggest not changing it in this PR. If that path turns out to be a real problem, we can handle it separately by switching the URL used for LLM-facing multimodal input from `sign_upload_file_preview_url` to another helper with the appropriate semantics. From what I can tell, the only problematic case there would be when `MULTIMODAL_SEND_FORMAT=url`.

Additional plugin-side context:

- Plugin tools / agent parameters already use a separate internal path: `File.to_plugin_parameter()` calls `generate_url(for_external=False)`, so that path is already explicitly modeled as plugin/internal communication.
- For plugin-backed LLMs, the behavior depends on the provider implementation:
  - OpenAI mostly forwards the image URL directly to the upstream provider.
  - Gemini and Anthropic fetch the URL inside the plugin first, then convert the content into the provider-specific format.
- So the LLM-plugin concern is real in principle, but it is a different semantic path from browser preview, and I do not think we need to conflate the two in this helper right now.

(This comment is generated and polished by Codex with GPT 5.4 High)

FFXN: Agreed and adopted the recommendations.

api/models/dataset.py
FFXN: I think this is inconsistent with the use case of sign_upload_file_preview_url, sign_upload_file_preview_url it is used to generate URLs with signed information.

QuantumGhost (approved):
LGTM

PR #31586, opened 2026-01-27
refactor: split docker-compose env config into separate files

macatizm:
Refactors docker-compose configuration to split environment variables into separate, organized files.

**Changes:**
- Split env vars into categorized files (global, server, security, database, redis, storage, vectorstore)
- Replace large YAML anchor with env_file references
- Update generation script to create env files
- Makes docker-compose.yaml cleaner and easier to maintain

Fixes #31572

PR #35939, opened 2026-05-08
revert: chore easier and simpler deploy

crazywoola:
## Summary

Reverts PR #35708 (`chore: easier and simpler deploy`).

This restores the previous Docker Compose startup flow, removes the `dify-compose` wrapper scripts and `.env.default`, and returns the Docker README and compose defaults to their prior behavior.

## Validation

- `git diff --check`
- `docker compose -f docker-compose.yaml config --quiet`

Note: `docker compose -f docker-compose-template.yaml config --quiet` fails on `HEAD` as well because the template is not valid as a standalone compose file (`yaml: map merge requires map or sequence of maps as the value`).


lyzno1 (approved):

laipz8200:
I'd like to include this refactoring in #31586.

PR #35976, opened 2026-05-09
refactor(web): converge overlay layering on dify-ui z-50

lyzno1:
## Summary
- converge dify-ui overlay primitives on the shared z-50 layer and keep toast at z-60
- migrate remaining drawer-plus callers to @langgenius/dify-ui/drawer and remove drawer-plus
- replace the migration guide with the current overlay best-practice guide
- remove stale high z-index overlay assertions and keep nested drawer dismissal/backdrop behavior wired through Base UI


CodingOnStar (approved):
LGTM

PR #36035, opened 2026-05-11
fix: regenerate document summary after update via API (#35950)

### LINKED ISSUE #35950: [Bug] Document summary not regenerated after update document throught service api
### Self Checks

### Dify version

1.13.2

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Update an exists document with summary by `GET /v1/datasets/{dataset_id}/documents/{document_id}/update-by-file`.
The document reindexed succeed, but summary not regenerated.

### ✔️ Expected Behavior

The summary regenerate after update document by service api.

### ❌ Actual Behavior

The [`document_indexing_update_task`](https://github.com/langgenius/dify/blob/1.14.0/api/tasks/document_indexing_update_task.py) not call `generate_summary_index_task` like in [`document_indexing_task`](https://github.com/langgenius/dify/blob/1.14.0/api/tasks/document_indexing_task.py#L122).

EvanYao826:
Fixes #35950

## Summary

`document_indexing_update_task` (the API update path) was missing the summary index generation step that `document_indexing_task` (the create path) performs after indexing completes. When a document was updated via the API, its summary was never regenerated even when the dataset had `summary_index_setting` enabled.

## Fix

Added summary generation logic to `api/tasks/document_indexing_update_task.py` after successful `IndexingRunner.run()`, matching the pattern in `document_indexing_task.py`:

- Checks for high_quality technique, enabled summary setting, COMPLETED status, non-QA doc_form, and need_summary=True before calling `generate_summary_index_task.delay()`
- Added a `return` in the exception handler so summary generation is skipped on indexing failure

## Tests

Added 11 unit tests in `test_document_indexing_update_task.py` covering:
- Happy path: summary queued when all conditions met (with and without segments)
- Skip paths: not high_quality, summary disabled, summary None, need_summary False, QA_INDEX form, indexing failure, dataset missing after indexing, document not completed
- Error swallowing: queue call exception doesn't propagate

### DIFF
--- a/api/tasks/document_indexing_update_task.py
+++ b/api/tasks/document_indexing_update_task.py
@@ -72,6 +72,7 @@ def document_indexing_update_task(dataset_id: str, document_id: str):
             segment_delete_stmt = delete(DocumentSegment).where(DocumentSegment.document_id == document_id)
             session.execute(segment_delete_stmt)
 
+    has_error = False
     try:
         indexing_runner = IndexingRunner()
         indexing_runner.run([document])
@@ -79,8 +80,12 @@ def document_indexing_update_task(dataset_id: str, document_id: str):
         logger.info(click.style(f"update document: {document.id} latency: {end_at - start_at}", fg="green"))
     except DocumentIsPausedError as ex:
         logger.info(click.style(str(ex), fg="yellow"))
+        has_error = True
     except Exception:
         logger.exception("document_indexing_update_task failed, document_id: %s", document_id)
+        has_error = True
+
+    if has_error:
         return
 
     # Trigger summary index generation for the updated document if enabled.
--- a/api/tasks/document_indexing_update_task.py
+++ b/api/tasks/document_indexing_update_task.py
@@ -7,10 +7,12 @@ from sqlalchemy import delete, select
 
 from core.db.session_factory import session_factory
 from core.indexing_runner import DocumentIsPausedError, IndexingRunner
+from core.rag.index_processor.constant.index_type import IndexStructureType, IndexTechniqueType
 from core.rag.index_processor.index_processor_factory import IndexProcessorFactory
 from libs.datetime_utils import naive_utc_now
 from models.dataset import Dataset, Document, DocumentSegment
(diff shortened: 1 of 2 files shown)


api/tasks/document_indexing_update_task.py
```diff
@@ -79,3 +81,39 @@ def document_indexing_update_task(dataset_id: str, document_id: str):
         logger.info(click.style(str(ex), fg="yellow"))
     except Exception:
         logger.exception("document_indexing_update_task failed, document_id: %s", document_id)
+        return
```
FFXN, line 89: Thanks for the fix — the overall direction looks right for #35950.

One concern is that the new return in the generic except makes the intent a bit ambiguous: the goal seems to be “skip summary generation if indexing did not complete successfully”, but the code expresses that as “exit the whole task immediately”.

Also, if this is the intended control flow, should DocumentIsPausedError also skip the summary path? Right now generic exceptions return immediately, but DocumentIsPausedError only logs and still continues into the summary eligibility check, which feels inconsistent.

Would it be clearer to mirror document_indexing_task and use an explicit flag here instead, e.g.:

```python
has_error = False
try:
    indexing_runner = IndexingRunner()
    indexing_runner.run([document])
    end_at = time.perf_counter()
    logger.info(click.style(f"update document: {document.id} latency: {end_at - start_at}", fg="green"))
except DocumentIsPausedError as ex:
    logger.info(click.style(str(ex), fg="yellow"))
    has_error = True
except Exception:
    logger.exception("document_indexing_update_task failed, document_id: %s", document_id)
    has_error = True

if has_error:
    return

# Trigger summary index generation...
```

I think that would make the intent more explicit and keep the paused/error paths consistent.

EvanYao826: Thanks for the detailed review! I've updated the code in commit 7723790e to use the `has_error` flag pattern as you suggested. Now `DocumentIsPausedError` also sets `has_error = True`, so both error paths consistently skip the summary generation. Could you take another look?

FFXN (approved):

PR #36238, opened 2026-05-15
fix(dify-agent): don't leave runs running after terminal event write failures

Epochex:
﻿## Summary

`AgentRunRunner` could leave a run stuck in `running` if a terminal event write failed. The scheduler would log the task exception, but the stored run state stayed in-progress, so API clients could keep polling forever.

This changes the runner so failed runs still persist a failed status even when `run_failed` itself cannot be appended. If `run_succeeded` cannot be appended, the run is treated as failed with `reason="event_sink"`, since the output and session snapshot never made it to event consumers.

I also fixed two Dify Agent test-suite issues that showed up while running the full suite locally: the packaged pydantic constraint now matches the existing packaging test, and snippet titles render POSIX-style paths on Windows.

## Testing

- `uv run --project . python -m ruff check .`
- `uv run --project . basedpyright --level error src examples tests`
- `uv run --project . python -m pytest tests` (`177 passed, 5 skipped`)


crazywoola:
Hi @Epochex, thank you for taking the time to contribute to Dify.

### Why this is being closed

### Next steps

Issue #35902, opened 2026-05-07
[Refactor/Chore] if isinstance to match case


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

<img width="555" height="372" alt="Image" src="https://github.com/user-attachments/assets/43bb805b-6db3-48ae-9b4b-d9d157692022" />

change to match case can

1. in case missing branch
2. better semantic meaning

asukaminato0721:
@dosu give me a file list and show me how to do it

EvanYao826:
Hi! I'd like to work on this. I'll start with the files listed by @dosubot and submit PRs incrementally.

Planned first batch:
- `api/core/tools/utils/message_transformer.py`
- `api/services/variable_truncator.py`
- `api/core/prompt/utils/prompt_message_util.py`
- `api/core/app/apps/common/workflow_response_converter.py`

puneetdixit200:
Hi maintainers, I’d like to work on this. I’ll first identify the current isinstance branches, convert only the safe cases to match/case without changing behavior, and run the relevant tests/lint before opening a PR. Please assign this to me if available.

coderback:
Hi, I see others asked to work on this wanted to ask if I could try my hand at it or collaborate with them?

puneetdixit200:
This looks implemented now: #35922 was merged with the title `refactor: convert isinstance chains to match/case (#35902)`, and the checked target files on `main` already use `match`/`case` for the main dispatch chains.

Should this issue be closed as completed, or is there a remaining follow-up scope you still want here?

liaoyl830:
Hi! The previously merged PR #35922 did not cover `api/extensions/otel/parser/base.py`. I have opened a follow-up PR #36534 to convert the remaining isinstance chains in that file to match/case. Please review, thanks!

[...]

PR #36284, opened 2026-05-18
feat: add new agent

zyssyz123:
> [!IMPORTANT]
>

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


api/clients/agent_backend/client.py
```diff
@@ -0,0 +1,126 @@
...
+        """Poll run status until terminal state and normalize client exceptions."""
+        try:
+            return self.client.wait_run_sync(run_id, timeout_seconds=timeout_seconds)
+        except Exception as exc:
+            raise _normalize_dify_agent_error(exc) from exc
+
+
+def _normalize_dify_agent_error(exc: Exception) -> AgentBackendError:
+    """Map public ``dify-agent`` client errors to API-side integration errors."""
+    if isinstance(exc, DifyAgentValidationError):
```
hjlarry: maybe we should change `isinstance` to `match case` ?
because there is a issue here: https://github.com/langgenius/dify/issues/35902

hjlarry (dismissed):
LGTM

hjlarry (dismissed):

MRZHUH (dismissed):
LGTM

api/models/agent.py
```diff
@@ -0,0 +1,223 @@
...
+    INLINE_AGENT = "inline_agent"
+
+
+class Agent(DefaultFieldsMixin, Base):
+    """Workspace-scoped Agent identity used by Agent Roster and workflow-only agents."""
+
+    __tablename__ = "agents"
+    __table_args__ = (
+        sa.PrimaryKeyConstraint("id", name="agent_pkey"),
+        UniqueConstraint("tenant_id", "roster_unique_name", name="agent_tenant_roster_name_unique"),
```
QuantumGhost: I suggest leaving index name empty and let alembic generate names from naming conventions.

api/models/agent.py
```diff
@@ -0,0 +1,223 @@
...
+        Index("agent_tenant_status_updated_at_idx", "tenant_id", "status", "updated_at"),
```
QuantumGhost: I would suggest removing `status` from `agent_tenant_status_updated_at_idx`, since the status enumeration only has two possible values and the selectivity is relatively low. If we found that only a small portion rows are `active`, we would introduce status to index later.

api/models/agent.py
```diff
@@ -0,0 +1,223 @@
...
+        Index("agent_tenant_scope_status_idx", "tenant_id", "scope", "status"),
```
QuantumGhost: Suggestion above about enumerations applies to agent_tenant_scope_status_idx, too.

api/models/agent.py
```diff
@@ -0,0 +1,223 @@
...
+        Index("agent_tenant_workflow_id_idx", "tenant_id", "workflow_id"),
+        Index("agent_tenant_app_id_idx", "tenant_id", "app_id"),
+        Index("agent_active_config_version_id_idx", "active_config_version_id"),
+    )
+
+    tenant_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    name: Mapped[str] = mapped_column(String(255), nullable=False)
+    description: Mapped[str] = mapped_column(LongText, nullable=False, default="")
+    icon_type: Mapped[str | None] = mapped_column(String(255), nullable=True)
```
QuantumGhost: Defining an enum for `icon_type` would be better.

api/models/agent.py
```diff
@@ -0,0 +1,223 @@
...
+    icon: Mapped[str | None] = mapped_column(String(255), nullable=True)
```
QuantumGhost: I suggest add a short comment clarifying the meaning of the `icon` column, as it depends on `icon_type`.

api/models/agent.py
```diff
@@ -0,0 +1,223 @@
...
+        sa.Computed("CASE WHEN scope = 'roster' AND status = 'active' THEN name ELSE NULL END"),
+        nullable=True,
+    )
+    created_by: Mapped[str | None] = mapped_column(StringUUID, nullable=True)
+    updated_by: Mapped[str | None] = mapped_column(StringUUID, nullable=True)
+    archived_by: Mapped[str | None] = mapped_column(StringUUID, nullable=True)
+    archived_at: Mapped[datetime | None] = mapped_column(DateTime, nullable=True)
+
+
+class AgentConfigVersion(Base):
```
QuantumGhost: I suggest inheriting from `DefaultFieldsMixin`.

api/models/agent.py
```diff
@@ -0,0 +1,223 @@
...
+        UniqueConstraint("agent_id", "version", name="agent_config_version_agent_version_unique"),
+        Index("agent_config_version_tenant_agent_created_at_idx", "tenant_id", "agent_id", "created_at"),
+        Index("agent_config_version_tenant_created_at_idx", "tenant_id", "created_at"),
+    )
+
+    id: Mapped[str] = mapped_column(StringUUID, primary_key=True, default=lambda: str(uuidv7()))
+    tenant_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    agent_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    version: Mapped[int] = mapped_column(sa.Integer, nullable=False)
+    config_snapshot: Mapped[str] = mapped_column(LongText, nullable=False, default="{}")
```
QuantumGhost: I would suggest referencing the Pydantic model for the `config_snapshot` field.

An even better approach would be to define a custom `JSONModelColumn` column type that accepts a Pydantic `BaseModel` subclass and handles validation, serialization, and deserialization internally.

api/models/agent.py
```diff
@@ -0,0 +1,255 @@
...
+
+    id: Mapped[str] = mapped_column(StringUUID, primary_key=True, default=lambda: str(uuidv7()))
+    tenant_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    agent_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    agent_config_version_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    revision: Mapped[int] = mapped_column(sa.Integer, nullable=False)
+    operation: Mapped[AgentConfigVersionOperation] = mapped_column(
+        EnumText(AgentConfigVersionOperation, length=64), nullable=False
+    )
+    config_snapshot: Mapped[str] = mapped_column(LongText, nullable=False)
```
QuantumGhost: The same suggestion above about `config_snapshot` applies to here, too.

api/models/agent.py
```diff
@@ -0,0 +1,255 @@
...
+        default=naive_utc_now,
+        server_default=func.current_timestamp(),
+    )
+
+    @property
+    def config_snapshot_dict(self) -> dict[str, Any]:
+        return json.loads(self.config_snapshot) if self.config_snapshot else {}
+
+
+class AgentConfigVersionRevision(Base):
```
QuantumGhost: The names `AgentConfigVersionRevision` and `AgentConfigVersion` are easy to confuse. I'd suggest renaming them to `AgentConfigSnapshot` and `AgentConfigRevision`, respectively.

api/models/agent.py
```diff
@@ -0,0 +1,255 @@
...
+    id: Mapped[str] = mapped_column(StringUUID, primary_key=True, default=lambda: str(uuidv7()))
+    tenant_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    agent_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    agent_config_version_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    revision: Mapped[int] = mapped_column(sa.Integer, nullable=False)
+    operation: Mapped[AgentConfigVersionOperation] = mapped_column(
+        EnumText(AgentConfigVersionOperation, length=64), nullable=False
+    )
+    config_snapshot: Mapped[str] = mapped_column(LongText, nullable=False)
+    previous_config_snapshot: Mapped[str | None] = mapped_column(LongText, nullable=True)
```
QuantumGhost: Given that `AgentConfigVersion` is immutable, storing a duplicate of the agent configuration inside `AgentConfigVersionRevision` seems redundant. Wouldn't it be cleaner to simply reference the corresponding version? I'd suggest replacing the `config_snapshot` and `previous_config_snapshot` fields with logical foreign keys to `AgentConfigVersion`.

api/models/agent.py
```diff
@@ -0,0 +1,255 @@
...
+        Index("workflow_agent_node_binding_workflow_idx", "tenant_id", "workflow_id", "workflow_version"),
+        Index("workflow_agent_node_binding_agent_idx", "tenant_id", "agent_id"),
+        Index("workflow_agent_node_binding_config_version_idx", "tenant_id", "agent_config_version_id"),
+        Index("workflow_agent_node_binding_app_idx", "tenant_id", "app_id"),
+    )
+
+    tenant_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    app_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    workflow_id: Mapped[str] = mapped_column(StringUUID, nullable=False)
+    workflow_version: Mapped[str] = mapped_column(String(255), nullable=False)
```
QuantumGhost: The `workflow_version` column seems unnecessary because the `Workflow` model is also immutable. Editing or publishing a workflow creates a new `Workflow` record.

api/models/agent.py
```diff
@@ -0,0 +1,255 @@
...
+    node_id: Mapped[str] = mapped_column(String(255), nullable=False)
+    binding_type: Mapped[WorkflowAgentBindingType] = mapped_column(
+        EnumText(WorkflowAgentBindingType, length=32), nullable=False
+    )
+    agent_id: Mapped[str | None] = mapped_column(StringUUID, nullable=True)
+    agent_config_version_id: Mapped[str | None] = mapped_column(StringUUID, nullable=True)
+    node_job_config: Mapped[str] = mapped_column(LongText, nullable=False, default="{}")
```
QuantumGhost: The suggestion above about Pydantic model applies to here, too.

api/models/agent.py
```diff
@@ -0,0 +1,255 @@
...
+    __table_args__ = (
+        sa.PrimaryKeyConstraint("id", name="workflow_agent_node_binding_pkey"),
+        UniqueConstraint(
+            "tenant_id",
+            "workflow_id",
+            "workflow_version",
+            "node_id",
+            name="workflow_agent_node_binding_node_unique",
+        ),
+        Index("workflow_agent_node_binding_workflow_idx", "tenant_id", "workflow_id", "workflow_version"),
```
QuantumGhost: The `"tenant_id", "workflow_id", "workflow_version"` index is already covered by the unique index `            "tenant_id", "workflow_id", "workflow_version", "node_id"` above.

api/clients/agent_backend/event_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+
+class AgentBackendRunCancelledInternalEvent(AgentBackendInternalEventBase):
+    """API-internal terminal cancellation event."""
+
+    type: Literal[AgentBackendInternalEventType.RUN_CANCELLED] = AgentBackendInternalEventType.RUN_CANCELLED
+    reason: str | None = None
+    message: str | None = None
+
+
+type AgentBackendInternalEvent = Annotated[
```
QuantumGhost, line 95: 👍

api/services/agent/roster_service.py
```diff
@@ -0,0 +1,333 @@
+import json
+from typing import Any
+
+from sqlalchemy import func, select
+from sqlalchemy.exc import IntegrityError
+
+from extensions.ext_database import db
```
QuantumGhost: I'd recommend injecting a `Session` / `sessionmaker` through the constructor rather than relying on a global `db` instance. This keeps dependencies explicit and makes the code far easier to test and maintain.

Ideally, we should not rely on web framework in business logic layers.

QuantumGhost (approved):

MRZHUH (approved):

Issue #32454, opened 2026-02-21
[Refactor/Chore] use Testcontainers to do sql test


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

now it's hard to do SQL improvement since all the query is patch.

Testcontainers will be much easier.

### Motivation

Improve DX

### Additional Context

Chat with llm https://g.co/gemini/share/abe5288b661a (chat is in Chinese).

asukaminato0721:
@dosu what is your opinion

KinomotoMio:
I'd like to work on this issue.

asukaminato0721:
@dosu give a list on all the file needs to migrate

asukaminato0721:
@dosu give a list on all the file needs to migrate

asukaminato0721:
@dosu give a list on all the file needs to migrate

asukaminato0721:
@dosu give a list on all the file needs to migrate

YB0y:
Hi @asukaminato0721, I think there are still several areas that need to be fixed. May I work on this issue? Could you assign it to me?

YB0y:
Thanks @asukaminato0721

Daedalus-Icarus:
@dosu give a list of files that are still not migrated.

asukaminato0721:
@dosu give a list of files that are still not migrated.

asukaminato0721:
https://github.com/langgenius/dify/issues/33891#issuecomment-4108679025

This can be prioritized.

asukaminato0721:
@dosu give a list on all the file needs to migrate

Daedalus-Icarus:
@YB0y could you help me with migrating those tests to Testcontainer first?

- `api/tests/unit_tests/controllers/web/test_conversation.py`
[...]

PR #36304, opened 2026-05-18
test(api): isolate container DB between tests

escape0707:
## Summary

Part of https://github.com/langgenius/dify/issues/32454#issuecomment-4417553947.

This is a diagnostic draft PR. It is expected to expose existing DB session leaks before the service fixes are applied.

The code change is the intended test infrastructure direction: after each container integration test that used the containerized Flask app, reset SQLAlchemy application tables with `TRUNCATE ... RESTART IDENTITY CASCADE`.

The cleanup also sets `SET LOCAL lock_timeout = '5s'` for the cleanup transaction. That is test-only behavior in `conftest.py`: if another session is holding a conflicting lock, CI should fail loudly instead of hanging for a long time.

## Why this exists

Container integration tests currently share one PostgreSQL container and many tests/services commit through different sessions. Without a central cleanup contract, DB state can leak between tests unless each test adds its own local cleanup workaround.

This draft PR records what happens when central DB isolation is enabled before fixing the existing session leaks. The follow-up service PR should close the leaked private sessions, and the final DB isolation PR should build on that.

## Validation

- `uv run --project api ruff check api/tests/test_containers_integration_tests/conftest.py`

### DIFF
--- a/api/tests/test_containers_integration_tests/conftest.py
+++ b/api/tests/test_containers_integration_tests/conftest.py
@@ -496,6 +496,51 @@ def db_session_with_containers(flask_app_with_containers: Flask) -> Generator[Se
             logger.debug("Database session closed")
 
 
+def _truncate_container_database(app: Flask) -> None:
+    """
+    Reset application tables after a container integration test.
+
+    Tests in this package share one PostgreSQL container for performance, while
+    application code may commit through db.session, Session(db.engine), or
+    session_factory-created sessions. Truncating after each test gives the suite
+    a central DB isolation contract that does not depend on which session a test used.
+    This only covers SQLAlchemy application tables in db.metadata for now;
+    Redis, object storage, and custom ad hoc metadata still need their own cleanup.
+    """
+    with app.app_context():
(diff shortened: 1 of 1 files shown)


escape0707:
Diagnostic result recorded.

This draft PR proved the intended failure mode: enabling central container DB cleanup before the service session fixes makes API Integration Tests fail loudly during cleanup instead of hanging.

Evidence from CI:

```text
ERROR api/tests/test_containers_integration_tests/services/test_attachment_service.py::TestAttachmentService::test_should_return_base64_when_file_exists - sqlalchemy.exc.OperationalError: (psycopg2.errors.LockNotAvailable) canceling statement due to lock timeout
[SQL: TRUNCATE TABLE ... upload_files ... RESTART IDENTITY CASCADE]
```

The job completed with `2409 passed, 8 skipped, 2 errors`, so this was not a setup failure. The errors were the expected leaked-session/cleanup-lock signal.

Follow-up PRs:

- Service session fixes: #36308
- Final DB isolation PR after those fixes: #36310

PR #36328, opened 2026-05-18
refactor: remove oclif

wylswz:
> [!IMPORTANT]
>

## Summary

- Removed oclif from the code base.
- Separated output rendering from the command layer.

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


cli/src/framework/run.ts
```diff
@@ -0,0 +1,83 @@
...
+    const Ctor = resolved.command
+    const cmd = new Ctor()
+    cmd._setArgv(argv.slice(resolved.path.length))
+    const output = await cmd.run()
+    if (output !== undefined)
+      process.stdout.write(stringifyOutput(output))
+  }
+  catch (err) {
+    if (err instanceof Error) {
+      process.stderr.write(`${err.message}\n`)
```
GareArc, line 50: Current handler drops hints, semantic exit codes and the --format=json error envelope. Consider restoring formatErrorForCli() + BaseError.exit() from the old DifyCommand.catch().

cli/src/commands/_shared/dify-command.ts
```diff
@@ -1,21 +1,13 @@
-import type { CommandError } from '@oclif/core/interfaces'
 import type { AuthedContext, AuthedContextOptions } from './authed-command.js'
-import { Command } from '@oclif/core'
-import { isBaseError } from '../../errors/base.js'
-import { formatErrorForCli } from '../../errors/format.js'
+import { Command } from '../../framework/command.js'
 import { buildAuthedContext } from './authed-command.js'
 
 export abstract class DifyCommand extends Command {
   protected outputFormat = ''
```
GareArc: outputFormat is set but never read

cli/src/framework/flags.ts
```diff
@@ -0,0 +1,161 @@
...
+
+      const def = meta.flags[name]
+      if (!def)
+        continue
+
+      if (def.type === 'boolean') {
+        flags[name] = rawValue === undefined ? true : coerceFlagValue(rawValue, def)
+      }
+      else if (rawValue !== undefined) {
+        flags[name] = coerceFlagValue(rawValue, def)
```
GareArc: `flags[name] = …` overwrites on each occurrence, repeatable flags will break. (e.g. `--file key=@path`, repeatable)

cli/src/framework/flags.ts
```diff
@@ -0,0 +1,161 @@
...
+        rawValue = token.slice(eqIdx + 1)
+      }
+      else {
+        name = token.slice(2)
+        rawValue = undefined
+      }
+
+      const def = meta.flags[name]
+      if (!def)
+        continue
```
lin-snow: @wylswz 

Unknown flags get silently swallowed — won't difyctl get app --typoflag my-id end up treating my-id as a positional arg?

cli/src/framework/run.ts
```diff
@@ -0,0 +1,83 @@
+import type { CommandTree } from './registry.js'
+import { formatHelp } from './help.js'
+import { stringifyOutput } from './output.js'
+import { findSuggestions, resolveCommand } from './registry.js'
+
+export async function run(tree: CommandTree, argv: string[]): Promise<void> {
+  if (argv.length === 0 || argv[0] === 'help' || argv.includes('--help') || argv.includes('-h')) {
```
lin-snow, line 7: @wylswz 

Any argv containing help gets intercepted as help-mode — doesn't this mean the help subcommand in the tree can never actually be reached? 🤔

cli/src/framework/registry.ts
```diff
@@ -0,0 +1,108 @@
...
+      const nextToken = argv[i + 1]
+      if (nextToken === undefined || nextToken.startsWith('-') || !(nextToken in node.subcommands))
+        return lastMatch
+    }
+  }
+
+  return lastMatch
+}
+
+export function findSuggestions(tree: CommandTree, argv: string[]): string[] {
```
GareArc: findSuggestions() enumerates siblings. Consider edit distance?

cli/src/commands/get/app/handlers.ts
lin-snow: dead code?

cli/src/framework/help.ts
```diff
@@ -0,0 +1,71 @@
...
+    lines.push('EXAMPLES')
+
+    for (const ex of ctor.examples) {
+      lines.push(`  $ ${ex.replace('<%= config.bin %>', bin)}`)
+    }
+
+    lines.push('')
+  }
+
+  const guide = (ctor as unknown as Record<string, unknown>).agentGuide
```
lin-snow: **Unsafe type assertion for `agentGuide`** (`framework/help.ts:65`) — the `as unknown as Record<string, unknown>` cast hides typos and breaks rename/go-to-def. Just declare `static agentGuide?: string` on the `Command` base class, then `ctor.agentGuide` works with no assertion.

cli/src/framework/command.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  abstract run(): Promise<CommandOutput | void>
+
+  static async run(argv: string[] = []): Promise<CommandOutput | void> {
+    const cmd = new (this as unknown as { new(): Command })()
+    cmd._setArgv(argv)
+    return cmd.run()
+  }
+
+  protected parse<C extends CommandConstructor>(ctor: C): ParseResult<C> {
```
lin-snow: `parse()` isn't async but everyone `await` it (`framework/command.ts:38`) — `parse()` returns `ParseResult<C>`, not `Promise<ParseResult<C>>`, but every call site does `await this.parse(...)` (e.g. `get/app/index.ts:38`). It works because `await` on a non-Promise is a no-op, but it's misleading. Either make `parse()` async or drop the `await` at call sites.

cli/src/framework/command.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export abstract class Command implements ICommand {
+  static description?: string
+  static flags: Record<string, FlagDefinition<string | boolean | number | string[] | undefined>> = {}
+  static args: Record<string, ArgDefinition<string | undefined>> = {}
+  static examples: string[] = []
+
+  private _argv: readonly string[] = []
+
+  _setArgv(argv: readonly string[]): void {
```
lin-snow: **`_setArgv` is a public method called externally** (`framework/command.ts:24-28`) — the underscore is just convention; `run.ts:44` does `new Ctor()` then `cmd._setArgv(...)`, which is two-phase construction (incomplete object between `new` and the setter) and lets any caller mutate argv. Use constructor injection instead: `constructor(argv: readonly string[] = [])`, then `new Ctor(argv.slice(...))` — `_argv` becomes `readonly`, no half-built window, and `_setArgv` disappears entirely.

cli/src/framework/run.ts
wylswz: Actually it's a good idea that help stays with the target command's definition instead of being a separated one.

I'll consider adding a custom help method to abstract command. By default, the help is derived from the argument definitions.

cli/src/framework/registry.ts
wylswz: Added a soft search using edit distance.

cli/src/commands/get/app/handlers.ts
wylswz: Yes.

cli/src/framework/command.ts
wylswz: I'll make argv an argument of run method instead.

PR #36332, opened 2026-05-18
fix(docker): harden default SSRF proxy egress

### LINKED ISSUE #36400: [Proposal] Harden Docker Compose SSRF proxy defaults
## Self Checks

## 1. Is this request related to a challenge you're experiencing? Tell me about your story.

Dify's Docker Compose deployment currently routes API, worker, sandbox, and related network features through the bundled Squid-based `ssrf_proxy`. The intended goal is to make network features work out of the box while reducing SSRF risk.

However, the current Squid configuration includes the Ubuntu Squid image's default `/etc/squid/conf.d/*.conf`. That packaged configuration allows `localnet` clients. In Docker Compose, Dify services connect from Docker private-network source addresses, so this effectively allows the proxy clients to reach arbitrary destinations, including loopback, RFC1918 private networks, link-local addresses, and cloud metadata endpoints.

This creates recurring SSRF concerns for community self-hosted deployments. At the same time, switching to a strict domain allowlist would break the expected out-of-the-box behavior for HTTP nodes, remote file ingestion, crawlers, tools, and other user-configured public network calls.

The proposed behavior is:

- Keep public outbound access working by default.
- Block loopback, private, CGN, link-local, ULA, multicast/reserved, and cloud metadata target networks by default.
- Preserve the existing sandbox reverse proxy path.
- Provide explicit opt-in allowlists for trusted private-network deployments that intentionally need private egress.

A draft implementation is available in #36332.

## 2. Additional context or comments

This proposal should be coordinated with #36397 before #36332 is moved forward.

The hardening in #36332 intentionally blocks private-network targets. That can regress local Docker deployments when a Dify-signed `/files/...` URL is passed back into Dify as a remote URL, because `FILES_URL` or `INTERNAL_FILES_URL` may resolve to `localhost`, a Docker-internal hostname, or another private address.

#36397 should centralize handling for Dify-owned signed file URLs so they are verified and resolved through an internal file/storage path instead of being treated as arbitrary external remote URLs through the SSRF proxy. After that path is in place, the SSRF proxy hardening can proceed without breaking local signed-file flows.

Related PR: #36332
Blocked by: #36397

## 3. Can you help us with this feature?

laipz8200:
## Summary

Closes #36400.

- Harden the Docker Compose SSRF proxy defaults by denying loopback, private, CGN, link-local, ULA, multicast, reserved, and cloud metadata target networks before general egress is allowed.
- Stop relying on the Ubuntu Squid image's packaged `conf.d` access rules so `allow localnet` cannot bypass Dify's policy.
- Preserve the sandbox reverse proxy path and add explicit `SSRF_PROXY_ALLOW_PRIVATE_IPS` / `SSRF_PROXY_ALLOW_PRIVATE_DOMAINS` escape hatches for trusted private-network deployments.
- Continue after #36397 was resolved by #36399, which centralizes Dify-signed file URL retrieval outside the generic SSRF proxy path.

## Screenshots

N/A

## Checklist


laipz8200:
Update: this PR should stay draft/paused until #36397 is resolved.

The proxy hardening in this PR blocks loopback, private, link-local, and metadata target networks by default. That is the intended security posture, but it also means local Docker deployments can regress when a Dify-signed `/files/...` URL is passed back into Dify as a remote URL. In those deployments, `FILES_URL` or `INTERNAL_FILES_URL` may resolve to `localhost`, a Docker-internal hostname, or another private address, so the current SSRF proxy policy would reject the request.

We should first land the centralized signed-file retrieval work in #36397, so Dify-owned signed file URLs are verified and resolved through an internal file/storage path instead of being treated as arbitrary external remote URLs through the SSRF proxy. Once that is in place, we can resume this PR without breaking local signed-file flows.

docker/ssrf_proxy/squid.conf.template
```diff
@@ -1,11 +1,23 @@
...
-acl localnet src fe80::/10      	# RFC 4291 link-local (directly plugged) machines
+acl client_localnet src 0.0.0.1-0.255.255.255	# RFC 1122 "this" network (LAN)
+acl client_localnet src 10.0.0.0/8		# RFC 1918 local private network (LAN)
+acl client_localnet src 100.64.0.0/10		# RFC 6598 shared address space (CGN)
+acl client_localnet src 169.254.0.0/16 	# RFC 3927 link-local (directly plugged) machines
+acl client_localnet src 172.16.0.0/12		# RFC 1918 local private network (LAN)
+acl client_localnet src 192.168.0.0/16		# RFC 1918 local private network (LAN)
+acl client_localnet src fc00::/7       	# RFC 4193 local private network range
+acl client_localnet src fe80::/10      	# RFC 4291 link-local (directly plugged) machines
+acl to_private_networks dst 0.0.0.0/32
```
QuantumGhost: ~~AFAIK, 0.0.0.0 should not be used as DST address.~~

Well, there is even a [vulnerability](https://www.oligo.security/blog/0-0-0-0-day-exploiting-localhost-apis-from-the-browser) for 0.0.0.0 address.

I would suggest blocking the entire `0.0.0.0/8`, since it should not be used as destination according to [IANA](https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml)

docker/ssrf_proxy/squid.conf.template
```diff
@@ -19,19 +31,27 @@ acl Safe_ports port 488		# gss-http
 acl Safe_ports port 591		# filemaker
 acl Safe_ports port 777		# multiling http
 acl CONNECT method CONNECT
+acl sandbox_reverse_proxy_port localport ${REVERSE_PROXY_PORT}
```
QuantumGhost: It seems that requests to dify-sandbox does not go through `ssrf_proxy`. I woud suggest removing the sandbox related configurations.

docker/ssrf_proxy/squid.conf.template
```diff
@@ -1,11 +1,23 @@
...
+acl to_private_networks dst 0.0.0.0/32
+acl to_private_networks dst 10.0.0.0/8
+acl to_private_networks dst 100.64.0.0/10
+acl to_private_networks dst 127.0.0.0/8
+acl to_private_networks dst 169.254.0.0/16
+acl to_private_networks dst 172.16.0.0/12
+acl to_private_networks dst 192.168.0.0/16
+acl to_private_networks dst 224.0.0.0/4
+acl to_private_networks dst 240.0.0.0/4
+acl to_private_networks dst ::1/128
```
QuantumGhost, line 19: `::/128` should also be blocked since Linux treats it as `::1/128`.

docker/ssrf_proxy/squid.conf.template
```diff
@@ -1,11 +1,23 @@
...
+acl to_private_networks dst fc00::/7
+acl to_private_networks dst fe80::/10
```
QuantumGhost, line 23: For IPv6 I suggest adding the following rules:

```
acl to_private_networks dst ::ffff:0:0/96    # IPv4-mapped
acl to_private_networks dst ::/96            # deprecated IPv4-compatible
```

QuantumGhost (commented):
`SSRF_SANDBOX_PROXY_TARGET_PORT` is not included in the environment variable example.

QuantumGhost (approved):

PR #36351, opened 2026-05-18
fix(workflow): persist STOPPED status on stop

### LINKED ISSUE #36204: Workflow run remains running after being terminated in v1.14.1
## Description

In both Dify Cloud and self-hosted Dify v1.14.1, after running a workflow and terminating the workflow run, the workflow run status remains `running` instead of transitioning to a terminal state.

## Affected versions

- Dify Cloud: v1.14.1
- Self-hosted Dify: v1.14.1

## Steps to reproduce

1. Open a workflow in Dify v1.14.1.
2. Run the workflow.
3. Terminate the running workflow.
4. Check the workflow run status.

## Actual behavior

The workflow run status stays as `running` after termination.

## Expected behavior

After the workflow run is terminated, its status should change to an appropriate terminal state, such as `stopped` or `failed`, instead of remaining `running`.

## Evidence

Video recording: 

https://github.com/user-attachments/assets/5ccb2d45-b44f-4294-ac98-6dc2bf852884

## Additional context

This behavior was observed in both the cloud-hosted version and a self-hosted deployment running v1.14.1.

Epochex:
Fixes #36204.

When a workflow run is terminated while the engine is waiting on external I/O (LLM/tool calls), the stream can end via QueueStopEvent before GraphEngine emits GraphRunAbortedEvent. In that case, the UI/log record can remain RUNNING even though the user stopped the run.

This updates WorkflowAppGenerateTaskPipeline to best-effort persist STOPPED status on QueueStopEvent (only when the record is still non-terminal), so the workflow_runs table reflects the stop action immediately.

Tests:
- uv run --project api pytest api/tests/unit_tests/core/app/apps/workflow/test_generate_task_pipeline.py -q


Epochex:
﻿Actions checks haven’t started for this fork PR yet (workflow runs show “Action required”).

I ran the focused unit test locally and it passes:

- uv run --project api --frozen pytest -q api/tests/unit_tests/core/app/apps/workflow/test_generate_task_pipeline.py

Could a maintainer please approve/run the workflows so CI can execute?

Issue #36386, opened 2026-05-19
OpenAI-API-compatible provider keeps broken Perplexity Sonar credentials and validation fails


fukayanozomifb4-maker:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.14.1

### Cloud or Self Hosted

Cloud

### Steps to reproduce

1. Add Perplexity Sonar models using OpenAI-API-compatible provider
2. Delete or modify credentials
3. Try reopening credential configuration
4. Reinstall provider
5. Try adding a new Sonar model

### ✔️ Expected Behavior

Credentials should delete correctly and deleted models should not reappear after reinstalling the provider.

Validation for newly added Sonar models should succeed with valid max token settings.

### ❌ Actual Behavior

Using the OpenAI-API-compatible provider with Perplexity Sonar models causes broken credential references.

Deleted credentials cannot be removed because Dify says they are still in use, but opening the credential configuration shows:
"Credential with id xxx not found"

Additionally, reinstalling the provider causes deleted Sonar models to reappear automatically.

Trying to add a new Sonar model with a new credential also fails validation incorrectly with:
"max_tokens must be at least 16 for sonar"

even though max tokens is set properly.

This issue may indicate orphaned credential references remaining in the workspace state after model/provider deletion.

<img width="1282" height="946" alt="Image" src="https://github.com/user-attachments/assets/e1dd1f1f-7bdb-4fff-8bab-fa171ec60541" />
<img width="1267" height="910" alt="Image" src="https://github.com/user-attachments/assets/d1443dbb-1496-467b-a3a0-481a3e410402" />
<img width="1272" height="966" alt="Image" src="https://github.com/user-attachments/assets/e914c50a-6678-46eb-a6cf-517ba58bf08e" />

fukayanozomifb4-maker:
[...]

Issue #36129, opened 2026-05-13
OpenAI-API-compatible LLM plugin - credentials unavailable


evitolo:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.14.1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

after update to 1.14.1 LLMs configured with OpenAI-API-compatible LLM plugin do not work anymore

### ✔️ Expected Behavior

Working LLM models

### ❌ Actual Behavior

1. Models credentials unavailable
<img width="483" height="248" alt="Image" src="https://github.com/user-attachments/assets/b6bf7cac-f56d-402c-b067-ef43e42ce0a3" />

2. Unable to remove models, message "Action succeed" but model not deleted

evitolo:
@dosu I have removed and reinstalled plugin but it did not help. After reinstalling I see the old models with missing auth configuration.

evitolo:
@dosu Dify version 1.14.1

Rickzhang512:
> [@dosu](https://github.com/dosu) Dify version 1.14.1

I encounter the same questions as well. Which version did you update from? I updated this from 1.11.2 to 1.14.0.

evitolo:
> > [@dosu](https://github.com/dosu) Dify version 1.14.1
> 
> I encounter the same questions as well. Which version did you update from? I updated this from 1.11.2 to 1.14.0.

One instance that I updated was version 13.3.3; the other two instances were version 1.14.0. All had the same result. Credentials are not available, models can not be removed.

evitolo:
[@dosu](https://go.dosu.dev/dosubot?utm_source=github&utm_medium=bot-comment&utm_campaign=github-comment-footer-20260415&utm_content=reply-with-mention&utm_term=langgenius%2Fdify).

I was able to add a new model 

[...]

PR #36124, opened 2026-05-13
chore(api): upgrade graphon to 0.4.0

laipz8200:
Fixes #36123

## Summary

- Upgrade the API Graphon dependency to `0.4.0` and refresh the lockfile.
- Migrate Dify workflow runtime adapters to Graphon's renamed LLM and code-node protocols.
- Update HITL human-input models to the new paragraph/select/file input configuration shape while preserving legacy text-input workflow compatibility.
- Regenerate the console OpenAPI markdown and TypeScript API contracts from the updated schema.

## Screenshots

| Before | After |
|--------|-------|
| N/A | N/A |

## Checklist

- [x] I understand that this PR may be closed in case there was no previous discussion or issues.

PR #36427, opened 2026-05-20
fix(api): fix legacy model_type query compatibility for provider reads

### LINKED ISSUE #36426: Legacy model_type values break custom model credential management

## Summary
Custom model credential management breaks for workspaces that still have legacy `model_type` values in provider-related tables. Requests with canonical query values such as `model_type=llm` can return `Credential with id ... not found`, which prevents users from loading, updating, switching, or deleting OpenAI-compatible custom model credentials.

## Root cause
The `EnumText(ModelType)` migration removed legacy query normalization from several read paths while keeping only result deserialization compatibility.

`ModelType.value_of("text-generation")` still deserializes legacy rows to `ModelType.LLM`, but SQL bind parameters are emitted as canonical values such as `llm`. Rows that still store legacy values such as `text-generation` or `embeddings` are therefore skipped by read queries.

The regression affects mixed datasets in tables including:
- `provider_models`
- `tenant_default_models`
- `provider_model_settings`
- `provider_model_credentials`
- `load_balancing_model_configs`

## Reproduction
1. Persist a custom model credential row with `provider_model_credentials.model_type = 'text-generation'`.
2. Open the custom model credential endpoint with `model_type=llm`.
3. Observe `Credential with id ... not found`.

## Impact
- Users cannot load, update, switch, or delete affected custom model credentials.
- Available credential lists can appear empty.
- Default model and load balancing lookups can also miss legacy rows.

## Proposed fix
Use a shared read-path helper that matches both canonical and legacy `model_type` values during queries, while keeping new writes on canonical enum values.

QuantumGhost:
**AI disclosure**: This PR was primarily drafted with Codex using GPT-5.4. I have reviewed and edited the final diff, and I am responsible for the content.

Closes #36426 #36129 #36386

## Summary
This fixes a regression in provider-side model credential reads for workspaces that still contain legacy `model_type` values in provider-related tables.

The important change in this revision is that the compatibility logic is now deterministic even when canonical and legacy rows coexist for the same logical scope:
- single-row lookups now query canonical first and only fall back to legacy when canonical is missing
- multi-row reads now fetch the raw persisted `model_type` string and prefer canonical rows during assembly; legacy rows are used only when the canonical counterpart does not exist for that scope

This keeps new writes on canonical enum values while making mixed datasets readable in a predictable way.

## Historical context
There are two layers of `ModelType` history that matter here:

1. **Definition-level semantics**
   - Before #1858, Dify still used provider-native strings such as `text-generation`, `embeddings`, and `speech2text` directly.
   - #1858 introduced the canonical internal `ModelType` values (`llm`, `text-embedding`, `rerank`, `speech2text`, `moderation`) plus compatibility helpers such as `value_of()` and `to_origin_model_type()`.
   - #2033 later completed that canonical set by formally adding `tts`.

2. **Persistence-layer semantics**
   - Even after canonical `ModelType` values existed in code, provider-related tables could still persist legacy/origin strings for compatibility.
   - #34300 was the real persistence switch: new writes for provider-related `model_type` columns started using canonical enum values instead of legacy strings.
   - That created a mixed-data period where old rows could still contain values like `text-generation`, `embeddings`, and `reranking`, while new rows used `llm`, `text-embedding`, and `rerank`.

The short version is:
- the semantic switch in code happened in #1858
- the semantic switch in provider-table persistence happened in #34300

## Cross references
- #34300 (`1344c3b280`) switched several provider tables to `EnumText(ModelType)` and removed the older `to_origin_model_type()` query normalization from controller/provider read paths.
- #34717 (`e138523123`) restored backward-compatible deserialization in `EnumText.process_result_value()`, so legacy persisted values could be loaded after a row was already selected.
- #36124 upgraded the standalone `graphon` package to `0.4.0`. That upgrade did not introduce a new `ModelType` semantic change, but it helped surface the remaining gap in Dify's query compatibility over mixed legacy/canonical rows.
- The standalone `graphon` repository itself is not the source of this semantic transition. Its visible releases inherit the already-stable canonical/legacy mapping that Dify had established earlier.

## Canonical vs legacy mapping
At the moment, the important compatibility pairs are:
- `llm` <-> `text-generation`
- `text-embedding` <-> `embeddings`
- `rerank` <-> `reranking`

The other currently supported values (`speech2text`, `moderation`, `tts`) did not undergo the same string transition.

## Root cause
`ModelType.value_of("text-generation")` still maps to `ModelType.LLM`, so once a row is fetched the in-memory value is fine.

The original failure happened earlier in SQL filtering:
- request/query paths passed canonical values such as `llm`
- SQLAlchemy bind parameters therefore emitted `model_type = 'llm'`
- legacy rows that still stored `text-generation` were skipped before deserialization ever ran

After the first compatibility patch, another issue remained: if both canonical and legacy rows existed for the same logical scope, an `OR`-style compatibility filter could still produce ambiguous or unstable results. This PR removes that ambiguity by splitting the handling for single-row and multi-row queries.

### Single-row queries
| Scenario | Location | Category | API |
| --- | --- | --- | --- |
| Read custom model record by `provider + model + model_type` | `ProviderConfiguration._get_custom_model_record()` | Single-row | `scalar_one_or_none()` |
| Read a specific custom model credential by `credential_id + model scope` | `ProviderConfiguration._get_specific_custom_model_credential()` | Single-row | `scalar_one_or_none()` |
| Read tenant default model by `model_type` | `ProviderManager.get_default_model()` | Single-row | `db.session.scalar()` |
| Read existing tenant default model before update | `ProviderManager.update_default_model_record()` | Single-row | `db.session.scalar()` |
| Read provider model setting by `model + model_type` | `ProviderConfiguration._get_provider_model_setting()` | Single-row | `first()` |
| Read load-balancing config detail by `config_id + model scope` | `ModelLoadBalancingService.get_load_balancing_config()` | Single-row | `db.session.scalar(...limit(1))` |
| Read model-level provider record for ops trace credential lookup | `ops_trace_manager._lookup_llm_credential_info()` | Single-row | `session.scalar()` |

### Single-row queries inside write flows
| Scenario | Location | Category | API |
| --- | --- | --- | --- |
| Reuse hidden secret from existing custom model credential during validation | `ProviderConfiguration.validate_custom_model_credentials()` | Single-row | `scalar_one_or_none()` |
| Read target custom model credential before update | `ProviderConfiguration.update_custom_model_credential()` | Single-row | `scalar_one_or_none()` |
| Read target custom model credential before delete | `ProviderConfiguration.delete_custom_model_credential()` | Single-row | `scalar_one_or_none()` |
| Read target custom model credential before adding it to model list | `ProviderConfiguration.add_model_credential_to_model()` | Single-row | `scalar_one_or_none()` |
| Read target custom model credential before switching active credential | `ProviderConfiguration.switch_custom_model_credential()` | Single-row | `scalar_one_or_none()` |
| Read custom model credential by `credential_id` while updating load-balancing configs | `ModelLoadBalancingService.update_load_balancing_configs()` | Single-row | `db.session.scalar(...limit(1))` |
| Read existing load-balancing config by `config_id` during validation | `ModelLoadBalancingService.validate_load_balancing_credentials()` | Single-row | `db.session.scalar(...limit(1))` |

### Multi-row queries
| Scenario | Location | Category | API |
| --- | --- | --- | --- |
| Read available custom model credentials for a `provider + model + model_type` scope | `ProviderManager.get_provider_model_available_credentials()` | Multi-row | `all()` |
| Read load-balancing configs for a `provider + model + model_type` scope | `ModelLoadBalancingService.get_load_balancing_configs()` | Multi-row | `all()` |
| Read current load-balancing configs before updating them | `ModelLoadBalancingService.update_load_balancing_configs()` | Multi-row | `all()` |
| Bulk-load all valid provider models for workspace assembly | `ProviderManager._get_all_provider_models()` | Multi-row | direct iteration over query results |
| Bulk-load all provider model settings for workspace assembly | `ProviderManager._get_all_provider_model_settings()` | Multi-row | direct iteration over query results |
| Bulk-load all provider model credentials for workspace assembly | `ProviderManager._get_all_provider_model_credentials()` | Multi-row | direct iteration over query results |
| Bulk-load all load-balancing configs for workspace assembly | `ProviderManager._get_all_provider_load_balancing_configs()` | Multi-row | direct iteration over query results |

### 1. Single-row queries: canonical-first fallback
For `scalar_one_or_none()`, `scalar()`, and `first()` style lookups, this PR no longer relies on a single `canonical OR legacy` filter.

Instead, the flow is explicit:
1. query using the canonical enum value
2. if no row is found, query using the legacy/origin string

This guarantees that when both canonical and legacy rows exist, the canonical row wins deterministically.

### 2. Multi-row queries: persisted-type-aware preference
For `all()` and bulk-iteration query paths, a pure fallback query is not sufficient because multiple rows may legitimately be returned.

For these paths, the query now includes an extra raw persisted type column using:
- `sa.type_coerce(model.model_type, sa.String())`

The assembly layer then applies a small, isolated compatibility rule:
- group rows by their logical scope
- if canonical rows exist in that scope, use only the canonical rows
- if canonical rows do not exist, use the legacy rows for that scope

This matters because once ORM deserialization runs, both `llm` and `text-generation` become `ModelType.LLM` in Python. The raw persisted string is therefore required to resolve precedence correctly.

### 3. Why the logic is isolated
The compatibility logic is kept in a dedicated helper module so it can be removed cleanly after the legacy rows are migrated away. The business modules only decide which query category they are in:
- single-row lookup -> canonical-first fallback helper
- multi-row assembly -> persisted-type-aware preference helper

## User-visible failure
This showed up in `CUS-1232` on OpenAI-compatible custom model credential flows:
- opening custom model credentials could fail with `Credential with id ... not found`
- saving or switching credentials could fail because the current credential/config row was not found
- available credential lists and related load balancing/default-model lookups could also miss legacy rows in mixed datasets
- when both canonical and legacy rows existed, some reads could also become ambiguous or nondeterministic

## Scope of the fix
This PR applies the deterministic compatibility strategy to provider-side read paths, including:
- custom model credential lookup
- available custom model credential lists
- provider model and default-model related reads
- load-balancing config and credential reads
- ops trace credential lookup
- provider-manager bulk assembly helpers

The important boundary is intentional:
- reads prefer canonical rows and only fall back to legacy when canonical is absent
- writes still use canonical enum values
- this PR does not perform a data migration or backfill

## Why this shape
I kept the fix focused on read-path compatibility because it addresses the production failure immediately without mixing the hotfix with a broader data cleanup. A later migration can still normalize old rows once operational risk is lower.

## Validation
- `make lint`
- `make type-check`
- `make test`

### DIFF
--- a/api/core/entities/provider_configuration.py
+++ b/api/core/entities/provider_configuration.py
@@ -403,6 +403,11 @@ class ProviderConfiguration(BaseModel):
                 ProviderModelCredential.tenant_id == self.tenant_id,
                 ProviderModelCredential.provider_name.in_(self._get_provider_names()),
                 ProviderModelCredential.model_name == model,
+                # While this is not a query returning one row, for
+                # API key name generation it's ok to keep both rows with canonical enum values
+                # and rows with legacy enum values.
+                #
+                # The `max()` call will ensure the result works as expected.
                 legacy_compatible_model_type_filter(ProviderModelCredential.model_type, model_type),
             ),
         )
--- a/api/core/entities/provider_configuration.py
+++ b/api/core/entities/provider_configuration.py
@@ -48,6 +48,7 @@ from models.provider import (
     TenantPreferredModelProvider,
 )
 from models.provider_ids import ModelProviderID
+from models.types import legacy_compatible_model_type_filter
 
 logger = logging.getLogger(__name__)
 
@@ -399,7 +400,7 @@ class ProviderConfiguration(BaseModel):
                 ProviderModelCredential.tenant_id == self.tenant_id,
                 ProviderModelCredential.provider_name.in_(self._get_provider_names()),
                 ProviderModelCredential.model_name == model,
-                ProviderModelCredential.model_type == model_type,
+                legacy_compatible_model_type_filter(ProviderModelCredential.model_type, model_type),
             ),
         )
 
@@ -748,7 +749,7 @@ class ProviderConfiguration(BaseModel):
             ProviderModel.tenant_id == self.tenant_id,
             ProviderModel.provider_name.in_(provider_names),
             ProviderModel.model_name == model,
-            ProviderModel.model_type == model_type,
+            legacy_compatible_model_type_filter(ProviderModel.model_type, model_type),
         )
 
         return session.execute(stmt).scalar_one_or_none()
@@ -773,7 +774,7 @@ class ProviderConfiguration(BaseModel):
                 ProviderModelCredential.tenant_id == self.tenant_id,
                 ProviderModelCredential.provider_name.in_(self._get_provider_names()),
                 ProviderModelCredential.model_name == model,
-                ProviderModelCredential.model_type == model_type,
+                legacy_compatible_model_type_filter(ProviderModelCredential.model_type, model_type),
             )
 
             credential_record = session.execute(stmt).scalar_one_or_none()
@@ -820,7 +821,7 @@ class ProviderConfiguration(BaseModel):
             ProviderModelCredential.tenant_id == self.tenant_id,
             ProviderModelCredential.provider_name.in_(self._get_provider_names()),
             ProviderModelCredential.model_name == model,
-            ProviderModelCredential.model_type == model_type,
+            legacy_compatible_model_type_filter(ProviderModelCredential.model_type, model_type),
             ProviderModelCredential.credential_name == credential_name,
         )
         if exclude_id:
@@ -895,7 +896,7 @@ class ProviderConfiguration(BaseModel):
                         ProviderModelCredential.tenant_id == self.tenant_id,
                         ProviderModelCredential.provider_name.in_(self._get_provider_names()),
                         ProviderModelCredential.model_name == model,
-                        ProviderModelCredential.model_type == model_type,
+                        legacy_compatible_model_type_filter(ProviderModelCredential.model_type, model_type),
                     )
                     credential_record = session.execute(stmt).scalar_one_or_none()
                     original_credentials = (
@@ -1031,7 +1032,7 @@ class ProviderConfiguration(BaseModel):
                 ProviderModelCredential.tenant_id == self.tenant_id,
                 ProviderModelCredential.provider_name.in_(self._get_provider_names()),
                 ProviderModelCredential.model_name == model,
-                ProviderModelCredential.model_type == model_type,
+                legacy_compatible_model_type_filter(ProviderModelCredential.model_type, model_type),
             )
             credential_record = session.execute(stmt).scalar_one_or_none()
             if not credential_record:
@@ -1075,7 +1076,7 @@ class ProviderConfiguration(BaseModel):
                 ProviderModelCredential.tenant_id == self.tenant_id,
                 ProviderModelCredential.provider_name.in_(self._get_provider_names()),
                 ProviderModelCredential.model_name == model,
-                ProviderModelCredential.model_type == model_type,
+                legacy_compatible_model_type_filter(ProviderModelCredential.model_type, model_type),
             )
             credential_record = session.execute(stmt).scalar_one_or_none()
             if not credential_record:
@@ -1108,7 +1109,7 @@ class ProviderConfiguration(BaseModel):
                     ProviderModelCredential.tenant_id == self.tenant_id,
                     ProviderModelCredential.provider_name.in_(self._get_provider_names()),
                     ProviderModelCredential.model_name == model,
-                    ProviderModelCredential.model_type == model_type,
+                    legacy_compatible_model_type_filter(ProviderModelCredential.model_type, model_type),
                 )
                 available_credentials_count = session.execute(count_stmt).scalar() or 0
                 session.delete(credential_record)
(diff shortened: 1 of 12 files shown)


QuantumGhost:
After a private discussion with @hjlarry, we concluded that a data migration is likely a simpler and more complete solution to this problem.

The main considerations are as follows:

1. For single-row queries (for example, `first()` or `scalar_one_or_none()`), changing the predicate from `model_type = :canonical_model_type` to `model_type IN (:canonical_model_type, :legacy_model_type)` looks straightforward at first glance. In practice, however, the result is not deterministic when canonical and legacy rows coexist for the same logical scope. Unless every such query also defines an explicit precedence, the selected row is effectively undefined.

2. For queries that return multiple rows, the approach above is not sufficient. We would need to fetch both canonical and legacy rows, preserve the persisted `model_type`, and then apply canonical-first deduplication in Python. Some of these read paths already perform additional grouping or assembly in Python, which makes the logic harder to implement correctly and reason about.

3. The original goal was to provide a hotfix. However, once we account for the query categories above, the required compatibility logic becomes fairly broad and touches multiple read paths. That makes the change scope larger than what we would normally want for a hotfix.

4. Even if implemented correctly, this would still be only a temporary workaround. As long as legacy values remain in the database, both existing code and future code will need to carry this additional complexity. Migrating legacy rows to canonical values is therefore a more thorough way to solve the problem and reduce long-term maintenance burden.

Considering the points above, we believe a data migration is the more robust approach.

PR #36429, opened 2026-05-20
chore(api): cap non-dev dependency major versions

### LINKED ISSUE #36428: [Refactor/Chore] Cap non-dev dependency ranges below next major versions
## Description

Several non-development dependency declarations allow automatic upgrades across major versions through open-ended `>=` constraints. This can let dependency resolution pull in major releases with breaking changes for runtime, optional runtime, provider, storage, and tool dependencies.

Update dependency metadata so non-dev dependencies keep their existing lower bounds but add upper bounds below the next major version. Development-only dependencies should remain exempt.

## Motivation

Major dependency upgrades should be deliberate because they can introduce breaking API, packaging, or runtime behavior changes. Adding upper bounds keeps routine dependency resolution safer while still allowing patch and minor upgrades.

## Additional Context

Scope includes backend API dependencies, Dify Agent runtime/server dependencies, provider package dependencies, uv override metadata, and generated lockfile metadata.

laipz8200:
Fixes #36428

## Summary

- Add upper bounds below the next major version for non-dev Python dependency constraints.
- Cover API runtime dependencies, storage/tools/VDB/trace groups, provider package dependencies, uv overrides, and Dify Agent runtime/server dependencies.
- Leave development-only dependency groups unchanged.
- Refresh uv lock metadata so resolved package metadata matches the updated constraints.

## Screenshots

N/A. Dependency metadata only.


api/providers/trace/trace-langsmith/pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "dify-trace-langsmith"
 version = "0.0.1"
 dependencies = [
-    "langsmith>=0.8.0",
+    "langsmith>=0.8.0,<1.0.0",
```
QuantumGhost: I would suggest changing the version requirement to `langsmith>=0.8.0,<0.9.0`, since semver allows breaking changes in minor versions for pre 1.0 version.

api/providers/trace/trace-weave/pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "dify-trace-weave"
 version = "0.0.1"
 dependencies = [
-    "weave>=0.52.36",
+    "weave>=0.52.36,<1.0.0",
```
QuantumGhost: ditto

api/providers/vdb/vdb-lindorm/pyproject.toml
```diff
@@ -4,7 +4,7 @@ version = "0.0.1"
 
 dependencies = [
     "opensearch-py==3.1.0",
-    "tenacity>=8.0.0",
+    "tenacity>=8.0.0,<10.0.0",
```
QuantumGhost: This should be `tenacity>=8.0.0,<9.0.0`.

api/pyproject.toml
```diff
@@ -5,24 +5,24 @@ requires-python = "~=3.12.0"
...
-    "google-api-python-client>=2.196.0",
-    "gunicorn>=26.0.0",
-    "psycogreen>=1.0.2",
-    "psycopg2-binary>=2.9.12",
-    "python-socketio>=5.13.0",
-    "redis[hiredis]>=7.4.0",
-    "sendgrid>=6.12.5",
-    "sseclient-py>=1.8.0",
+    "flask-cors>=6.0.2,<7.0.0",
+    "gevent>=26.4.0,<27.0.0",
```
QuantumGhost: Gevent uses [CalVer](https://www.gevent.org/development/release_process.html#release-cadence-and-versions). I would suggest changing this to `gevent>=26.4.0,<26.5.0`

QuantumGhost (commented):
For packages currently at 0.x versions, should we restrict the allowed version range to patch updates only? Since SemVer 2.0 permits breaking changes in minor releases for 0.x versions—and both FastAPI and opendal have indeed introduced breaking changes in their 0.x releases—this seems like a prudent consideration.

Update: I have examined Semver 2.0 and found that patch release could contain breaking changes, unless package provide extra compatibility guarantee.

api/pyproject.toml
```diff
@@ -103,8 +103,8 @@ dify-trace-weave = { workspace = true }
 default-groups = ["storage", "tools", "vdb-all", "trace-all"]
 package = false
 override-dependencies = [
-    "litellm>=1.83.10",
-    "pyarrow>=18.0.0",
+    "litellm>=1.83.10,<2.0.0",
+    "pyarrow>=18.0.0,<24.0.0",
```
QuantumGhost: I have examined that the current lock file uses `pyarrow==23.0.1`. Since pyarrow follows Semver, should we change this to `pyarrow>=23.0.1,<24.0.0` to prevent accidental downgrade?

dify-agent/pyproject.toml
```diff
@@ -3,23 +3,23 @@ name = "dify-agent"
...
-    "pydantic-ai-slim>=1.85.1",
-    "typing-extensions>=4.12.2",
+    "pydantic-ai-slim>=1.85.1,<2.0.0",
+    "typing-extensions>=4.12.2,<5.0.0",
 ]
 
 [project.optional-dependencies]
 server = [
-    "fastapi>=0.136.0",
+    "fastapi>=0.136.0,<1.0.0",
```
QuantumGhost: FastAPI [explicitly states](https://fastapi.tiangolo.com/deployment/versions/#:~:text=Breaking%20changes%20and%20new%20features%20are%20added%20in%20%22MINOR%22%20versions.) that breaking change will happen in minor releases. This should be changed to `fastapi>=0.136.0,<0.137.0`.

dify-agent/pyproject.toml
```diff
@@ -3,23 +3,23 @@ name = "dify-agent"
...
     "graphon~=0.2.2",
-    "jsonschema>=4.23.0",
-    "pydantic-ai-slim[anthropic,google,openai]>=1.85.1",
-    "pydantic-settings>=2.12.0",
-    "redis>=5",
-    "uvicorn[standard]>=0.38.0",
+    "jsonschema>=4.23.0,<5.0.0",
+    "pydantic-ai-slim[anthropic,google,openai]>=1.85.1,<2.0.0",
+    "pydantic-settings>=2.12.0,<3.0.0",
+    "redis>=5,<8.0.0",
```
QuantumGhost: [The current used version of Redis is v7.4.0](https://github.com/langgenius/dify/blob/6e917c9c018f3a3cc27f1652a96d69ee0468afd5/dify-agent/uv.lock#L2617-L2619). I suggest changing this line to `redis>=7.4.0,<8.0.0` to avoid accidental downgrade.

QuantumGhost (commented):
Other changes LGTM

QuantumGhost (approved):

PR #36434, opened 2026-05-20
fix(dataset): handle missing hit testing document name

sacahan:
## Summary
- avoid crashing hit testing result rendering when the returned segment document name is null
- apply the same fallback in the chunk detail modal
- update the hit testing document type and add regression coverage

## Verification
- git diff --cached --check
- docker build -t dify-web:1.14.1-ttri-null-doc-name -f web/Dockerfile .


sacahan:
Closing this PR for now; the patch will be kept in a fork branch while deployment strategy is finalized.

Issue #35557, opened 2026-04-24
try clean the unnecessary | None in code


asukaminato0721:
They are annoying.

We can check the real case to

1. Give a proper default value 
2. Just remove | None if possible

Ella-horizon:
Hi @asukaminato0721
I am so interested in this issue.
I would like to work on your issue.
Thank you!

asukaminato0721:
@dosu give me a file list and example on where to change.

asukaminato0721:
@dosu compare to | None, I want to give them a default value.

ninad-kawade:
"Hi, I'd like to work on this issue. Can I take it?"

PR #36460, opened 2026-05-20
refactor: clean unnecessary `| None` in services

sisyphusse1-ops:
## Summary

Remove unnecessary `| None` type annotations in service layer files where parameters are only checked with truthiness (`if param:`) and never with `is None`.

Relates to #35557

## Changes

- **api/services/tag_service.py**: `keyword: str | None = None` → `keyword: str = ""`
- **api/services/feedback_service.py**: `from_source`, `rating`, `start_date`, `end_date`: `str | None = None` → `str = ""`
- **api/controllers/console/tag/tags.py**: Updated `TagListQueryParam.keyword` field to `str = ""` to match
- **api/controllers/console/app/message.py**: Coalesce `None` to `""` at call site for `FeedbackService.export_feedbacks`

## Verification

- Each parameter was verified to only be used in truthiness checks (`if param:`), never `is None`
- No callers pass explicit `None` without coalescing
- All files pass `ruff check` and AST parsing
- `has_comment: bool | None = None` in feedback_service.py was intentionally preserved because it uses `is not None` check

## What was not changed

- `api/services/errors/enterprise.py`: The single `int | None = None` for `status_code` was preserved because `None` carries semantic meaning (no HTTP status code available) and there is no natural empty integer default

### DIFF
--- a/api/controllers/console/app/message.py
+++ b/api/controllers/console/app/message.py
@@ -387,11 +387,11 @@ class MessageFeedbackExportApi(Resource):
         try:
             export_data = FeedbackService.export_feedbacks(
                 app_id=app_model.id,
-                from_source=args.from_source,
-                rating=args.rating,
+                from_source=args.from_source or "",
+                rating=args.rating or "",
                 has_comment=args.has_comment,
-                start_date=args.start_date,
-                end_date=args.end_date,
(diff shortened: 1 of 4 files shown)


sisyphusse1-ops:
Closing this PR. Thanks.

PR #36517, opened 2026-05-22
fix: migrate workflow plugin refs on upgrade

### LINKED ISSUE #36215: Plugin upgrade does not update existing workflow plugin_unique_identifier references, causing silent tool failure
### Self Checks

### Dify version

 Dify version: 1.14.1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

  1. Install a plugin from Marketplace (e.g., bowenliang123/md_exporter v2.1.0)
  2. Create a Chatflow/Workflow, add a tool node using that plugin, publish
  3. Upgrade the plugin to a newer version (v2.1.0 → v3.6.9), either via
     Dify's plugin auto-upgrade mechanism or manual upgrade
  4. Run the workflow again

### ✔️ Expected Behavior

 One of the following:
  - Dify auto-migrates existing workflow nodes to reference the upgraded plugin version
  - Dify keeps the old plugin version runtime available for existing workflows
  - At minimum, the tool node throws an explicit error when its referenced plugin version cannot be found

### ❌ Actual Behavior

  - The workflow node's plugin_unique_identifier still points to the old version (e.g., bowenliang123/md_exporter:2.1.0@<old_hash>), while the installed plugin is at the new
  version (bowenliang123/md_exporter:3.6.9@<new_hash>)
  - The workflow executes without any error
  - The tool node produces no output (silent failure)
  - Downstream answer nodes referencing {{#tool_node.files#}} receive empty values — the text prompt "DOCX file has been generated, available in this reply's attachments" is
  shown, but no file is attached
  - User sees incomplete output with no indication of what failed

  Root Cause:

  When a plugin is upgraded, Dify does not scan or update plugin_unique_identifier references in existing workflow DSLs. The old version runtime is removed, and workflow tool
   nodes can no longer resolve their plugin. However, no error is raised — the tool silently produces empty output.

  Note: The plugin_id field (e.g., bowenliang123/md_exporter) is version-agnostic and could serve as a fallback for version resolution, but Dify currently matches only on the
   full plugin_unique_identifier.

  Related existing issues: #20756, #26991, #19032

lord-Rheagar:
## Summary

- migrate stored workflow graph references from the old plugin `plugin_unique_identifier` to the upgraded plugin identifier after marketplace or GitHub plugin upgrades are accepted
- update all matching draft and published workflow rows for the tenant, while only replacing exact `plugin_unique_identifier` fields and leaving incidental text values untouched
- add regression coverage for nested graph rewrites, false-positive graph matches, and both upgrade entry points invoking the workflow migration

Fixes #36215

## Tests

- `python -m py_compile services/plugin/plugin_service.py tests/unit_tests/services/plugin/test_plugin_service.py`
- `pytest -o addopts='' tests/unit_tests/services/plugin/test_plugin_service.py`
- `pytest -o addopts='' tests/unit_tests/controllers/console/workspace/test_plugin.py tests/unit_tests/services/plugin/test_plugin_service.py`
- `git diff --check`

Note: `ruff` is not installed in this local API virtualenv (`python -m ruff` reports `No module named ruff`).


WH-2099:
Thanks for the contribution! Please resolve the conflict in the current PR first.

Additionally, I think the idea of ​​"updating all workflows" here is too radical; it might be acceptable if it were only used for drafting workflows. @laipz8200, could you please help me look into this direction?

lord-Rheagar:
Hi @WH-2099!! The merge conflict has been resolved now. Please check once

WH-2099 (changes requested):
Thank you for working on this issue. The underlying problem is real, and I agree that workflows should not silently fail after a plugin upgrade. However, I do not think this implementation is safe enough to merge under our reliability and enterprise-grade security standards.

The main concern is the migration timing. `upgrade_plugin(...)` starts an install/upgrade task; it does not guarantee that the new plugin version is already installed and usable. This PR rewrites persisted workflow graphs immediately after the task is created. If the task is still pending or eventually fails, existing workflows can be changed to reference a plugin version that is not available, which may turn a recoverable upgrade failure into persisted workflow corruption.

The second concern is migration scope. The current query rewrites every matching workflow row for the tenant, including historical published workflow snapshots. Published workflow versions are part of the execution and audit history, so changing them implicitly is a much broader behavioral change than a safe upgrade repair. We need an explicit product and data-retention decision here, for example draft-only migration or draft plus the currently active published workflow, rather than updating all historical versions.

The third concern is that this is an implicit data rewrite without the safeguards we would expect for this kind of change: no post-success upgrade hook, no durable migration state tied to the upgrade task, no clear rollback/failure behavior, no concurrency guard, and no DB-backed coverage for tenant isolation, false positives, and draft/published scoping. The current `py_compile`-level validation is also not an acceptable quality gate for a change that mutates persisted workflow definitions; syntax validation does not prove the migration behavior is safe.

A safer direction would be to perform the migration only after the upgrade task is confirmed successful, limit the affected workflow set according to an explicit policy, make the operation idempotent and observable, and add database-backed tests for the upgrade lifecycle and workflow-version scope.

For these reasons, I am requesting changes and closing this PR for now. Thanks again for surfacing and attempting to address the issue; the problem is worth solving, but it needs a more conservative design before we can accept it.

WH-2099:
Closing this PR based on the review above. The issue is valid, but this implementation needs a safer design before it can be accepted.

PR #36520, opened 2026-05-22
feat(api): introduce model-type migration script

### LINKED ISSUE #36426: Legacy model_type values break custom model credential management

QuantumGhost:
**AI disclosure**: This PR was primarily drafted with Codex using GPT-5.4. I have reviewed and edited the final diff, and I am responsible for the content.

This PR adopts a relatively verbose style intentionally, so that the script can be reviewed carefully and any issues can be spotted easily.

**Please review this PR with extra care: it modifies production credential records. Any script errors could cause data loss or a major outage.**

## Summary

This migration normalizes legacy provider-related `model_type` values to the canonical enum values:

- `text-generation` -> `llm`
- `embeddings` -> `text-embedding`
- `reranking` -> `rerank`

The main goal is not just to update enum values in place, but to do so safely when a legacy row and a canonical row would collapse onto the same logical key after normalization.

### How conflicts are handled

For each table, the migration groups rows by the **normalized business key** (that is, the key after mapping legacy `model_type` values to their canonical value). If multiple rows end up in the same logical group:

- the migration keeps a single **winner** row
- deletes the **loser** rows
- updates the winner’s `model_type` to the canonical value if needed

The winner is chosen deterministically as the most recently updated row (`updated_at`), with `id` as the tiebreaker.

This is how the migration avoids creating unique-key or logical-key conflicts during canonicalization.

#### `provider_models`

Rows are grouped by the logical key behind `unique_provider_model_name`:

- `(tenant_id, provider_name, model_name, model_type)`

If legacy and canonical rows collide after normalization, they are merged into one winner row.

#### `tenant_default_models`
Rows are grouped by the logical key behind `unique_tenant_default_model_type`:

- `(tenant_id, model_type)`

Again, legacy/canonical duplicates are merged before final canonicalization.

#### `provider_model_settings`
This table does not enforce the key at the DB level, but the migration treats the following as the logical uniqueness boundary:

- `(tenant_id, provider_name, model_name, model_type)`

So duplicates are merged the same way as the indexed tables.

#### `provider_model_credentials`

This table also uses a logical uniqueness rule rather than relying on a DB unique index:

- `(tenant_id, provider_name, model_name, credential_name, model_type)`

When duplicates are merged, the migration also rewrites references from loser credentials to the winner credential in:

- `provider_models`
- `load_balancing_model_configs`

Only after references are rewritten are loser credential rows removed.

#### `load_balancing_model_configs`

This table is handled differently.

For normal rows, canonicalization is mostly row-by-row.

For `name="__inherit__"` rows, the migration first performs a **pre-deduplication step** by normalized business key:

- `(tenant_id, provider_name, model_name, model_type)`

This is necessary because a legacy `__inherit__` row and a canonical `__inherit__` row could otherwise collide after the `model_type` update. The migration resolves that duplicate first, then canonicalizes remaining rows.

### Operational behavior

- The command is **dry-run by default**
- `--apply` is required to persist changes
- It supports scoping by table, tenant, and canonical model type
- It runs in small transactions and skips lock-conflicted groups/rows instead of blocking the whole migration
- Use the `--output` flag to capture a complete log of migration changes and execution context for debugging or recovery.
- This migration is idempotent. It is safe to run multiple times, and it is also safe to rerun after a partial or interrupted execution.

**Because this migration updates and deletes a large volume of data, we strongly recommend that you:**

1. test and validate the migration in a staging environment before applying it to production;
2. preserve the output logs for a period of time after execution so that any migration-related issues can be diagnosed and resolved.

Continuation of #36427.

Closes #36426 #36129 #36386

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


wylswz (dismissed):

GareArc (dismissed):

wylswz (dismissed):

hjlarry (dismissed):
LGTM

hjlarry (approved):

GareArc (approved):

wylswz (approved):

PR #36651, opened 2026-05-26
feat(openapi,cli): workspace switch + member management

lin-snow:
## Summary

Adds **workspace switching** and **member management** to the bearer-authenticated `/openapi/v1` surface, plus the matching `difyctl` commands. These endpoints are role-gated against the caller's membership in the target workspace, so they are safe to expose to programmatic (CLI / API) clients.

### New `/openapi/v1` endpoints

| Method | Route | Access | Purpose |
|--------|-------|--------|---------|
| `POST` | `/workspaces/{id}/switch` | any member | Switch the bearer's active workspace |
| `GET` | `/workspaces/{id}/members` | any member | List members (paginated: `page` / `limit`) |
| `POST` | `/workspaces/{id}/members` | owner / admin | Invite a member |
| `PUT` | `/workspaces/{id}/members/{member_id}/role` | owner / admin | Change a member's role |
| `DELETE` | `/workspaces/{id}/members/{member_id}` | owner / admin | Remove a member |

A new `@require_workspace_role(*roles)` decorator centralizes the gate. It returns **404** for non-members (parity with `GET /workspaces/{id}` — no cross-tenant ID leak) and **403** for members whose role is not allowed. Member invites additionally pass through **edition-aware quota gates**.

### Request flow

```mermaid
sequenceDiagram
    participant CLI as difyctl
    participant API as /openapi/v1
    participant Gate as require_workspace_role
    CLI->>API: Bearer dfoa_… + workspace_id
    API->>API: validate_bearer → set_auth_ctx (ContextVar)
    API->>Gate: accept_subjects(ACCOUNT)
    Gate->>Gate: try_get_auth_ctx() → account_id
    Gate->>Gate: lookup TenantAccountJoin(workspace_id, account_id)
    alt not a member
        Gate-->>CLI: 404
    else role not allowed
        Gate-->>CLI: 403
    else
        Gate->>API: invoke handler → 200
    end
```

### New `difyctl` commands

| Command | Notes |
|---------|-------|
| `difyctl use workspace` | Set the active workspace for the current host |
| `difyctl get member` | List members; `--page` / `--limit` pagination |
| `difyctl create member` | Invite a member |
| `difyctl set member` | Update a member's role |
| `difyctl delete member` | Remove a member |

`create` / `set` / `delete member` accept `-o json|yaml|name|text`; `auth devices list` also gained `--page` / `--limit`.

### Included fix

`require_workspace_role` and the member handlers now read the caller identity from the openapi auth **ContextVar** (`try_get_auth_ctx()` / `get_auth_ctx()`), the slot `validate_bearer` actually publishes — not `flask.g.auth_ctx`, which production never writes. The earlier `flask.g` read made every role-gated request raise `RuntimeError` → 500 while unit tests stayed green (they seeded `g` directly). A regression test now seeds **only** `g`, leaves the ContextVar empty, and asserts the gate still rejects — pinning the identity source.

## Test plan

- Backend: `api/tests/unit_tests/controllers/openapi/auth/test_role_gate.py` (gate: 404 / 403 / pass, query scoping, wiring-bug `RuntimeError`s) and `test_workspaces_members.py` (endpoint behaviors).
- CLI: unit tests per command against a stub server (`*.test.ts`).
- Manual: exercised all endpoints against a live EE server via `curl` and the real `difyctl` binary — switch / list / invite / role-change / remove all return expected results.

## Checklist


api/controllers/openapi/auth/role_gate.py
```diff
@@ -0,0 +1,83 @@
...
+                raise RuntimeError(
+                    "require_workspace_role called without account-bearer context; "
+                    "stack validate_bearer + accept_subjects(SubjectType.ACCOUNT) above it"
+                )
+
+            workspace_id = kwargs.get("workspace_id")
+            if not workspace_id:
+                raise RuntimeError("require_workspace_role expects a 'workspace_id' route parameter")
+
+            join = db.session.execute(
```
wylswz: Consider moving queries to service layer.

lin-snow: @wylswz Good call — moved the membership/role lookup into TenantService.get_account_role_in_tenant (session-injected, mirrors account_belongs_to_tenant). The gate is now pure policy (404 for non-member, 403 for out-of-set role) and the SQL-scoping assertion moved down to the service tests.

api/controllers/openapi/workspaces.py
```diff
@@ -57,6 +176,194 @@ def get(self, workspace_id: str):
...
+        account = _load_account(ctx.account_id)
+
+        try:
+            TenantService.switch_tenant(account, workspace_id)
+        except AccountNotLinkTenantError:
+            # Membership existed at gate time but Tenant.status != NORMAL or
+            # the row was just removed — treat as not-found.
+            raise NotFound("workspace not found")
+
+        row = db.session.execute(
```
wylswz: All db.session.xxx usages in controller layer need to be refactored.

wylswz (approved):

lin-snow:
😡😡😡 what the fking merge queue doing....

PR #36693, opened 2026-05-26
feat(openapi): redesign auth pipeline with per-token-type routing

GareArc:
## Summary

- Introduces a `TokenType` discriminator (`oauth_account`, `oauth_external_sso`) on bearer tokens, enabling correct routing for future token kinds (e.g. PATs) that share the same subject type
- Each token type now has its own auth pipeline with distinct prepare and verification steps, replacing a single shared pipeline that required branching on token identity inside each step
- Edition gating (CE/EE/SaaS) is enforced at the router level rather than scattered across individual steps, so pipelines themselves are edition-agnostic
- Adds `WORKSPACE_READ` scope used by the `/workspaces` endpoint

## Test plan

- [ ] `pytest api/tests/unit_tests/controllers/openapi/` — all pipeline, condition, and endpoint tests
- [ ] `pytest api/tests/unit_tests/libs/test_oauth_bearer_rate_limit_ordering.py api/tests/unit_tests/libs/test_workspace_member_helper.py`
- [ ] Manually verify account token flow: `GET /openapi/v1/account` returns account info
- [ ] Manually verify external SSO token rejected on account-only endpoints with `403 unsupported_token_type`
- [ ] Verify external SSO token rejected on EE-only endpoints when running CE with `403 external_sso_requires_ee`


api/controllers/openapi/auth/prepare.py
```diff
@@ -0,0 +1,78 @@
...
+from core.app.entities.app_invoke_entities import InvokeFrom
+from extensions.ext_database import db
+from models.account import TenantStatus
+from services.account_service import AccountService, TenantService
+from services.app_service import AppService
+from services.end_user_service import EndUserService
+from services.enterprise.enterprise_service import EnterpriseService, WebAppAccessMode
+
+
+def build_external_identity(builder: dict) -> None:
```
wylswz: Could this builder be a `TypedDict`?

wylswz:
There are some conflicts.

wylswz (approved):

PR #36711, opened 2026-05-27
refactor(cli/http): replace ky with a self-contained HTTP client

lin-snow:
## Summary

Replace the CLI's `ky`-based HTTP layer with a small, self-contained client and migrate every call site to it. Internal transport refactor — no behavioral change to CLI commands.

Dependency-wise this swaps the `ky` HTTP **wrapper** for `undici` (the engine that already powers Node's `fetch`), pulled in for one reason: honoring `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` via `EnvHttpProxyAgent` (Node's global `fetch` silently ignores them). It is externalized at build, not bundled.

```mermaid
flowchart LR
  subgraph Before
    A["~30 call sites"] -->|"KyInstance leak"| K["ky + middleware/"]
  end
  subgraph After
    B["~30 call sites"] --> H["createHttpClient → HttpClient"]
    H --> D["recursive dispatch + fetch + hooks"]
    D --> P["proxy dispatcher (EnvHttpProxyAgent)"]
  end
```

**`HttpClient` surface**

| Method | Returns | Notes |
|---|---|---|
| `get` / `post` / `put` / `patch` / `delete<T>` | `Promise<T>` | typed JSON, throws on non-2xx; `204`/empty body → `undefined` |
| `fetch` | `Promise<Response>` | raw Response (OAuth) |
| `stream` | `Promise<Response>` | SSE — no retry / no timeout |
| `extend` | `HttpClient` | re-run factory with merged opts |

`cli/src/http/` ships focused, unit-tested modules: `url` (query building), `body` (encoding), `retry` (policy), `hooks` (lifecycle), `proxy` (env-driven proxy dispatcher), `types` (contract). Auth / UA / logging / error-classification are composable **hooks** compiled in at construction; `AbortSignal.any([user, timeout])` replaces the old timing cast and the `{ throwHttpErrors: false, skipClassify }` escape hatch. The `middleware/` layer and the `ky` dependency are removed; `testHttpClient` dedupes client setup across the suite.

**Proxy support**

`proxyDispatcher()` resolves a single `EnvHttpProxyAgent` once per process when a proxy env var is present and passes it as the per-request `dispatcher`; with no proxy var set, `fetch` keeps its default dispatcher untouched. Routing was verified end-to-end (dead-proxy → connection refused at the proxy) on **both** Node (via the dispatcher) and the `bun build --compile` binary (native proxy support + dispatcher, no regression).

## Verification

| Check | Result |
|---|---|
| `pnpm test` | 87 files / **893 tests** pass |
| `pnpm type-check` + `pnpm lint` | clean |
| `ky` / `KyInstance` refs in `cli/src` | none |
| proxy routing (Node + bun binary) | honored |

## Screenshots

N/A — internal refactor, no user-facing UI change.

## Checklist

- [x] I ran `cd cli && pnpm type-check && pnpm lint && pnpm test` to appease the lint gods


cli/src/http/client.ts
```diff
@@ -1,63 +1,232 @@
...
+    request,
+    options: resolved,
+    attempt,
+    meta: new Map(),
+  }
+
+  await runHooks(state.hooks.onRequest, ctx)
+
+  try {
+    ctx.response = await fetch(ctx.request, { signal })
```
GareArc: Proxy env vars HTTP_PROXY/http_proxy silently ignored. Consider setting up an EnvHttpProxyAgent in dispatcher?

GareArc (dismissed):
lgtm

cli/src/http/client.ts
```diff
@@ -1,63 +1,241 @@
...
+
+  return {
+    get: <T>(p: string, o?: RequestOptions) => typedCall<T>('GET', p, o),
+    post: <T>(p: string, o?: RequestOptions) => typedCall<T>('POST', p, o),
+    put: <T>(p: string, o?: RequestOptions) => typedCall<T>('PUT', p, o),
+    patch: <T>(p: string, o?: RequestOptions) => typedCall<T>('PATCH', p, o),
+    delete: <T>(p: string, o?: RequestOptions) => typedCall<T>('DELETE', p, o),
+    fetch: rawFetch,
+    stream: streamFetch,
+    extend,
```
GareArc, line 242: One forward-looking ask: we're planning to adopt the generated oRPC contract in the CLI. oRPC's OpenAPILink builds a standard Request and calls a fetch-shaped fn `(req: Request, init?) => Promise<Response>`. To route that through this client and reuse all transport policy, we'd need a low-level entrypoint that executes a pre-built Request:

e.g. `readonly request: (req: Request, init?: RequestInit) => Promise<Response>`

eslint.config.mjs
```diff
@@ -22,6 +22,10 @@ export default antfu(
       '!packages/**',
       '!web/**',
       '!e2e/**',
+      '!cli/**',
+      'cli/context/**',
+      'cli/docs/**',
+      'cli/oclif.manifest.json',
```
wylswz: Plz clean up unused file patterns.

cli/src/http/client.ts
lin-snow: @GareArc  Agreed, let's land this as a follow-up when we adopt oRPC — no change needed here to unblock it.

wylswz (approved):

PR #36724, opened 2026-05-27
feat(api): agent backend session lifecycle for workflow agent nodes

zyssyz123:
## Summary

Implements the [`dify-agent` run lifecycle contract](https://langgenius.feishu.cn/wiki/UcaIwlWhwinbaMkocYvcrpbande) for workflow Agent nodes so a single workflow run can fan out into multiple Agent backend runs that share an Agent session via persisted snapshots, and so a HITL pause routes through a workflow-level pause instead of being modeled as an Agent-run pause.

This is the API-side half of Stage 5 groundwork — Stage 0-4 features (already on `main` via #36644 + #36686) are untouched.

## What changed

**Persistence**
- New table `workflow_agent_runtime_sessions` (migration `a7b9c8d1e2f3`) keyed by `(tenant_id, workflow_run_id, node_id, binding_id, agent_id)`. Stores the `CompositorSessionSnapshot` plus the replayable `composition_layer_specs`.
- `composition_layer_specs` is the key fix for a trap we hit in E2E: agenton's compositor rejects any session snapshot whose layer names don't match the active composition. Persisting the specs lets a future cleanup-only run replay the exact non-plugin layer graph.

**Request builder** — `api/clients/agent_backend/request_builder.py`
- Default `on_exit.default = suspend` (was `delete`) so layers survive between runs in the same Agent session.
- Default `include_history = True` adds a `pydantic_ai.history` layer that carries chat history across runs.
- `extract_cleanup_layer_specs(composition)` projects the in-flight composition into the persistable cleanup shape, dropping credential-bearing plugin layers (`dify.plugin.llm`, `dify.plugin.tools`).
- `build_cleanup_request(*, session_snapshot, composition_layer_specs, ...)` replays the persisted specs and filters the snapshot to match so the agenton snapshot-vs-composition name-order validator succeeds.

**Node runtime** — `api/core/workflow/nodes/agent_v2/agent_node.py`
- On entry: `load_active_snapshot(scope)` and thread it into the runtime request.
- On `run_succeeded`: `save_active_snapshot` with `composition_layer_specs` extracted from the just-built request.
- On `run_paused` (HITL): persist snapshot and yield workflow-level `PauseRequestedEvent` — the agent run is **terminal**, the outer workflow run is what pauses, per the lifecycle skill.
- On `run_failed` / `run_cancelled`: `mark_cleaned` immediately so a workflow loop cannot resume from stale state.
- `_save_session_snapshot` now logs persistence failures with `exc_info` and adds session-store cleanup metadata.

**Workflow-terminal cleanup** — `api/core/workflow/nodes/agent_v2/session_cleanup_layer.py`
- New `WorkflowAgentSessionCleanupLayer` (`GraphEngineLayer`) attached on both `WorkflowAppRunner` and `AdvancedChatAppRunner`. Triggers on `GraphRunSucceeded / GraphRunPartialSucceeded / GraphRunFailed / GraphRunAborted` (deliberately *not* `GraphRunPaused`).
- The HTTP cleanup path is fully built — composition replay, snapshot filtering, `wait_run` gated mark_cleaned — but is **kept behind `_HTTP_CLEANUP_SUPPORTED = False`** because the current `dify-agent::runner._run_agent` always requires an LLM layer and has no `purpose=cleanup` branch yet. Until that lands the layer just retires the local row so the snapshot can't be reused, and lets the agent backend's retention TTL release suspended layers. Flipping the flag enables the full contract.

**Tests** (`130 passing`)
- `test_session_cleanup_layer.py` — 11 cases including all 4 terminal events, non-terminal events skipping cleanup, HTTP-disabled production default, `wait_run` failure paths, fan-out across multiple ACTIVE sessions.
- `test_agent_node.py` — new cases for failed-run `mark_cleaned`, paused-run snapshot persistence, success-run spec persistence (no plugin layers leak in).
- `test_request_builder.py` — `build_cleanup_request` shape, empty-specs rejection, `extract_cleanup_layer_specs` dropping plugin types.
- **`test_cleanup_composition_compositor_integration.py`** — new file. Runs the real `agenton.Compositor.from_config + _create_run` against the built cleanup request so the snapshot-vs-composition mismatch trap (invisible to `FakeAgentBackendRunClient`) is caught at unit-test time. Includes a negative case that confirms the validator does raise when the shapes diverge.

## End-to-end verified

Started agent backend (5050), API (5001), celery (all queues), drove `POST /apps/{id}/workflows/draft/run` via real cookies:

| Time | DB observation |
|---|---|
| t=2s | workflow `running`, agent-1 in-flight |
| t=4s | workflow `succeeded`, session row `cleaned` with `snap=2063 bytes`, `specs=8014 bytes`, `cleaned_at` set |

`agent_backend_request` composition has the 7 expected layers (`agent_soul_prompt`, `workflow_node_job_prompt`, `workflow_user_prompt`, `execution_context`, `history`, `llm`, `output`). No `run_failed` agent backend records for cleanup (the trap is closed).

## Test plan

- [ ] `pnpm test:e2e` not applicable (backend-only change)
- [x] `uv run --project api pytest api/tests/unit_tests/core/workflow/nodes/agent_v2/ api/tests/unit_tests/clients/agent_backend/` → 130 passing
- [x] `uv run --project api ruff check api/` → clean
- [x] `uv run --project api ruff format --check ...` → clean
- [x] Real-user E2E against 4 services (agent backend + API + celery + plugin daemon) → workflow + cleanup succeeded in 4s

## Follow-ups

- Once `dify-agent` adds a `purpose=cleanup` branch in `runner._run_agent`, flip `_HTTP_CLEANUP_SUPPORTED = True` and the cleanup-only HTTP run will start releasing suspended layers eagerly instead of relying on retention TTL.
- Memory isolation modes (`global / by_app / by_run / end_user`) per PRD are not addressed here — the current keying is `by_run` (workflow-scoped). Future Stage 5 work.


MRZHUH (approved):
LGTM

PR #36687, opened 2026-05-26
refactor(cli): add kvstore and platform interface

wylswz:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary
- Provided a `Store` interface for local k-v storage.
- Added a `Platform` interface for os-related behaviors.

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [ ] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [ ] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods

PR #36726, opened 2026-05-27
refactor(cli): use Store interface as token storage

wylswz:
> [!IMPORTANT]
>

## Summary
- refactor TokenStore to use Store interface

see #36687 

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


cli/src/commands/auth/login/login.ts
lin-snow, line 140: The bearer is unconditionally written into `HostsBundle.tokens.bearer` and persisted to `hosts.yml` by `saveHosts(bundle)`, while every read path (`_shared/authed-command.ts:55`, `whoami`, `status`, `logout`) reads `bundle.tokens.bearer` and never goes through the `Store` we just wrote to on line 64 — so the new `keychain` mode does not actually keep the bearer off disk, and `file` mode double-writes it to both `tokens.yml` and `hosts.yml`.

cli/src/commands/auth/logout/logout.ts
```diff
@@ -53,18 +52,11 @@ function revokeAllowed(bearer: string): boolean {
   return REVOCABLE_PREFIXES.some(p => bearer.startsWith(p))
 }
 
-async function clearLocal(configDir: string, bundle: HostsBundle, store: TokenStore): Promise<void> {
```
lin-snow, line 56: `clearLocal` is byte-identical in both files (and the `accountId` fallback chain also lives inline as `accountKey()` in `login.ts:165-171`) — extract one `clearLocalSession(bundle, store)` helper so the three sites can't drift.

cli/src/store/store.ts
lin-snow: No `stale` option is passed to `lockfile.lockSync`, so any `SIGKILL`'d or crashed `difyctl` invocation leaves the `.lock` file forever and every subsequent command throws `ConcurrentAccessError` until the user reads the hint and deletes it manually — pass `{ stale: 30_000 }` so `lockfile` reclaims it automatically.

cli/src/store/store.ts
```diff
@@ -96,18 +94,44 @@ abstract class FileBasedStore implements Store {
...
 
   set<T>(key: Key<T>, value: T) {
     this.withLock(() => {
+      this.load()
       this.doSet(key, value)
       this.flush()
     })
   }
 
+  unset<T>(key: Key<T>): void {
```
lin-snow, line 133: When `doUnset` early-returns because the key isn't present, `raw_content` is unchanged but `flush()` still rewrites the file — make `doUnset` return a `boolean` and skip `flush()` when nothing changed.

cli/src/store/manager.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      map.set(key.key, value)
+    },
+    unset<T>(key: Key<T>): void {
+      map.delete(key.key)
+    },
+  }
+}
+
+describe('getTokenStore', () => {
+  it('returns keychain store when probe succeeds', async () => {
```
lin-snow: `getTokenStore` became sync in commit `a2cea7f86d`, so the `await` in every `it` block is misleading dead syntax — drop it and make the test bodies sync.

cli/src/store/store.ts
lin-snow: `new YamlStore(path)` synchronously `mkdirSync`s the parent dir, so `getConfigurationStore()` / `getHostStore()` / `getCache()` now do filesystem I/O just to return a handle — move the `mkdirSync` into `flush()` so reads against a missing file don't materialize a directory tree.

wylswz: makes sense.

cli/src/store/store.ts
wylswz: getter / setter and dirty flag were added.

cli/src/commands/auth/login/login.ts
wylswz: This could be done in another pull request.

GareArc (approved):
lgtm

PR #36743, opened 2026-05-27
fix(web): pass currentProvider/currentTool to second ToolForm so form: 'form' dynamic-select fires

### LINKED ISSUE #36518: Tool plugin dynamic-select dropdown does not fire _fetch_parameter_options on Cloud 1.14.1 (regression from 1.6.0)
### Self Checks

- [x] This is only for bug report.
- [x] I have searched for existing issues including closed ones. Closest match is #22147 — the title and the merged fix PR #23323 only address the "data lost on tab switch" half of that thread; a separate symptom reported by @gao-xiangyang on 2025-07-12 inside the same thread ("After Dify was upgraded to version 1.6.0, the dynamic-select was not triggered") was never addressed. This issue is a focused report of that second symptom, reproduced on Cloud 1.14.1 (2026-05-22).
- [x] [FOR CHINESE USERS] N/A

### Dify version

1.14.1 (Cloud — confirmed via `GET https://cloud.dify.ai/console/api/version?current_version=1.0.0`).
Self-hosted not yet tested for this specific symptom — happy to add if requested.

### Cloud or Self Hosted

Cloud

### Steps to reproduce

A Tool plugin parameter declared as `type: dynamic-select` should populate its dropdown by Dify calling the plugin's `_fetch_parameter_options(<parameter_name>)` hook when the dropdown is opened. On Cloud, that call **never happens**: opening the dropdown produces no network request from the browser, no daemon log entry, and no traffic to the plugin runtime. The dropdown stays empty forever.

Minimum reproduction (YAML + Python, both excerpted from a Tool plugin we shipped — full history at `ShuntaroOkuma/convoprobe-plugin` commit 42528f5^):

`tools/<tool>.yaml`:
```yaml
identity:
  author: <author>
  name: run_scenario
  label:
    en_US: Run Scenario

extra:
  python:
    source: tools/run_scenario.py

parameters:
  - name: scenario_id
    type: dynamic-select          # <-- the dropdown that never fires
    form: form
    required: true
    label:
      en_US: Scenario
    human_description:
      en_US: Select a scenario you have created in ConvoProbe Web UI.
    llm_description: ConvoProbe scenario identifier (UUID).
```

`tools/<tool>.py`:
```python
from dify_plugin import Tool
from dify_plugin.entities import I18nObject, ParameterOption

class RunScenarioTool(Tool):
    def _fetch_parameter_options(self, parameter: str) -> list[ParameterOption]:
        # This method is never called by Dify Cloud 1.14.1 when the user
        # opens the `scenario_id` dropdown in the workflow node editor.
        # Returning a fixed list here for the simplest repro:
        return [
            ParameterOption(value="A", label=I18nObject(en_US="Scenario A")),
            ParameterOption(value="B", label=I18nObject(en_US="Scenario B")),
        ]
```

Steps:

1. Build the above as a `.difypkg` (or use Remote Debug).
2. Install on Dify Cloud, add it to a workflow node.
3. Open the node and click the `Scenario` dropdown.
4. Observe: the dropdown shows no options, no spinner. Browser DevTools shows zero network requests. Plugin process (in Remote Debug mode) shows no incoming call.

For our plugin specifically the live confirmation came from:
- Backend logs at `https://convoprobe-production.up.railway.app` during the dropdown click: zero hits to the corresponding `/api/internal/plugin/scenarios` endpoint that `_fetch_parameter_options` proxies to.
- Plugin process running `INSTALL_METHOD=remote` against `debug.dify.ai:5003`: zero received dispatches when the dropdown is opened. The TCP connection itself is healthy and the plugin is fine with other operations.

### ✔️ Expected Behavior

Same behavior as Dify 1.5.1 (per #22147 screenshots from 2025-07-12): opening a `dynamic-select` dropdown triggers Dify api → plugin daemon → plugin runtime → `_fetch_parameter_options(<param>)`, the returned `ParameterOption[]` populates the dropdown.

### ❌ Actual Behavior

Opening the dropdown is a no-op. The hook is never invoked. The dropdown stays empty. The user has no way to know the failure mode without checking three layers (browser network panel, plugin daemon logs, plugin process stderr).

### Why this matters

`dynamic-select` is the standard mechanism for any Tool plugin whose parameter values come from the plugin's own external API (saved scenarios, projects, channels, datasets, lists of records, etc.). With it broken on Cloud, **all such plugins must currently fall back to a free-text input** (the workaround we shipped in commit 42528f5 — `type: string` with a human "paste a UUID" instruction), which:

- destroys UX (users must copy a UUID from a web page and paste it),
- destroys discoverability (no list of available values),
- destroys validation (typos surface only at run time, deep in the plugin),
- and forces the plugin author to ship docs explaining how to find the value out-of-band.

### Additional context

- The related thread #22147 already reported this same symptom in a comment dated **2025-07-12** but the fix PR (#23323) only changed `web/app/components/workflow/nodes/_base/components/variable/var-reference-picker.tsx` and addressed the "selected value is lost when switching tabs" half. The "dropdown never fires" half was never touched.
- **I would like to be assigned to this issue and attempt the fix PR myself** (per CONTRIBUTING.md: "Before you draft a PR, please create an issue to discuss the changes you want to make"). Best guess so far: the workflow node editor's React component that mounts dynamic-select dropdowns may have lost the `onOpen` → `fetchOptions` wiring in the 1.6.0 refactor — would appreciate a pointer to the right file from anyone who already knows the area.
- Public reproduction code (with full git history of the dynamic-select implementation before we had to remove it): https://github.com/ShuntaroOkuma/convoprobe-plugin — see commit `42528f5` and its parent for the working `_fetch_parameter_options` implementation and the YAML.

ShuntaroOkuma:
## Summary

`panel.tsx` for the Tool node renders `<ToolForm>` twice — once for `toolInputVarSchema` (`form: 'llm'`) with `currentProvider`/`currentTool`, and once for `toolSettingSchema` (`form: 'form'`) **without** them. `form: 'form'` is the typical case for parameters the workflow author sets in the node UI, so almost every realistic `dynamic-select` parameter falls into the second `<ToolForm>` and the downstream guard in `form-input-item.tsx`'s `useEffect` (`isDynamicSelect && currentTool && currentProvider && ...`) silently bails out. The result: `_fetch_parameter_options` is never called and the dropdown stays empty.

This PR passes `currentProvider`/`currentTool` to the second `<ToolForm>` too. No other change.

## Verification

Local docker-compose Dify 1.14.2 with a Tool plugin exposing a `dynamic-select` `form: 'form'` parameter (`soitto/convoprobe`).

- **Before**: dropdown empty, no `dynamic-options` request fires, plugin `_fetch_parameter_options` never invoked.
- **After**: `dynamic-options` request fires correctly with `provider_type=tool` and the expected params, plugin returns options, dropdown populates.

Fixes #36518


ShuntaroOkuma:
Friendly nudge — this is a 2-line / 1-file diff (CI green) fixing #36518. Could a maintainer take a look when convenient? Happy to address any feedback.

ShuntaroOkuma:
@iamjoel @zxhlyh — gentle ping. Tiny diff (+2/-0), CI green, fixes #36518. Could you take a look when you have a moment?

PR #36830, opened 2026-05-29
refactor(cli): unify token storage behind Store + add host/account switching

GareArc:
## Overview

Unifies how `difyctl` reads and writes credentials, adds first-class multi-host/account switching, replaces the split output formatter system with a single protocol, and adds `auth list`.

Resolves [WTA-356](https://linear.app/dify/issue/WTA-356).

### Token storage
- All credential reads/writes go through the `Store` interface; keychain and file backends share one code path.
- Host registry stripped of token fields and token-read logic; `tokens.yml` is authoritative.

### Auth commands
- `difyctl auth list` — table of all authenticated host+account pairs, supports `-o json/yaml/name`. Replaces `auth status` (redundant with `auth whoami` for single-context view).
- `difyctl use host [--domain <domain>]` — switch the active Dify instance.
- `difyctl use account [--email <email>]` — switch active account on the current host.
- Interactive picker when flag is omitted in a TTY; fails with a hint in non-TTY (CI/agent) contexts.

### Login UX
- Blue spinner during device-flow authorization poll.
- Omitting `--host` in a non-TTY context now errors with a usage hint instead of hanging.

### Output formatter unification
- Single source of truth for all CLI output: `framework/output.ts` (`TablePrintable`, `FormattedPrintable`) and `framework/stream.ts` (`StreamPrinter`).
- `src/printers/` deleted entirely (15 files). `AppRunObject` now implements `FormattedPrintable` directly; mode-dispatch (chat/completion/workflow) internalized.
- `StreamPrinter` type moved to `framework/stream.ts`; concrete implementations stay command-scoped.
- All `get/*` handlers migrated from `printers/format-table` type imports to `framework/output`.

## Testing

- `pnpm type-check`, `pnpm lint`, `pnpm tree:check` — clean.
- `pnpm test` — 786 passing, including coverage for the interactive picker, `use host`/`use account` flows, and `auth list` output shapes.


cli/src/auth/hosts.ts
```diff
@@ -25,42 +27,152 @@ export const ExternalSubjectSchema = z.object({
...
+export type RegistryData = z.infer<typeof RegistrySchema>
 
-export function loadHosts(): HostsBundle | undefined {
-  const raw = getHostStore().getTyped<Record<string, unknown>>()
-  if (raw === null)
-    return undefined
-  return HostsBundleSchema.parse(raw)
+export type ActiveContext = {
+  readonly host: string
+  readonly email: string
```
wylswz, line 56: Should `email` be a property of `ctx` and `scheme` be a property of `host`?

cli/src/auth/hosts.ts
```diff
@@ -25,42 +27,152 @@ export const ExternalSubjectSchema = z.object({
...
+  resolveActive(): ActiveContext | undefined {
+    const host = this.data.current_host
+    if (host === undefined || host === '')
+      return undefined
+    const entry = this.data.hosts[host]
+    if (entry === undefined)
+      return undefined
+    const email = entry.current_account
+    if (email === undefined || email === '')
+      return undefined
```
wylswz, line 101: This could be simplified to
```
const email = entry?.current_account
if (!email)
    return undefined
```

cli/src/auth/hosts.ts
GareArc: ActiveContext is extracted from context and host as a flattened view used for convenience at call sites by design.

wylswz (approved):

PR #36839, opened 2026-05-30
feat(api): add MCP user-identity forwarding

CourTeous33:
## Summary
- Adds end-user SSO identity forwarding to outbound MCP requests. When an MCP provider has `forward_user_identity` enabled, the API calls dify-enterprise's `/inner/api/mcp/issue-token` to mint a fresh per-user OIDC `id_token` (or OAuth2 `access_token`) for the calling workflow user and stamps it as `Authorization: Bearer <token>` on every outbound MCP request — so the MCP server can act on behalf of the verified end user rather than seeing only "Dify is calling."
- Two new columns on `tool_mcp_providers`:
  - `forward_user_identity` (bool, default `false`) — master gate exposed to the UI as a toggle.
  - `identity_mode` (varchar(32), default `"off"`) — `"off"` or `"idp_token"`. Forward-compatible so a future RFC 8693 token-exchange mode lands without UI/DB churn.
- Errors map cleanly to workflow halts: `428` → `MCPNoRefreshTokenError` (user has no stored SSO refresh token), `401` → `MCPIdentityRefreshError` (IdP rejected the refresh). No silent downgrade to anonymous.

## Test plan
- [ ] `flask db upgrade` applies the new migration on a fresh DB.
- [ ] Existing MCP providers continue to work unchanged (`forward_user_identity` defaults to `false`).
- [ ] With a workspace using SSO + an MCP provider with `forward_user_identity=true`, invoking a tool from a workflow forwards an audience-bound JWT (verifiable via a JWKS-checking MCP server).
- [ ] With `forward_user_identity=true` but the calling user has no stored refresh token, the workflow fails with `MCPNoRefreshTokenError` instead of silently going anonymous.
- [ ] GET `/console/api/workspaces/current/tools/mcp` returns the two new fields in each provider entry.

This is the server-side half. The UI toggle ships in a separate PR (feat/mcp-token-transfer-web).


api/models/tools.py
```diff
@@ -350,6 +350,21 @@ class MCPToolProvider(TypeBase):
     # encrypted headers for MCP server requests
     encrypted_headers: Mapped[str | None] = mapped_column(LongText, nullable=True, default=None)
 
+    # M2 (MCP user-identity forwarding) — master switch per provider. When True
+    # AND identity_mode is "idp_token", workflows that invoke tools on this
+    # provider will have the caller's SSO id_token stamped on the outbound
+    # request as `Authorization: Bearer …`. Off by default so existing
+    # providers retain pre-M2 behaviour.
+    forward_user_identity: Mapped[bool] = mapped_column(
```
wylswz: is `forward_user_identity: false` equivalent to `identity_mode: off`?

api/core/tools/mcp_tool/tool.py
```diff
@@ -280,9 +317,38 @@ def invoke_remote_mcp_tool(self, tool_parameters: dict[str, Any]) -> CallToolRes
...
+
+        Errors are surfaced as ToolInvokeError so the workflow halts with a
+        clear message instead of silently dropping identity and hitting the
+        MCP server unauthenticated.
+        """
+        from services.enterprise.base import MCPTokenError
+        from services.enterprise.enterprise_service import EnterpriseService
+
+        try:
+            token, _expires_at = EnterpriseService.issue_mcp_token(
```
wylswz, line 353: all enterprise related must be gated using `dify_config.ENTERPRISE_ENABLED`, including UI and backend invocations.

api/services/tools/mcp_tools_manage_service.py
```diff
@@ -194,6 +198,8 @@ def update_provider(
         configuration: MCPConfiguration,
         authentication: MCPAuthentication | None = None,
         validation_result: ServerUrlValidationResult | None = None,
+        forward_user_identity: bool | None = None,
+        identity_mode: IdentityMode | None = None,
```
wylswz: Maybe we can just keep one parameter,
`identity_mode: IdentityMode = IdentityMode.OFF`, then None check is unnecessary

api/core/tools/mcp_tool/tool.py
```diff
@@ -280,9 +317,38 @@ def invoke_remote_mcp_tool(self, tool_parameters: dict[str, Any]) -> CallToolRes
...
+                user_id=user_id,
+                tenant_id=self.tenant_id,
+                app_id=app_id,
+                audience=audience,
+            )
+        except MCPTokenError as e:
+            raise ToolInvokeError(f"Failed to obtain forwarded identity token: {e}") from e
+        headers["Authorization"] = f"Bearer {token}"
```
wylswz: Maybe consider using a custom header like `x-dify-sso-access-token` to avoid conflict? The Authorization may be used for workspace scoped credential.

api/core/tools/mcp_tool/tool.py
CourTeous33: Most of the route are gated already. 

One possible non-gated senario is that someone curl to the controller with identity_mode=idp_token on a non-EE install. should we consider to gate such attack?

api/core/tools/mcp_tool/tool.py
CourTeous33: Accepted. 

This would require extra implementation in customers MCP service. will update document

api/models/tools.py
CourTeous33: Yes. forward_user_identity field is redundant, will romove

api/services/tools/mcp_tools_manage_service.py
CourTeous33: yes forward_user_identity can be removed

but PUT /workspaces/current/tool-provider/mcp can have the identity_mode = None if unchanged, so I would rather keep the none check.

api/core/tools/mcp_tool/tool.py
wylswz: Yes. Both API and backend implementation should be gated.

api/services/tools/mcp_tools_manage_service.py
wylswz: API related behavior could be handled by controller layer

wylswz (dismissed):

MRZHUH (dismissed):

MRZHUH (approved):
LGTM

PR #36545, opened 2026-05-23
chore: dep inject for sql session

asukaminato0721:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

#36544

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [ ] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [ ] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods

Issue #36544, opened 2026-05-23
dep inject in current flask


asukaminato0721:
For anyone wants to pick some files, make sure also migrate corresponding tests to test container.

see #32454

asukaminato0721:
@dosu based on https://github.com/langgenius/dify/pull/36545 , tell me how to improve the codebase.

asukaminato0721:
@dosu give me a file list need to update

m1ng-wym:
Hi @asukaminato0721 ,

I’m a newcomer to this repo. I’ve used Dify a lot before, and now I’d like to start contributing.

Since this issue touches many files, I’d like to handle it in batches. For the first PR, I plan to focus on the `console/app` controllers:

- `api/controllers/console/app/site.py`
- `api/controllers/console/app/model_config.py`
- `api/controllers/console/app/mcp_server.py`
- `api/controllers/console/app/workflow_draft_variable.py`
- `api/controllers/console/app/workflow_run.py`
- `api/controllers/console/app/conversation.py`
- `api/controllers/console/app/message.py`

This keeps the first change limited to one area and extends the `@with_session` pattern from #36545.

If this scope looks good, could you please assign this issue to me? I’ll add/update tests as needed and follow up on review feedback.

myshkin451:
Hi @asukaminato0721,

I’d like to help with this migration while avoiding overlap with @m1ng-wym’s proposed batch.

Would it be okay if I take a small non-overlapping slice around the remaining `console/app` usage?

Proposed scope:

- `api/controllers/console/app/generator.py`: apply `@with_session(write=False)` in `InstructionGenerateApi`
- `api/controllers/console/app/app.py`: migrate the remaining read-only `db.session.execute(...)` in `AppListApi`
- `api/services/workflow_service.py`: let `get_draft_workflow` optionally reuse an injected session
- update the related unit tests

If this scope sounds useful, could you assign this part to me? Thanks!

Quantum0uasar:
I'd like to pick up the `api/controllers/service_api/` slice (the Service API layer controllers) to avoid overlap with the Console and Inner API batches others are working on. Will open a PR shortly!

WUMIKE233:
Hi @asukaminato0721, I'd like to take a very small non-overlapping slice of this migration: pi/controllers/inner_api/wraps.py, specifically replacing the enterprise_inner_api_user_auth direct db.session.get(EndUser, user_id) lookup with an explicit session_factory.create_session() scope.

[...]

PR #36841, opened 2026-05-30
refactor: use explicit session in explore banner & auth controllers (#36544)

EvanYao826:
## Summary

Migrate `db.session` usage to explicit request-scoped sessions in three controller areas as part of #36544.

### Changes

- **`explore/banner.py`** — use `@with_session(write=False)` for read-only banner queries
- **`auth/activate.py`** — use one explicit session for invitation lookup, membership creation, and tenant switching
- **`auth/forgot_password.py`** — use one explicit session for account lookup, merge, workspace setup, and commit
- update affected unit/integration tests to inject the same session directly instead of patching global `db.session`

The auth handlers use `write=False` only to obtain a plain request-scoped session because the existing account/tenant services and reset flow commit explicitly. Wrapping those calls in `Session.begin()` would conflict with their internal commits.

### Verification

- Ruff on all changed files
- Mypy on the three changed controller files
- 23 focused tests covering banner, activation, and password reset paths

### Related

- Part of #36544
- Follows the `@with_session` pattern introduced in #36545


asukaminato0721 (changes requested):
fix ci, remove the patch

EvanYao826:
Addressed the requested changes: removed global-session patching, fixed all remaining mixed `db.session` usage, and avoided `Session.begin()` conflicts with service-level commits. Ruff, mypy, focused tests, and independent review now pass. Could you please take another look?

asukaminato0721:
conflict

asukaminato0721 (changes requested):
^

Issue #36898, opened 2026-06-01
[Refactor/Chore] Move localStorage/window.localStorage to @/hooks/use-local-storage


lyzno1:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

Move direct `localStorage` / `window.localStorage` / `globalThis.localStorage` usage in frontend code to `@/hooks/use-local-storage`.

Keep each PR small: migrate one vertical area or one storage concern at a time, instead of mixing unrelated keys/features in the same PR.

### Pattern

For normal JSON-backed values:

```ts
const [value, setValue] = useLocalStorage<boolean>('some-key', false)
```

For existing raw string values:

```ts
const [value, setValue] = useLocalStorage<string>('some-key', '0', { raw: true })
```

For setter-only usage:

```ts
const setValue = useSetLocalStorage<string>('some-key', { raw: true })
```

lyzno1:
@dosubot give a best practice example for this migration.

adityabhatkar23:
@lyzno1 can i work on this issue?

lyzno1:
> @lyzno1 can i work on this issue?
> 

sure, pls keep it within 3–5 call sites per PR.

myshkin451:
Hi @lyzno1, I'd like to take a small slice of this migration around the workflow block selector. I can start with `featured-tools.tsx` and `featured-triggers.tsx` if that sounds good.

kuishou68:
I've submitted a PR to migrate `NOTE_SHOW_AUTHOR_STORAGE_KEY` to use the `useLocalStorage`/`useSetLocalStorage` hooks: https://github.com/langgenius/dify/pull/36915

This covers the workflow note-node and operator hooks, replacing direct `localStorage.getItem`/`setItem` calls with the hook-based pattern.

myshkin451:
[...]

PR #36908, opened 2026-06-01
refactor(web): migrate NEED_REFRESH_APP_LIST_KEY to useLocalStorage/useSetLocalStorage

ZongrongLi:
## Summary

Migrates all direct `localStorage.getItem/setItem/removeItem` calls for `NEED_REFRESH_APP_LIST_KEY` to `useLocalStorage` and `useSetLocalStorage` hooks from `@/hooks/use-local-storage`.

This is part of the broader refactor tracked in #36898 — one storage concern at a time.

## Changes

**9 source files** touched, all using the same `{ raw: true }` option for raw string storage:

### Setter-only (useSetLocalStorage)
- `hooks/use-import-dsl.ts` — 2 occurrences
- `app/components/app-sidebar/app-info/use-app-info-actions.ts`
- `app/components/app/create-from-dsl-modal/index.tsx` — 2 occurrences
- `app/components/app/create-app-modal/index.tsx`
- `app/components/app/create-app-dialog/app-list/index.tsx`
- `app/components/app/switch-app-modal/index.tsx`
- `app/components/apps/app-card.tsx`
- `app/(commonLayout)/app/(appDetailLayout)/[appId]/overview/card-view.tsx`

### Getter + remover (useLocalStorage)
- `app/components/apps/list.tsx` — migrated the `getItem` + `removeItem` check-and-clear pattern to `useLocalStorage<string>(key, '0', { raw: true })`, calling `setNeedRefresh(null)` to clear.

## Checklist

- [x] TypeScript compiles with zero errors
- [x] All `localStorage.*NEED_REFRESH_APP_LIST_KEY` references in source files have been migrated

Part of #36898


lyzno1:
I think this changes the behavior of `NEED_REFRESH_APP_LIST_KEY` in `web/app/components/apps/list.tsx`.

Before this PR, the app list only performed a one-shot check on mount:

```ts
if (localStorage.getItem(NEED_REFRESH_APP_LIST_KEY) === '1') {
  localStorage.removeItem(NEED_REFRESH_APP_LIST_KEY)
  refetch()
}
```

After the change, `useLocalStorage` turns the key into subscribed state:

```ts
const [needRefresh, setNeedRefresh] = useLocalStorage<string>(NEED_REFRESH_APP_LIST_KEY, '0', { raw: true })
```

That changes two things:

1. When the key is missing, the hook writes the default `'0'` into localStorage. The old code did not write anything.
2. The list now reacts to same-tab storage updates from `useSetLocalStorage('1')`, because the hook dispatches/subscribes to its custom storage event. The old code only checked the flag once on mount. In cases like app duplication from an already-mounted app list, this can add another `refetch()` path on top of the existing `onRefresh={refetch}` path.

The writer-side replacements with `useSetLocalStorage(..., { raw: true })` look fine. The issue is only the reader migration in `apps/list.tsx`: this key is a one-shot refresh signal, not a persisted UI preference. To keep behavior equivalent, please keep the one-shot read/remove behavior here, or use a storage boundary that does not write a default value and does not subscribe to live updates.

lyzno1 (approved):

PR #36931, opened 2026-06-02
fix(api): enforce workspace membership + role checks in auth pipeline

GareArc:
## Summary

- Moves workspace ownership validation from handler-level decorators into the OpenAPI auth pipeline, closing the gap where a valid token could pass an arbitrary `workspace_id` and operate in a workspace the caller doesn't belong to.
- Adds `guard_workspace()` on `PipelineRouter` which sets `workspace_membership` and `allowed_roles` on `RequestContext`. New pipeline steps (`load_tenant_from_request`, `check_workspace_mismatch`, `check_workspace_role`) enforce membership and role before the handler runs.
- Migrates `apps.py` and `workspaces.py` endpoints from ad-hoc `require_workspace_role` decorator to pipeline-native enforcement. Covers CE/EE/SaaS editions.

Closes WTA-256

## Test plan

- [x] 136 unit tests pass (`pytest api/tests/unit_tests/controllers/openapi/auth/`)
- [x] New tests for `load_tenant_from_request` (6 tests), `check_workspace_mismatch` (4 tests), `check_workspace_role` (7 tests), `guard_workspace` (2 tests), conditions (6 tests), data fields (6 tests)
- [x] Lint clean (`dev/reformat` — ruff, pyrefly, import-linter, dotenv-linter)
- [ ] Verify non-member account gets 404 on workspace-scoped endpoints
- [ ] Verify wrong-role account gets 403 on admin-only endpoints
- [ ] Verify existing happy paths unchanged for member with correct role


api/controllers/openapi/auth/verify.py
```diff
@@ -30,6 +31,30 @@ def check_membership(data: AuthData) -> None:
     )
 
 
+def check_workspace_mismatch(data: AuthData) -> None:
```
lin-snow, line 32: No wired endpoint triggers this: it runs only under `WORKSPACE_MEMBERSHIP_REQUIRED & PATH_HAS_APP_ID`, but no `guard_workspace` route has an `app_id` path
  param (AppList uses `workspace_id` as a query param; workspaces.py has no `app_id`). So it currently never executes. Confirm it's intentional forward-looking
  code (and note that), or wire/remove it.

api/controllers/openapi/auth/verify.py
lin-snow: "404 (not 403) ... so workspace IDs don't leak across tenants"？

api/controllers/openapi/auth/verify.py
GareArc: Redesigned the pipeline conditions to make use of it.

wylswz (approved):

gausi-sudo:
## Pentest PASSED

<details>
<summary>3 tests passed</summary>

- `TC-941F848B` **HIGH** — User A must not list apps in a workspace they are not a member of via the /openapi/v1/apps endpoint. (`http://172.16.0.2`)
- `TC-BA74EF93` **HIGH** — User A must not access workspace member management endpoints (members list, invite, delete, role update) for a workspace they are not a member of. (`http://172.16.0.2`)
- `TC-F075BF7C` **HIGH** — A workspace member without Owner or Admin role must not invite members, remove members, or update member roles. (`http://172.16.0.2`)

</details>

_Pentested by [Niro](https://apxlabs.ai/niro)_

PR #36989, opened 2026-06-03
fix(api): tighten agent v2 generated contracts

lyzno1:
## Summary
- replace loose Agent v2 soul/node-job/candidate DTOs with explicit Pydantic schema models
- type Agent App presentation-feature request payloads and keep runtime dumps as plain JSON dicts
- regenerate console agents/apps contracts so Agent v2 operations no longer carry generated-contract deprecated markers

## Root Cause
The deprecated markers were not caused by Agent v2 endpoints calling old APIs internally. They came from the contracts generator marking operations as unsafe when OpenAPI request/response schemas contained loose object shapes such as dict[str, Any], list[dict[str, Any]], or additionalProperties: true without typed properties. Agent v2 exposed several future/legacy config slots through AgentSoulConfig, WorkflowNodeJobConfig, composer candidates, and agent-features, so the generated contracts were conservatively marked deprecated.

## Verification
- uv run --project api pytest api/tests/unit_tests/controllers/common/test_schema.py api/tests/unit_tests/commands/test_generate_swagger_specs.py api/tests/unit_tests/controllers/test_swagger.py
- uv run --project api pytest api/tests/unit_tests/core/workflow/nodes/agent_v2 api/tests/unit_tests/clients/agent_backend/test_request_builder.py
- pnpm --dir packages/contracts gen-api-contract
- pnpm --dir packages/contracts type-check
- uv run --project api ruff check api/controllers/console/app/agent_app_feature.py api/core/app/apps/agent_app/app_config_manager.py api/core/app/apps/agent_app/runtime_request_builder.py api/core/workflow/nodes/agent_v2/runtime_feature_manifest.py api/core/workflow/nodes/agent_v2/runtime_request_builder.py api/core/workflow/nodes/agent_v2/validators.py api/fields/agent_fields.py api/models/agent_config_entities.py
- node readiness check for Agent v2 generated operations

### DIFF
--- a/api/controllers/console/app/agent_app_feature.py
+++ b/api/controllers/console/app/agent_app_feature.py
@@ -23,6 +23,7 @@ from controllers.console.wraps import (
     with_current_user,
 )
 from events.app_event import app_model_config_was_updated
+from libs.helper import dump_response
 from libs.login import login_required
 from models import Account
 from models.agent_config_entities import (
@@ -35,7 +36,7 @@ from models.model import App, AppMode
 from services.agent_app_feature_service import AgentAppFeatureConfigService
 
 
-class AgentAppFeaturesRequest(BaseModel):
+class AgentAppFeaturesPayload(BaseModel):
     """Presentation features configurable on an Agent App.
 
     All fields are optional; an omitted field is reset to its disabled/empty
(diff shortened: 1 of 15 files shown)


zyssyz123 (approved):
LGTM

PR #37022, opened 2026-06-03
fix: strip local file defaults from app DSL exports

### LINKED ISSUE #36996: [Bug] Workflow stuck in "Running" forever after cross-workspace import: `upload_file_id` not filtered in DSL export + error not propagated via SSE
### Self Checks

### Dify version

1.14.2

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

> **Note:** This issue was originally reported as #33822 (v1.10.1, closed as stale without a fix). The root cause is **still present and reproducible in v1.14.1 and v1.14.2**. This report adds a second compounding bug (SSE error not propagated) and confirms the DSL export issue persists across major versions.

1. In **Workspace A**, create a Workflow app with a `file` or `file-list` type variable in the Start node
2. Upload a local file and set it as the **default value** for that variable
3. Export the app DSL via `GET /console/api/apps/{app_id}/export?include_secret=false`
4. Import the DSL into **Workspace B** (a different tenant) — import succeeds with no warning
5. Open the imported workflow in Studio → click **Test Run** → click **Start Run** (default file is auto-filled from the saved graph)
6. **Observe:** The UI shows `Test Run (Running)` indefinitely. No nodes execute, no error is displayed.

### Evidence: DSL contains `upload_file_id` (confirmed in v1.14.2)

The exported YAML at `workflow.graph.nodes[start].data.variables[].default` contains:

```yaml
variables:
  - type: file
    variable: contract_file
    default:
      name: example.pdf
      transfer_method: local_file
      type: document
      upload_file_id: aabbccdd-1234-5678-abcd-workspaceafile   # ← Workspace A's file ID, not filtered
      uploadedId: aabbccdd-1234-5678-abcd-workspaceafile       # ← camelCase duplicate
```

The `upload_file_id` is **not stripped during DSL export** and is persisted in the `graph` column. Every time the Test Run panel opens, it auto-populates this stale cross-tenant file ID.

### ✔️ Expected Behavior

1. **On export/import**: `upload_file_id` / `uploadedId` should be stripped from file variable defaults in the DSL — file IDs are tenant-private and meaningless across workspaces
2. **On runtime error**: When `ValueError("Invalid upload file")` is raised, the user should see a clear error message instead of the workflow freezing in `Running` state

### ❌ Actual Behavior

**Two bugs compound to produce this failure:**

### Bug 1 — DSL export does not filter cross-tenant file IDs (regression from #33822)

- Both `upload_file_id` (snake_case, read by `_build_from_local_file`) and `uploadedId` (camelCase, stored by the frontend) are serialized into the exported DSL
- After import into another tenant, these stale IDs are stored in the workflow `graph` column
- The Start node's Test Run panel auto-populates the default file on every open, making it impossible to simply refresh to fix the problem

**Affected file:** `api/core/app/apps/workflow/app_generator.py` → DSL export path
**DSL version:** 0.6.0

### Bug 2 — `ValueError` raised before `queue_manager` is initialized, preventing SSE error delivery

```
workflow_execute_task
  └─ WorkflowAppGenerator.generate()
       └─ _prepare_user_inputs()         ← called BEFORE queue_manager is created
            └─ file_factory.build_from_mappings()
                 └─ _build_from_local_file()
                      └─ UploadFile.tenant_id == tenant_id  ← cross-tenant mismatch → not found
                           └─ raise ValueError("Invalid upload file")
```

Because the exception is raised **before** `queue_manager` (the SSE message queue) is initialized, the error **cannot be pushed to the frontend via SSE**. The SSE connection hangs for the full 300-second timeout, leaving the UI permanently stuck in `Running` with no feedback to the user.

**Worker log:**
```
Task tasks.app_generate.workflow_execute_task.workflow_based_app_execution_task[xxx]
  raised unexpected: ValueError('Invalid upload file')
```

**Affected files:**
- `api/factories/file_factory/builders.py` → `_build_from_local_file()` (~line 135–141)
- `api/core/app/apps/workflow/app_generator.py` → `generate()` (~line 175)
### Short-term (fix SSE propagation — Bug 2)

In `app_generator.py` `generate()`, wrap `_prepare_user_inputs()` in a try/except placed **after** `queue_manager` is initialized, so errors can be surfaced to the user:

```python
# Current (broken): exception thrown before queue_manager exists
inputs = self._prepare_user_inputs(...)   # ~line 175

# Fix: move after queue_manager init, or catch and publish
try:
    inputs = self._prepare_user_inputs(...)
except ValueError as e:
    queue_manager.publish_error(e, PublishFrom.APPLICATION_MANAGER)
    return
```

### Long-term (fix DSL export — Bug 1)

Strip `upload_file_id` and `uploadedId` from file/file-list variable defaults during DSL export. File IDs are tenant-scoped storage references that are meaningless in another workspace. Optionally emit a warning during import if stale file IDs are detected.

## Additional context

- **Related issue:** #33822 — same root cause, reported on v1.10.1, closed as stale without a fix being merged. This report confirms the bug persists in v1.14.1 and v1.14.2.
- Both `upload_file_id` (snake_case, backend) and `uploadedId` (camelCase, frontend) exist simultaneously in the exported DSL, meaning both the frontend storage path and backend read path are affected
- Confirmed environment: Dify v1.14.1, Self Hosted (Docker Compose), App DSL version 0.6.0

he-yufeng:
## Summary
- clear tenant-local file defaults from workflow app DSL exports
- remove local-file entries from file-list defaults while preserving portable remote-url entries
- add a unit test covering snake_case and camelCase local file ids

## Why
Workflow Start node file defaults can contain `upload_file_id` / `uploadedId` values that only exist in the source workspace. Exporting those ids into DSL lets the import succeed in another workspace, but the imported app later auto-fills a stale file reference and fails at runtime with `Invalid upload file` or related file-mapping errors.

This keeps tenant-private upload file references out of exported app DSL instead of preserving a default value that cannot work in another workspace.

Fixes #36996.
Related stale report: #33822.

## Validation
- `uv sync --group dev --no-default-groups`
- `.venv\\Scripts\\python.exe -m pytest -o addopts="" tests\\unit_tests\\services\\test_app_dsl_service.py -q`
- `.venv\\Scripts\\ruff.exe check services\\app_dsl_service.py tests\\unit_tests\\services\\test_app_dsl_service.py`
- `.venv\\Scripts\\python.exe -m py_compile services\\app_dsl_service.py tests\\unit_tests\\services\\test_app_dsl_service.py`
- `git diff --check`

Note: I used `-o addopts=""` for the targeted pytest run because the default coverage plugin hit a Windows `.coverage` file lock after the test itself had passed.


api/services/app_dsl_service.py
```diff
@@ -599,6 +603,51 @@ def _append_workflow_export_data(
             )
         ]
 
+    @classmethod
+    def _strip_tenant_file_defaults_from_workflow_dict(cls, workflow_dict: dict[str, Any]) -> None:
+        for node in workflow_dict.get("graph", {}).get("nodes", []):
+            node_data = node.get("data", {})
```
QuantumGhost, line 722: I suggest only processing nodes with `start` type and skip other nodes.

api/services/app_dsl_service.py
```diff
@@ -599,6 +603,51 @@ def _append_workflow_export_data(
...
+            if not isinstance(node_data, dict):
```
QuantumGhost, line 723: I suggest building a [VariableEntity](https://github.com/langgenius/graphon/blob/39d5d474d9cb0646b0dd5eb56a323d4ac34363d2/src/graphon/variables/input_entities.py#L32) from the DSL and operate on the VariableEntity. This approach makes the code easier to maintain.

api/services/app_dsl_service.py
he-yufeng: Good point. Right now the loop only acts on nodes that carry a `variables` list with `file`/`file-list`-typed entries, so non-input nodes are skipped in practice — but I can make that explicit with a `start`-type check if you would rather not lean on the `variables` shape. Happy to tighten it.

api/services/app_dsl_service.py
he-yufeng: I went back and forth here. This pass runs on the already-serialized `workflow_dict` right before export and only needs to null out local-file default ids, so I kept it on the raw dict to avoid coupling the exporter to the variable-entity model for a sanitization step. If you would prefer the `VariableEntity` route for maintainability I am glad to switch — just flagging the tradeoff. Rebased on `main`, so this is mergeable again (kept your import-size tests alongside the strip test).

he-yufeng:
Closing this one myself. The export path has moved twice since July (app deployment v2 in #39829, then the ownership-aware lifecycle work), and this diff no longer applies cleanly. The underlying leak is still present on current main (there is no strip helper in app_dsl_service.py), so if a maintainer wants the fix I am happy to re-cut it fresh against the new flow. The branch stays around for reference.

PR #37036, opened 2026-06-04
feat(cli): difyctl release pipeline + tokenless installers

GareArc:
## Summary

Reworks the `difyctl` release pipeline and installers so binaries are published as **public GitHub Releases** and installed **without authentication**. Replaces the previous flow, which built per-run Actions artifacts and required a token to download them.

Closes WTA-426

### Release workflow (`.github/workflows/cli-release.yml`)

- **Trigger-agnostic entry.** Runs from `workflow_dispatch`, `workflow_call`, `repository_dispatch` (`difyctl-release`), and `push` of a `difyctl-v*` tag. The version is always sourced from `cli/package.json`, never inferred from the trigger.
- **Two-job split: `validate` → `release`.**
  - `validate` reads the manifest, checks the manifest schema, enforces that a pushed tag equals `difyctl-v<version>`, and runs an idempotency guard (`gh release view`) so an already-published version is skipped.
  - `release` builds all targets, generates checksums, and publishes the GitHub Release (creating the `difyctl-v<version>` tag).
- **Immutable releases.** A published version is never overwritten; the idempotency guard makes re-runs no-ops. Fixing a shipped bug means a patch bump.
- **Serialized concurrency.** A single fixed concurrency group with `cancel-in-progress: false` ensures releases never run in parallel and an in-flight release is never cancelled mid-publish.

### Build tooling

- **`cli/.bun-version` (new).** Pins Bun for reproducible `bun build --compile` output across all targets.

## Before / after

| | Before | After |
|---|---|---|
| Artifact source | Actions workflow artifacts (per run) | Public GitHub Releases |
| Install auth | `GH_TOKEN` required | None |
| Triggers | dispatch + tag | dispatch + call + repository_dispatch + tag |
| Idempotency | none | `gh release view` guard, immutable |
| Concurrency | per-ref, cancel-in-progress | single group, never cancelled |
| Windows install | not supported | `install.ps1` |
| Checksum verify | — | fail-closed `sha256` (sh + ps1) |

## Release flow

```mermaid
flowchart TD
    A[workflow_dispatch] --> V
    B[workflow_call] --> V
    C[repository_dispatch: difyctl-release] --> V
    D[push tag difyctl-v*] --> V

    subgraph validate
        V[read version from cli/package.json] --> G1[validate manifest]
        G1 --> G2{tag == difyctl-v&lt;version&gt;?}
        G2 -- no --> X[fail]
        G2 -- yes / non-tag --> G3{release already exists?}
    end

    G3 -- yes --> S[skip: immutable]
    G3 -- no --> R

    subgraph release
        R[build all targets] --> CK[generate sha256 checksums]
        CK --> P[publish GitHub Release + tag]
    end
```

Targets: `linux-x64`, `linux-arm64`, `darwin-x64`, `darwin-arm64`, `windows-x64.exe`, plus a `difyctl-v<version>-checksums.txt`.

## Install flow

```mermaid
flowchart TD
    I[install-cli.sh / install.ps1] --> Q[resolve highest tag via git/matching-refs]
    Q --> C{channel}
    C -- stable --> ST[highest non-prerelease]
    C -- rc --> RC[highest -rc.N]
    ST --> DL[download asset + checksums.txt]
    RC --> DL
    DL --> Vfy{sha256 matches?}
    Vfy -- no --> Fail[abort]
    Vfy -- yes --> Inst[install to PREFIX/bin/difyctl]
```

## Testing

- `vitest run cli/scripts/install-cli.test.ts` — resolver unit tests pass (channel filter, `rc` ordering).


cli/scripts/install.ps1
```diff
@@ -0,0 +1,107 @@
+# install.ps1 — one-line difyctl installer for Windows. difyctl ships as assets
```
wylswz, line 1: Is there a test for ps1?

GareArc: Tested with arm pwsh 7.6.2.

wylswz (approved):

PR #37048, opened 2026-06-04
feat: redact workflow secrets and restrict secret-bearing DSL export

### LINKED ISSUE #37047: Prevent workflow secret values from leaking via logs, traces, and DSL export
### Self Checks

When a workflow node references a secret (e.g. an API key stored as a secret variable), the plaintext value can currently leak through several surfaces even though the editor UI masks it:

- **Execution persistence / logs / traces**: secret values (and values tainted by them) are written verbatim into `workflow_node_executions` inputs/outputs, so anyone with log/trace access can read them.
- **DSL export**: exporting an app with `include_secret=True` returns the plaintext secret to any user who can hit the export endpoint, regardless of role.
- **Node configuration UI**: secret variables show up as selectable/visible config in some nodes.

The intended behavior is: **an app builder can *use* a secret in a node, but must not be able to *see* or *exfiltrate* its plaintext value.**

Proposed scope (redaction + export hardening only, no new ACL/visibility feature):

1. Redact secret values from execution inputs/outputs, logs, and traces. Real secrets are masked by substring (min length 8); values tainted by a secret are masked by exact match; very short taints (<4 chars) are excluded to avoid over-masking.
2. Restrict `include_secret=True` DSL export to workspace owners/admins; non-managers get a secret-free export.
3. Filter secret variables out of node configuration UI.

Out of scope (explicitly deferred): runtime exfiltration where a raw secret is passed into user-controlled Code/HTTP node logic. This is a known limitation and not addressed here.

r-hashi01:
## Summary

Fixes #37047

A workflow node can reference a secret (e.g. an API key stored as a secret variable), but its plaintext value could leak through execution persistence/logs/traces and through DSL export. This PR closes those plaintext surfaces while keeping the builder able to *use* the secret.

Changes:

- **Redact secrets from execution outputs/logs/traces.** New `api/core/workflow/secret_redaction.py` masks sensitive values before they are persisted to `workflow_node_executions`. Real secrets are masked by substring (min length 8); values tainted by a secret are masked by exact match; taints shorter than 4 chars are excluded to avoid over-masking. Wired in via the persistence layer and the workflow/advanced-chat/pipeline runners.
- **Restrict secret-bearing DSL export to workspace owners/admins.** `AppDslService.assert_secret_export_allowed` raises `Forbidden` for non-managers requesting `include_secret=True`; the export endpoint enforces it. Non-managers receive a secret-free export. Internal `include_secret=True` paths (app copy, RAG templating) are unaffected since they are not returned to the user.
- **Filter secret variables out of node configuration UI** (answer/end nodes).

Out of scope (explicitly deferred): runtime exfiltration where a raw secret is passed into user-controlled Code/HTTP node logic. This is a known limitation, tracked in #37047.

## Screenshots

| Before | After |
|--------|-------|
| Secret value appears in node execution output / exported DSL | Value redacted in execution output; non-managers get secret-free export (`value: ''`) |

## Checklist


Kota-Maeda:
Great work! This is a much-needed feature for security.
I'm really looking forward to seeing this merged.

dparkmit24:
Reproduced on 1.16.1 (self-hosted, Docker) while building a workflow, so adding a data point and a scoping question. Thanks for picking this up — the taint-tracking for code-node outputs is a nice touch and the exact-match threshold reasoning in `secret_redaction.py` is well argued.

**Reproduction — HTTP Request node, not covered by the existing tests here**

1. ENV variable `GITHUB_TOKEN`, type `Secret`. Confirmed masked in the ENV panel (`github***********Rq`).
2. HTTP Request node, `Authorization` header value = `Bearer ` + the secret reference.
3. Run the node.

`PROCESS DATA` renders the full request line with the token in plaintext:

```
"request": "GET /search/issues?... HTTP/1.1\r\nHost: api.github.com\r\n...\r\nAuthorization: Bearer github_pat_11BG4..."
```

Same on a full Test Run. The tests in this PR cover Code-node inputs/outputs and error strings; the HTTP node reaches `process_data` by a different route, and the secret arrives embedded in a larger string rather than as a standalone value — which your `_MIN_SUBSTRING_REDACTION_LENGTH` handling looks like it covers, but there's no test pinning it.

To be clear this is distinct from the runtime-exfiltration case you've deferred — I'm not sending the secret anywhere the workflow didn't intend, just observing that the request line gets recorded verbatim.

**Scoping question**

As I read it, redaction here happens at persistence: `WorkflowPersistenceLayer._update_node_execution` and `WorkflowService._populate_execution_result`. Both would cover what I reproduced above.

`api/core/app/apps/common/workflow_response_converter.py` isn't in this changeset, and `workflow_node_finish_to_stream_response` builds the `node_finished` payload as:

```python
process_data, process_data_truncated = self._truncate_mapping(event.process_data)
```

— truncation, no redaction — and passes `error=error_message` through unchanged. `workflow_node_retry_to_stream_response` has the same shape. And when `invoke_from == InvokeFrom.SERVICE_API`, `DummyVariableTruncator` means that payload isn't truncated either.

If I'm reading that right, after this lands `workflow_node_executions` would be clean while the SSE stream still carries the raw secret to the client. Is the streaming path deliberately out of scope for this PR, or would it be worth routing the converter through `redact_sensitive_values` as well?

Happy to test a patch against my reproduction if that's useful.

PR #37057, opened 2026-06-04
fix(api): return agent timestamps as epoch seconds

lyzno1:
## Summary
- Return agent roster, snapshot, and revision timestamps as epoch seconds instead of ISO strings
- Update agent response schemas to document timestamp fields as integers
- Cover non-null timestamp serialization in agent service tests

## Tests
- uv run --project api --dev ruff check api/fields/agent_fields.py api/services/agent/roster_service.py api/services/agent/composer_service.py api/tests/unit_tests/services/agent/test_agent_services.py
- uv run --project api --dev pytest api/tests/unit_tests/services/agent/test_agent_services.py api/tests/unit_tests/controllers/console/agent/test_agent_controllers.py api/tests/unit_tests/controllers/common/test_schema.py api/tests/unit_tests/commands/test_generate_swagger_specs.py api/tests/unit_tests/controllers/test_swagger.py

Note: generated frontend contracts are intentionally not updated in this PR; CI/schema generation should handle the downstream contract refresh.

### DIFF
--- a/api/fields/agent_fields.py
+++ b/api/fields/agent_fields.py
@@ -37,7 +37,7 @@ class AgentConfigSnapshotSummaryResponse(ResponseModel):
     summary: str | None = None
     version_note: str | None = None
     created_by: str | None = None
-    created_at: str | None = None
+    created_at: int | None = None
 
 
 class AgentRosterResponse(ResponseModel):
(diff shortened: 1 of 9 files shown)


zyssyz123 (approved):
LGTM

PR #37090, opened 2026-06-05
feat(cli): adopt generated oRPC contract for unary endpoints

lin-snow:
## Summary

Adopt the generated oRPC contract in difyctl: a low-level `client.request` routes oRPC's `OpenAPILink` through the existing HTTP transport (retry / timeout / bearer / error-map), and every clean unary endpoint now calls the contract. Streaming / multipart / auth-bootstrap stay on the raw client.

Backend (`/openapi/v1` surface only): add the missing response annotations so `stop` / `submit` / `health` / `sessions` / `apps.describe` lose their `deprecated` flag. The console contract is unchanged.

## Checklist

- [x] CLI type-check + lint + 927 tests green; backend ruff clean.

---

## One engine, multiple facades

```mermaid
graph TD
    C[Command layer<br/>commands/*]
    C --> O[Domain wrapper layer<br/>api/*.ts]
    O -->|"this.orpc (contract facade)"| ORPC[oRPC + OpenAPILink]
    O -->|"this.http (raw entry)<br/>SSE · multipart · auth bootstrap"| RAW[http.stream / http.fetch]
    ORPC --> REQ[http.request]
    REQ --> EX[execute<br/>retry · timeout · +UA · +bearer]
    RAW --> EX
    EX --> NET[global fetch · network]
```

`this.orpc` and `this.http` both funnel into the same `execute`. The only difference: the path-style `get/post` go through `buildRequest`, which does `joinURL(baseURL, path)`; `requestFetch` **skips the join** because oRPC already produced an absolute URL. UA / bearer are `execute`'s `onRequest` hooks (`client.ts:56-58`), independent of who built the `Request`, so oRPC calls inherit them for free.

---

## Error handling: two seams

```mermaid
graph LR
    RESP[Response] --> Q{res.ok?}
    Q -->|"non-2xx"| S1["seam ① fetch wrapper<br/>classifyResponse(req,res)<br/>→ BaseError"]
    Q -->|"2xx"| D{oRPC can decode?}
    D -->|yes| OK[typed result]
    D -->|no| S2["seam ② link wrapper<br/>mapOrpcError<br/>→ unknownError"]
```

Both seams sit at the transport boundary; together they guarantee an error has **already become a `BaseError`** by the time it leaves the client. Why split into two layers:

| seam | location | why it must be here |
| -- | -- | -- |
| ① fetch wrapper | the `fetch` option in `orpc.ts` | here we hold both `req` + `res`, so we can preserve method/url + the raw body (`ORPCError` can't reach url/method) |
| ② link wrapper | wrapped around `link.call` | decode failure happens **after** oRPC has the Response, out of the fetch layer's reach |

### DIFF
--- a/api/controllers/openapi/account.py
+++ b/api/controllers/openapi/account.py
@@ -6,6 +6,7 @@ from flask import request
 from flask_restx import Resource
 from werkzeug.exceptions import NotFound
 
+from controllers.common.schema import query_params_from_model
 from controllers.openapi import openapi_ns
 from controllers.openapi._models import (
     MAX_PAGE_LIMIT,
@@ -13,6 +14,7 @@ from controllers.openapi._models import (
     AccountResponse,
     PaginationEnvelope,
     RevokeResponse,
+    SessionListQuery,
     SessionListResponse,
     SessionRow,
     WorkspacePayload,
@@ -70,6 +72,7 @@ class AccountSessionsSelfApi(Resource):
 
 @openapi_ns.route("/account/sessions")
 class AccountSessionsApi(Resource):
+    @openapi_ns.doc(params=query_params_from_model(SessionListQuery))
     @openapi_ns.response(200, "Session list", openapi_ns.models[SessionListResponse.__name__])
     @auth_router.guard(scope=Scope.FULL, allowed_token_types=frozenset({TokenType.OAUTH_ACCOUNT}))
     def get(self, *, auth_data: AuthData):
--- a/api/controllers/openapi/account.py
+++ b/api/controllers/openapi/account.py
@@ -4,12 +4,12 @@ from datetime import UTC, datetime
 
 from flask import request
 from flask_restx import Resource
-from werkzeug.exceptions import NotFound
+from pydantic import ValidationError
+from werkzeug.exceptions import NotFound, UnprocessableEntity
 
 from controllers.common.schema import query_params_from_model
 from controllers.openapi import openapi_ns
 from controllers.openapi._models import (
-    MAX_PAGE_LIMIT,
     AccountPayload,
     AccountResponse,
     PaginationEnvelope,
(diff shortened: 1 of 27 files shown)


api/controllers/openapi/account.py
```diff
@@ -70,13 +72,21 @@ def delete(self, *, auth_data: AuthData):
...
+    @openapi_ns.doc(params=query_params_from_model(SessionListQuery))
     @openapi_ns.response(200, "Session list", openapi_ns.models[SessionListResponse.__name__])
     @auth_router.guard(scope=Scope.FULL, allowed_token_types=frozenset({TokenType.OAUTH_ACCOUNT}))
     def get(self, *, auth_data: AuthData):
+        # Validate page/limit through the same model the contract advertises (extra='forbid',
+        # page>=1, 1<=limit<=MAX_PAGE_LIMIT) so the server actually enforces those bounds rather
+        # than silently coercing (e.g. page=0 -> empty slice). Mirrors AppDescribeQuery.
+        try:
+            query = SessionListQuery.model_validate(request.args.to_dict(flat=True))
+        except ValidationError as exc:
```
GareArc, line 84: This block is likely to repeat across the openapi controllers. Feel like it is worth folding into decorator layers.

### Propose mental model:
`@expect(query=, body=, files=)` : one per method; validates the request, injects typed kwargs, emits the param/body/formData schema. Single decorator owns the whole request so it can enforce the swagger in one place.
`@response(status, Model)` :  one per status; emits the response schema and serializes the 2xx return through the model.
`@doc(description=, summary=, …)` : prose only.

And the surface becomes something like `def get(self, *, auth_data: AuthData, query: SessionListQuery):`

lin-snow: Agreed — worth a decorator. Worth noting the repeated blocks are also inconsistent: inline ones raise 422, the _validate_body/_validate_json helpers raise 400 for the same case. A @validate_request(query=, body=) would unify that + emit the swagger params in one place (same kwargs[...] injection auth_router.guard already uses).

I'd do it as a focused follow-up rather than grow this PR — it touches every openapi endpoint and is orthogonal to the oRPC migration. Keeping the inline block here (covered by the new test_account.py cases).

GareArc (dismissed):
lgtm

wylswz (approved):

PR #37091, opened 2026-06-05
fix(web): attach Amplitude user ID before firing registration event

CodingOnStar:
`user_registration_success` was sent to Amplitude before a user ID was attached, so it landed on an anonymous profile and broke cohort attribution. The user ID is only set in AppContextProvider via setUserId() once the profile loads after the post-registration redirect, while the event was fired earlier at registration time (both the email and OAuth flows).

Neither producer has the user ID at fire time (the email-register response carries no email/id; the OAuth initializer only has a boolean flag), so the event is now deferred via sessionStorage (rememberRegistrationSuccess) and replayed in AppContextProvider right after setUserId runs (flushRegistrationSuccess) — guaranteeing setUserId -> track ordering so the event lands on the identified profile.

GA events are intentionally left firing immediately: gtag manages its own identity and is unaffected by this ordering issue.

> [!IMPORTANT]
>

## Checklist


zyssyz123 (approved):
LGTM

PR #37115, opened 2026-06-05
fix: tolerate missing dataset icon info

he-yufeng:
## Summary

- make the app dataset selector tolerate missing or null `icon_info`
- reuse the same default dataset icon shape used elsewhere in the web app
- add a regression test for datasets that render with `icon_info: null`

## Context

`DatasetNav` already normalizes missing dataset icon metadata on current `main`, but the app configuration dataset selector still reads `item.icon_info.icon_type` and related fields directly. A dataset row with null icon metadata can still crash that selector before the user can pick a dataset.

This follows the existing fallback convention instead of changing the API contract.

## Validation

- `git diff --check`
- secret scan on the changed files

## To verify

- `pnpm --filter dify-web test web/app/components/app/configuration/dataset-config/select-dataset/__tests__/index.spec.tsx`

I could not complete the targeted pnpm test locally because dependency installation repeatedly timed out while downloading packages from the npm registry. The lockfile policy check passed on retry, but tarball downloads kept aborting before a usable `node_modules` was created.

### DIFF
--- a/web/app/components/app/configuration/dataset-config/select-dataset/__tests__/index.spec.tsx
+++ b/web/app/components/app/configuration/dataset-config/select-dataset/__tests__/index.spec.tsx
@@ -9,6 +9,12 @@ import { DatasetPermission } from '@/models/datasets'
 import { RETRIEVE_METHOD } from '@/types/app'
 import SelectDataSet from '../index'
 
+const mockAppIcon = vi.hoisted(() => vi.fn(() => null))
+
+vi.mock('@/app/components/base/app-icon', () => ({
+  default: mockAppIcon,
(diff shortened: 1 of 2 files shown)


crazywoola:

We require every PR to be associated with a prior discussion or issue, and the PR description should link it using GitHub's closing syntax, for example `Fixes #<issue number>`, `Closes #<issue number>`, or `Resolves #<issue number>`.

I'm closing this PR for now because I couldn't find a linked issue in the description. Please open or find the related issue first, then reopen or submit a new PR that links it properly. Typos and very small documentation fixes may be exceptions.

PR #37165, opened 2026-06-08
fix(rag): use IMAGE_QUERY when reranking attachment-only hybrid search

### LINKED ISSUE #37116: Attachment-only hybrid retrieval is reranked as TEXT_QUERY instead of IMAGE_QUERY
Self-checks: searched existing issues, reproduced on latest `main`.

### Summary
In `RetrievalService._retrieve` (`api/core/rag/datasource/retrieval_service.py`), the `HYBRID_SEARCH` branch derives the rerank `query_type` like this:

```python
query = query or attachment_id
if not query:
    return
all_documents_item = data_post_processor.invoke(
    ...,
    query_type=QueryType.TEXT_QUERY if query else QueryType.IMAGE_QUERY,
)
```

The ternary is evaluated **after** `query = query or attachment_id`. For an attachment-only retrieval the caller passes `query=None, attachment_id=<upload_file_id>`, so by the time of the ternary `query` is the (truthy) upload-file id — and the `if not query: return` guard guarantees `query` is truthy here. The `IMAGE_QUERY` branch is therefore **unreachable**, and attachment-only hybrid retrieval is always reranked as `TEXT_QUERY`.

### Impact
In `RerankModelRunner.fetch_multimodal_rerank` (`core/rag/rerank/rerank_model.py`), `IMAGE_QUERY` loads the upload file from storage and base64-encodes the image as the rerank query, whereas `TEXT_QUERY` passes the query string to the reranker as literal text. With this bug, an image/attachment-driven hybrid retrieval using a vision rerank model sends the **raw upload-file-id string** to the reranker as text instead of the image, producing incorrect rerank ordering.

### Sibling reference
The non-hybrid embedding path in the same file already distinguishes them correctly — `query_type=QueryType.TEXT_QUERY` for the text future and `query_type=QueryType.IMAGE_QUERY` for the attachment future.

### Fix
Compute `query_type` from the original `query` before it is overwritten with `attachment_id`. A PR with a regression test follows.

caoergou:
## Summary

In `RetrievalService._retrieve`, the `HYBRID_SEARCH` branch used `query = query or attachment_id` before deriving `query_type`, making `IMAGE_QUERY` permanently unreachable. Attachment-only hybrid retrieval always passed a raw file-id string to the reranker as a text query instead of routing to the vision rerank path.

Fixes #37116

## Changes

- Compute `rerank_query` and `query_type` independently before any variable is overwritten, mirroring the sibling `embedding_search` path (lines 737/754) which already handles text vs. image queries separately
- Add parametrized regression test covering both `query=""` and `query=None` entry points

## Test plan

- `pytest api/tests/unit_tests/core/rag/retrieval/test_dataset_retrieval.py` — 35 passed
- `ruff check` — no violations


caoergou:
Hi, just checking in — happy to address any feedback or make adjustments if needed. Thanks for your time! 🙏

crazywoola:
Thanks for the contribution. We already have an earlier PR for this issue and follow FIFO for duplicate fixes, so review will continue in #37117.

PR #37211, opened 2026-06-09
fix: preserve file preview content type

### LINKED ISSUE #37198: File preview returns `Content-Type: application/octet-stream` on v1.14.2 with local storage, causing browser download instead of inline preview
### Self Checks

### Dify version

1.14.2

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

### ✔️ Expected Behavior

### ❌ Actual Behavior

```http
HTTP/1.1 200 OK
Content-Length: 818288
Connection: keep-alive
Server: gunicorn
X-Version: 1.14.2
X-Env: PRODUCTION
Vary: Cookie
Content-Disposition: inline

myshkin451:
## Summary

Fixes #37198

This keeps the uploaded file MIME type for inline `/file-preview` responses. The response was already created with `upload_file.mime_type`, but `Content-Type` was later overwritten to `application/octet-stream` for every file, which made browsers download PDFs/images instead of previewing them inline.

The change now only sets `application/octet-stream` for explicit attachment downloads. The existing HTML safety path still forces HTML previews to download via `enforce_download_for_html()`.

## Checklist

Validation:

- `pytest api/tests/unit_tests/controllers/files/test_image_preview.py`
- `ruff check api/controllers/files/image_preview.py api/tests/unit_tests/controllers/files/test_image_preview.py`


FFXN:
Thanks for the fix. The direction looks right to me: inline `/file-preview` should preserve the original MIME type for PDFs/images, and only explicit attachment downloads should force `application/octet-stream`.

One concern: the previous unconditional `application/octet-stream` behavior seems to have been introduced for a security reason, specifically to avoid loading executable content in SVG previews. Since `svg` is included in `IMAGE_EXTENSIONS`, this change may allow `image/svg+xml` files to be rendered inline again.

Could you please also keep SVG previews forced as downloads, similar to the existing HTML safety path? For example, when `mime_type == "image/svg+xml"` or the extension/filename is `.svg`, set:

- `Content-Disposition: attachment`
- `Content-Type: application/octet-stream`
- `X-Content-Type-Options: nosniff`

It would also be good to add a unit test covering SVG, alongside the new PDF/HTML tests. That should preserve the intended PDF/image inline preview behavior without regressing the earlier SVG security protection.

myshkin451:
Thanks for catching this, @FFXN — and sorry for the slow follow-up. I've updated the PR so SVG previews are still forced to download when SVG is detected by the MIME type, filename, or stored extension. SVG responses now use `Content-Disposition: attachment`, `Content-Type: application/octet-stream`, and `X-Content-Type-Options: nosniff`, with tests covering all three detection paths.

PDF and other ordinary previews still keep their original MIME type. The branch is rebased onto the current `main`, and all checks are passing.

FFXN (dismissed):

FFXN (approved):

wylswz (approved):

PR #37212, opened 2026-06-09
feat(api,cli): strict UUID validation for app-id and workspace-id

GareArc:
## Summary

Root cause of WTA-493: `load_app` in the auth pipeline called `session.get(App, app_id)` with a non-UUID string against a PostgreSQL UUID column → `DataError` → unhandled 500.

**Backend**
- `load_app()` (`auth/prepare.py`): validate UUID before DB call; non-UUID app_id now returns 404 instead of 500
- `AppDescribeQuery`: remove `workspace_id` field — describe endpoint is UUID-only; callers sending it get 422 (`extra=forbid`)
- `AppListQuery.workspace_id`: `str` → `UUIDStr` (new strict type, rejects empty string and non-UUID)
- `AppDescribeApi.get()`: call `_load(app_id)` without workspace_id; UUID gate is in the auth pipeline
- `UUIDStr`: new strict type alias in `libs/helper.py` alongside existing `UUIDStrOrEmpty`
- `AppReadResource._load()`: kept intact for future name-based lookup via subclasses

**CLI**
- `resolveWorkspaceId()`: validate `--workspace` flag and `DIFY_WORKSPACE_ID` env var as UUID; exports `isValidUuid()`
- `DescribeApp.run()`: validate positional app UUID client-side before API call
- `AppsClient.describe()`: remove `workspaceId` param — no longer sent to API
- `AppMetaClient.get()`: remove `workspaceId` param
- `run/resume/describe/get app`: drop `wsId` from `meta.get()` calls; `run`/`resume` no longer call `resolveWorkspaceId` (it was only feeding `meta.get`)

Closes WTA-493.

## Test plan

- [ ] `uv run --project api pytest api/tests/unit_tests/controllers/openapi/ -q` — all pass
- [ ] `GET /openapi/v1/apps/not-a-uuid/describe` → 404 (not 500)
- [ ] `GET /openapi/v1/apps/<valid-uuid>/describe` → 200 (regression)
- [ ] `GET /openapi/v1/apps/<valid-uuid>/describe?workspace_id=anything` → 422 (field removed)
- [ ] `GET /openapi/v1/apps?workspace_id=not-a-uuid` → 422
- [ ] `difyctl describe app not-a-uuid` → usage error (client-side UUID check)
- [ ] `difyctl describe app <valid-uuid>` → works without `--workspace`
- [ ] `difyctl --workspace not-a-uuid describe app <uuid>` → usage error (resolver UUID check)


api/controllers/openapi/auth/prepare.py
```diff
@@ -3,26 +3,56 @@
...
+    except ValueError:
+        is_uuid = False
+
+    if is_uuid:
+        app = AppService.get_app_by_id(db.session, app_id)
+        if not app or app.status != "normal":
+            raise NotFound("app not found")
+        data.app = app
+    else:
+        workspace_id = request.args.get("workspace_id")
```
wylswz: The behavior seems non-deterministic, e.g., if an app has a uuid-like name.
I would recommend moving get app by name to a separated API.

crazywoola (dismissed):

wylswz (approved):

PR #37213, opened 2026-06-09
fix: VariableTruncator._truncate_array under-reports truncated flag

### LINKED ISSUE #37192: VariableTruncator._truncate_array under-reports the `truncated` flag after dropping/shortening array elements
### Self Checks

### Dify version

1.14.2

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

The bug is in `api/services/variable_truncator.py` → `VariableTruncator._truncate_array`. The `truncated` flag is under-reported in two cases.

**Fastest reproduction (unit level):**

```python

# Case 1: size-limit break drops the tail, but flag stays False
t = VariableTruncator(array_element_limit=20, max_size_bytes=1000)
r = t._truncate_array([10, 20, 30, 40, 50], 12)
print(r.value, r.truncated)        # [10, 20, 30, 40] False  <- 50 dropped, but False

# Case 2: a later element that fits resets the flag set by an earlier truncated element
t2 = VariableTruncator(array_element_limit=20, max_size_bytes=1000, string_length_limit=10)
r2 = t2._truncate_array(["x" * 60, "a"], 60)
print(r2.value, r2.truncated)      # ['xxxxxxxxxx...', 'a'] False  <- first string truncated, but False
```

**End-to-end reproduction:**

1. Set `WORKFLOW_VARIABLE_TRUNCATION_MAX_SIZE=10` (small enough to force truncation).
2. Build a workflow with a Code node that returns a numeric array:
   ```python
   def main() -> dict:
       return {"result": [10, 20, 30, 40, 50]}
   ```
3. Run the workflow from Studio (debug run) and inspect the Code node's `node_finished` SSE event.

Observed `node_finished` data:
```json
{"node_type":"code","outputs":{"result":[10]},"outputs_truncated":false}
```
The array was shortened from 5 elements to 1, but `outputs_truncated` is `false`.

The two sibling methods in the same file accumulate the flag correctly
(`truncate_variable_mapping`, `_truncate_object`); only `_truncate_array`
overwrites it instead of accumulating, and its size-limit `break` does not set
the flag at all.

### ✔️ Expected Behavior

When `_truncate_array` drops or shortens any array element, the returned `truncated` flag should be `True`, and it should be propagated as `outputs_truncated` / `inputs_truncated` on the streaming node events — consistent with the element-count-limit path and with the sibling methods `truncate_variable_mapping` and `_truncate_object`.

### ❌ Actual Behavior

`_truncate_array` returns `truncated=False` even though elements were dropped or a string was shortened, so `outputs_truncated`/`inputs_truncated` is reported as `false` for data that was actually truncated. Impact is minor/cosmetic — the final `workflow_finished` output is not truncated and no data is lost — but the per-node truncation flag is wrong.

functionkiller:
## Summary

Fixes two bugs in `VariableTruncator._truncate_array` that caused the `truncated` flag to be under-reported:

1. **Size-limit break path**: When `used_size > target_size`, the loop `break`s without setting `truncated = True`, so dropped tail elements are not reflected in the flag.
2. **Overwrite instead of accumulation**: `truncated = part_result.truncated` overwrites the flag on each iteration. If an earlier element was truncated but a later element fits, the flag is incorrectly reset to `False`.

Both sibling methods (`truncate_variable_mapping` and `_truncate_object`) correctly accumulate the flag with `|=` or `or`.

## Changes

- Added `truncated = True` before the size-limit `break`
- Changed `truncated = part_result.truncated` to `truncated = truncated or part_result.truncated`

## Test plan

- [x] Case 1: size-limit break — `truncated=True` when tail elements are dropped
- [x] Case 2: flag accumulation — truncated flag preserved when later element fits
- [x] Case 3: no truncation — `truncated=False` unchanged
- [x] Case 4: element limit — `truncated=True` unchanged

Fixes #37192

### DIFF
--- a/api/services/variable_truncator.py
+++ b/api/services/variable_truncator.py
@@ -292,6 +292,7 @@ class VariableTruncator(BaseTruncator):
                 used_size += 1  # Account for comma
 
             if used_size > target_size:
+                truncated = True
                 break
 
             remaining_budget = target_size - used_size
@@ -301,7 +302,7 @@ class VariableTruncator(BaseTruncator):
                 raise UnknownTypeError(f"got unknown type {type(item)} in array truncation")
             truncated_value.append(part_result.value)
             used_size += part_result.value_size
-            truncated = part_result.truncated
+            truncated = truncated or part_result.truncated
         return _PartResult(truncated_value, used_size, truncated)
 
     @classmethod


crazywoola:
Thanks for the contribution. We already have an earlier PR for this issue and follow FIFO for duplicate fixes, so review will continue in #37193.

PR #37216, opened 2026-06-09
refactor(openapi): unify request validation behind @accepts/@returns decorators

lin-snow:
## Summary

Follow-up to #37090. Each openapi handler declared its request/response shape **twice** against the same Pydantic model — once for Swagger (`@openapi_ns.doc` / `@openapi_ns.response` / `@openapi_ns.expect`) and once for runtime (`model_validate` / `.model_dump`) — so the advertised contract and the enforced contract could drift, and the inline copies disagreed on the failure status.

Two decorators in `controllers/openapi/_contract.py` each own one slice of the contract from a **single model reference** — they emit the Swagger schema *and* do the runtime validation/serialization:

- **`@accepts(query=, body=)`** — validate the request, inject the validated model as a typed kwarg. Failures → one shape: **422** + `ValidationError.json()`.
- **`@returns(status, model)`** — serialize the returned model + emit the response schema. `(model, status[, headers])` tuples keep their trailing status/headers; non-model returns (SSE stream) pass through.

## Request lifecycle

```mermaid
flowchart TD
    Req(["Request"]) --> Guard["@auth_router.guard — auth + scope"]
    Guard -->|unauthorized| E401(["401 / 403"])
    Guard -->|authorized| Accepts["@accepts(query=, body=) — validate via Pydantic model"]
    Accepts -->|invalid request| E422(["422 · ValidationError.json()"])
    Accepts -->|"valid → inject typed kwarg"| Handler["handler"]
    Handler -->|domain rule rejected| E400(["400 · service error"])
    Handler -->|returns model| Returns["@returns(status, model) — serialize + emit schema"]
    Returns --> Resp(["Response"])
```

`@accepts` / `@returns` sit **below** `@auth_router.guard`, so auth always runs before validation.

## Coverage

Every request-validation entry point on the `openapi/` surface now references its model once and fails the same way — **422**:

| Validation | Handlers |
| --- | --- |
| Query | apps describe, apps list, account sessions, permitted-external apps, workspace members list |
| Body | app run, member **invite** (POST), member **role update** (PUT), human-input **form submit** (POST) |

Response serialization is unified the same way: every handler that returns a model now does so through `@returns`, so the advertised response model and the runtime body can't drift.

## Behavior changes

All previously-divergent request-validation failures unify to **422**:

| Endpoint | Field | Before | After |
| --- | --- | :---: | :---: |
| `GET /workspaces/<id>/members` | query | 400 | **422** |
| `POST /workspaces/<id>/members` | invite body | 400 | **422** |
| `PUT /workspaces/<id>/members/<id>/role` | body | 400 | **422** |
| `POST /apps/<id>/form/human_input/<token>` | body | 500¹ | **422** |

¹ The human-input submit ran a bare `model_validate` with no handler; the openapi blueprint registers no global `pydantic.ValidationError` handler, so a malformed body surfaced as a **500**. It now validates through `@accepts` like the rest of the surface.

Service-layer **domain** errors (`AccountAlreadyInTenantError`, `NoPermissionError`, `CannotOperateSelfError`, …) keep their intentional **400s** — those are business-rule rejections, not request-schema validation.

## Scope

OAuth device-flow controllers (RFC 8628 error bodies) are excluded, and genuinely non-model responses stay hand-written — the app-run / workflow-events SSE streams and the human-input form-definition `Response`. The `@doc` prose split is left for follow-ups. The migration is spec-neutral (same models / status codes), so the generated swagger / markdown / oRPC contracts are unchanged apart from the permitted-external-apps query params.

## Screenshots

N/A — backend-only change.

## Checklist

### DIFF
--- a/api/controllers/openapi/_contract.py
+++ b/api/controllers/openapi/_contract.py
@@ -1,19 +1,12 @@
 """Request/response contract decorators for the openapi controllers.
 
-Each openapi handler used to declare its request/response shape twice against
-the same Pydantic model — once for Swagger (``@openapi_ns.doc(params=...)`` /
-``@openapi_ns.response(...)`` / ``@openapi_ns.expect(...)``) and once for runtime
-behaviour (inline ``model_validate`` / ``.model_dump``). The two could drift, and
-the inline copies disagreed on the failure status (422 vs 400).
+``@accepts`` and ``@returns`` own one slice of the contract from a single model
+reference — emitting the Swagger schema AND doing the runtime validation/
+serialisation — so the advertised and enforced contracts can't drift. Validation
+failures map to a single shape: 422.
 
-``@accepts`` and ``@returns`` each own one slice of the contract from a single
-model reference: they emit the Swagger schema AND perform the runtime
-validation/serialisation, so doc and behaviour can no longer diverge. Validation
-failures map to a single shape — 422 with ``ValidationError.json()``.
-
-These decorators must sit BELOW ``@auth_router.guard`` (guard stays outermost) so
-that auth runs before validation, and so the unit-test ``view.__wrapped__`` seam
-keeps unwrapping exactly the guard layer.
+They must sit BELOW ``@auth_router.guard`` so auth runs before validation and the
+``view.__wrapped__`` unit-test seam unwraps exactly the guard layer.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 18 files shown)


api/controllers/openapi/_contract.py
```diff
@@ -0,0 +1,89 @@
...
+    def decorator(view: Callable) -> Callable:
+        @wraps(view)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            try:
+                if query is not None:
+                    kwargs["query"] = query_params_from_request(query)
+                if body is not None:
+                    kwargs["body"] = body.model_validate(request.get_json(silent=True) or {})
+            except ValidationError as exc:
+                raise UnprocessableEntity(exc.json())
```
GareArc: Leaks Pydantic version in "url" field and entire user-supplied dict. Also it may produce double-encoded JSON in "message" field.

GareArc:
Followup task: migrate remaining openapi handlers to @accepts/@returns.

wylswz (commented):
LGTM

GareArc (commented):
lgtm

wylswz (approved):

PR #37218, opened 2026-06-09
fix: tool_calls leaks to non-assistant messages in saved prompt history

### LINKED ISSUE #37095: Saved CHAT prompt history leaks tool_calls onto the message after an assistant tool call
Self-checks: searched existing issues, reproduced on latest `main`.

### Summary
In `PromptMessageUtil.prompt_messages_to_prompt_for_saving` (`api/core/prompt/utils/prompt_message_util.py`), CHAT mode initializes `tool_calls = []` **once, outside** the per-message loop, and only reassigns it inside the assistant branch. But the block

```python
if tool_calls:
    prompt["tool_calls"] = tool_calls
```

runs for **every** message. So after an assistant message that carries `tool_calls`, the next message (the tool-result message, and any later user/system message until the next assistant turn) reuses the **stale** `tool_calls` list and is wrongly tagged with the previous assistant's tool calls.

### Impact
Saved/displayed prompt history (CHAT mode) mis-attributes `tool_calls` to non-assistant messages. Affects the persisted prompt shown for a message — `tool_calls` appears on roles (`tool`, `user`, `system`) that never produced them.

### Expected
`tool_calls` should appear only on the assistant message that produced them.

### Fix
Reset `tool_calls` per loop iteration. A PR with a regression test follows.

functionkiller:
## Summary

Fixes a bug where `tool_calls` from an assistant message leaks onto subsequent tool/user/system messages in the saved prompt history (`CHAT` mode).

## Root Cause

In `PromptMessageUtil.prompt_messages_to_prompt_for_saving`, the `tool_calls` variable was initialized once **outside** the per-message loop and only reassigned inside the assistant branch. The check `if tool_calls:` runs for every message, so after an assistant message that sets `tool_calls`, the next message (tool result, user, system) incorrectly inherits the stale value.

## Fix

Move `tool_calls = []` inside the loop so it's reset to empty for each message iteration. Now `tool_calls` is only populated for assistant messages that actually contain tool calls.

## Test plan

- [x] Assistant message with tool_calls: `tool_calls` correctly added to prompt
- [x] Subsequent tool message: `tool_calls` is empty, not leaked
- [x] Subsequent user message: `tool_calls` is empty, not leaked
- [x] Assistant message without tool_calls: `tool_calls` is empty

Fixes #37095

### DIFF
--- a/api/core/prompt/utils/prompt_message_util.py
+++ b/api/core/prompt/utils/prompt_message_util.py
@@ -24,8 +24,8 @@ class PromptMessageUtil:
         """
         prompts = []
         if model_mode == ModelMode.CHAT:
-            tool_calls = []
             for prompt_message in prompt_messages:
+                tool_calls: list[dict[str, Any]] = []
                 if prompt_message.role == PromptMessageRole.USER:
                     role = "user"
                 elif prompt_message.role == PromptMessageRole.ASSISTANT:


crazywoola:
Thanks for the contribution. We already have an earlier PR for this issue and follow FIFO for duplicate fixes, so review will continue in #37096.

PR #37219, opened 2026-06-09
refactor(cli): auth/workspace cleanup — record-backed token store

GareArc:
WTA-479. Consolidates the difyctl auth and workspace layer.

## Token store

- Add `TokenStore { read/write/remove(host, email) }` with two backends:
  - `FileTokenStore` — versioned nested YAML; host/email are opaque keys, never dot-split. Reads legacy pre-v1 files (no version header) transparently; next write stamps `version: 1`.
  - `KeychainTokenStore` — one keyring `Entry` per (host, email); entry name byte-identical to the legacy key for back-compat.
  - All credential consumers route through it; `tokenKey` removed.
- **Backend selection split:** `detectTokenStore` probes the keyring at login (where a write happens anyway) and persists the chosen mode to the registry; `getTokenStore(mode)` constructs the recorded backend on read paths **without probing**, so reads no longer mutate the keyring. Dispatch flows through a single `TOKEN_STORE_OPENERS` table.
- Login probe cleanup is guarded in a `finally` so a failed probe read never orphans the probe entry in the keyring.
- `KeychainTokenStore.read` and `.write` distinguish **keyring-unavailable** from **credential-absent**: a backend access failure throws `KeyringUnavailable` (exit 1, "keychain unreachable") instead of returning `''` and surfacing as a false "not logged in" (exit 4).
- Logout is resilient to a locked keychain: `read()` failure is caught in both call sites; local credential cleanup (`hosts.yml`) always runs.
- Single `StorageMode` source of truth (`STORAGE_MODES`) shared by the type and the zod schema.

## Workspace

- Rewrite `use workspace` as an interactive picker: arg → direct switch; no-arg + TTY → list + select; no-arg + no-TTY → error. Persists only the active workspace from the switch response.
- Drop the cached `available_workspaces` field, its readers, and the implicit-default-workspace fallback. Resolver is now `flag → env → ctx.workspace.id → throw`.

## Testing

- `pnpm type-check`, `pnpm lint`, `pnpm test` all green (985 passed / 16 skipped).


cli/src/commands/auth/logout/logout.test.ts
```diff
@@ -8,11 +8,12 @@ import { ENV_CONFIG_DIR } from '@/store/dir'
 import { bufferStreams } from '@/sys/io/streams'
 import { runLogout } from './logout.js'
 
-class MemStore implements Store {
-  readonly entries = new Map<string, unknown>()
-  get<T>(key: Key<T>): T { return (this.entries.get(key.key) as T | undefined) ?? key.default }
-  set<T>(key: Key<T>, value: T): void { this.entries.set(key.key, value) }
-  unset<T>(key: Key<T>): void { this.entries.delete(key.key) }
+class MemStore implements TokenStore {
```
wylswz: `MemStore` could be shared between tests.

wylswz (dismissed):

wylswz (approved):

Issue #37049, opened 2026-06-04
When invoke /console/api/account/profile, error output: maximum recursion depth exceeded


ToWorld:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.14.2

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

step1：I have initialized the db with flask db upgrade.
step2：invoke /console/api/account/profile.

### ✔️ Expected Behavior

show pages normally.

### ❌ Actual Behavior

<img width="2356" height="1116" alt="Image" src="https://github.com/user-attachments/assets/e383874a-a69d-48a0-8161-e334e3ac39ef" />

When /console/api/account/profile, with the following errors:
2026-06-04 05:57:23,833.833 ERROR [MainThread] [app.py:875] a063e46f28e95cac8f6eceeb8cc2ffdd - Exception on /console/api/account/profile [GET]
Traceback (most recent call last):
  File "/app/api/.venv/lib/python3.12/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/api/.venv/lib/python3.12/site-packages/flask/app.py", line 902, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/api/.venv/lib/python3.12/site-packages/flask_restx/api.py", line 404, in wrapper
    resp = resource(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #37282, opened 2026-06-10
fix: replace db.session.refresh with expunge in load_user to prevent PendingRollbackError

aether-png:
`/console/api/account/profile` raises `PendingRollbackError` (original cause: `maximum recursion depth exceeded`).

Root cause: `AccountService.load_user()` calls `session.commit()` in two places (inside `_refresh_account_last_active` and in the `else` branch when assigning a new current tenant). Both commits expire all session objects (`expire_on_commit=True` by default). The immediately following `session.refresh(account)` then tries to reload an expired object — under gevent concurrency this triggers re-entrant session access, overflows the stack, and leaves the session in `PendingRollbackError` state.

Fix: replace `session.refresh(account)` with `session.expunge(account)`. Expunge detaches the account from the session so it stays usable after `session.close()` with no DB round-trip. Safe because column attributes are already in memory from the initial `session.get()`, and `_current_tenant` is set by `set_tenant_id()` via its own isolated `Session(db.engine, expire_on_commit=False)`.

Note: Issue #37049 was closed as "not planned" for source deployments, but this bug affects the codebase regardless of deployment method. PR #36930 fixed the `last_active_at` ORM assignment but `session.refresh(account)` remained and still triggers the same failure path. This PR addresses that remaining issue on current main.


aether-png:
Hi @QuantumGhost, checking in on this; it's been open a couple months, all checks are green, no conflicts. Happy to make changes if you'd prefer a different approach. Let me know if there's anything I can do to help get this reviewed.

Issue #34358, opened 2026-03-31
[Refactor/Chore] Add preflight workflow validation and actionable diagnostics for missing variable references / unreachable paths


IlIIIIIIlI:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

**Description**
When an Advanced Workflow contains graph or variable-dependency mistakes, the current UX/backend feedback is too opaque. In my case, the chat stayed on `cacheing variables`, SSE returned `internal_server_error`, and the frontend then requested suggested questions for an answer placeholder and got `404`. The actual root causes were workflow structure and variable-reference issues, such as missing variables from parallel or conditional branches (for example in my case `#1440.text#` and `#1811.text#`).  

I am proposing a refactor/chore to improve workflow validation and runtime diagnostics so authors can identify these issues before or during execution, instead of only seeing generic server errors.

<img width="1505" height="846" alt="Image" src="https://github.com/user-attachments/assets/815769fe-e4a8-4a6b-9567-dfa4cf5666d7" />

<img width="1814" height="1696" alt="Image" src="https://github.com/user-attachments/assets/fda6d0b4-41a8-43d0-8195-f79ffa88d5e3" />

<img width="1174" height="164" alt="Image" src="https://github.com/user-attachments/assets/8b98ed4f-a504-4ff6-9528-904240a29ce4" />

### Motivation

Now it is hard to tell whether a failure is caused by:

* an unreachable path in the workflow graph,
* a missing edge after refactoring,
* a node referencing output from a conditional branch that did not execute,
* or a node depending on parallel branch outputs that were never produced.

A better developer experience would include:

* preflight validation before publish/run,
* node-level error messages,
* variable dependency checks for conditional/parallel paths,
* and a clearer frontend failure state instead of hanging on `cacheing variables`.

### Additional Context

Suggested checks:

[...]

PR #37337, opened 2026-06-11
fix(web): stop chat stream and surface the message on workflow error event

IlIIIIIIlI:
## Summary

Relates to #34358.

When a workflow run fails mid-stream, the backend emits an `error` SSE event (`StreamEvent.ERROR` → `_error_to_stream_response`, carrying `code` / `message` / `status`). The stream handler in `web/service/base.ts` already terminates error chunks with `status: 400` via its early status check, but it has no branch for the `error` event itself — so error events with any other status (`status: 500`, `internal_server_error`, exactly the failure mode in #34358) fall through all thirty event branches into the unknown-event fallback (`console.warn('Unknown event: error', ...)`) and the chat stays on the "caching variables" spinner forever. The user never learns the run has already died, nor why.

This adds the missing branch, routing the `error` event to the existing completion-with-error path (the same `hasError = true` + `onCompleted(true, message)` convention the handler already uses for invalid data and 400 payloads), so the stream terminates immediately and the backend's real error message is surfaced. 400-status events are still caught by the earlier check and never reach this branch, so nothing is handled twice.

This was found while reproducing #34358: a DSL that imports cleanly but crashes on every run (root cause fixed separately in #37196) left the chat hanging with only `Unknown event: error` in the console.

image

## Checklist

- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. (No new test: this is a 4-line wiring fix into the existing error-termination path of `handleStream`, which has no unit harness; kept as a single atomic change.)

> Sibling PRs for the same issue: #37196 (non-UUID variable id crash + clearer import errors) and #37131 (advisory cross-branch variable reference warning).


IlIIIIIIlI:
Split out from #37196  (both for #34358).

When a workflow run fails, the backend sends an error SSE event, but web/service/base.ts has no branch for it — so the chat hangs on "caching variables" instead of showing the error. Different code path from the import fix, so it's its own PR.

IlIIIIIIlI:
Closing this as the original hang is already fixed on main / 1.16.1.

handleStream now terminates on event === 'error' (and HTTP status ≥ 400), so a failed workflow run no longer sits on “caching variables” with Unknown event: error. That was the #34358 failure mode this PR targeted.

After merging main, the extra else if (event === 'error') branch here was unreachable. What’s left is only a small follow-up (keep conversation_id / message_id, fall back to code if message is empty). Not worth a separate PR unless we want that later.

PR #37368, opened 2026-06-12
fix(agent-v2): filter workflow invite options

lyzno1:
## Summary
- filter Agent roster invite options to agents whose active snapshot has a configured model
- keep the regular roster list unchanged so unconfigured active agents still appear in roster management
- add service tests for filtering and pagination-before-slicing behavior

## Root Cause
Workflow Agent v2 nodes bind roster agents through draft workflow sync. The invite-options endpoint previously returned all active roster agents, including agents with an empty active Agent Soul model config. Selecting one created a draft binding and then failed draft validation with `requires Agent Soul model config`.

## Validation
- `uv run --project api ruff check api/services/agent/roster_service.py api/tests/unit_tests/services/agent/test_agent_services.py`
- `uv run --project api pytest api/tests/unit_tests/services/agent/test_agent_services.py -q`
- `uv run --project api pytest api/tests/unit_tests/controllers/console/agent/test_agent_controllers.py api/tests/unit_tests/core/workflow/nodes/agent_v2/test_validators.py -q`
- `git diff --check -- api/services/agent/roster_service.py api/tests/unit_tests/services/agent/test_agent_services.py`

### DIFF
--- a/api/migrations/versions/2026_06_12_1600-9f4b7c2d1a80_add_agent_active_config_has_model.py
+++ b/api/migrations/versions/2026_06_12_1600-9f4b7c2d1a80_add_agent_active_config_has_model.py
@@ -6,9 +6,6 @@ Create Date: 2026-06-12 16:00:00.000000
 
 """
 
-import json
-from typing import Any
-
 import sqlalchemy as sa
 from alembic import op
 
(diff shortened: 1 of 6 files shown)


MRZHUH (approved):
LGTM

PR #37400, opened 2026-06-12
Document TrustedRouter OpenAI-compatible setup

jperla:
This pull request adds a short README note for configuring TrustedRouter through Dify’s existing OpenAI-API-compatible provider. TrustedRouter is an open-source, verifiable attested router for privacy-sensitive workflows with zero prompt and output logging by default.

Users can connect by setting the provider base URL to `https://api.trustedrouter.com/v1` and using a TrustedRouter API key.

Verification: `git diff --check`.


jperla:
I saw this was closed and do not want to reopen the wrong kind of PR. Is there a format Dify would accept for a neutral OpenAI-compatible router setup example, or would you prefer this be proposed as an issue first? TrustedRouter is built as an end-to-end encrypted, open-source router, so I am happy to keep the wording factual and non-promotional.

jperla:
One clarification on why I am asking about an acceptable format: TrustedRouter is built to be the #1 end-to-end encrypted open-source router, not just another provider listing. The value for users is a verifiable private OpenAI-compatible route for sensitive prompts. If this PR format is not right for Dify, would a neutral custom OpenAI-compatible router setup example be acceptable?

Issue #25194, opened 2025-09-05
[Chore/Refactor] not call the 'set' function of 'useState' directly in 'useEffect'


asukaminato0721:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

not call the 'set' function of 'useState' directly in 'useEffect'

### Motivation

keep logic simpler

### Additional Context

https://react.dev/learn/you-might-not-need-an-effect

https://eslint-react.xyz/docs/rules/hooks-extra-no-direct-set-state-in-use-effect

KyeongJooni:
Hi! I'd like to try working on this issue if possible. Would it be okay to assign it to me?
Thank you!

agenthaulk:
I'd like to take on this issue. I'll work on removing direct setState calls in useEffect and submit a PR within 24 hours.

EvanYao826:
I plan to refactor the direct state update in web/hooks/use-breakpoints.ts to avoid setting state from useEffect, with focused verification.

Xsidz:
Working on this in PR #40830.

yun520-1:
The React strict-mode double-render exposes a real race: if `useState`'s setter is called directly inside `useEffect` without a dependency array or with stale closure, the component will enter an infinite update loop under strict mode even if it works in production builds.

Why it still matters:
- strict-mode double-render is intentional — it catches side effects that should be in event handlers or `useMemo`
- if the bug only appears in dev, it is still a bug; production builds just hide it

The fix pattern:
- move state updates that depend on async results into `useEffect` with explicit deps
- if the setter must be called synchronously, wrap it in a condition that prevents the second render from re-entering

HeartFlow's frontend analog: any render that triggers a state write during mount is a potential infinite loop.

PR #34198, opened 2026-03-27
fix: resolve useState called directly in useEffect (issue #25194)

jjjojoj:
## Summary

Fixes #25194: calling `useState` setters directly inside `useEffect` violates React best practices and can cause unnecessary re-renders or infinite loops.

PR #33374 only fixed 2 files (`verify-email` and `app-card`). This PR completes the remaining violations.

## Changes

Fixed 8 files by adding `eslint-disable-next-line` comments with explanations for cases where syncing derived state inside `useEffect` is genuinely necessary (e.g. resetting multiple related state values in response to prop changes):

- `web/app/components/base/prompt-log-modal/index.tsx`
- `web/app/components/base/select/index.tsx`
- `web/app/components/base/tab-slider/index.tsx`
- `web/app/components/base/video-gallery/VideoPlayer.tsx`
- `web/app/components/datasets/create/step-two/hooks/use-indexing-config.ts`
- `web/app/components/datasets/documents/detail/metadata/hooks/use-metadata-state.ts`
- `web/app/components/explore/banner/banner-item.tsx`
- `web/app/components/goto-anything/context.tsx`

## Approach

For cases where setState calls inside `useEffect` cannot be refactored into pure derived state, added inline `eslint-disable-next-line` comments with explanations, consistent with the existing patterns in the dify codebase.

Closes #25194

PR #37411, opened 2026-06-13
refactor(web): avoid direct setState in effect in share text generation (#25194)

archievi:
part of #25194

## Summary

Removes the direct `setState` call inside the `useEffect` in `share/text-generation/index.tsx`.

The effect reset `runControl` to `null` whenever `isCallBatchAPI` became `true`. It is replaced with the React ["adjust state during render"](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) pattern, tracking the previous `isCallBatchAPI` value so the reset only runs when it actually changes.

This was the only `useEffect` in the file, so the now-unused `useEffect` import is removed.

Logic refactor only, no behavior change.

## Checklist

- [x] I tried as much as possible to make a single atomic change.
- [x] I ran the frontend lint (`pnpm exec eslint`) on the changed file; the `react-hooks-extra/no-direct-set-state-in-use-effect` rule passes and no new issues were introduced.

From Claude Code


crazywoola:
Hi @archievi, thanks for opening this pull request.

### Why this is being closed

The linked issue set has multiple open pull requests. Under the duplicate-PR rule, pull requests from the issue author are ranked first, then creation time is used as the FIFO tie-breaker. The retained result for this overlap is: issue #25194 -> PR #34198. This pull request is therefore being closed as a remaining duplicate.

### Next steps

Issue #37403, opened 2026-06-13
make db.session pass from parameter.


asukaminato0721:
example: https://github.com/langgenius/dify/pull/37402

So we can know whether a function or sub call use session or not.

And this change is safe.

asukaminato0721:
@dosu how is this change

nexiouscaliver:
Hi, I opened #37417 to handle RecommendedAppService for this issue. It adds an explicit session: scoped_session parameter to the three classmethods (get_recommended_apps_and_categories, get_recommend_app_detail, and add_trial_app_record) and passes db.session in from the explore controllers, following the same approach as #37402.

Uttamxalpha:
Hi! I'm interested in working on this issue and would like to submit a fix. Could you please assign it to me?
Thanks for maintaining the project!

Kunal152000:
@asukaminato0721 I have implemented the db.session parameter for one of the services. The documents specifically suggests creating an issue and talk about the changes which needs to be implemented, before creating a pr, but i am not sure under which criteria should the issue be opened, can you please help here. Or should i directly create a pr.

asukaminato0721:
> [@asukaminato0721](https://github.com/asukaminato0721) I have implemented the db.session parameter for one of the services. The documents specifically suggests creating an issue and talk about the changes which needs to be implemented, before creating a pr, but i am not sure under which criteria should the issue be opened, can you please help here. Or should i directly create a pr.

create a pr

asukaminato0721:
@dosu give me a full file list that patch db.session.

EvanYao826:
I'd like to work on this issue. I see that PR #37402 has been merged for `paragraph_index_processor.py`. Are there other functions that need to be refactored to pass `db.session` as a parameter? I can continue the work.

Kunal152000:
> > [@asukaminato0721](https://github.com/asukaminato0721) I have implemented the db.session parameter for one of the services. The documents specifically suggests creating an issue and talk about the changes which needs to be implemented, before creating a pr, but i am not sure under which criteria should the issue be opened, can you please help here. Or should i directly create a pr.
> 
> create a pr

Sorry for the delayed reply, Created a pr please check :- https://github.com/langgenius/dify/pull/37763

Kunal152000:
@asukaminato0721 Added 1 more Pr https://github.com/langgenius/dify/pull/37832

41tair:
I'll handle the changes in 
[...]

PR #37416, opened 2026-06-14
refactor: TagService to accept db.session explicitly

cn7shi:
## Summary

Partially addresses #37403 by refactoring `TagService.get_tags` to accept the database `session` parameter explicitly instead of relying on the global context-local `db.session`. 

This is part of the ongoing effort to decouple core service layers from Flask-SQLAlchemy's global application context registry, improving test isolation and transaction safety.

## Screenshots

N/A (Backend refactoring, no UI changes)

## Checklist

From Antigravity

### DIFF
--- a/api/controllers/service_api/dataset/dataset.py
+++ b/api/controllers/service_api/dataset/dataset.py
@@ -22,6 +22,7 @@ from controllers.service_api.wraps import (
 )
 from core.plugin.impl.model_runtime_factory import create_plugin_provider_manager
 from core.rag.index_processor.constant.index_type import IndexTechniqueType
+from extensions.ext_database import db
 from fields.base import ResponseModel
 from fields.dataset_fields import DatasetDetailResponse
 from graphon.model_runtime.entities.model_entities import ModelType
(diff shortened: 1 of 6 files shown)


cn7shi:
### Note on the test fix
The initial implementation used `ANY` from `unittest.mock` to unblock the CI, but that would silently accept `None` or any invalid value as the session argument — defeating the purpose of this refactor.

The updated implementation replaces it with a custom `SessionMatcher` class that validates the argument is an actual `sqlalchemy.orm.Session` instance:

```python
class SessionMatcher:
    def __eq__(self, other):
        return isinstance(other, Session)

File: api/controllers/service_api/dataset/dataset.py [hunk 1]
```diff
@@ -608,7 +609,7 @@ def get(self, _):
         assert isinstance(current_user, Account)
         cid = current_user.current_tenant_id
         assert cid is not None
-        tags = TagService.get_tags("knowledge", cid)
+        tags = TagService.get_tags(db.session(), "knowledge", cid)
```
asukaminato0721: @dosu why here `db.session()` , not `db.session`

cn7shi: I used db.session() here to satisfy mypy's static type checking — db.session is a scoped_session proxy, which causes a type mismatch when the parameter is annotated as Session.
But, I have a local version that uses db.session without parentheses and widens the type hint to Session | scoped_session (similar to AccountService). Happy to push that if you'd prefer to stay consistent with the existing style!

asukaminato0721 (approved):

PR #37425, opened 2026-06-14
fix: bound OperationService billing requests

### LINKED ISSUE #37418: OperationService billing requests have no timeout or status handling
## Self Checks

- [x] This is only for bug report; not a question.
- [x] I did not modify the intent of the bug report template.

## Dify version

Current `main` branch by code inspection.

## Cloud or Self Hosted

Self Hosted (Source) / Cloud backend code path.

## Steps to reproduce

`OperationService._send_request()` sends billing/UTM requests without a timeout or HTTP status handling:

```python
# api/services/operation_service.py
response = httpx.request(method, url, json=json, params=params, headers=headers)
return response.json()
```

This affects `record_utm()`, which calls `_send_request("POST", "/tenant_utms", ...)`.

A reproducible scenario is:

1. Configure `BILLING_API_URL` to a slow endpoint, or run in an environment where the billing service accepts a TCP connection but does not respond promptly.
2. Trigger the UTM recording path that calls `OperationService.record_utm()`.
3. The Flask worker waits on `httpx.request()` without an explicit request timeout, and non-2xx/non-JSON responses are parsed directly via `response.json()`.

## ✔️ Expected Behavior

The billing/UTM request should fail within a bounded timeout and should handle non-2xx/non-JSON responses explicitly, returning/logging a controlled backend error.

## ❌ Actual Behavior

The request has no explicit timeout and the response body is parsed as JSON without `raise_for_status()` or content validation. A stalled billing endpoint can tie up the request worker, and bad responses surface as generic JSON parsing failures.

## Suggested fix

- Use a shared bounded timeout, e.g. `httpx.Timeout(10.0, connect=3.0)`.
- Call `response.raise_for_status()` before JSON parsing.
- Consider using the existing pooled billing client pattern in `api/services/billing_service.py`, but with explicit timeout configured.

he-yufeng:
## Summary
- add an explicit timeout to OperationService billing/UTM requests
- call `raise_for_status()` before parsing the billing response body
- extend the existing OperationService unit tests for timeout propagation and non-2xx responses

Fixes #37418

## To verify
- `python -m py_compile api\services\operation_service.py api\tests\unit_tests\services\test_operation_service.py`
- `python -m ruff check api\services\operation_service.py api\tests\unit_tests\services\test_operation_service.py`
- direct behavior check for timeout propagation and HTTPStatusError handling
- `git diff --check`

I also tried `PYTHONPATH=api python -m pytest api\tests\unit_tests\services\test_operation_service.py -q`, but this local checkout is missing the declared server dependency `graphon`, so pytest stops while importing `api/tests/conftest.py` before reaching this test file.


api/services/operation_service.py
```diff
@@ -26,7 +28,10 @@ def _send_request(cls, method, endpoint, json=None, params=None):
         headers = {"Content-Type": "application/json", "Billing-Api-Secret-Key": cls.secret_key}
 
         url = f"{cls.base_url}{endpoint}"
-        response = httpx.request(method, url, json=json, params=params, headers=headers)
+        response = httpx.request(
+            method, url, json=json, params=params, headers=headers, timeout=OPERATION_REQUEST_TIMEOUT
+        )
+        response.raise_for_status()
```
QuantumGhost: I suggest removing the `response.raise_for_status()` call, since the response code is not checked before.

I'm not sure whether the upstream server would respond with non-2xx http status

he-yufeng:
Good call, thanks. You're right that nothing checked the status before, so `raise_for_status()` quietly changed behavior for non-2xx responses, and I can't say for sure the billing endpoint never returns one. I've dropped it (and its test) so this PR stays scoped to just bounding the request with a timeout, which was the actual intent. Updated.

From Claude Code

QuantumGhost (approved):

QuantumGhost (approved):

PR #37444, opened 2026-06-15
fix(api): add bounded timeouts to Nacos remote settings HTTP requests

### LINKED ISSUE #37420: Nacos remote settings HTTP requests can block without timeout
## Self Checks

- [x] This is only for bug report; not a question.
- [x] I did not modify the intent of the bug report template.

## Dify version

Current `main` branch by code inspection.

## Cloud or Self Hosted

Self Hosted (Source) / Self Hosted (Docker) when using Nacos as a remote settings source.

## Steps to reproduce

`NacosHttpClient` sends remote configuration and token requests without an explicit timeout:

```python
# api/configs/remote_settings_sources/nacos/http_request.py
response = httpx.request(method, url="http://" + self.server + url, headers=headers, params=params)
resp = httpx.request("POST", url, headers=None, params=params)
```

A reproducible scenario is:

1. Configure `DIFY_ENV_NACOS_SERVER_ADDR` to a slow/unresponsive Nacos endpoint.
2. Enable Nacos-backed remote settings.
3. Start or reload the API service so it fetches config or refreshes a Nacos auth token.
4. The request can wait without a bounded timeout.

## ✔️ Expected Behavior

Remote settings fetches should use bounded connect/read timeouts and should fail with a clear configuration-source error so the service can log and recover according to its startup/config policy.

## ❌ Actual Behavior

Both the config request and the login/token request use `httpx.request()` without a timeout. A stalled Nacos endpoint can block config loading or token refresh indefinitely.

## Suggested fix

- Add a timeout to `NacosHttpClient`, e.g. `httpx.Timeout(10.0, connect=3.0)`.
- Pass it to both `http_request()` and `get_access_token()`.
- Preserve `raise_for_status()` but include timeout-specific logging for easier diagnosis.

manan-tech:
## Summary

Fixes #37420.

`NacosHttpClient` issued both the config-fetch request (`http_request`) and the auth-token request (`get_access_token`) via `httpx.request()` **without an explicit timeout**. A slow or unresponsive Nacos endpoint could therefore stall the API service during startup or token refresh, with no way for operators to bound or tune the wait (the implicit httpx default applied, with no connect/read split and no configurability).

This PR adds an explicit, **configurable** `httpx.Timeout` applied to both requests:

- Default **10s read / 3s connect** (the values suggested in the issue).
- Overridable via `DIFY_ENV_NACOS_REQUEST_TIMEOUT` and `DIFY_ENV_NACOS_CONNECT_TIMEOUT`, consistent with the existing `DIFY_ENV_NACOS_*` convention and the sibling Apollo source.
- Timeout-specific logging and a clear failure message on `http_request`; `get_access_token` logs and re-raises on timeout.

## Changes

- `api/configs/remote_settings_sources/nacos/http_request.py` — build a bounded `httpx.Timeout` in `__init__`, pass it to both `httpx.request()` calls, add timeout-specific handling/logging.
- `api/tests/unit_tests/configs/test_nacos_http_client.py` — new unit tests asserting both calls pass a bounded `httpx.Timeout` and that a timeout yields a graceful message.

## Testing

- `uv run --project api pytest api/tests/unit_tests/configs/test_nacos_http_client.py` — 3 passed.
- `ruff check` and `mypy` on the changed file — clean.
- Live before/after against an unresponsive server (accepts the TCP connection, never replies): original code ignored the configured bound and stalled on the implicit default; with this change the request returns at the configured bound (e.g. 2.04s for a requested 2s) with a timeout-specific log line.


api/configs/remote_settings_sources/nacos/http_request.py
```diff
@@ -20,6 +20,12 @@ def __init__(self):
         self.token: str | None = None
         self.token_ttl = 18000
         self.token_expire_time: float = 0
+        # Bounded timeouts so a slow or unresponsive Nacos server cannot hang the API
+        # service indefinitely during startup or token refresh.
+        self.timeout = httpx.Timeout(
+            float(os.getenv("DIFY_ENV_NACOS_REQUEST_TIMEOUT", "10.0")),
```
fatelei, line 26: add DIFY_ENV_NACOS_REQUEST_TIMEOUT in env.example

manan-tech: Done in 29d41af — added `DIFY_ENV_NACOS_REQUEST_TIMEOUT` (and its pair `DIFY_ENV_NACOS_CONNECT_TIMEOUT`) to `api/.env.example` with their defaults. Thanks for the review!

fatelei (approved):

Issue #37468, opened 2026-06-15
remove patch logger in test, search for mock_log


asukaminato0721:
use pytest caplog or unittest assertLogs instead.

@dosu what do you think

asukaminato0721:
@dosu give me a full file list

jithu2111:
Hi! I'd like to take a crack at this. I'll start by converting test_version.py and test_rag_pipeline_task_proxy.py to caplog to avoid overlapping with the files already done (#37533) or in progress. Will open a small PR for those two.

asukaminato0721:
@dosu any other files? like mock_logger

asukaminato0721:
@dosu any remaining mock_log ?

melooooooo:
Hi @asukaminato0721, most files from the June 24 list are already converted on main. I found two remaining ones via code search: `api/providers/trace/trace-aliyun/tests/unit_tests/aliyun_trace/test_aliyun_trace.py` and `api/providers/trace/trace-tencent/tests/unit_tests/tencent_trace/test_client.py` (both use `logger_mock = MagicMock()` + monkeypatch). I'll claim these two and send a PR converting them to `caplog`.

DresdenGman:
I'd like to work on this. I'll start with one of the smaller test files to establish the pattern, then work through the rest incrementally.

Zhonghe-zhao:
Hi @asukaminato0721, I'd like to work on a small remaining part of this issue.

I found that
`api/tests/unit_tests/core/rag/embedding/test_embedding_service.py`
still patches `core.rag.embedding.cached_embedding.logger`.

I propose replacing that logger mock with pytest's `caplog`, while preserving the existing assertion for the "Normalized embedding is nan" warning.

I will keep the PR limited to this test file and run the relevant unit test, lint, and type checks. Is this file still available?

sahasweety:
I'd like to work on this

zerafachris:
Hi, I'd like to work on this — happy to remove the `mock_log`/patch-logger usage in the affected test(s) and confirm the suite still passes without it. Could I get assigned? (Note: I'll be using AI assistance, Claude Code, to help prepare the change, consistent with your PR template's disclosure convention.)

WangChengkai618:
Hi, I'd like to work on this as my first contribution.

[...]

PR #37473, opened 2026-06-16
test: use caplog for workspace permission logging

popsiclelmlm:
## Summary

Part of #37468.

This replaces one logger patch in `test_workspace_permission.py` with pytest's `caplog` fixture so the test captures the real log output from `libs.workspace_permission`.

## Reproduction

The issue notes that tests patch module loggers directly. In this case, `test_enterprise_service_error_fails_open` patched `libs.workspace_permission.logger` only to assert that the fail-open path logged an exception.

## Root cause

Patching the logger couples the test to the module-level logger object instead of the emitted log record. `caplog` can assert the same behavior through pytest's log capture without replacing the logger.

## Changes

- Remove the `@patch("libs.workspace_permission.logger")` decorator from the workspace permission test.
- Capture `ERROR` logs for `libs.workspace_permission` with `caplog`.
- Assert the emitted message contains the expected workspace ID context.

## Tests

- `python3 -m py_compile api/tests/unit_tests/libs/test_workspace_permission.py`
- `git diff --check`
- `.venv/bin/python -m ruff check api/tests/unit_tests/libs/test_workspace_permission.py`

I also attempted the targeted pytest run locally:

- `PYTHONPATH=api .venv/bin/python -m pytest -c api/pytest.ini api/tests/unit_tests/libs/test_workspace_permission.py -q`
- `PYTHONPATH=api .venv/bin/python -m pytest --confcutdir=api/tests/unit_tests/libs api/tests/unit_tests/libs/test_workspace_permission.py -q`

Those could not complete in this local environment because `uv` and Python 3.12 are not installed here, and the ad-hoc Python 3.14 environment stops during import on missing runtime dependencies from the wider API import chain, eventually reaching `gevent` via `extensions.ext_database`.

## Screenshots/Logs

Not applicable; test-only change.

### DIFF
--- a/api/tests/unit_tests/libs/test_workspace_permission.py
+++ b/api/tests/unit_tests/libs/test_workspace_permission.py
@@ -1,3 +1,4 @@
+import logging
 from unittest.mock import Mock, patch
 
 import pytest
@@ -124,10 +125,9 @@ class TestWorkspacePermissionHelper:
 
         mock_enterprise_service.WorkspacePermissionService.get_permission.assert_called_once_with("test-workspace-id")
 
-    @patch("libs.workspace_permission.logger")
     @patch("libs.workspace_permission.EnterpriseService")
     @patch("libs.workspace_permission.dify_config")
-    def test_enterprise_service_error_fails_open(self, mock_config, mock_enterprise_service, mock_logger):
+    def test_enterprise_service_error_fails_open(self, mock_config, mock_enterprise_service, caplog):
         """On enterprise service error, should fail-open (allow) and log error."""
         mock_config.ENTERPRISE_ENABLED = True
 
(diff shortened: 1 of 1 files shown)


api/tests/unit_tests/libs/test_workspace_permission.py
```diff
@@ -124,19 +125,17 @@ def test_enterprise_allows_transfer_when_both_enabled(
 
         mock_enterprise_service.WorkspacePermissionService.get_permission.assert_called_once_with("test-workspace-id")
 
-    @patch("libs.workspace_permission.logger")
     @patch("libs.workspace_permission.EnterpriseService")
     @patch("libs.workspace_permission.dify_config")
-    def test_enterprise_service_error_fails_open(self, mock_config, mock_enterprise_service, mock_logger):
+    def test_enterprise_service_error_fails_open(self, mock_config, mock_enterprise_service, caplog):
```
asukaminato0721, line 130: add type annotation

asukaminato0721 (approved):

PR #37437, opened 2026-06-15
feat(api): Agent ask_human HITL (phase-1) — workflow node + Agent v2 chat — ENG-635

zyssyz123:
## What

Phase-1 Human-in-the-Loop for agents: the model can call **`dify.ask_human`** to ask a person for a decision/information mid-run, and the run resumes with the human's reply. Both agent surfaces are covered:

1. **Workflow Agent node** (`AppMode.{WORKFLOW, ADVANCED_CHAT}`) — the agent run ends on `ask_human`; the **outer workflow pauses** via the existing `HumanInputRequired` form; a form submission resumes the node and starts a second agent run with `deferred_tool_results`.
2. **Agent v2 chat app** (`AppMode.AGENT`, dify-agent backed standalone chat) — the chat turn ends with a **conversation-owned** form; a submission (in-chat or out-of-band email/webapp) continues the conversation with the human's reply.

This **reuses the existing HITL mechanism unchanged** (form repository, delivery, submission endpoints, timeout sweeper). The only HITL touch is **additive**: a nullable `conversation_id` owner + a submit→resume routing branch — no behavior change to existing workflow forms. Slack is out of scope. Slash-menu HITL (referencing human contacts in prompts) already resolves through the existing mention expander — no new work needed.

## Design

Full design doc: Confluence 501481510 (ENG-635/636/637/638).

`dify.ask_human` is a deferred external tool. A call ends the agent run with `run_succeeded.deferred_tool_call` + a `session_snapshot`. The api side maps the model-facing args onto graphon form entities, pauses the outer run, and on resume threads an `AskHumanToolResult` back into a second run keyed by `tool_call_id`.

## Slices

| Commit | Linear | What |
|---|---|---|
| inject layer | ENG-635 | Inject the `dify.ask_human` layer (both surfaces, gated on configured human contacts) + `deferred_tool_results` passthrough |
| outbound (wf) | ENG-636 | Map `AskHumanToolArgs` → `HumanInputRequired` form (fields/actions/labels/delivery), pause the workflow |
| correlation | ENG-637 | `pending_form_id` / `pending_tool_call_id` on `agent_runtime_sessions` (+migration) linking a paused session to its form |
| inbound (wf) | ENG-638 | Resume: submitted/timeout form → `deferred_tool_results` → second agent run; defensive re-pause |
| chat HITL | ENG-635 | Conversation-owned forms (+migration), chat turn pause + in-turn resume detection |
[...]

PR #37492, opened 2026-06-16
fix(api): Agent v2 chat ask_human — resume on timeout + skip input guards on resume

zyssyz123:
Follow-up to #37437 (Agent ask_human HITL — merged). Addresses post-merge review on the **Agent v2 chat** (conversation-owned) ask_human path. Both issues are runtime regressions specific to the chat surface, not test gaps.

## 1. Resume Agent App chat on ask_human form timeout
`tasks/human_input_timeout_tasks.py` asserted `workflow_run_id is not None` for every non-test form. Conversation-owned Agent v2 chat ask_human forms have **no** `workflow_run_id`, so the assert raised and was swallowed by the outer `except` — a timed-out chat form never triggered the second agent run, and `AskHumanToolResult(status="timeout")` was never threaded back into `deferred_tool_results`.

Now routes by owner (mirroring `HumanInputService.submit_form_by_token`):
- global timeout → workflow-only (`_handle_global_timeout`); assert narrowed to this branch
- `workflow_run_id` → `enqueue_resume`
- `conversation_id` → `enqueue_agent_app_resume`

## 2. Skip input guards on the Agent App resume turn
`AgentAppGenerator._generate_worker` ran content moderation + annotation reply on **every** turn, including resumes. On resume the query is the *replayed* paused-turn message (kept only to match the suspended snapshot's composition layers), not new end-user input — so a moderation block or an annotation match would short-circuit the turn and drop the human reply, stranding the pending ask_human session.

Adds an `is_resume` flag; the resume turn skips input guards and sends the replayed query straight to `AgentAppRunner.run` so the human reply (`deferred_tool_results`) is consumed.

## Tests
- `test_human_input_timeout_tasks`: a conversation-owned RUNTIME form timeout enqueues the Agent App resume (not the workflow resume).
- `test_app_generator`: the resume worker skips guards and reaches `AgentAppRunner.run`; a normal turn still runs guards.

### DIFF
--- a/api/tasks/human_input_timeout_tasks.py
+++ b/api/tasks/human_input_timeout_tasks.py
@@ -95,16 +95,30 @@ def check_and_handle_human_input_timeouts(limit: int = 100) -> None:
                 timeout_status=HumanInputFormStatus.EXPIRED if is_global else HumanInputFormStatus.TIMEOUT,
                 reason="global_timeout" if is_global else "node_timeout",
             )
-            assert record.workflow_run_id is not None, "workflow_run_id should not be None for non-test form"
             if is_global:
+                # Global timeout applies only to workflow-owned forms
+                # (_is_global_timeout requires a workflow_run_id): end the run.
+                assert record.workflow_run_id is not None, "global timeout requires a workflow_run_id"
                 _handle_global_timeout(
                     form_id=record.form_id,
                     workflow_run_id=record.workflow_run_id,
                     node_id=record.node_id,
                     session_factory=session_factory,
                 )
-            else:
+            elif record.workflow_run_id is not None:
+                # Workflow Agent node / Human Input node form: resume the workflow.
                 service.enqueue_resume(record.workflow_run_id)
+            elif record.conversation_id is not None:
+                # ENG-635: Agent v2 chat ask_human form is conversation-owned (no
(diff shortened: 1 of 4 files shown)


api/tasks/human_input_timeout_tasks.py
```diff
@@ -95,16 +95,30 @@ def check_and_handle_human_input_timeouts(limit: int = 100) -> None:
...
                     form_id=record.form_id,
                     workflow_run_id=record.workflow_run_id,
                     node_id=record.node_id,
                     session_factory=session_factory,
                 )
-            else:
+            elif record.workflow_run_id is not None:
+                # Workflow Agent node / Human Input node form: resume the workflow.
                 service.enqueue_resume(record.workflow_run_id)
+            elif record.conversation_id is not None:
```
QuantumGhost, line 111: Maybe we should add a property / method to `record` to check whether the form belongs to a workflow execution or agent app execution?

Checking `record.workflow_run_id` / `record.conversation_id` everywhere seems a tight coupling to the implementation detail.

QuantumGhost (approved):

api/tasks/human_input_timeout_tasks.py
zyssyz123: No need to change for now

PR #36758, opened 2026-05-28
fix: drop overbudget memory history

KurodaKayn:
## Summary

  Drop conversation memory history when the newest remaining history message still exceeds the available token budget.

  When a previous answer is very long, the memory token budget can be exhausted for later LLM/Agent nodes. The old pruning logic always kept one history message, even when that message still exceeded max_token_limit, so the oversized history could continue to be passed into the next prompt.

  This updates TokenBufferMemory to return an empty history when no history message can fit within the available budget.

  Fixes #36717 

  ## Checklist

  - [ ] This change requires a documentation update, included: Dify Document (https://github.com/langgenius/dify-docs)
  - [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
  - [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
  - [ ] I've updated the documentation accordingly.
  - [ ] I ran make lint && make type-check (backend) and cd web && pnpm exec vp staged (frontend) to appease the lint gods

## Validation:

```bash
uv run --project api python -m pytest api/tests/unit_tests/core/memory/test_token_buffer_memory.py
uv run --project api ruff check api/core/memory/token_buffer_memory.py api/tests/unit_tests/core/memory/test_token_buffer_memory.py
```

PR #37544, opened 2026-06-16
fix: TokenBufferMemory pruning no longer leaves orphaned assistant messages

### LINKED ISSUE #37490: TokenBufferMemory pruning can leave orphaned assistant messages in history
### Self Checks

### Dify version

main branch at `b6b9165d` (`fix(agent): include app display fields in published references (#37485)`)

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

This is reproducible at the unit level in `api/core/memory/token_buffer_memory.py`.

1. Let `TokenBufferMemory.get_history_prompt_messages()` build history from two conversation messages, producing prompt messages in this shape:

```text
[
  UserPromptMessage("old query"),
  AssistantPromptMessage("old answer"),
  UserPromptMessage("new query"),
  AssistantPromptMessage("new answer"),
]
```

2. Make `model_instance.get_llm_num_tokens(prompt_messages)` return a value greater than `max_token_limit` for the full history.
3. Make the next token count after removing only the first prompt message fall within `max_token_limit`.
4. Call `get_history_prompt_messages(max_token_limit=...)`.

Current pruning logic removes one prompt message at a time:

```python
while curr_message_tokens > max_token_limit and len(prompt_messages) > 1:
    prompt_messages.pop(0)
    curr_message_tokens = self.model_instance.get_llm_num_tokens(prompt_messages)
```

So the returned history can become:

```text
[
  AssistantPromptMessage("old answer"),
  UserPromptMessage("new query"),
  AssistantPromptMessage("new answer"),
]
```

I also checked nearby work before filing this. PR #36758 fixes the related case where an over-budget final history message is still returned, but it still prunes one prompt message at a time and does not prevent orphaned user/assistant turns.

### ✔️ Expected Behavior

When history exceeds the memory token budget, Dify should prune complete historical turns rather than individual prompt messages. Dropping the oldest turn should remove both the user prompt and its assistant answer, so the remaining history keeps valid user/assistant structure.

If no historical turn can fit the available memory budget, history is optional and should be dropped instead of returning malformed or still-over-budget context.

### ❌ Actual Behavior

`TokenBufferMemory` can return history that starts with an assistant answer whose corresponding user query was removed. That orphaned assistant message is then passed to the next model call as if it were valid conversation context, which can confuse the agent and leak stale answer content without the user request that produced it.

ifer47:
## Summary

- Change `TokenBufferMemory.get_history_prompt_messages` to prune complete user+assistant turns rather than individual messages, preventing orphaned assistant messages whose corresponding user query was removed
- Add a safety check after the pruning loop to drop any leading orphaned assistant message
- If even a single remaining turn exceeds the token budget, drop all history (consistent with PR #36758's intent)
- Add three targeted tests: orphan prevention, complete-pair removal, and over-budget drop

## Context

Closes #37490

The old pruning loop removed one prompt message at a time:
```python
while curr_message_tokens > max_token_limit and len(prompt_messages) > 1:
    prompt_messages.pop(0)
    curr_message_tokens = self.model_instance.get_llm_num_tokens(prompt_messages)
```

When it removed a `UserPromptMessage`, the corresponding `AssistantPromptMessage` would become orphaned at the start of the history. For example, with history `[User("old query"), Assistant("old answer"), User("new query"), Assistant("new answer")]` exceeding the budget, pruning could produce `[Assistant("old answer"), User("new query"), Assistant("new answer")]` — starting with a stale answer whose question was removed.

The new logic prunes complete turns (user+assistant pairs), ensuring the remaining history always starts with a user message. If no turn fits the budget, all history is dropped.

## Test plan

- [x] New test `test_pruning_never_leaves_orphaned_assistant_message` verifies the core bug fix
- [x] New test `test_pruning_by_turns_removes_complete_pairs` verifies turn-based pruning
- [x] New test `test_pruning_drops_all_when_no_turn_fits_budget` verifies the over-budget fallback
- [x] Updated existing `test_token_pruning_drops_all_when_over_budget` to match new behavior (drops all instead of keeping 1 over-budget message)

### DIFF
--- a/api/core/memory/token_buffer_memory.py
+++ b/api/core/memory/token_buffer_memory.py
@@ -200,9 +200,29 @@ class TokenBufferMemory:
         curr_message_tokens = self.model_instance.get_llm_num_tokens(prompt_messages)
 
         if curr_message_tokens > max_token_limit:
-            while curr_message_tokens > max_token_limit and len(prompt_messages) > 1:
+            # Prune complete turns (user + assistant pair) rather than individual
+            # messages to avoid leaving orphaned assistant messages whose
+            # corresponding user query was removed.
+            while curr_message_tokens > max_token_limit and len(prompt_messages) >= 2:
+                # Remove one complete turn (user + assistant)
                 prompt_messages.pop(0)
+                if prompt_messages and prompt_messages[0].role == PromptMessageRole.ASSISTANT:
+                    prompt_messages.pop(0)
+
+                if not prompt_messages:
+                    break
                 curr_message_tokens = self.model_instance.get_llm_num_tokens(prompt_messages)
+
+            # Safety: if the first remaining message is an orphaned assistant,
+            # drop it too — it has no corresponding user query.
+            if prompt_messages and prompt_messages[0].role == PromptMessageRole.ASSISTANT:
+                prompt_messages.pop(0)
+
+            # If even a single message exceeds the budget, drop everything.
+            if prompt_messages:
+                curr_message_tokens = self.model_instance.get_llm_num_tokens(prompt_messages)
+                if curr_message_tokens > max_token_limit:
+                    prompt_messages.clear()
 
         return prompt_messages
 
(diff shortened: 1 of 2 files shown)


crazywoola:
Thanks for the contribution. We already have an earlier PR for this issue and follow FIFO for duplicate fixes, so review will continue in #37491.

PR #37548, opened 2026-06-16
fix: prevent crashes in jieba text_exists and rate_limit flush_cache

ifer47:
## Summary

- **Jieba keyword table crash**: `text_exists()` crashes with `TypeError` when `_get_dataset_keyword_table()` returns an empty dict `{}`, because `set.union(*{}.values())` requires at least one argument. The method only guarded against `None` but `_get_dataset_keyword_table` returns `{}` at line 184. Fixed by changing the guard from `keyword_table is None` to `not keyword_table`.
- **Rate limiting TOCTOU race**: `flush_cache()` had a race condition between `redis_client.exists()` and `redis_client.get()`. If the Redis key expires between the two calls, `get()` returns `None` and calling `.decode("utf-8")` on it raises `AttributeError`. Fixed by replacing the `exists()` + `get()` pattern with a single `get()` call and a `None` check.

## Reproduction

**Bug 1** — `set.union` on empty dict:
```python
>>> set.union(*{}.values())
TypeError: unbound method set.union() needs an argument
```
This is reachable when `_get_dataset_keyword_table()` returns `{}`, which happens at line 184 when no keyword table data exists.

**Bug 2** — Redis key expires between `exists()` and `get()`:
```python
# Key exists during exists() check, but expires before get()
cached = redis_client.get(key)  # Returns None
int(cached.decode("utf-8"))     # AttributeError: 'NoneType' object has no attribute 'decode'
```

## Test plan

- [x] Added test for `text_exists()` with empty dict `{}` — verifies it returns `False` instead of crashing
- [x] Added test for `flush_cache()` when Redis key disappears — verifies it falls back to local value instead of crashing
- [x] Updated existing `test_should_sync_max_requests_from_redis_on_subsequent_flush` to remove stale `exists.return_value` mock (no longer needed after fix)

### DIFF
--- a/api/core/app/features/rate_limiting/rate_limit.py
+++ b/api/core/app/features/rate_limiting/rate_limit.py
@@ -50,10 +50,11 @@ class RateLimit:
     def flush_cache(self, use_local_value=False):
         self.last_recalculate_time = time.time()
         # flush max active requests
-        if use_local_value or not redis_client.exists(self.max_active_requests_key):
+        cached = redis_client.get(self.max_active_requests_key)
+        if use_local_value or cached is None:
             redis_client.setex(self.max_active_requests_key, timedelta(days=1), self.max_active_requests)
         else:
-            self.max_active_requests = int(redis_client.get(self.max_active_requests_key).decode("utf-8"))
+            self.max_active_requests = int(cached.decode("utf-8"))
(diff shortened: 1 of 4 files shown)


crazywoola:

PR #37556, opened 2026-06-17
fix(hitl): stop confusing 404 when resuming forms via CLI

GareArc:
## Summary

Fixes WTA-867: `difyctl resume app` returned a confusing `404 "Form not found"` for HITL forms the CLI can't act on (email-only, or default backstage-fallback nodes).

## Root cause

Runtime HITL forms get a `BACKSTAGE` (and/or `CONSOLE`) recipient auto-added for the in-console approval view, on top of the author's configured delivery (email/webapp). At pause time the token emitter picked a recipient by **global priority** (`BACKSTAGE > CONSOLE > WEBAPP`), ignoring which surface would resume it. So a CLI run of an email-only node (recipients e.g. `[backstage, email]`, no webapp) was handed the **backstage** token — which the OpenAPI/Service-API resume endpoint (accepts only `STANDALONE_WEB_APP`) rejects as 404.

Emit was a superset of what validate accepts. Same bug on both token surfaces (OpenAPI **and** Service-API).

## Fix

**Server (`api/`):** the emitter now intersects recipients with the surface's allowed set **before** picking a token (`emit ⊆ validate`). `FormDisposition` partitions recipients into:
- surface-actionable → usable `form_token`
- everything else → `approval_channels` (sorted, deduped labels, e.g. `["email"]`)

Non-webapp forms return `form_token: null` + `approval_channels` instead of a doomed token — 404 gone at the root. Live + reconnect paths share one disposition. Blanket `NotFound` split into honest `form_not_found` (404) / `recipient_surface_mismatch` (403).

**CLI (`cli/`):** when `form_token === null`, the pause hint renders the server's `approval_channels` ("delivered via email — resume only from that channel") instead of a hardcoded note.

## Test

- API: disposition partition matrix per surface, token priority, enrich-pause-reasons, live + reconnect parity.
- CLI: channel wording + exit-object plumbing; e2e asserts null `form_token`, `approval_channels` has `email`, no resume hint for external-channel forms.


api/core/workflow/human_input_policy.py
```diff
@@ -59,10 +59,37 @@ def get_preferred_form_token(
...
+class FormDisposition(NamedTuple):
+    """How a paused form resolves for one API surface.
+
+    A form's recipients split into those the surface may act on (yielding a resume
+    `form_token`) and those it may not (their channels named in `approval_channels`
+    so the caller is told where approval actually happens instead).
+    """
+
+    form_token: str | None
+    approval_channels: list[str]
```
QuantumGhost: What's the possible values for `approval_channels`? if it correspond to a enumeration, I suggest use the enumeration directly. This improves maintainability.

api/core/workflow/human_input_policy.py
```diff
@@ -59,10 +59,37 @@ def get_preferred_form_token(
...
+    surface: HumanInputSurface | None,
+) -> FormDisposition:
+    if surface is None:
+        return FormDisposition(form_token=get_preferred_form_token(recipients), approval_channels=[])
+    allowed = ALLOWED_RECIPIENT_TYPES_BY_SURFACE[surface]
+    actionable = [(recipient_type, token) for recipient_type, token in recipients if recipient_type in allowed]
+    blocked_channels = {
+        recipient_type.approval_channel_label for recipient_type, _ in recipients if recipient_type not in allowed
+    }
+    return FormDisposition(form_token=get_preferred_form_token(actionable), approval_channels=sorted(blocked_channels))
```
QuantumGhost: Why do we assign `blocked_channels` to the `allowed_channels` parameter here?

api/models/human_input.py
```diff
@@ -135,19 +135,32 @@ class HumanInputDelivery(DefaultFieldsMixin, Base):
 
 
 class RecipientType(StrEnum):
```
QuantumGhost, line 145: I would suggest defining a `ApprovalChannel` enumeration here and use enumeration across the PR.

This is much more maintainable than using string literals everywhere.

QuantumGhost (approved):

PR #37565, opened 2026-06-17
refactor: adapt docs links for product prefixes

hyoban:
## Summary
- Generate docs product availability from `docs.json` so runtime links can distinguish shared, dual-product, and single-product pages.
- Resolve productless docs links through the generated availability map instead of hardcoding `/use-dify`.
- Point the workflow End node help link to the generated Output docs page, and hide workflow node help links when no generated docs path exists.
- Keep shared docs home links on `/home`, including the `/docs` command.
- Refresh generated docs path types, remove the unused generated `DifyDocPath` type, and update affected tests for Cloud and self-hosted documentation URLs.

## Why
The docs navigation now separates product-specific pages under `/cloud` and `/self-host`, while shared docs keep their original public paths. Some productless app links are available in both products and should follow the current edition, but single-product links should stay on the only product path that actually exists. For example, `/use-dify/workspace/subscription-management#dify-for-education` is Cloud-only and should not be rewritten to a self-hosted URL.

Workflow node help links were also deriving docs paths from node metadata defaults. The End node's type-derived `/use-dify/nodes/end` path does not exist; it should use `/use-dify/nodes/output`. Other nodes without generated docs pages should not expose a Help Link.

## Validation
- `pnpm -C web test context/i18n.spec.ts app/components/goto-anything/actions/commands/__tests__/direct-commands.spec.ts app/components/integrations/__tests__/page.spec.tsx app/components/header/account-setting/api-based-extension-page/__tests__/empty.spec.tsx app/components/workflow/block-selector/__tests__/main.spec.tsx app/components/tools/provider/__tests__/empty.spec.tsx app/components/tools/provider/__tests__/custom-create-card.spec.tsx app/components/datasets/external-knowledge-base/create/__tests__/index.spec.tsx`
- `pnpm -C web test app/components/workflow-app/hooks/__tests__/use-available-nodes-meta-data.spec.ts`
- `pnpm -C web test context/i18n.spec.ts app/components/workflow-app/hooks/__tests__/use-available-nodes-meta-data.spec.ts`
- `pnpm -C web type-check`


hyoban:
@copilot resolve the merge conflicts in this pull request

lyzno1 (approved):

PR #37567, opened 2026-06-17
fix(agent-v2): include workflow references in agent list

lyzno1:
## Summary
- add Agent App list response fields for published workflow reference count and references
- reuse AgentRosterService published-reference loading so the roster list and Access reference data share the same source
- regenerate the console agent contract for the updated /agent response

## Why
The Agent V2 roster card needs to show whether an agent is used by published workflows. The existing /agent list response only returned app-backed agent metadata, so the frontend would fall back to 0 even when /agent/{agent_id}/referencing-workflows had data.

## Validation
- uv run --project api pytest api/tests/unit_tests/controllers/console/agent/test_agent_controllers.py -q
- uv run --project api ruff check api/controllers/console/agent/roster.py api/services/agent/roster_service.py api/tests/unit_tests/controllers/console/agent/test_agent_controllers.py
- pnpm -C packages/contracts type-check

### DIFF
--- a/api/controllers/console/agent/roster.py
+++ b/api/controllers/console/agent/roster.py
@@ -64,10 +64,17 @@ class AgentAppUpdatePayload(UpdateAppPayload):
     role: str | None = Field(default=None, description="Agent role", max_length=255)
 
 
+class AgentAppPublishedReferenceResponse(BaseModel):
+    app_id: str
+    app_name: str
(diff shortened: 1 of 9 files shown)


zyssyz123 (dismissed):
LGTM

zyssyz123 (approved):
LGTM

PR #37579, opened 2026-06-17
fix(notion): add bounded timeouts to Notion API requests

he-yufeng:
## What

`NotionExtractor` makes five calls to the Notion API — database query, block children, block read, table rows, and last-edited-time — all via `httpx` with no `timeout`. This adds a module-level `NOTION_REQUEST_TIMEOUT` (`httpx.Timeout(30.0, connect=10.0)`) and passes it to each call.

## Why

`httpx` has no default timeout, so a call with no `timeout=` waits forever. If Notion is slow or unresponsive, a document import or sync can hang indefinitely and tie up the worker handling it — there's no upper bound on how long these requests block.

This is the same failure mode the recently merged bounded-timeout fixes addressed for the Nacos remote-settings client (#37444) and the Marketplace client (#37424); the Notion extractor is another data-source path that was still unbounded.

## Test

Added `TestNotionRequestTimeouts` to `test_notion_extractor.py`, covering the `httpx.post` (database query) and `httpx.request` (last-edited-time) paths and asserting each call receives the bounded `NOTION_REQUEST_TIMEOUT`. The tests follow the existing `mocker.patch("httpx.post"/"httpx.request")` + `_mock_response` patterns already used in that file.

Note: I couldn't run the full backend test suite locally (it pulls in the complete workflow stack via `conftest.py`, which needs dependencies I don't have set up here), so I'm relying on CI for the test run. `ruff check`/`ruff format --check` pass on both changed files and the module compiles cleanly.

From Claude Code

### DIFF
--- a/api/core/rag/extractor/notion_extractor.py
+++ b/api/core/rag/extractor/notion_extractor.py
@@ -21,6 +21,10 @@ SEARCH_URL = "https://api.notion.com/v1/search"
 
 RETRIEVE_PAGE_URL_TMPL = "https://api.notion.com/v1/pages/{page_id}"
 RETRIEVE_DATABASE_URL_TMPL = "https://api.notion.com/v1/databases/{database_id}"
+
+# Bound how long Notion API requests wait so a slow or unresponsive Notion
+# endpoint cannot stall document import or sync indefinitely.
+NOTION_REQUEST_TIMEOUT = httpx.Timeout(30.0, connect=10.0)
 # if user want split by headings, use the corresponding splitter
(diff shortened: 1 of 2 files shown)


he-yufeng:
Bumping this gently. It adds bounded timeouts to the Notion API calls so a hung request cannot stall ingestion indefinitely. Happy to tweak the timeout defaults if you would prefer different values.

he-yufeng:
Covered upstream: #39833 landed the same bounded timeout for the Notion extractor ( on all five call sites), so this is redundant now. Closing.

PR #37598, opened 2026-06-17
fix(datasets): restore cross-page document selection for batch actions

### LINKED ISSUE #37559: [Bug] Cross-page selection breaks select-all behavior and batch download state in documents list
### Dify version
Observed on a self-hosted Docker deployment based on `main` at commit `a8236499343f34cdb665b80c90296f221bc16d60`.

### Cloud or Self Hosted
Self Hosted (Docker)

### Steps to reproduce
1. Open a dataset with enough documents to span multiple pages.
2. On page 1, select one or more documents.
3. Go to another page.
4. Select one or more additional documents.
5. Observe the header select-all checkbox and the batch action area.

### ✔️ Expected Behavior
- The header select-all checkbox should reflect only the current page subset.
- The batch download action should stay enabled when the full cross-page selection still contains downloadable file documents.

### ❌ Actual Behavior
- The header select-all checkbox reflects the full cross-page `selectedIds` incorrectly for the current page.
- The batch download action is enabled/disabled based only on the current page subset, even when the full cross-page selection still contains downloadable file documents.

### Additional context
- I can reproduce two regressions reliably:
  1. The header select-all checkbox behaves incorrectly after selecting documents across pages.
  2. The batch download action state becomes incorrect after selecting documents across pages.
- This is observed on a Docker-based self-hosted deployment running on Kubernetes. The previous issue was closed because I mistakenly labeled the deployment mode as `Self Hosted (Source)`: #37476.
- I have not observed any related backend error logs so far. This currently looks like a frontend selection-state issue.
- I also noticed a likely related problem around `hasErrorDocumentsSelected` only checking the current page documents, which may affect the batch re-index action after cross-page selection. I have not reproduced that one locally yet because I do not currently have `error` documents in my dataset, so I am intentionally not including it in this report's primary scope.

tsushanth:
## Summary

Fixes #37559.

When users select documents across multiple pages of a dataset, two regressions surface on the documents list (the issue's repro reliably hits both on a multi-page dataset):

1. **Header select-all checkbox renders the wrong state.** `CheckboxGroup` is passed `value={selectedIds}` (full cross-page selection) and `allValues={documentIds}` (current page only), so its checked / indeterminate computation reflects the cross-page selection on every page instead of the current page's subset.

2. **Batch download disables incorrectly after pagination.** `useDocumentSelection` derives `hasErrorDocumentsSelected` and `downloadableSelectedIds` by filtering the current page's `documents` array. Any selection made on a previous page becomes invisible to these gates the moment the user navigates away, even though `selectedIds` itself is intentionally preserved across pagination (see `use-documents-page-state.ts`).

`@dosubot` independently confirmed the same two locations in the issue thread and noted that prior PR #23000 used a similar shape (passing the full id set separately from the current-page slice) to fix the equivalent bug for batch metadata editing.

### `components/list.tsx`

- Clip `value` passed to `CheckboxGroup` to the intersection of `selectedIds` and the current page's `documentIds`, and merge the off-page selection back in inside the `onValueChange` handler. A "select all" on page 2 now only mutates page 2's selection, leaving page 1's intact.

### `components/document-list/hooks/use-document-selection.ts`

- Remember, per selected id, the bits we need for batch-action gating (`isFile`, `isError`) inside the hook.
- Expose `updateSelectionFromCurrentPage(nextIds)` as the single write-path the list calls when the visible-page selection changes. The hook looks up newly-selected ids in the current page's `documents` to capture their metadata, and drops metadata for ids that left the selection.
- `hasErrorDocumentsSelected` and `downloadableSelectedIds` now read from the persisted map, so they remain accurate after pagination.
- `clearSelection` also wipes the map so a fresh selection starts clean.

### Tests

- New cases verify the cross-page behaviour explicitly: error gating survives navigating away from the page that contained the error doc, downloadable selections survive pagination, metadata for unselected ids is dropped, status updates are reflected, and `clearSelection` wipes both the ids and the metadata.
- Existing `list.spec.tsx` mock for `useDocumentSelection` updated to return the new `updateSelectionFromCurrentPage` (it just forwards to `onSelectedIdChange` for the select-all test).
- Full suite (`pnpm test app/components/datasets/documents`) passes: **165 files, 3,535 tests**.
- Targeted lint passes on all four touched files.

## Why this shape

Persisting selection metadata is the only way to keep gating accurate without forcing a backend call to look up the metadata of off-page selections. The hook is the only place that sees both the visible page's documents and the active selection, so it owns the bookkeeping. Stale entries are bounded (at most O(unique ids selected during this mount) — pruned on the next selection change) and harmless because the derive step iterates `selectedIds` first.

## Out of scope

The bug also mentions a probable third gap around `hasErrorDocumentsSelected` and the batch re-index action — `@isku-dere` couldn't reproduce it without `error`-status documents, so the report intentionally leaves it for a follow-up. With this PR's persisted metadata in place, that third gap is also fixed automatically: any `error`-status doc selected on a previous page will keep `hasErrorDocumentsSelected = true` after pagination, which is exactly what the batch re-index action checks.

### DIFF
--- a/web/app/components/integrations/__tests__/page.spec.tsx
+++ b/web/app/components/integrations/__tests__/page.spec.tsx
@@ -1,4 +1,5 @@
-import { fireEvent, screen, within } from '@testing-library/react'
+import { screen, within } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import { renderWithNuqs } from '@/test/nuqs-testing'
 import IntegrationsPage from '../page'
 
@@ -476,10 +477,11 @@ describe('IntegrationsPage', () => {
     ['extension', 'empty marketplace', '/plugins/extension'],
   ] as const)(
     'opens the %s marketplace path from integrations',
-    (section, buttonName, marketplacePath) => {
+    async (section, buttonName, marketplacePath) => {
+      const user = userEvent.setup()
       renderIntegrationsPage({ section })
 
-      fireEvent.click(screen.getByRole('button', { name: buttonName }))
+      await user.click(screen.getByRole('button', { name: buttonName }))
 
       expect(mockWindowOpen).toHaveBeenCalledWith(
         expect.stringContaining(`${marketplacePath}?source=`),
@@ -490,11 +492,12 @@ describe('IntegrationsPage', () => {
     },
   )
 
-  it('passes marketplace platform paths to external marketplace callbacks', () => {
+  it('passes marketplace platform paths to external marketplace callbacks', async () => {
+    const user = userEvent.setup()
     const onSwitchToMarketplace = vi.fn()
     renderIntegrationsPage({ section: 'trigger' }, { onSwitchToMarketplace })
 
-    fireEvent.click(screen.getByRole('button', { name: 'empty marketplace' }))
+    await user.click(screen.getByRole('button', { name: 'empty marketplace' }))
 
     expect(onSwitchToMarketplace).toHaveBeenCalledWith('/plugins/trigger')
     expect(mockRouterPush).not.toHaveBeenCalled()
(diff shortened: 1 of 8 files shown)


tsushanth:
Rebased onto current main.

tsushanth:
Rebased onto current main.

PR #37641, opened 2026-06-18
refactor(openapi/cli): split app usage-face from studio-app build-face

GareArc:
## Summary

Splits the app command surface into two faces by auth subject:

- **`app`** (usage face) — `run` / `get` / `describe` / `resume`, dual-subject (account + external-SSO), each subject reading from its own backend surface.
- **`studio-app`** (build face) — account-only `export` / `import`.

Renames `export app` / `import app` → `export studio-app` / `import studio-app`.

Closes WTA-888

## Highlights

- `describe` is now a public projection only — drops `author`/`tags` to close a cross-tenant identity leak.
- New external-SSO public `describe` route.
- Subject-driven app reader dispatch on the CLI.
- Help/guide text distinguishing the two nouns.

## Breaking

- CLI: `export app` / `import app` removed in favour of `studio-app`.
- HTTP: `GET /apps/{id}/describe` no longer returns `author` / `tags`.

## Verification

All CI green — API, CLI (3 OS), Web, E2E, CodeQL, style.


cli/src/commands/get/app/run.ts
```diff
@@ -102,9 +109,9 @@ function describeToEnvelope(desc: AppDescribeResponse, wsId: string, wsName: str
       name: desc.info.name,
       description: desc.info.description,
       mode: desc.info.mode as AppMode,
-      tags: desc.info.tags,
+      tags: [],
       updated_at: desc.info.updated_at,
-      created_by_name: desc.info.author === '' ? undefined : desc.info.author,
+      created_by_name: undefined,
```
lin-snow: This (and `tags: []` above) is where single `get app <id>` drops author/tags, while
the list path keeps them from `AppListRow`. Let's document the divergence or route
single-by-id through a tenant-scoped lookup so both views agree.

lin-snow (dismissed):
Verified the security fix: account `describe` goes through
`AppService.get_visible_app_by_id` (a bare `session.get(App, id)`, no tenant filter),
so dropping `author`/`tags` is a real cross-tenant leak fix. LGTM overall

cli/src/commands/get/app/run.ts
GareArc: Agreed. App is use-face only so we should drop them entirely here in the model.

wylswz (approved):

PR #37653, opened 2026-06-18
fix(web): match trace entries by execution id in parallel workflow branches

### LINKED ISSUE #37645: [Bug] Workflow node trace misattributed to wrong parallel branch when parallel_id differs between start and finish events
### Self Checks

### Dify version

v1.14.2 (latest main, confirmed via commit d74e84c)

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

1. Deploy Dify v1.14.2 from source (or Docker Compose).
2. Open Studio and create a new **Workflow** app.
3. Add a **Parallel Branch** node. Inside each branch, add an **LLM** or **Tool** node.
4. Open the **Debug & Preview** panel and run the workflow **at least twice** in the same debug session without closing the panel.
5. Inspect the **Trace / Execution Log** panel in the right sidebar.
6. Observe per-node trace entries for nodes inside parallel branches — they are misattributed or overwritten across runs.

**Affected code** (confirmed via PR #33215 Copilot code review):
- `web/app/components/workflow/utils/top-level-tracing.ts` → `findParallelTraceIndex()`
- `web/app/components/workflow/hooks/use-workflow-run-event/use-workflow-node-started.ts`

```typescript
// Current buggy logic in findParallelTraceIndex():
const matchedByIdIndex = tracing.findIndex((item) => {
  if (item.id !== data.id)
    return false

  const existingParallelId = item.execution_metadata?.parallel_id ?? item.parallel_id
  if (!existingParallelId || !incomingParallelId)
    return true

  return existingParallelId === incomingParallelId  // ← BUG: rejects a valid exact id match
})
```

When `item.id === data.id` (exact `node_execution_id` match), the function **still validates** `parallel_id` consistency. If `parallel_id` is populated differently between `node_started` and `node_finished` events (e.g. one event uses `execution_metadata.parallel_id` while the other uses the top-level `parallel_id` field), this secondary check fails and the exact match is rejected. The code then falls back to `node_id + parallel_id` matching, which attaches the update to a **different parallel branch's trace entry**.

### ✔️ Expected Behavior

When a workflow node's `node_execution_id` (the `id` field on a `NodeTracing` entry) uniquely identifies an execution, any subsequent event carrying the **same** `node_execution_id` should be matched to that exact trace entry — regardless of how `parallel_id` is sourced or populated across different SSE events.

The Trace / Execution Log panel should:
- Correctly display each parallel branch's per-node execution log in isolation.
- Preserve all trace entries when the same workflow is run multiple times in the same Debug & Preview session.
- Route agent logs to the node execution that actually produced them.

The fix is straightforward — once `item.id === data.id` is confirmed, the secondary `parallel_id` consistency check should be skipped:
```typescript
// Correct fix (as suggested by Copilot review on PR #33215, Mar 25 2026):
const matchedByIdIndex = tracing.findIndex(item => item.id === data.id)
```

### ❌ Actual Behavior

In `findParallelTraceIndex()` (`web/app/components/workflow/utils/top-level-tracing.ts`), after finding a candidate where `item.id === data.id`, the function **still** validates `parallel_id` consistency. When `parallel_id` is populated differently between the `node_started` event (`execution_metadata.parallel_id`) and the `node_finished` event (top-level `parallel_id` field), this secondary check fails and the exact match is **rejected**.

The code then falls back to a `node_id + parallel_id` lookup, which attaches the update to a **different** parallel branch's trace entry.

**Observed symptoms:**
- In the Trace panel, a node's status/timing/output appears under the wrong parallel branch.
- Agent logs are misrouted to an incorrect node execution (also affects `use-workflow-agent-log.ts`).
- On repeated runs in the same Debug & Preview session, earlier trace entries are overwritten or duplicated instead of being preserved with their unique execution IDs.
- The bug is **silent** — no error is thrown, making it very hard to diagnose without inspecting SSE event payloads manually.

**Root cause confirmed** via Copilot automated code review on PR #33215 (comment dated Mar 25, 2026):
> "Since NodeTracing.id is the execution identifier, prefer matching by id alone once ids match (and only use parallel_id logic as a fallback when id is missing)."

This issue has **never been filed as a standalone bug** and the root-cause fix was not applied in the current PR. Affects all users running workflows with Parallel Branch nodes in Debug & Preview mode.

citizen204:
## Summary

Fixes #37645

In workflows with parallel branches, the same logical node (same `node_id`) can run concurrently across multiple branches in a single execution, and can also run again in subsequent executions within the same Debug & Preview session. Two places in the tracing state update code used `node_id` as the lookup key, causing trace entries to be misattributed or overwritten across parallel branches or repeated runs.

The `id` field on `NodeTracing` is the unique `node_execution_id` per SSE event — once you match on `id`, you've identified exactly the right trace entry. The secondary `parallel_id` consistency check in `onNodeFinished` was not only redundant but actively harmful: when `parallel_id` is populated differently between the `node_started` and `node_finished` SSE events (one using `execution_metadata.parallel_id`, the other the top-level `parallel_id` field), the secondary check incorrectly rejects a valid exact-id match.

**Changes:**
- `use-workflow-run-event/use-workflow-node-started.ts`: change `item.node_id === data.node_id` → `item.id === data.id`
- `panel/debug-and-preview/hooks.ts` (`onNodeStarted` in `updateChatTreeNode`): same `node_id` → `id` fix
- `panel/debug-and-preview/hooks.ts` (`onNodeFinished` in `updateChatTreeNode`): collapse 5-line conditional (with redundant `parallel_id` check) to `item.id === nodeFinishedData.id`

## Test plan

- [ ] Create a Workflow app with a **Parallel Branch** node; add at least one **LLM** or **Tool** node inside each branch
- [ ] Open Debug & Preview and run the workflow **twice** without closing the panel
- [ ] Verify the Trace panel shows each branch's per-node log isolated to its own branch (no cross-branch misattribution)
- [ ] Verify trace entries from the second run are not overwriting first-run entries for the same `node_id`
- [ ] Verify agent logs appear under the correct node execution

### DIFF
--- a/web/app/components/workflow/hooks/use-workflow-run-event/__tests__/use-workflow-agent-log.spec.ts
+++ b/web/app/components/workflow/hooks/use-workflow-run-event/__tests__/use-workflow-agent-log.spec.ts
@@ -71,6 +71,27 @@ describe('useWorkflowAgentLog', () => {
     expect((log[0] as unknown as { text: string }).text).toBe('new')
   })
 
+  it('routes by node_execution_id so parallel branches with the same node_id do not collide', () => {
+    const { result, store } = renderWorkflowHook(() => useWorkflowAgentLog(), {
+      initialStoreState: {
+        workflowRunningData: baseRunningData({
+          tracing: [
+            { id: 'exec-a', node_id: 'n1', execution_metadata: {} },
+            { id: 'exec-b', node_id: 'n1', execution_metadata: {} },
+          ],
+        }),
+      },
+    })
+
+    result.current.handleWorkflowAgentLog({
+      data: { node_id: 'n1', node_execution_id: 'exec-b', message_id: 'm1' },
+    } as AgentLogResponse)
+
+    const tracing = store.getState().workflowRunningData!.tracing!
+    expect(tracing[0]!.execution_metadata!.agent_log).toBeUndefined()
+    expect(tracing[1]!.execution_metadata!.agent_log).toHaveLength(1)
+  })
+
   it('creates execution_metadata when it does not exist', () => {
     const { result, store } = renderWorkflowHook(() => useWorkflowAgentLog(), {
       initialStoreState: {
--- a/web/app/components/workflow/hooks/use-workflow-run-event/__tests__/use-workflow-node-started.spec.ts
+++ b/web/app/components/workflow/hooks/use-workflow-run-event/__tests__/use-workflow-node-started.spec.ts
@@ -77,4 +77,30 @@ describe('useWorkflowNodeStarted', () => {
     expect(tracing).toHaveLength(2)
     expect(tracing[1]!.status).toBe(NodeRunningStatus.Running)
   })
+
+  it('matches parallel-branch entries by execution id, not node_id', () => {
+    const { result, store } = renderViewportHook(() => useWorkflowNodeStarted(), {
+      initialStoreState: {
+        workflowRunningData: baseRunningData({
+          tracing: [
+            { id: 'exec-a', node_id: 'n1', status: NodeRunningStatus.Succeeded } as never,
+          ],
+        }),
+      },
+    })
+
+    act(() => {
+      result.current.handleWorkflowNodeStarted(createNodeStartedResponse({
+        data: {
+          id: 'exec-b',
+          node_id: 'n1',
+          execution_metadata: { parallel_id: 'p1' },
+        } as never,
+      }), containerParams)
+    })
+
+    const tracing = store.getState().workflowRunningData!.tracing!
+    expect(tracing).toHaveLength(2)
+    expect(tracing[1]!.status).toBe(NodeRunningStatus.Running)
+  })
 })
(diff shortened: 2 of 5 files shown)


citizen204:
Pushed a follow-up: the `onNodeStarted` run-event hook (`use-workflow-run-event/use-workflow-node-started.ts`) still matched purely by execution `id`, which both broke its unit test and was inconsistent with the reviewer-requested change in `debug-and-preview/hooks.ts`. It now uses the same conditional — match by `id` only when `execution_metadata.parallel_id` is present, otherwise fall back to `node_id`.

This restores the original non-parallel behavior (fixing the failing `use-workflow-node-started.spec.ts` case), and I added a test covering the parallel-branch case where the same `node_id` under a different execution id must not overwrite a sibling tracing entry.

rkfshakti:
Hi @citizen204 — I had a parallel fix (#39313) for the same issue that was closed as a duplicate of this PR. My fix also included the agent-log handler (`use-workflow-agent-log.ts`) which had the same `node_id`-based matching bug — it matched trace entries by `node_id` instead of `node_execution_id`, causing agent logs to be misrouted across parallel branches.

If you're interested, I can open a PR against your branch with that complementary fix, or share the diff here. The change is small:

In `use-workflow-agent-log.ts`, match by `data.node_execution_id` (the unique execution identifier) instead of `data.node_id`, with a fallback to `node_id` for older events that may not carry an execution id.

Let me know if that would be helpful!

citizen204:
Thanks for the heads up @laipz8200 — rebased onto current main and added the complementary fix for `use-workflow-agent-log.ts`: it now matches by `data.node_execution_id` when present (falling back to `node_id` for older events without one), same pattern as the tracing-entry fix, with a new test covering two parallel branches sharing a `node_id`. Let me know if you'd prefer it split into a separate PR instead.

citizen204:
(Correction: meant to tag @rkfshakti above, not @laipz8200 — apologies for the mix-up.)

PR #37647, opened 2026-06-18
fix(api/services): set annotation app cache keys to prevent duplicate jobs

Harsh23Kashyap:
## Summary
Set app-level annotation processing cache keys at enqueue time in enable_app_annotation and disable_app_annotation to prevent duplicate jobs from concurrent requests.

Fixes #37646

## Test plan
- Targeted unit tests for both cache-miss enqueue paths pass.
- Local containers integration run attempted; blocked here because Docker daemon is unavailable.

PR #37668, opened 2026-06-19
fix: set app-level cache key to prevent duplicate annotation jobs

### LINKED ISSUE #37646: Bug: enable/disable app annotation cache key is checked but never set, allowing duplicate jobs
### Self Checks
- [x] This is a bug report (not a question/discussion).
- [x] I am using English for this report.
- [x] I did not modify the issue template sections.

### Dify version
main branch (observed on current upstream as of June 2026)

### Cloud or Self Hosted
N/A — code-level bug in API service layer

### Steps to reproduce
1. Call `AppAnnotationService.enable_app_annotation(...)` concurrently for the same `app_id`.
2. In `api/services/annotation_service.py`, `enable_app_annotation` checks `enable_app_annotation_{app_id}` via `redis_client.get(...)`.
3. On cache miss, it only sets `enable_app_annotation_job_{job_id}` with `setnx`, but does **not** set `enable_app_annotation_{app_id}`.
4. A second concurrent call still sees cache miss and enqueues another task.

Same pattern exists for `disable_app_annotation` with `disable_app_annotation_{app_id}`.

Relevant current behavior:
- service checks app-level keys (`enable_app_annotation_{app_id}` / `disable_app_annotation_{app_id}`)
- tasks delete those app-level keys in `finally` (`enable_annotation_reply_task` / `disable_annotation_reply_task`)
- but service never sets those app-level keys, so dedupe gate is ineffective.

### Expected Behavior
When the first request enqueues a job, app-level processing key should be set (e.g. `setex(..., job_id)`), so subsequent requests return `{"job_id": <existing>, "job_status": "processing"}` instead of scheduling duplicate jobs.

### Actual Behavior
Concurrent requests can enqueue duplicate annotation enable/disable tasks for the same app because the checked app-level cache key is never written.

### Additional context
Sibling behavior in tasks suggests intended lifecycle already exists (task `finally` deletes app-level key), but the service-side key write is missing.

Happy to send a PR. Could you assign this issue to me?

EvanYao826:
## Problem

`enable_app_annotation` and `disable_app_annotation` check app-level redis keys (`enable_app_annotation_{app_id}` / `disable_app_annotation_{app_id}`) via `redis_client.get()` but **never set them** on cache miss. This means concurrent requests all see a cache miss and enqueue duplicate Celery tasks.

The task `finally` blocks already delete these keys, confirming the intended lifecycle:

1. Service sets app-level key → indicates "processing"
2. Task runs
3. Task `finally` deletes app-level key

But step 1 was missing.

## Fix

Add `redis_client.setex(enable_app_annotation_key, 3600, job_id)` after the cache-miss check and before enqueuing the task. Same for `disable_app_annotation`. TTL of 3600s provides a safety net if the task fails to clean up.

Fixes #37646

### DIFF
--- a/api/services/annotation_service.py
+++ b/api/services/annotation_service.py
@@ -167,6 +167,9 @@ class AppAnnotationService:
 
         # async job
         job_id = str(uuid.uuid4())
+        # Set the app-level cache key so concurrent requests see the in-progress job.
+        # The task deletes this key in its finally block.
+        redis_client.setex(enable_app_annotation_key, 3600, job_id)
         enable_app_annotation_job_key = f"enable_app_annotation_job_{str(job_id)}"
         # send batch add segments task
         redis_client.setnx(enable_app_annotation_job_key, "waiting")
@@ -192,6 +195,9 @@ class AppAnnotationService:
 
         # async job
         job_id = str(uuid.uuid4())
+        # Set the app-level cache key so concurrent requests see the in-progress job.
+        # The task deletes this key in its finally block.
+        redis_client.setex(disable_app_annotation_key, 3600, job_id)
         disable_app_annotation_job_key = f"disable_app_annotation_job_{str(job_id)}"
         # send batch add segments task
         redis_client.setnx(disable_app_annotation_job_key, "waiting")


crazywoola:
Thanks for the contribution. We already have an earlier PR for this issue and follow FIFO for duplicate fixes, so review will continue in #37647.

PR #37669, opened 2026-06-19
fix: skip empty tool entries in legacy dataset config extraction

he-yufeng:
### Summary

`DatasetConfigManager.extract_dataset_config_for_legacy_compatibility` iterates `agent_mode.tools` and reads `list(tool.keys())[0]` without checking that the tool dict is non-empty. A malformed empty tool entry (`{}`) therefore raises `IndexError: list index out of range` before any dataset is resolved.

The sibling method `convert()` already guards this exact case with `if len(tool) == 1:`. This change skips empty tool entries in the legacy-compatibility path the same way, so a stray `{}` is ignored instead of crashing.

```python
config = {"agent_mode": {"enabled": True, "strategy": PlanningStrategy.ROUTER, "tools": [{}]}}
DatasetConfigManager.extract_dataset_config_for_legacy_compatibility("t", AppMode.CHAT, config)
# before: IndexError: list index out of range
# after:  the empty entry is skipped, config returned unchanged
```

### Testing

Added `TestExtractDatasetConfigForLegacyCompatibility::test_skips_empty_tool_entry` (the method previously had no direct coverage for this path). I confirmed the `IndexError` and the fix with a standalone repro of the slicing logic; the full unit test runs under CI, since the backend test environment isn't fully installable on my machine.

From Claude Code

### DIFF
--- a/api/core/app/app_config/easy_ui_based_app/dataset/manager.py
+++ b/api/core/app/app_config/easy_ui_based_app/dataset/manager.py
@@ -213,6 +213,11 @@ class DatasetConfigManager:
             PlanningStrategy.REACT_ROUTER,
         }:
             for tool in config.get("agent_mode", {}).get("tools", []):
+                if not tool:
+                    # Skip malformed empty tool entries; list(tool.keys())[0]
+                    # would otherwise raise IndexError. The sibling convert()
+                    # already guards this with `if len(tool) == 1`.
+                    continue
                 key = list(tool.keys())[0]
                 if key == "dataset":
                     # old style, use tool name as key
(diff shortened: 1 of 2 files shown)


asukaminato0721:
@dosu instead of add special check, how about use pydantic for it.

asukaminato0721:
ref https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

asukaminato0721:
@dosu no, only validate config

he-yufeng:
Appreciate the pointer — parse-don't-validate is the right instinct in general. I kept this one as a local guard for two reasons:

- `convert()` in this same file already handles the identical `agent_mode.tools` shape with a plain `if len(tool) == 1` check (line 55), so skipping a malformed entry here matches the pattern that's already established right next to it.
- Putting a Pydantic model in front of `validate_and_set_defaults()` would mean modeling the whole legacy easy-UI config, which is a much wider change across the legacy-compat path than this one IndexError warrants, and risks shifting behavior for existing stored configs.

Happy to take the Pydantic route as a separate, focused refactor if the maintainers would prefer that direction — but for the crash itself the minimal guard keeps the blast radius small and consistent with the neighboring code.

From Claude Code

asukaminato0721 (approved):

PR #37671, opened 2026-06-19
fix: reject empty user_input_form items instead of raising IndexError

he-yufeng:
## Summary

`BasicVariablesConfigManager` reads each `user_input_form` entry's type with `list(item.keys())[0]`, with no guard for an empty object. So a config like `{"user_input_form": [{}]}` raises a bare `IndexError: list index out of range` in two places:

- `validate_variables_and_set_defaults` (the validation entry point) — every other malformed shape here is turned into a clear `ValueError`, but an empty item slips past and crashes with `IndexError`.
- `convert` — same `[0]` access while building the entities.

`user_input_form` comes from client-supplied app config, so `[{}]` is reachable malformed input.

## Fix

- In `validate_variables_and_set_defaults`, raise a clear `ValueError` for an empty item, consistent with the other validation errors in that method.
- In `convert`, skip empty items (mirrors how the method already skips entries it can't use).

This is the same class of bug as the legacy dataset/manager fix in #37669, in a different config path.

## Test

Added two cases to `test_variables_manager.py` (matching the existing style there): `validate_variables_and_set_defaults({"user_input_form": [{}]})` raises `ValueError`, and `convert({"user_input_form": [{}]})` returns empty lists. Both fail on `main` (they raise `IndexError`) and pass with this change.

I couldn't run the suite locally (the module's import chain pulls in `graphon`, which isn't installable here), so I verified the boundary logic in isolation — `list({}.keys())[0]` raises `IndexError`, the guarded paths raise `ValueError` / skip — and ran `ruff check` / `ruff format --check` clean. The unit tests run in CI.

From Claude Code.

### DIFF
--- a/api/core/app/app_config/easy_ui_based_app/variables/manager.py
+++ b/api/core/app/app_config/easy_ui_based_app/variables/manager.py
@@ -45,6 +45,8 @@ class BasicVariablesConfigManager:
 
         # variables and external_data_tools
         for variables in config.get("user_input_form", []):
+            if not variables:
+                continue
             variable_type = list(variables.keys())[0]
             if variable_type == VariableEntityType.EXTERNAL_DATA_TOOL:
                 variable = variables[variable_type]
@@ -112,6 +114,8 @@ class BasicVariablesConfigManager:
 
         variables = []
         for item in config["user_input_form"]:
+            if not item:
+                raise ValueError("Each item in user_input_form must be a non-empty object")
             key = list(item.keys())[0]
             # if key not in {"text-input", "select", "paragraph", "number", "external_data_tool"}:
             if key not in {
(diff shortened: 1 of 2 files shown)


he-yufeng:
Gentle ping — this has been sitting since mid-June with green CI. `BasicVariablesConfigManager` reads each `user_input_form` entry's type via `list(item.keys())[0]` with no empty-object guard, so a config containing `[{}]` throws a bare `IndexError: list index out of range` in two spots instead of a clear validation error. The PR turns those into a proper rejection. Let me know if you'd like it scoped differently.

he-yufeng:
Still mergeable with green CI. This rejects empty user_input_form items at validation instead of letting them raise deeper in the flow. Happy to adjust anything.

he-yufeng:
Still reproduces on current main: api/core/app/... app_config/easy_ui_based_app/variables/manager.py line 115 still does `list(...keys())[0]` on the entries without an emptiness guard, so an empty {} entry takes the same 500 path. The branch is rebased and the unit tests cover both the empty-entry skip/reject and the normal entries.

PR #37685, opened 2026-06-20
fix(watercrawl): don't disable request timeouts with timeout=None

he-yufeng:
### Summary

`WaterCrawlAPIClient` builds its `httpx.Client` with `timeout=None`. In httpx that disables timeouts entirely (infinite), so a WaterCrawl endpoint that stalls hangs the calling worker forever. Every crawl/scrape call (`create_crawl_request`, `get_crawl_request`, `download_crawl_request`, `get_crawl_request_results`, …) goes through that client, and none of them can ever time out.

`services/enterprise/base.py` already calls this exact pitfall out and deliberately avoids passing `timeout=None`:

```python
# - In httpx, passing timeout=None disables timeouts (infinite) and overrides the library default.
```

### Fix

- `init_session` no longer passes `timeout=None`, so regular requests fall back to httpx's default timeout.
- The crawl-status SSE stream (`monitor_crawl_request`) legitimately needs to stay open for the whole crawl, so the streaming request keeps an unbounded read while still capping the connect: `_STREAM_TIMEOUT = httpx.Timeout(None, connect=10.0)`.

Net effect: normal API calls can no longer hang indefinitely on a stalled endpoint, and long crawl monitoring is unchanged.

### Tests

In `tests/unit_tests/core/rag/extractor/watercrawl/test_watercrawl.py`:

- a regression test that the session's default timeout is now bounded (it was `None`);
- the existing stream/non-stream test now also asserts the streaming request keeps an unbounded read with a bounded connect.

I verified the changed code path locally against the real client module; the full API test environment doesn't install cleanly on my machine, so I'm relying on CI for the complete suite.

DCO signed off.

From Claude Code.

### DIFF
--- a/api/core/rag/extractor/watercrawl/client.py
+++ b/api/core/rag/extractor/watercrawl/client.py
@@ -71,9 +71,7 @@ class BaseAPIClient:
         stream = kwargs.pop("stream", False)
         url = urljoin(self.base_url, endpoint)
         if stream:
-            request = self.session.build_request(
-                method, url, params=query_params, json=data, timeout=_STREAM_TIMEOUT
-            )
+            request = self.session.build_request(method, url, params=query_params, json=data, timeout=_STREAM_TIMEOUT)
             return self.session.send(request, stream=True, **kwargs)
 
         return self.session.request(method, url, params=query_params, json=data, **kwargs)
--- a/api/core/rag/extractor/watercrawl/client.py
+++ b/api/core/rag/extractor/watercrawl/client.py
@@ -14,6 +14,12 @@ from core.rag.extractor.watercrawl.exceptions import (
 
 WATERCRAWL_REQUEST_TIMEOUT: httpx.Timeout = httpx.Timeout(30.0, connect=5.0)
 
+# The crawl-status stream is a long-lived SSE connection that can stay open for
+# the whole duration of a crawl, so it keeps an unbounded read while still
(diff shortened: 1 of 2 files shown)


he-yufeng:
Rebased onto current `main`, which since this PR opened has merged its own fix for the `timeout=None` issue (`WATERCRAWL_REQUEST_TIMEOUT = httpx.Timeout(30.0, connect=5.0)` on the client). I kept that as the regular-request timeout.

What this PR still adds on top: the crawl-status endpoint is consumed as a long-lived SSE stream (`monitor_crawl_request` → `_request(..., stream=True)`), and a single crawl can legitimately stay open well past 30s between events. With only the 30s client timeout, that stream now gets cut off mid-crawl by a read timeout. So `_request` overrides it for the streaming case with `_STREAM_TIMEOUT = httpx.Timeout(None, connect=10.0)` — unbounded read for the stream, but still a bounded connect so a dead endpoint can't hang the worker. Regular requests are unchanged (they keep the merged `WATERCRAWL_REQUEST_TIMEOUT`).

Net diff is now just the `_STREAM_TIMEOUT` constant and the stream-path override, plus a test. I couldn't run the full api suite locally (the unit test transitively imports the app and a dependency isn't installed here), so I'm relying on CI for the suite; the change itself is type-checked and the two touched files compile clean.

From Claude Code

crazywoola (approved):

PR #37728, opened 2026-06-22
fix(cli): align difyctl help text with actual flags and error envelope

lin-snow:
## Summary

Two pieces of `difyctl` help text had drifted from the actual CLI behavior:

- **`help external`** instructed users to run `difyctl auth login --external --token "$DIFY_TOKEN"`, but neither flag exists — `auth login` only accepts `--host` / `--no-browser` / `--insecure`. The real mechanism for external-SSO bearers is the `DIFY_TOKEN` env var, so the step now shows `export DIFY_TOKEN="<your-token>"` (consistent with the topic's own Notes section).
- **`help agent`** documented a `request` field in the structured JSON error envelope, but the envelope never emits `request` — it emits `method` / `url` / `raw_response` / `server`. The `request:` string only exists in the human-readable render. The contract shape now matches `HttpClientError.toEnvelope()`.

A guard test (`cli/src/help/contract.test.ts`) asserts that every key the real envelope emits is documented in the contract shape, so the two cannot silently drift again.

Help text and tests only — no CLI flag or error-rendering behavior changes.

## Screenshots

N/A — CLI help text only.

## Checklist

- [x] I ran the `cli` package gates (lint, type-check, and unit tests) to appease the lint gods.

### DIFF
--- a/cli/src/help/contract.test.ts
+++ b/cli/src/help/contract.test.ts
@@ -1,0 +1,26 @@
+import type { ErrorBody } from '@dify/contracts/api/openapi/types.gen'
+import { describe, expect, it } from 'vitest'
+import { HttpClientError, newError } from '@/errors/base'
+import { ErrorCode } from '@/errors/codes'
+import { CONTRACT } from './contract'
+
+describe('errorEnvelope contract', () => {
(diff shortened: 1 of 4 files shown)


GareArc (approved):
lgtm

PR #37736, opened 2026-06-22
fix(cli): apply --think filtering to workflow app outputs

lin-snow:
## Summary

The `--think` flag documents: *"Strips `<think>...</think>` blocks silently by default; with `--think`, thinking is printed to stderr."* This held for `chat` / `agent-chat` / `advanced-chat` / `completion` because the filter was wired to the `answer` field. **Workflow apps have no `answer`** — their output lives in `data.outputs` — so the flag was a silent no-op for them: `<think>` blocks leaked into stdout in both the streaming and non-streaming paths, and likewise on `resume app` for workflow-type apps.

This wires the existing think-filter into the workflow output path so the flag behaves consistently across **all** app modes, both commands (`run app` / `resume app`), both stream modes, and both `text` / `json` formats.

Resolves WTA-950.

### Behavior

| App mode | `--think` before | after (default) | after (`--think`) |
|---|---|---|---|
| chat / agent-chat / advanced-chat / completion | ✅ strips / routes | unchanged | unchanged |
| **workflow** — `run`, stream & non-stream | ❌ no-op, `<think>` leaks to stdout | `<think>` stripped silently | reasoning routed to stderr, stdout clean |
| **workflow** — `resume` | ❌ no-op | inherits fix | inherits fix |

```mermaid
flowchart LR
    O["workflow data.outputs"] --> F["filterThinkInOutputs(outputs, think)"]
    F -->|clean outputs| OUT["stdout"]
    F -.->|reasoning · only with --think| ERR["stderr"]
```

### What changed

- **`think-filter.ts`** — new `filterThinkInOutputs(outputs, showThink)`: strips `<think>` from top-level string fields of a workflow `outputs` object and collects the reasoning. Reuses the existing `extractThinkBlocks`.
- **`stream-handlers.ts`** — `WorkflowStreamPrinter` now receives the `think` flag and filters outputs in `onEnd` (live streaming).
- **`streaming-structured.ts`** — a workflow branch filters `data.outputs` before rendering (non-streaming + `--stream -o json`), mirroring the existing `answer` handling. Filtering happens before formatting, so `json` is cleaned too — consistent with how chat already strips `answer` in `json`.
- `renderWorkflow` and `resume/app/run.ts` are **unchanged**: the renderer receives already-clean outputs, and resume reuses the same strategy/printer, so both inherit the fix.

### Scope / follow-up

- Covers `tagged` reasoning (inline `<think>`), which is the default path.
- `separated`-mode reasoning over the new `reasoning_chunk` SSE event (depends on #37460) is a separate follow-up.
- Nested strings inside arrays/objects within `outputs` are intentionally not recursed.

## Tests

Added unit coverage for the helper, the streaming printer, and the structured path (incl. a `workflow-think` mock scenario). `pnpm test`, `pnpm type-check`, and `pnpm lint` are all green in `cli/`.

## Checklist

- [x] I ran lint & type-check (`pnpm lint && pnpm type-check && pnpm test` in `cli/`) to appease the lint gods.

From Claude Code

### DIFF
--- a/cli/src/sys/io/think-filter.ts
+++ b/cli/src/sys/io/think-filter.ts
@@ -14,6 +14,28 @@ export function extractThinkBlocks(s: string): { clean: string, thinking: string
   return { clean, thinking: parts.join('\n---\n') }
 }
 
+// Workflow outputs carry their answer text in top-level string fields rather than
+// a single `answer`, so think filtering navigates the outputs object. Nested
+// strings (inside arrays/objects) are left untouched.
+export function filterThinkInOutputs(
+  outputs: Record<string, unknown>,
+  showThink: boolean,
+): { outputs: Record<string, unknown>, thinking: string } {
+  const thoughts: string[] = []
+  const clean: Record<string, unknown> = {}
+  for (const [key, value] of Object.entries(outputs)) {
+    if (typeof value !== 'string') {
+      clean[key] = value
+      continue
+    }
+    const extracted = extractThinkBlocks(value)
+    clean[key] = extracted.clean
+    if (showThink && extracted.thinking !== '')
+      thoughts.push(extracted.thinking)
+  }
+  return { outputs: clean, thinking: thoughts.join('\n---\n') }
+}
+
 function splitAtPotentialTag(s: string, tag: string): [string, string] {
   const maxHold = tag.length - 1
   for (let len = Math.min(maxHold, s.length); len > 0; len--) {
--- a/cli/src/commands/run/app/_strategies/streaming-structured.ts
+++ b/cli/src/commands/run/app/_strategies/streaming-structured.ts
@@ -1,14 +1,14 @@
 import type { RunContext, RunStrategy } from './index'
 import type { SseEvent } from '@/http/sse'
 import { buildRunBody } from '@/api/app-run'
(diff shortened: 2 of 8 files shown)


cli/src/sys/io/think-filter.ts
```diff
@@ -14,6 +14,28 @@ export function extractThinkBlocks(s: string): { clean: string, thinking: string
...
+// Workflow outputs carry their answer text in top-level string fields rather than
+// a single `answer`, so think filtering navigates the outputs object. Nested
+// strings (inside arrays/objects) are left untouched.
+export function filterThinkInOutputs(
+  outputs: Record<string, unknown>,
+  showThink: boolean,
+): { outputs: Record<string, unknown>, thinking: string } {
+  const thoughts: string[] = []
+  const clean: Record<string, unknown> = {}
+  for (const [key, value] of Object.entries(outputs)) {
```
GareArc, line 26: Could outputs ever be an array here? cuz this will turn array into plain object.

lin-snow: Good catch. In practice data.outputs is always a variable-name→value object, never an array — but the guard used typeof raw === 'object', which doesn't exclude arrays, so a stray array would have been reshaped into a plain object. Hardened both call sites with !Array.isArray(raw) so arrays pass through untouched, and added a test.

GareArc (approved):
lgtm

PR #37738, opened 2026-06-22
fix(api): split resource-maintainers migration; batch the backfill

### LINKED ISSUE #37706: [api] Migration a7c4e9d2f681 runs unbatched UPDATE on apps/datasets hot tables and blocks writes for the full backfill duration

## Dify version

Current `main` (commit `3a3ad6ad7c`); migration head `a7c4e9d2f681`.

## Cloud or Self Hosted

- [ ] Cloud
- [x] Self Hosted (Docker)
- [x] Self Hosted (Source)

(Reproducible on any PG-backed deployment that runs this migration against a non-trivial dataset; the operational impact grows with row counts.)

## Steps to reproduce

1. Apply the Alembic migration `2026_06_15_1200-a7c4e9d2f681_add_resource_maintainers.py` against a PostgreSQL database where the `apps` and `datasets` tables already contain rows (typical SaaS tenant scale, e.g. 1 M+ rows in `apps`).
2. While the migration is running, observe `pg_stat_activity` and `pg_locks` from a separate session.
3. From the application, perform any concurrent write against `apps` or `datasets` (e.g. `POST /console/api/apps` or `POST /console/api/datasets`).

The relevant code is at `api/migrations/versions/2026_06_15_1200-a7c4e9d2f681_add_resource_maintainers.py` lines 30-31:

```python
op.execute(sa.text("UPDATE apps SET maintainer = created_by WHERE maintainer IS NULL"))
op.execute(sa.text("UPDATE datasets SET maintainer = created_by WHERE maintainer IS NULL"))
```

Both `UPDATE` statements run with no `LIMIT`, no id-range batching, and no intermediate commits. On PostgreSQL this acquires a row lock per updated row for the entire backfill duration.

## Expected Behavior

- The backfill completes within a bounded wall-clock window, proportional to per-batch size rather than to total table size.
- Concurrent application reads/writes against `apps` and `datasets` remain responsive throughout.
- Each batch commits independently so a failure in the middle of the backfill can be resumed from a known id range rather than re-running the entire migration.

## Actual Behavior

- The single `UPDATE apps ...` and `UPDATE datasets ...` hold row locks for the full backfill duration. Concurrent `INSERT` / `UPDATE` / `DELETE` against either table blocks behind the row lock, producing 504 / timeout errors on the application path.
- A failure in the middle of one `UPDATE` leaves the table in a partially backfilled state. The `WHERE maintainer IS NULL` clause still matches the remaining rows, so correctness is preserved on re-run, but the write-stall is unavoidable on large tenants and the migration is not restartable from a known checkpoint.

The `apps` and `datasets` tables are the hottest domain tables in Dify, touched on virtually every application and dataset API call. Blocking writes against them is operationally untenable on multi-tenant deployments.

## Suggested fix

Split the schema change from the data backfill, and batch the backfill in id ranges:

1. **Migration A** (this revision, split): add the nullable `maintainer` column and its `(tenant_id, maintainer)` index. Use `postgresql_concurrently=True` inside `op.get_context().autocommit_block()` for the index; provide a MySQL fallback without the keyword.
2. **Migration B** (separate revision): backfill in id-range batches with explicit commits between batches; each step independently restartable from the last successful batch id.
3. **Migration C** (later, after the feature is GA): add `NOT NULL` and drop the default in a third revision, gated on a feature flag.

The sibling migration `4474872b0ee6` in this codebase already demonstrates the established `autocommit_block + postgresql_concurrently=True` pattern that should be applied to the new index here.

AlexMultiAgent:
## Summary

Two changes that together address the operational impact of running migration `a7c4e9d2f681` on populated `apps` and `datasets` tables:

1. The composite index creation in `a7c4e9d2f681` now uses `autocommit_block` + `postgresql_concurrently=True` on PostgreSQL, so the two `(tenant_id, maintainer)` indexes are built without taking a `SHARE` lock on the hot tables. Mirrors the pattern in migrations `4474872b0ee6` and `6b5f9f8b1a2c`. `downgrade()` matches with a `CONCURRENTLY` drop.
2. The data backfill is split out into a new revision (`b7c8d9e0f1a2`) and rewritten to scan in id-range batches, with each batch running as its own transaction via `autocommit_block`. Row locks are released between batches instead of being held for the entire backfill. The `WHERE maintainer IS NULL` filter keeps the migration idempotent and restartable from the last committed id range on partial failure.

## Changes

- `api/migrations/versions/2026_06_15_1200-a7c4e9d2f681_add_resource_maintainers.py`
  - Add `_is_pg(conn)` helper.
  - Move the two `create_index` calls out of `batch_alter_table` and into `op.get_context().autocommit_block()` with `postgresql_concurrently=True` on PG; MySQL fallback preserved.
  - Remove the two unbatched `UPDATE ... SET maintainer = created_by WHERE maintainer IS NULL` statements (moved to the new revision below).
  - `downgrade()`: drop the indexes via `autocommit_block` + `postgresql_concurrently=True` on PG before dropping the columns.
- `api/migrations/versions/2026_06_22_0900-b7c8d9e0f1a2_backfill_resource_maintainers.py` (new)
  - `revision = b7c8d9e0f1a2`, `down_revision = 4f7b2c8d9a10` (current linear head — no branch introduced; verified via static graph analysis).
  - Idempotent batched UPDATE for `apps` and `datasets`; batch size 10_000 id range per iteration.
  - Reversible: `downgrade()` clears `maintainer` for rows that were backfilled by this migration, batched identically.

## Test plan

- [x] 2 files changed, 141 insertions(+), 6 deletions(-)
- [x] Python syntax OK on both files (UTF-8 safe)
- [x] Module imports cleanly; `revision`, `down_revision`, `upgrade`, `downgrade` all present
- [x] Migration graph remains linear with exactly one head (`b7c8d9e0f1a2`)
- [x] No other migration references `b7c8d9e0f1a2` as parent (verified)
- [x] Follows the established `autocommit_block + postgresql_concurrently=True` pattern from `4474872b0ee6` and `6b5f9f8b1a2c`

Fixes #37706

### DIFF
--- a/api/migrations/versions/2026_06_15_1200-a7c4e9d2f681_add_resource_maintainers.py
+++ b/api/migrations/versions/2026_06_15_1200-a7c4e9d2f681_add_resource_maintainers.py
@@ -18,24 +18,69 @@ branch_labels = None
 depends_on = None
 
 
+def _is_pg(conn) -> bool:
+    return conn.dialect.name == "postgresql"
+
+
 def upgrade() -> None:
     with op.batch_alter_table("apps", schema=None) as batch_op:
         batch_op.add_column(sa.Column("maintainer", models.types.StringUUID(), nullable=True))
-        batch_op.create_index("app_tenant_maintainer_idx", ["tenant_id", "maintainer"], unique=False)
 
     with op.batch_alter_table("datasets", schema=None) as batch_op:
         batch_op.add_column(sa.Column("maintainer", models.types.StringUUID(), nullable=True))
-        batch_op.create_index("dataset_tenant_maintainer_idx", ["tenant_id", "maintainer"], unique=False)
 
-    op.execute(sa.text("UPDATE apps SET maintainer = created_by WHERE maintainer IS NULL"))
-    op.execute(sa.text("UPDATE datasets SET maintainer = created_by WHERE maintainer IS NULL"))
+    # `CREATE INDEX CONCURRENTLY` cannot run within a transaction; wrap it in
+    # `autocommit_block` on PostgreSQL so the composite indexes are built
+    # without blocking concurrent INSERT/UPDATE/DELETE on the hot `apps` and
+    # `datasets` tables. The data backfill itself is split into a separate
+    # revision (b7c8d9e0f1a2) so it can run in id-range batches with explicit
+    # commits between batches, instead of holding row locks for the full
+    # backfill duration. See migration 4474872b0ee6 for the established
+    # concurrent-index pattern.
+    conn = op.get_bind()
+    if _is_pg(conn):
+        with op.get_context().autocommit_block():
+            op.create_index(
+                op.f("app_tenant_maintainer_idx"),
+                "apps",
+                ["tenant_id", "maintainer"],
+                unique=False,
+                postgresql_concurrently=True,
+            )
+            op.create_index(
+                op.f("dataset_tenant_maintainer_idx"),
+                "datasets",
+                ["tenant_id", "maintainer"],
+                unique=False,
+                postgresql_concurrently=True,
+            )
+    else:
+        op.create_index(
+            op.f("app_tenant_maintainer_idx"),
+            "apps",
+            ["tenant_id", "maintainer"],
+            unique=False,
+        )
+        op.create_index(
+            op.f("dataset_tenant_maintainer_idx"),
+            "datasets",
+            ["tenant_id", "maintainer"],
+            unique=False,
+        )
 
 
 def downgrade() -> None:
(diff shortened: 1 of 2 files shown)


AlexMultiAgent:
Closing. Editing already-shipped migration a7c4e9d2f681 (dated 2026-06-15) to switch the backfill to batched is an alembic anti-pattern — production databases that already ran the original `UPDATE apps SET maintainer = created_by` won't re-run the new code. If the synchronous backfill is still a problem on large tables, the right approach is a follow-up migration that batches the remaining work. Happy to open that separately.

PR #37740, opened 2026-06-22
fix(cli): make auth devices revoke --yes a real flag

GareArc:
## Summary

`auth devices revoke` is a destructive command but had **no** confirmation prompt — so its `--yes` flag ("skip confirmation prompt") was dead, doing nothing. `runDevicesRevoke` accepted `opts.yes` but never read it and revoked unconditionally (`devices.ts:99-100`). By contrast `delete member` (also destructive) has a real TTY-gated confirm.

This makes the flag honest by adding the missing prompt:

- Extract a shared `promptConfirm` into `sys/io/prompt.ts`; migrate `delete member` onto it (removes a duplicate copy).
- `runDevicesRevoke` now confirms before revoking on a TTY (`if (opts.yes !== true && opts.io.isErrTTY)`), prompting `Revoke N session(s)? [y/N]` and aborting on decline. `--yes` skips it. This guards `--all` in particular.
- Add `-y` short flag to `revoke` for parity with `delete member -y`.

Behavior: interactive terminals get the prompt; non-TTY (pipes/scripts/CI) get no prompt and revoke directly — no breaking change for automation.

Linear: WTA-957

## Checklist

- [ ] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods.

<sub>CLI gate only: `pnpm type-check`, `pnpm lint`, and 30 targeted `pnpm test` all green. Backend/frontend gates N/A (CLI-only change).</sub>

From Claude Code

### DIFF
--- a/cli/src/commands/auth/devices/_shared/devices.test.ts
+++ b/cli/src/commands/auth/devices/_shared/devices.test.ts
@@ -146,6 +146,43 @@ describe('runDevicesRevoke', () => {
     expect(saved?.hosts[mock.url]).toBeUndefined()
   })
 
+  it('TTY without --yes: prompts and aborts on decline (no revoke)', async () => {
+    const base = bufferStreams('n\n')
+    const io = { ...base, isErrTTY: true }
+    const store = new MemStore()
+    const { reg, active } = buildRegistry(mock.url, 'tester@dify.ai', 'tok-1')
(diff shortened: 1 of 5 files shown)


lin-snow (approved):
LGTM

PR #37766, opened 2026-06-22
fix: isolate agent debug conversations by account

zyssyz123:
## Summary
- Store Agent App debug conversations per account instead of on the shared agent row.
- Return the current user's debug_conversation_id from Agent App list/detail responses.
- Force Agent App debug chat and message history APIs to the current user's debug conversation, rejecting cross-user conversation ids.

## Validation
- Deployed commit to deploy/agent and ran the deploy-agent workflow successfully.
- Updated the agent.dify.dev database schema manually to match the edited pre-release migration.
- Verified on https://agent.dify.dev with two users in the same tenant:
  - each user receives a different debug_conversation_id for the same Agent App;
  - own GET /console/api/agent/{agent_id}/chat-messages returns 200;
  - cross-user GET and POST using another user's conversation_id return 404.
- Local checks:
  - ruff check on modified files: passed;
  - pyrefly check on modified business files: 0 errors;
  - compileall on modified business files: passed.


lyzno1 (approved):

MRZHUH (approved):
LGTM

PR #37771, opened 2026-06-22
fix: include signed content in hit testing results

### LINKED ISSUE #36532: Hit testing returns null sign_content while segment list returns signed content, causing embedded files to fail preview
### Self Checks

### Dify version

main

### Cloud or Self Hosted

Self Hosted (Docker), Self Hosted (Source)

### Steps to reproduce

Steps to reproduce:
1. Create/open a dataset document.
2. Upload an image in the chunk editor. Example uploaded file:
   `395ba4b2-1940-4f2d-84b3-bd46d7532870`
3. Edit a chunk content to include an internal file preview URL:
   ```markdown
   repro-sign-content needle

   ![head](/files/395ba4b2-1940-4f2d-84b3-bd46d7532870/file-preview)
   ```
4. Save the chunk.
5. Confirm the segment list endpoint returns `sign_content` with a signed URL:
   `/console/api/datasets/ecbf3ac0-e5c4-464e-b440-0289d8365934/documents/fed9c931-2b3f-4f32-ba50-13a0bcd76e6c/segments?page=1&limit=10&keyword=&enabled=all`
6. Go to dataset Hit Testing and query:
   ```text
   needle
   ```
7. Inspect the hit-testing response from:
   `/console/api/datasets/ecbf3ac0-e5c4-464e-b440-0289d8365934/hit-testing`

### ✔️ Expected Behavior

Hit-testing results should include usable `segment.sign_content` when segment content contains Dify file preview URLs, consistent with the segment list/detail API. The frontend renders `sign_content || content`, so embedded file previews should use signed URLs.

### ❌ Actual Behavior

The segment list endpoint returns `sign_content`, but hit-testing returns `segment.sign_content: null`. The frontend falls back to raw `segment.content`, which contains an unsigned URL such as:

```text
/files/395ba4b2-1940-4f2d-84b3-bd46d7532870/file-preview
```

The browser then requests that URL directly and receives `400` because required signature params (`timestamp`, `nonce`, `sign`) are missing.

Note, we have several security implications: 

- A fix should not blindly sign every `/files/<uuid>/...` string in arbitrary segment text, because signed preview URLs are short-lived bearer URLs.
- The fix should preserve tenant/dataset/file authorization boundaries before minting signed URLs.
- Signed URLs are operationally sensitive because they may appear in API responses, browser/network logs, screenshots, or referrers, although expiry limits exposure.

vicksiyi:
## Summary
- Add a shared hit-testing retrieval record dump helper that preserves `DocumentSegment.sign_content`.
- Keep injected dataset document metadata behavior unchanged.
- Add regression coverage for file preview URLs in hit-testing records.

Fixes #36532

## Tests
- `python -m compileall -q services/hit_testing_service.py tests/test_containers_integration_tests/services/test_hit_testing_service.py`
- Commit hook Ruff checks passed

## Notes
- `python -m pytest -q tests/test_containers_integration_tests/services/test_hit_testing_service.py -k "signed_segment_content or dump_retrieval_records"` could not run in this local shell because `pytest` is not installed.

### DIFF
--- a/api/services/hit_testing_service.py
+++ b/api/services/hit_testing_service.py
@@ -58,8 +58,9 @@ class HitTestingService:
     @staticmethod
     def _dump_retrieval_record(record: RetrievalSegments) -> dict[str, Any]:
         dumped_record = record.model_dump()
-        if record.segment and isinstance(dumped_record.get("segment"), dict):
-            dumped_record["segment"]["sign_content"] = record.segment.sign_content
+        dumped_segment = dumped_record.get("segment")
+        if record.segment and isinstance(dumped_segment, dict):
+            dumped_segment["sign_content"] = record.segment.sign_content
         return dumped_record
 
     @classmethod
--- a/api/services/hit_testing_service.py
+++ b/api/services/hit_testing_service.py
@@ -55,6 +55,13 @@ class HitTestingService:
             "doc_metadata": document.doc_metadata,
         }
 
+    @staticmethod
+    def _dump_retrieval_record(record: RetrievalSegments) -> dict[str, Any]:
+        dumped_record = record.model_dump()
+        if record.segment and isinstance(dumped_record.get("segment"), dict):
+            dumped_record["segment"]["sign_content"] = record.segment.sign_content
+        return dumped_record
+
     @classmethod
     def _dump_retrieval_records(cls, records: list[RetrievalSegments]) -> list[dict[str, Any]]:
         document_ids = {
@@ -65,7 +72,7 @@ class HitTestingService:
             if isinstance(document_id, str) and document_id
         }
         if not document_ids:
-            return [record.model_dump() for record in records]
+            return [cls._dump_retrieval_record(record) for record in records]
 
         documents = {
             document.id: cls._dump_dataset_document(document)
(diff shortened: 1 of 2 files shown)


api/services/hit_testing_service.py
```diff
@@ -55,6 +55,13 @@ def _dump_dataset_document(document: DatasetDocument) -> dict[str, Any]:
             "doc_metadata": document.doc_metadata,
         }
 
+    @staticmethod
+    def _dump_retrieval_record(record: RetrievalSegments) -> dict[str, Any]:
+        dumped_record = record.model_dump()
+        if record.segment and isinstance(dumped_record.get("segment"), dict):
```
asukaminato0721: why not

```suggestion
        if record.segment and isinstance(record.segment, dict):
```

vicksiyi: Thanks for the suggestion. Addressed in be48eb219d by extracting the dumped segment into a local variable and keeping the dict guard on the serialized `model_dump()` result. `record.segment` is a `DocumentSegment`, so checking it as a dict would make the `sign_content` access invalid. Verified with `python -m py_compile services/hit_testing_service.py`.

vicksiyi: record.segment is the original segment object, not the serialized segment dict. The value that needs the extra field is dumped_record["segment"], so the isinstance check has to stay on that dumped value before mutating it.

crazywoola:
Hi @vicksiyi, thanks for opening this pull request.

### Why this is being closed

Issue #36532 has multiple open pull requests. Under the duplicate-PR rule, pull requests from the issue author are ranked first, then creation time is used as the FIFO tie-breaker. PR #36570 is the retained pull request, so this one is being closed.

### Next steps

Please coordinate on PR #36570 if you would like to contribute further. Thank you for understanding.

PR #37775, opened 2026-06-22
fix: skip empty tool entries in agent_chat config validation

he-yufeng:
`validate_agent_mode_and_set_defaults` iterates `agent_mode["tools"]` and immediately does `key = list(tool.keys())[0]`, which raises `IndexError` when a tool entry is an empty dict `{}` (e.g. from a malformed or legacy/migrated config), crashing config validation.

This adds the same `if not tool: continue` guard that was already merged for the sibling `dataset/manager.py` (#37669) and is present in `variables/manager.py` (#37671), so an empty tool entry is skipped instead of crashing. Added a unit test in `test_agent_chat_app_config_manager.py` covering the empty-entry case (it raised `IndexError` before, passes after).

From Claude Code.

### DIFF
--- a/api/core/app/apps/agent_chat/app_config_manager.py
+++ b/api/core/app/apps/agent_chat/app_config_manager.py
@@ -200,6 +200,11 @@ class AgentChatAppConfigManager(BaseAppConfigManager):
             raise ValueError("tools in agent_mode must be a list of objects")
 
         for tool in agent_mode["tools"]:
+            if not tool:
+                # Skip malformed empty tool entries; list(tool.keys())[0]
+                # would otherwise raise IndexError (same guard as the sibling
+                # dataset/manager.py and variables/manager.py).
+                continue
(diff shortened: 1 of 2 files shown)


he-yufeng:
Gentle ping here — this just drops empty tool entries in the agent-chat completion path so they don't slip through as blank tool calls. Small and contained; let me know if you'd prefer a different spot for the guard or want a test added.

he-yufeng:
One more nudge on this small one. It just drops empty tool entries in the agent_chat completion path instead of raising on them; still mergeable and green on current main.

he-yufeng:
Still reproduces on current main: api/core/app/... apps/agent_chat/app_config_manager.py line 207 still does `list(...keys())[0]` on the entries without an emptiness guard, so an empty {} entry takes the same 500 path. The branch is rebased and the unit tests cover both the empty-entry skip/reject and the normal entries.

PR #37833, opened 2026-06-23
fix(trace): add explicit ids to Langfuse generation

### LINKED ISSUE #37824: Langfuse: message_trace generations dropped (no cost/latency) for Chatbot/Agent/Completion apps since 1.14.0 — generation created without an observation id
### Self Checks

### Dify version

1.14.2 (root cause also present on `main`; latest commit to the affected file at time of writing: `f15a8f02e`)

### Cloud or Self Hosted

Self Hosted (Docker / Kubernetes)

### Steps to reproduce

1. Self-host Dify 1.14.x.
2. Create any **non-workflow** app — a **Chatbot** (basic *or* Agent mode) or a **Completion** app.
3. Enable Langfuse tracing for that app (Monitoring → Tracing → Langfuse; public key, secret key, host).
4. Send a chat message so a `message` record is produced.
5. Open the resulting trace in Langfuse.

### ✔️ Expected Behavior

The trace contains an `llm` **generation** observation carrying `model` + token `usage`, so Langfuse computes **cost** and **latency** — the same way Workflow / Advanced-Chat (chatflow) apps do.

### ❌ Actual Behavior

The trace has **only the top-level `message` node** — no child generation. Token counts appear only as plain fields inside the trace **input** payload (`message_tokens`, `answer_tokens`), not as a generation's `usage`. Because there is no generation observation, **Total Cost is blank and Latency shows 0.00s**. Workflow / Advanced-Chat (chatflow) apps are **not** affected.

---

#### Root cause

`api/providers/trace/trace-langfuse/src/dify_trace_langfuse/langfuse_trace.py`

`message_trace()` creates the trace (with `id=trace_id`) and then an `llm` generation — but the generation is built **without an `id`**:

```python
langfuse_generation_data = LangfuseGeneration(
    name="llm",
    trace_id=trace_id,
    start_time=trace_info.start_time,
    ...
    # <-- no id= is passed
)
self.add_generation(langfuse_generation_data)
```

`LangfuseGeneration.id` defaults to `None`. `add_generation()` then sends the event through the **low-level ingestion API**, and `filter_none_values` strips the `None` id:

```python
data = filter_none_values(langfuse_generation_data.model_dump())  # drops the None "id"
body = CreateGenerationBody(id=data.get("id"), ...)               # id is None
event = IngestionEvent_GenerationCreate(body=body, ...)
self.langfuse_client.api.ingestion.batch(batch=[event])
```

So the generation event reaches Langfuse with **no observation `id`** and is dropped, while the trace (which has its own id) is accepted — hence "trace but no generation, no cost, 0.00s latency".

`workflow_trace()` is **not** affected because it always passes an explicit id:

```python
node_generation_data = LangfuseGeneration(id=node_execution_id, ...)
```

#### Why it regressed

Before 1.14 (e.g. `1.13.3`), `add_generation()` used the **high-level SDK**, which auto-generates a client-side observation id when none is supplied:

```python
self.langfuse_client.generation(**format_generation_data)
```

The trace-provider refactor (#35144) / Langfuse v3-SDK migration switched to the raw ingestion API, which does **not** auto-generate an id — but `message_trace` was never updated to supply one. This broke generation export for every `message_trace` app type (Chatbot, Agent, Completion) from 1.14.0 onward, and it is still present on `main`.

#### Suggested fix

Pass an explicit `id` to the `LangfuseGeneration` created in `message_trace` (and to the other `message`-path observations in this file that currently omit it — `suggested_question_trace`, `moderation_trace`, `dataset_retrieval_trace`, `tool_trace`), e.g.:

```python
langfuse_generation_data = LangfuseGeneration(
    id=str(uuid.uuid4()),
    name="llm",
    trace_id=trace_id,
    ...
)
```

#### Notes

- The same `MessageTraceInfo` flow is shared by other trace providers (e.g. Opik), so it's worth checking whether they have the same missing-id pattern on their low-level ingestion paths.
- Possibly related: #35421 (incorrect token counts on the same `message_trace` path).

MEGHADESAI5:
## Summary

This PR adds explicit observation IDs to Langfuse generation events created in:

* `message_trace`
* `suggested_question_trace`

## Problem

`add_generation()` sends generation events through the Langfuse ingestion API. When `LangfuseGeneration.id` is not provided, `filter_none_values()` removes the field and the resulting `CreateGenerationBody` is created with `id=None`.

This can result in generation observations being dropped by Langfuse, causing traces to be created without corresponding generation records.

## Changes

* Added `id=str(uuid.uuid4())` to generation creation in `message_trace`
* Added `id=str(uuid.uuid4())` to generation creation in `suggested_question_trace`

## Related

Fixes #37824

### DIFF
--- a/api/providers/trace/trace-langfuse/src/dify_trace_langfuse/langfuse_trace.py
+++ b/api/providers/trace/trace-langfuse/src/dify_trace_langfuse/langfuse_trace.py
@@ -357,6 +357,7 @@ class LangFuseDataTrace(BaseTraceInstance):
         )
 
         langfuse_generation_data = LangfuseGeneration(
+            id=str(uuid.uuid4()),
             name="llm",
             trace_id=trace_id,
             start_time=trace_info.start_time,
@@ -405,6 +406,7 @@ class LangFuseDataTrace(BaseTraceInstance):
         )
 
         generation_data = LangfuseGeneration(
+            id=str(uuid.uuid4()),
             name=TraceTaskName.SUGGESTED_QUESTION_TRACE,
             input=trace_info.inputs,
             output=str(trace_info.suggested_question),


brendangooden (approved):
LGTM! @QuantumGhost  can we please get this merged in ASAP? This is causing massive issues for us not being able to correctly report cost in Langfuse

brendangooden:
Following up on my approval above, @QuantumGhost — this is a live production issue for us on `hotfix-1.17.0-fix.14`: the message-level trace lands with no generation, so Langfuse reports `$0` cost and `0s` latency for ~half of all message traffic.

Full repro + numbers are in #37824, including that **advanced-chat / chatflow apps are affected too** — not just Chatbot / Agent / Completion (contrary to the current recap there). Happy to test a build against our instance. Any chance of a review this week?

PR #37841, opened 2026-06-23
fix(web): register audio listeners on real media events

### LINKED ISSUE #37840: AudioPlayer registers listeners on non-existent media events ('paused', 'loaded', 'loadeddate')
### Self Checks

- [x] 【中文用户 & Non English User】请使用英语提交。

### Dify version

main (latest)

### Cloud or Self Hosted

N/A — source-level bug in `web/app/components/base/audio-btn/audio.ts`.

### Steps to reproduce

`AudioPlayer.setCallback` registers media event listeners with names that `HTMLMediaElement` never emits:

```ts
this.audio.addEventListener('paused', ...)     // real event is 'pause'
this.audio.addEventListener('loaded', ...)     // no such media event
this.audio.addEventListener('loadeddate', ...) // typo of 'loadeddata'
```

Because the names are wrong, these listeners never fire in a real browser. The unit test (`audio.spec.ts`) hides this because the mock emits the same misspelled names.

### ✔️ Expected Behavior

Listeners are registered on real media events (`pause`, `loadeddata`), so the callback fires when the audio is paused or its data loads — e.g. a pause not triggered by the button updates the UI state.

### ❌ Actual Behavior

The listeners are dead: the `pause`/`loadeddata` callbacks never run from the audio element. Pause currently only works because `pauseAudio()` calls the callback manually.

eldardada:
## Summary

`AudioPlayer.setCallback` registered three media event listeners with names that `HTMLMediaElement` never emits, so they never fired in a real browser:
- `'paused'` → real event is `pause`
- `'loaded'` → no such media event; the load event is `loadeddata`
- `'loadeddate'` → a typo of `loadeddata`

This fixes the two event names while keeping the callback strings the consumers already handle (`'paused'`, `'loaded'`), so `audio-btn` / `new-audio-button` behave the same but now actually react to the audio element. The `'loadeddate'` listener is removed: it is a dead duplicate of the (now corrected) `loadeddata` listener, and its callback string `'loadeddate'` is handled by no consumer. The `timeupdate` / `canplay` listeners are left untouched — they fire on valid events and changing them would alter behavior.

The unit test mock and its event-name type are updated to emit the real event names; without that the test would keep passing against the misspelled names and hide the bug.

Fixes #37840

## Screenshots

N/A — no visual change; this restores event wiring in `web/app/components/base/audio-btn/audio.ts`.

## Checklist

- [x] I ran the audio-btn test suites (50 passed) and the frontend lint

From Claude Code


eldardada:
For maintainers: this is the original PR for #37840, opened on 2026-06-23 ~6 minutes after I filed the issue. It registers the listeners on the real media events (`pause`, `loadeddata`) instead of the invalid `paused`/`loaded`/`loadeddate` names, with tests updated accordingly.

A later duplicate (#37946, 2026-06-25) targets the same fix — could review consolidate on this earlier PR? Happy to fold in any suggested improvements here. Thanks! 🙏

crazywoola (dismissed):

crazywoola (approved):

PR #37892, opened 2026-06-24
feat(mcp): support MCP protocol 2025-06-18 for workflow-as-MCP server (version negotiation + structured output)

CourTeous33:
## Summary

Upgrades the published-workflow MCP server (`/mcp/server/<server_code>/mcp`) from the hardcoded `2024-11-05` protocol revision to negotiate up to **MCP `2025-06-18`**, and adds the headline 2025-06-18 feature — **structured tool output** — while keeping older clients (e.g. Claude on `2024-11-05`) byte-for-byte compatible.

### What changed

- **Version negotiation** (`handle_initialize`): echoes the client's requested version when supported (`2024-11-05`, `2025-03-26`, `2025-06-18`); otherwise falls back to the server's latest. Previously the response was hardcoded to `2024-11-05`.
- **`MCP-Protocol-Version` header** (`controllers/mcp/mcp.py`): post-initialize requests are validated against the supported set. Absent header defaults to `2025-03-26` (per the spec's back-compat rule); an unsupported value returns a JSON-RPC `INVALID_REQUEST` error echoing the request id (notifications are still accepted). The negotiated version is threaded through to the handlers.
- **Structured tool output** (gated on negotiated `>= 2025-06-18`): `tools/list` advertises `outputSchema` and `title`; `tools/call` returns `structuredContent` (the workflow `outputs` mapping, or `{"answer": ...}` for chat modes) alongside the existing `TextContent`.
- **Backward compatibility**: for `2024-11-05` / absent-header clients the new fields stay `None` and are stripped by the existing `exclude_none` serialization, so responses are unchanged.

No type-layer or DB changes — the MCP pydantic types already carried the 2025-06-18 fields.

## Test Plan

- [x] `uv run pytest tests/unit_tests/core/mcp/ -q` — **415 passed**. New/updated unit tests cover: initialize echo/fallback for every supported version (plus non-string input), `MCP-Protocol-Version` header resolution, structured-output gating including the `2025-03-26` boundary, version threading through the `handle_mcp_request` dispatcher, legacy serialization pinned at the `model_dump` level, and `extract_structured_output` across app modes and malformed workflow responses.
- [x] `uv run pytest tests/test_containers_integration_tests/controllers/mcp/test_mcp.py` — **34 passed** (runs in CI; locally needs Docker). 16 new integration tests drive the real controller with per-request `MCP-Protocol-Version` headers:
  - initialize echoes each supported body version, falls back for unknown ones, and never rejects on header;
  - unsupported header → JSON-RPC `INVALID_REQUEST` echoing the request id (missing id → `null`); notifications still return 202;
  - the header value (and the absent-header default) is threaded into `handle_mcp_request`;
  - serialized JSON contract: modern clients see `outputSchema`/`title`/`structuredContent`; legacy `tools/list` responses are pinned to the exact pre-upgrade key set, so any field that starts leaking to old clients fails loudly.
- [x] `ruff check` and `ruff format --check` clean on all changed files.

### DIFF
--- a/api/controllers/mcp/mcp.py
+++ b/api/controllers/mcp/mcp.py
@@ -1,6 +1,6 @@
 from typing import Any, Union
 
-from flask import Response
+from flask import Response, request
 from flask_restx import Resource
 from pydantic import BaseModel, Field, ValidationError
 from sqlalchemy import select
@@ -9,7 +9,7 @@ from sqlalchemy.orm import Session, sessionmaker
 from controllers.common.schema import register_schema_model
 from controllers.mcp import mcp_ns
 from core.mcp import types as mcp_types
-from core.mcp.server.streamable_http import handle_mcp_request
+from core.mcp.server.streamable_http import handle_mcp_request, negotiate_protocol_version
 from extensions.ext_database import db
 from graphon.variables.input_entities import VariableEntity, VariableEntityType
 from libs import helper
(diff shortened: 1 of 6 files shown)


wylswz (dismissed):

wylswz:
Tested locally

wylswz (approved):

PR #37875, opened 2026-06-24
fix(api): fail fast for unsupported code executor languages

Harsh23Kashyap:
## Summary

- Validate the sandbox runtime mapping at the start of `CodeExecutor.execute_code`.
- Raise `CodeExecutionError` for unsupported languages before calling the remote sandbox service.
- Add a unit test covering the unsupported-language path in `execute_code`.

Fixes #37874

## Problem

`execute_code` previously sent `"language": null` to the sandbox when `code_language_to_running_language.get(language)` returned `None`, producing a cryptic remote error. `execute_workflow_code_template` already fails fast for unsupported languages; this aligns `execute_code` with that behavior.

## Test plan

- [x] `uv run --group dev pytest tests/unit_tests/core/helper/code_executor/test_code_executor.py -q` (10 passed)
- [x] `uv run --group dev ruff check core/helper/code_executor/code_executor.py tests/unit_tests/core/helper/code_executor/test_code_executor.py`

PR #37900, opened 2026-06-24
fix: CodeExecutor.execute_code raises CodeExecutionError for unsupported languages

### LINKED ISSUE #37874: Bug: CodeExecutor.execute_code sends null language to sandbox for unsupported languages
## Description

`CodeExecutor.execute_code` maps `CodeLanguage` to the sandbox runtime via `code_language_to_running_language.get(language)`. When the language is unsupported, `.get()` returns `None` and the request is sent with `"language": null`, which leads to a cryptic remote sandbox error instead of a clear local failure.

`execute_workflow_code_template` already fails fast with `CodeExecutionError("Unsupported language ...")` when no transformer exists, but `execute_code` itself has no equivalent guard.

## Expected behavior

`execute_code` should raise `CodeExecutionError` with an explicit unsupported-language message before calling the sandbox service.

## Proposed fix

Validate the mapped running language at the start of `execute_code` and raise `CodeExecutionError` when missing. Add a focused unit test for the unsupported-language path.

## Context

This is a defensive boundary fix for direct callers of `execute_code` (for example builtin code tools) and keeps error handling consistent with `execute_workflow_code_template`.

Oxygen56:
## Summary

Previously, `execute_code` used `code_language_to_running_language.get(language)` which returned `None` for unsupported languages, sending `"language": null` to the sandbox service and producing a cryptic remote error.

This change adds an explicit guard that raises `CodeExecutionError('Unsupported language ...')` before the sandbox call, matching the existing behavior of `execute_workflow_code_template`.

Closes #37874

## Changes

- **api/core/helper/code_executor/code_executor.py**: Add guard clause for unsupported language at the start of `execute_code`
- **api/tests/unit_tests/core/helper/code_executor/test_code_executor.py**: Add test `test_execute_code_raises_for_unsupported_language`

## Testing

All existing tests pass. New test verifies the unsupported language path raises `CodeExecutionError`.

### DIFF
--- a/api/core/helper/code_executor/code_executor.py
+++ b/api/core/helper/code_executor/code_executor.py
@@ -74,12 +74,16 @@ class CodeExecutor:
         :param code: code
         :return:
         """
+        running_language = cls.code_language_to_running_language.get(language)
+        if running_language is None:
+            raise CodeExecutionError(f"Unsupported language {language}")
+
         url = code_execution_endpoint_url / "v1" / "sandbox" / "run"
 
         headers = {"X-Api-Key": dify_config.CODE_EXECUTION_API_KEY}
 
         data = {
-            "language": cls.code_language_to_running_language.get(language),
+            "language": running_language,
             "code": code,
             "preload": preload,
             "enable_network": True,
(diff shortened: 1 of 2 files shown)


crazywoola:
Thanks for the contribution! We are following FIFO for duplicate fixes linked to the same issue, so we will continue review on the earlier PR #37875 for #37874 and close this one to avoid parallel review. Really appreciate you taking the time to help here.

Issue #37885, opened 2026-06-24
[Bug]: Global db.session.commit() in cloud_edition_billing_rate_limit_check decorator breaks transaction isolation and commits dirty data


QiuYucheng2003:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

Latest main branch

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

This is an architectural transaction management issue identified via code audit in api/controllers/service_api/wraps.py. The execution path that triggers the vulnerability is as follows:

1. A user initiates an API request that routes through the cloud_edition_billing_rate_limit_check decorator (e.g., a knowledge base operation).

2. Prior to evaluating this specific rate limit decorator, the Flask request lifecycle or upstream interceptors make pending modifications to the global SQLAlchemy session (db.session), such as updating a user's last_active timestamp or initializing temporary session states.

3. The user's request exceeds the rate limit threshold (request_count > knowledge_rate_limit.limit).

4. Inside the interceptor, the code executes:
rate_limit_log = RateLimitLog(...)
db.session.add(rate_limit_log)
db.session.commit() # Flaw occurs here
raise Forbidden(...)

5. The request is aborted by the Forbidden exception before reaching the main route controller.

### ✔️ Expected Behavior

The rate limit interceptor should isolate its logging transaction from the main request's SQLAlchemy session context. To adhere to AOP (Aspect-Oriented Programming) isolation principles, it should write the RateLimitLog using a completely separate database connection/session or push the logging task to an async queue (like Celery), ensuring that the main request's transaction lifecycle remains untouched before aborting.

### ❌ Actual Behavior

The explicit db.session.commit() inside the decorator forces a flush of all uncommitted, pending "dirty" objects currently held in the Flask request's db.session.

[...]

PR #37975, opened 2026-06-25
fix: use isolated session in _on_query to prevent premature commit

### LINKED ISSUE #37886: [Bug]: Premature db.session.commit() in DatasetRetrieval._on_query breaks transaction isolation and invalidates rollbacks
### Self Checks

### Dify version

Latest main branch

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

This is an architectural transaction management issue identified via code audit in api/core/rag/retrieval/dataset_retrieval.py. The execution path that triggers the vulnerability is as follows:

1. An agent triggers a RAG retrieval task, routing through either single_retrieve or multiple_retrieve.

2. During the retrieval process, the system calls the _on_query method to persist a DatasetQuery audit log for the operation.

3. Inside _on_query, the code executes db.session.add_all(...) immediately followed by an explicit, global db.session.commit().

4. If a subsequent step in the overarching workflow fails (e.g., vector database timeout, subsequent LLM generation failure, or sensitive word trigger), the top-level orchestrator catches the exception and attempts to execute db.session.rollback() to maintain data consistency.

### ✔️ Expected Behavior

The _on_query method should append audit logs without interfering with the main request's transaction lifecycle. To adhere to transaction isolation, it should write the DatasetQuery using a completely separate database connection/session or push the logging task to an async queue (like Celery). The global db.session should only be committed by the top-level controller once the entire workflow succeeds.

### ❌ Actual Behavior

The explicit db.session.commit() in the side-channel _on_query method forces a flush of all uncommitted, pending "dirty" objects currently held in the Flask request's db.session.

This causes a "transaction split-brain". It violently truncates the atomic transaction boundary of the main business flow. If the workflow encounters an error downstream, the subsequent db.session.rollback() will fail to revert the system state because intermediate modifications (e.g., token deductions, partial node executions) have already been irreversibly committed to the physical database, destroying the agent's backtracking capabilities and leaving dirty data.

Jefsky:
## Problem

`_on_query` calls `db.session.commit()` on the Flask-scoped SQLAlchemy session, which commits **all** pending dirty state from the current request, not just the `DatasetQuery` audit rows. If the downstream workflow fails, the subsequent `db.session.rollback()` cannot revert already-committed modifications, leaving dirty data (e.g. partially recorded token deductions, half-executed node state).

This is the same class of issue as [#37885](https://github.com/langgenius/dify/issues/37885) (billing decorator global commit).

## Root Cause

In `api/core/rag/retrieval/dataset_retrieval.py`, the `_on_query` method:
1. Used `db.session.add_all()` — adding to the shared request-scoped session.
2. Called `db.session.commit()` — flushing all pending changes globally.
3. Called `add_all()` inside the loop instead of after it, re-adding previously accumulated rows on each dataset_id iteration (a latent logic error).

## Fix

Follow the same pattern already demonstrated by `_on_retrieval_end` in the same file: use an **independent session** via `sessionmaker(bind=db.engine).begin()` so the audit log write is fully isolated from the request's main transaction. The context manager auto-commits on success and auto-rolls back on exception.

## Verification

- ✅ The existing `_on_retrieval_end` method (lines ~882-990) already uses `sessionmaker(bind=db.engine).begin()` — this fix applies the same proven pattern.
- ✅ No imports needed — `sessionmaker` is already imported at the top of the file.
- ✅ No behavioral change to callers — `_on_query` signature is unchanged.

Fixes #37886

### DIFF
--- a/api/core/rag/retrieval/dataset_retrieval.py
+++ b/api/core/rag/retrieval/dataset_retrieval.py
@@ -1030,6 +1030,9 @@ class DatasetRetrieval:
     ):
         """
         Persist dataset query audit rows for retrieval requests.
+
+        Uses an independent session to avoid committing the request-scoped
+        db.session, which would break transaction isolation for the caller.
         """
         if not query and not attachment_ids:
             return
@@ -1059,9 +1062,9 @@ class DatasetRetrieval:
                     created_by=created_by,
                 )
                 dataset_queries.append(dataset_query)
-            if dataset_queries:
-                db.session.add_all(dataset_queries)
-        db.session.commit()
+        if dataset_queries:
+            with sessionmaker(bind=db.engine).begin() as session:
+                session.add_all(dataset_queries)
 
     def _retriever(
         self,


crazywoola:
Thanks for the contribution! We are following FIFO for duplicate fixes linked to the same issue, so we will continue review on the earlier PR #37895 for #37886 and close this one to avoid parallel review. Really appreciate you taking the time to help here.

PR #38002, opened 2026-06-26
perf(memory): batch-load message files in TokenBufferMemory to remove N+1 queries

### LINKED ISSUE #38001: perf(memory): N+1 MessageFile queries in TokenBufferMemory.get_history_prompt_messages
### Summary

`TokenBufferMemory.get_history_prompt_messages` loads message files with **two `MessageFile` queries per message** (one for user files, one for assistant files), i.e. **2N+1** database round-trips for an N-message history (N capped at 500).

This method runs on every chat / agent-chat / advanced-chat turn that uses conversation memory, so long conversations cause a large number of DB round-trips on a hot path.

### Where

`api/core/memory/token_buffer_memory.py` → `get_history_prompt_messages`:

```python
for message in messages:
    user_files = db.session.scalars(
        select(MessageFile).where(
            MessageFile.message_id == message.id,
            (MessageFile.belongs_to == "user") | (MessageFile.belongs_to.is_(None)),
        )
    ).all()
    ...
    assistant_files = db.session.scalars(
        select(MessageFile).where(
            MessageFile.message_id == message.id, MessageFile.belongs_to == "assistant"
        )
    ).all()
```

### Proposal

Batch-load the files with two `IN` queries keyed by `message_id` (constant 3 queries total: 1 messages + 2 files), grouping in Python. Filter semantics stay identical (user files still include `belongs_to IS NULL`); the returned prompt messages are unchanged.

This is in the spirit of prior perf work (e.g. #29483, #33014).

I have a small patch ready (~+64/-9, including a regression test that asserts a constant query count) and will open a PR linking this issue.

### Environment

- Affected: 1.14.x and current `main`

weijun-xia:
## Summary

`TokenBufferMemory.get_history_prompt_messages` loaded message files with **two `MessageFile` queries per message** (2N+1 round-trips for an N-message history, N capped at 500). This method runs on every chat / agent-chat / advanced-chat turn that uses conversation memory, so long conversations issue a large number of DB round-trips on a hot path.

This PR replaces the per-message queries with **two batched `IN` queries** keyed by `message_id` (a constant 3 queries: 1 messages + 2 files), grouping the results in Python. The SQL filter semantics are unchanged (user files still include rows whose `belongs_to IS NULL`) and the returned prompt messages are identical.

### Why batched `IN` instead of `selectinload`

There is no `Message → MessageFile` ORM relationship to eager-load (`Message.message_files` is a property that runs its own query and returns processed info, not the raw rows the method needs). Adding a relationship + backref would be a larger, riskier change. Batched `IN` achieves the same 2N → 2 reduction with a minimal diff.

## Test plan

- [x] Added `test_message_files_loaded_with_constant_query_count` — asserts a constant 3 queries for a 5-message thread (previously 1 + 2*5 = 11).
- [x] Updated two existing tests for the batched call pattern (mock files now carry `message_id`).
- [x] `uv run pytest tests/unit_tests/core/memory/test_token_buffer_memory.py` → **57 passed**.
- [x] `ruff check` and `ruff format --check` clean.

Fixes #38001

### DIFF
--- a/api/core/memory/token_buffer_memory.py
+++ b/api/core/memory/token_buffer_memory.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from collections.abc import Sequence
 
 from sqlalchemy import select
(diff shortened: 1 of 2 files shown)


itxaiohanglover:
Excellent optimization! The batch IN queries reduce 2N+1 round-trips to 3. Using defaultdict for grouping is clean. The filter semantics are preserved exactly, and the test mock updates with message_id are correct. This will make a big difference on long conversations.

fatelei (approved):

PR #38030, opened 2026-06-26
fix(web): download markdown file-preview links as attachments

uxgnod:
## Summary

Fixes #34006

This PR fixes Markdown-rendered Dify file-preview links so downloaded files keep their original filename.

Workflow output can render file variables as Markdown links, for example:

```md
[file.pdf](/files/:id/file-preview?timestamp=...&nonce=...&sign=...)
```

Without `as_attachment=true`, browsers may download the file using the route name `file-preview` instead of the original filename. File-card downloads already use attachment mode, so this aligns Markdown file links with the existing file download behavior.

Changes:
- Append `as_attachment=true` to Markdown-rendered `/files/:id/file-preview` links by default.
- Avoid duplicating `as_attachment` when it already exists.
- Add an opt-out prop for Markdown contexts that intentionally need preview behavior.
- Add frontend tests for the default behavior, opt-out behavior, and duplicate prevention.

Validation:
- `pnpm test app/components/base/markdown-blocks/__tests__/link.spec.tsx app/components/base/markdown/__tests__/index.spec.tsx app/components/workflow/run/__tests__/result-text.spec.tsx`
- `pnpm exec vp check --fix app/components/base/markdown-blocks/link.tsx app/components/base/markdown/index.tsx app/components/base/markdown/streamdown-wrapper.tsx app/components/base/markdown-blocks/__tests__/link.spec.tsx`

## Screenshots

| Before | After |
|--------|-------|
| Markdown file link downloads as `file-preview` | Markdown file link downloads with the original filename |

No visual UI change.

## Checklist

From Codex

### DIFF
--- a/web/app/components/base/markdown-blocks/__tests__/link.spec.tsx
+++ b/web/app/components/base/markdown-blocks/__tests__/link.spec.tsx
@@ -170,6 +170,57 @@ describe('Link component', () => {
     expect(link).toHaveAttribute('rel', 'noopener noreferrer')
   })
 
+  it('adds attachment mode to file-preview links by default', () => {
+    mockIsValidUrl.mockReturnValue(true)
+
+    const node = {
+      properties: {
+        href: 'http://localhost:5001/files/123/file-preview?timestamp=1&nonce=2&sign=3',
+      },
+    }
+
+    render(<Link node={node}>doc.pdf</Link>)
+
+    expect(screen.getByText('doc.pdf')).toHaveAttribute(
+      'href',
+      'http://localhost:5001/files/123/file-preview?timestamp=1&nonce=2&sign=3&as_attachment=true',
+    )
+  })
+
+  it('does not duplicate attachment mode for file-preview links', () => {
+    mockIsValidUrl.mockReturnValue(true)
+
+    const node = {
+      properties: {
+        href: 'http://localhost:5001/files/123/file-preview?timestamp=1&nonce=2&sign=3&as_attachment=true',
+      },
+    }
+
+    render(<Link node={node}>doc.pdf</Link>)
+
+    expect(screen.getByText('doc.pdf')).toHaveAttribute(
+      'href',
+      'http://localhost:5001/files/123/file-preview?timestamp=1&nonce=2&sign=3&as_attachment=true',
+    )
+  })
+
+  it('keeps protocol-relative file-preview links protocol-relative', () => {
+    mockIsValidUrl.mockReturnValue(true)
+
+    const node = {
+      properties: {
+        href: '//localhost:5001/files/123/file-preview?timestamp=1#page',
+      },
+    }
+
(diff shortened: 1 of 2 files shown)


jashwanth-reddy-g (approved):
LGTM! The markdown file-preview attachment logic is solid. Parsing the path carefully to append `as_attachment=true` while avoiding duplicate params and handling hashes is very safe and well tested. Approved!

Blackoutta (approved):
Hi @uxgnod  LGTM, I compressed your code into fewer lines and kept the implementation idea the same. Glad to have your contribution ;)

uxgnod:
Thanks @Blackoutta and @gmrnlg1971 for the review and merge!

I originally added the opt-out prop because I wasn’t sure whether every Markdown rendering context should force `file-preview` links into attachment mode. The simplified version makes sense to me now that the behavior is limited to Dify file-preview URLs and covered by tests.

Glad to contribute!

jashwanth-reddy-g (commented):
Great work on adding this feature. The test coverage is solid and the regular expressions accurately target the correct routes.

However, there's a potential edge case with how relative paths are handled that could cause the dummy base URL (`http://dify.local`) to leak into the UI.

### Issue: Dummy Base Leak on Relative Paths
If `href` is a relative URL like `/files/123/file-preview` (which could happen depending on how `isValidUrl` evaluates or if its behavior changes in the future), `href.startsWith('//')` evaluates to `false`. The function will return `url.href` outright, effectively injecting the dummy base and resulting in `http://dify.local/files/123/file-preview?as_attachment=true`.

### Suggested Fix
We can make this logic fully robust for relative paths, protocol-relative paths, and absolute URLs by inspecting `url.origin`. I also highly recommend wrapping the `URL` instantiation in a `try-catch` to avoid front-end crashes on maliciously malformed user input that bypassed prior validation:

```typescript
const withFilePreviewAttachment = (href: string) => {
  if (!filePreviewPathPattern.test(href) || asAttachmentParamPattern.test(href)) {
    return href
  }

  try {
    const dummyBase = 'http://dify.local'
    const url = new URL(href, dummyBase)
    url.searchParams.set('as_attachment', 'true')

    // If the origin matches the dummy base, the original URL was relative
    if (url.origin === dummyBase) {
      return url.pathname + url.search + url.hash
    }

    return href.startsWith('//') ? url.href.replace(url.protocol, '') : url.href
  } catch (e) {
    // Safe fallback if URL parsing fails
    return href
  }
}
```

This drop-in replacement ensures we defensively handle any path type and prevent potential unhandled exceptions. Let me know if you have any questions!

uxgnod:
Thanks for the detailed review, @gmrnlg1971

I understand your concern now. The current `isValidUrl` guard prevents relative paths from reaching this helper today, but I agree that the helper itself could be made more defensive in case that validation changes later.

Do you think it would be better for me to open a small follow-up PR to handle this case?

PR #36998, opened 2026-06-03
fix(api): skip uuidv7() creation when PostgreSQL 18 provides it natively

zeeshan56656:
## Summary

On PostgreSQL 18 a fresh start fails during database migration. Revision `1c9ba48be8e4` runs `CREATE FUNCTION uuidv7()`, but PostgreSQL 18 already ships a native `pg_catalog.uuidv7()`. Our create collides with the built-in, and the unqualified `COMMENT ON FUNCTION uuidv7` that follows becomes ambiguous, so the migration aborts and the API container never finishes booting. Reported against PostgreSQL 18.3 in #36907.

This PR guards the create. In `upgrade()` we look up `pg_proc`/`pg_namespace` for a `uuidv7` function in the `pg_catalog` schema before creating our own. When the server already provides one (PostgreSQL 18), we skip both the `CREATE FUNCTION uuidv7()` and its `COMMENT`, and call sites that use `uuidv7()` then resolve to the native implementation, which has a compatible zero-argument signature. The `uuidv7_boundary(timestamptz)` helper is still created unconditionally because PostgreSQL 18 does not provide it. This reuses the same `_is_pg(conn)` dialect-guard style that #28188 added to this file.

`downgrade()` now uses `DROP FUNCTION IF EXISTS public.uuidv7()` and `DROP FUNCTION IF EXISTS public.uuidv7_boundary(timestamptz)`. The `IF EXISTS` plus the explicit `public` schema make a rollback a no-op on PostgreSQL 18 (where the public copy was never created) and keep it from ever touching the built-in.

Behavior on PostgreSQL 13 to 17 is unchanged: there is no native `uuidv7`, so both functions are created exactly as before. The change is scoped to this one migration file.

Fixes #36907

## Tests

Added `api/tests/unit_tests/migrations/test_uuidv7_pg18_migration.py`. It mocks the Alembic bind and covers each branch separately:

- native `uuidv7` absent (PostgreSQL 13 to 17): both `uuidv7()` and `uuidv7_boundary()` are created
- native `uuidv7` present (PostgreSQL 18): `uuidv7()` is skipped, `uuidv7_boundary()` is still created, and the probe queries the `pg_catalog` schema
- `downgrade()` emits the `IF EXISTS` public-scoped drops
- non-PostgreSQL dialects stay a no-op for both `upgrade()` and `downgrade()`

```
python -m pytest api/tests/unit_tests/migrations/test_uuidv7_pg18_migration.py -v
```

5 passed.

## Checklist

[...]

PR #38031, opened 2026-06-26
fix(migration): avoid AmbiguousFunction on uuidv7() under PostgreSQL 18

shaba:
## Summary

PostgreSQL 18 ships a built-in `pg_catalog.uuidv7()`. The migration
`2025_07_02_2332-1c9ba48be8e4_add_uuidv7_function_in_sql` unconditionally
creates `public.uuidv7()` and then runs `COMMENT ON FUNCTION uuidv7 ...`
**without** an argument signature. On PG 18 both `public.uuidv7()` and the
built-in `pg_catalog.uuidv7()` are visible, so the unqualified `COMMENT`
resolves ambiguously and the migration aborts with:

```
psycopg2.errors.AmbiguousFunction: function name "uuidv7" is not unique
```

This blocks `flask db upgrade` (fresh installs and upgrades) on PostgreSQL 18.

## Fix

- Create `public.uuidv7()` only when no 0-argument `uuidv7` already exists
  (guarded by a `DO` block that checks `pg_proc`).
- Qualify the `COMMENT` with the explicit signature `uuidv7()`.
- Make `downgrade()` symmetric and safe:
  `DROP FUNCTION IF EXISTS public.uuidv7()` /
  `public.uuidv7_boundary(timestamptz)`.

Behaviour on PostgreSQL < 18 is unchanged; the migration is now idempotent
and PG 18 compatible.

## How tested

- `flask db upgrade` on a fresh PostgreSQL 18 database — previously failed
  with `AmbiguousFunction`, now succeeds.
- PostgreSQL < 18 — unchanged (function and comment created as before).

### DIFF
--- a/api/migrations/versions/2025_07_02_2332-1c9ba48be8e4_add_uuidv7_function_in_sql.py
+++ b/api/migrations/versions/2025_07_02_2332-1c9ba48be8e4_add_uuidv7_function_in_sql.py
@@ -47,28 +47,43 @@ def upgrade():
     conn = op.get_bind()
     
     if _is_pg(conn):
-        # PostgreSQL: Create uuidv7 functions
+        # PostgreSQL: Create uuidv7 functions.
+        # PostgreSQL 18 ships a built-in pg_catalog.uuidv7(). Creating
+        # public.uuidv7() is fine, but `COMMENT ON FUNCTION uuidv7` without an
+        # argument signature becomes ambiguous (public + pg_catalog) and fails
+        # with AmbiguousFunction. Create ours only if no 0-arg uuidv7 exists yet
+        # and qualify the COMMENT with the explicit signature.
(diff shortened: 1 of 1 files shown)


jashwanth-reddy-g (approved):
LGTM! The solution correctly addresses the `AmbiguousFunction` error that occurs in PostgreSQL 18 due to the built-in `pg_catalog.uuidv7()`. Using a `DO` block with explicit signatures (`uuidv7()`) is a solid fix for the migration script and handles rollback gracefully too. Approved!

nicolaerosia:
@wylswz is there a blocker here from merging this? we need this as well

wylswz:
> @wylswz is there a blocker here from merging this? we need this as well

I think we have already merged one, please see #36998

wylswz:
Closing as duplicated.

PR #38032, opened 2026-06-26
perf: batch MessageFile queries to fix N+1 in TokenBufferMemory

### LINKED ISSUE #38001: perf(memory): N+1 MessageFile queries in TokenBufferMemory.get_history_prompt_messages

EvanYao826:
## Summary

`TokenBufferMemory.get_history_prompt_messages` was issuing **2N+1** DB round-trips for N messages (two `MessageFile` queries per message via `db.session.scalars`). On the hot path of every chat turn, this causes significant overhead for long conversations.

## Fix

Replace the per-message `MessageFile` queries with a single batch `IN` query before the loop, then group results by `message_id` and `belongs_to` in Python.

**Before:** 2N+1 queries (1 message query + 2 per message)
**After:** 2 queries (1 message query + 1 batch file query)

## Changes

- `api/core/memory/token_buffer_memory.py`: Batch-fetch `MessageFile` records, group by `(message_id, belongs_to)`, look up from dicts in the loop
- `api/tests/unit_tests/core/memory/test_token_buffer_memory.py`: New test file with 3 tests:
  - Verifies only 2 DB queries are issued (not 2N+1)
  - Verifies files are correctly grouped by message_id and belongs_to (user/None vs assistant)
  - Verifies no file query is issued when there are no messages

## Related Issues

Fixes #38001

### DIFF
--- a/api/core/memory/token_buffer_memory.py
+++ b/api/core/memory/token_buffer_memory.py
@@ -156,9 +156,7 @@ class TokenBufferMemory:
         # Batch-fetch all MessageFile records to avoid N+1 queries
         message_ids = [m.id for m in messages]
         if message_ids:
-            all_files = db.session.scalars(
-                select(MessageFile).where(MessageFile.message_id.in_(message_ids))
-            ).all()
+            all_files = db.session.scalars(select(MessageFile).where(MessageFile.message_id.in_(message_ids))).all()
         else:
             all_files = []
 
--- a/api/core/memory/token_buffer_memory.py
+++ b/api/core/memory/token_buffer_memory.py
@@ -153,16 +153,29 @@ class TokenBufferMemory:
 
         messages = list(reversed(thread_messages))
 
+        # Batch-fetch all MessageFile records to avoid N+1 queries
+        message_ids = [m.id for m in messages]
+        if message_ids:
+            all_files = db.session.scalars(
+                select(MessageFile).where(MessageFile.message_id.in_(message_ids))
+            ).all()
+        else:
+            all_files = []
+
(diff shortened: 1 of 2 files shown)


crazywoola:
See https://github.com/langgenius/dify/issues/38001

Issue #38047, opened 2026-06-26
[1.15.0] Under certain interactions, objects underneath a toast become unclickable


kurokobo:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.15.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

I’ll describe a case where it’s easy to reproduce.
I haven’t investigated other toasts, but if they share the same codebase, I think the issue may also be reproducible there.

1. Open any app in Studio.
2. Go to Publish -> Publish Update.
3. Click the X button on the displayed toast.
4. Before the toast finishes fading out and completely disappears, click the center area of the toast again.

<img width="805" height="286" alt="Image" src="https://github.com/user-attachments/assets/b4ef91c9-7c03-473d-b9a8-31ba1e208baf" />

<img width="792" height="324" alt="Image" src="https://github.com/user-attachments/assets/9292b725-9d19-4d27-ad22-ec3d2c7a6b2b" />

### ✔️ Expected Behavior

The toast disappears, and Features and the Publish button appear to become usable again.

### ❌ Actual Behavior

Although the toast disappears, the Features and Publish buttons that were underneath it become unclickable.

<img width="806" height="345" alt="Image" src="https://github.com/user-attachments/assets/e469efec-07dc-476c-9e59-f4106abd10fe" />

Looking at it in DevTools, it seems like a leftover fragment of the toast remains on the upper layer.

<img width="1763" height="669" alt="Image" src="https://github.com/user-attachments/assets/a4516675-9971-4bfa-afad-6e14b1b2eae8" />

vicksiyi:
I opened a small fix for this in https://github.com/langgenius/dify/pull/38049.

[...]

PR #38049, opened 2026-06-26
fix: ignore exiting toast pointer events

vicksiyi:
Summary\n- Disable pointer events on toast roots while Base UI marks them as exiting.\n- Cover the toast root class so invisible exiting toasts cannot block underlying controls.\n\nTests\n- pnpm --filter @langgenius/dify-ui test -- src/toast/__tests__/index.spec.tsx\n\nFixes #38047


lyzno1:
Thanks for working on this and for taking the time to investigate the issue.

After looking into it, I think this needs to be fixed in a slightly different way to better fit the existing architecture / avoid edge cases. I’ve opened #38063 with the approach I’d like to move forward with.

I’m going to close this PR in favor of that one, but I really appreciate your contribution here.

PR #38070, opened 2026-06-27
fix: remove premature db.session.commit/close in utility methods

### LINKED ISSUE #37884: [Bug]: Premature db.session.close() in ToolEngine._create_message_files causes DetachedInstanceError
### Self Checks

### Dify version

Latest main branch

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Since this is a code-level architectural issue, here is the execution path that triggers the bug:

1. An agent invokes a multimodal tool that returns binary files (e.g., image generation, audio synthesis).

2. The framework processes the response in api/core/tools/tool_engine.py.

3. The execution flow calls the static method ToolEngine._create_message_files(...) to persist the attachments.

4. Inside this method, after the db.session.add(...), commit(), and refresh(...) operations, the code explicitly executes db.session.close().

5. The execution flow returns to the main agent scheduler.

6. Subsequent operations in the same request/task lifecycle attempt to update the agent's reasoning history, billing logs, or access lazy-loaded ORM properties.

### ✔️ Expected Behavior

The _create_message_files method should perform its specific persistence tasks (add, commit) but must not close the session. The db.session lifecycle (Scoped Session) should be entirely managed by the global Flask/FastAPI request context or Celery task teardown logic.

### ❌ Actual Behavior

The explicit db.session.close() call within a local utility method violently severs the database connection bound to the current thread/context.

When the execution returns to the main scheduling loop, any further database interactions using the global db.session, or any attempts to access attributes of ORM objects queried prior to this tool invocation, will instantly throw a sqlalchemy.orm.exc.DetachedInstanceError or underlying connection errors. This not only aborts the current agent execution and destroys subsequent persistence operations but also disrupts the connection pool's standard check-in/check-out lifecycle.

### LINKED ISSUE #37886: [Bug]: Premature db.session.commit() in DatasetRetrieval._on_query breaks transaction isolation and invalidates rollbacks

EvanYao826:
## Summary

Two utility methods in the core codebase improperly manage the global `db.session` lifecycle, breaking transaction isolation and causing runtime errors.

### 1. `DatasetRetrieval._on_query` — premature `db.session.commit()` (Fixes #37886)

The `_on_query` method persists `DatasetQuery` audit logs with `db.session.add_all(...)` followed by an explicit `db.session.commit()`. This forces a flush of **all** pending dirty objects in the session, not just the audit logs. If a downstream step fails and the caller attempts `db.session.rollback()`, the rollback is ineffective because intermediate state has already been irreversibly committed.

**Fix**: Remove the explicit `db.session.commit()`. The audit logs are added to the session and will be committed when the caller's transaction completes.

### 2. `ToolEngine._create_message_files` — premature `db.session.close()` (Fixes #37884)

After persisting `MessageFile` records, the method calls `db.session.close()`, which severs the database connection bound to the current thread. Subsequent operations in the same request lifecycle that attempt to access ORM objects queried before this call raise `DetachedInstanceError`.

**Fix**: Remove `db.session.close()`. The scoped session lifecycle is managed by the Flask request context.

## Tests

- Updated 3 test assertions that verified the now-removed `commit`/`close` calls
- All 135 tests in the affected test files pass

### DIFF
--- a/api/core/rag/retrieval/dataset_retrieval.py
+++ b/api/core/rag/retrieval/dataset_retrieval.py
@@ -1061,7 +1061,6 @@ class DatasetRetrieval:
                 dataset_queries.append(dataset_query)
             if dataset_queries:
                 db.session.add_all(dataset_queries)
-        db.session.commit()
 
     def _retriever(
         self,
--- a/api/core/rag/retrieval/dataset_retrieval.py
+++ b/api/core/rag/retrieval/dataset_retrieval.py
@@ -1060,7 +1060,9 @@ class DatasetRetrieval:
                 )
                 dataset_queries.append(dataset_query)
             if dataset_queries:
-                db.session.add_all(dataset_queries)
+                with session_factory.create_session() as session:
+                    session.add_all(dataset_queries)
+                    session.commit()
 
     def _retriever(
         self,
--- a/api/core/tools/tool_engine.py
+++ b/api/core/tools/tool_engine.py
@@ -379,6 +379,4 @@ class ToolEngine:
 
             result.append(message_file.id)
 
-        db.session.close()
-
         return result
--- a/api/tests/unit_tests/core/rag/retrieval/test_dataset_retrieval.py
+++ b/api/tests/unit_tests/core/rag/retrieval/test_dataset_retrieval.py
@@ -3888,7 +3888,12 @@ class TestDatasetRetrievalAdditionalHelpers:
         trace_manager.add_trace_task.assert_not_called()
 
     def test_on_query(self, retrieval: DatasetRetrieval) -> None:
-        with patch("core.rag.retrieval.dataset_retrieval.db.session") as mock_session:
+        with patch(
+            "core.rag.retrieval.dataset_retrieval.session_factory.create_session"
+        ) as mock_create:
+            mock_session = MagicMock()
+            mock_create.return_value.__enter__.return_value = mock_session
+
             retrieval._on_query(
                 query=None,
                 attachment_ids=None,
@@ -3908,9 +3913,15 @@ class TestDatasetRetrievalAdditionalHelpers:
                 user_id="u1",
             )
             mock_session.add_all.assert_called()
+            mock_session.commit.assert_called()
 
     def test_on_query_normalizes_workflow_end_user_role(self, retrieval: DatasetRetrieval) -> None:
-        with patch("core.rag.retrieval.dataset_retrieval.db.session") as mock_session:
+        with patch(
+            "core.rag.retrieval.dataset_retrieval.session_factory.create_session"
+        ) as mock_create:
+            mock_session = MagicMock()
+            mock_create.return_value.__enter__.return_value = mock_session
+
             retrieval._on_query(
                 query="python",
                 attachment_ids=None,
--- a/api/tests/unit_tests/core/rag/retrieval/test_dataset_retrieval.py
+++ b/api/tests/unit_tests/core/rag/retrieval/test_dataset_retrieval.py
@@ -3908,7 +3908,6 @@ class TestDatasetRetrievalAdditionalHelpers:
                 user_id="u1",
             )
             mock_session.add_all.assert_called()
-            mock_session.commit.assert_called()
 
     def test_on_query_normalizes_workflow_end_user_role(self, retrieval: DatasetRetrieval) -> None:
         with patch("core.rag.retrieval.dataset_retrieval.db.session") as mock_session:
@@ -3926,7 +3925,6 @@ class TestDatasetRetrievalAdditionalHelpers:
 
             assert len(added_queries) == 1
             assert added_queries[0].created_by_role == CreatorUserRole.END_USER
-            mock_session.commit.assert_called_once()
 
     def test_handle_invoke_result(self, retrieval: DatasetRetrieval) -> None:
         usage = LLMUsage.empty_usage()
--- a/api/tests/unit_tests/core/tools/test_tool_engine.py
+++ b/api/tests/unit_tests/core/tools/test_tool_engine.py
@@ -142,7 +142,6 @@ def test_create_message_files_and_invoke_generator():
 
     assert ids == ["mf-1", "mf-2"]
     assert mock_db.session.add.call_count == 2
-    mock_db.session.close.assert_called_once()
 
     tool = _build_tool()
     invoked = list(ToolEngine._invoke(tool, {"a": 1}, user_id="u"))


itxaiohanglover:
Good fix! Removing premature db.session.commit/close prevents partial commits and connection churn. The test updates correctly remove the commit assertions.

api/core/rag/retrieval/dataset_retrieval.py
```diff
@@ -1061,7 +1061,6 @@ def _on_query(
                 dataset_queries.append(dataset_query)
             if dataset_queries:
                 db.session.add_all(dataset_queries)
-        db.session.commit()
```
fatelei, line 1064: i think remove it is not right

db.session.commit() flushes all pending dirty objects in the shared session, not just the audit log — which can defeat a caller's rollback. But the real problem is that the audit log shares the caller's db.session transaction at all. Blindly
deleting the commit fixes the "don't flush unrelated state" half while breaking the "audit log must persist" half.

Suggested fix — decouple via a dedicated session

if dataset_queries:
   with session_factory.create_session() as session:
       session.add_all(dataset_queries)
       session.commit()

EvanYao826:
Thanks for the thorough review @fatelei! You raise a valid point — removing `db.session.commit()` does shift the audit log's persistence timing from "immediately" to "when the caller commits." 

A few considerations for the current approach:

1. **`_on_query` audit context**: This method logs a `DatasetQuery` record during a dataset search operation. If the caller's operation fails and rolls back, not persisting this audit log is arguably correct — the query never completed successfully. This is the same semantic that most web frameworks follow for request-scoped audit logs.

2. **`_create_message_files` is the clearer case**: Removing `db.session.close()` is unambiguously correct — closing the scoped session mid-request causes `DetachedInstanceError` for subsequent ORM access.

3. **Dedicated session approach**: You're absolutely right that using a separate session (e.g. `with db.session.begin(subtransactions=True)` or a dedicated `Session` instance) would be the more rigorous solution. It would:
   - Guarantee the audit log persists regardless of caller's transaction outcome
   - Avoid polluting the caller's session state

However, introducing a dedicated session pattern in these utility methods would be a larger architectural change that might be better scoped as a follow-up PR with broader alignment on Dify's session management conventions. The immediate fix here addresses the two concrete bugs (#37886 and #37884) without introducing a new pattern.

If you think the dedicated session approach is better even for this minimal fix, I'm happy to revise. What do you think?

fatelei:
> Thanks for the thorough review @fatelei! You raise a valid point — removing `db.session.commit()` does shift the audit log's persistence timing from "immediately" to "when the caller commits."
> 
> A few considerations for the current approach:
> 
> 1. **`_on_query` audit context**: This method logs a `DatasetQuery` record during a dataset search operation. If the caller's operation fails and rolls back, not persisting this audit log is arguably correct — the query never completed successfully. This is the same semantic that most web frameworks follow for request-scoped audit logs.
> 2. **`_create_message_files` is the clearer case**: Removing `db.session.close()` is unambiguously correct — closing the scoped session mid-request causes `DetachedInstanceError` for subsequent ORM access.
> 3. **Dedicated session approach**: You're absolutely right that using a separate session (e.g. `with db.session.begin(subtransactions=True)` or a dedicated `Session` instance) would be the more rigorous solution. It would:
>    
>    * Guarantee the audit log persists regardless of caller's transaction outcome
>    * Avoid polluting the caller's session state
> 
> However, introducing a dedicated session pattern in these utility methods would be a larger architectural change that might be better scoped as a follow-up PR with broader alignment on Dify's session management conventions. The immediate fix here addresses the two concrete bugs (#37886 and #37884) without introducing a new pattern.
> 
> If you think the dedicated session approach is better even for this minimal fix, I'm happy to revise. What do you think?

I understand the request-scoped transaction semantics, but my concern is that without `db.session.commit()`, `db.session.add()` only
  stages the `DatasetQuery` in the current session. The audit log is not actually persisted unless the outer caller later commits. If the
  caller does not commit, or if the transaction rolls back, the audit record will be lost.

EvanYao826:
Thanks for the review, @fatelei! I've updated the PR based on your suggestion.

Instead of simply removing `db.session.commit()`, the audit log now uses a **dedicated session** via `session_factory.create_session()`:

```python
if dataset_queries:
    with session_factory.create_session() as session:
        session.add_all(dataset_queries)
        session.commit()
```

This way:
1. ✅ The audit log persists immediately and independently (your "must persist" concern)
2. ✅ The caller's transaction is unaffected — no premature flush of unrelated dirty objects
3. ✅ No `DetachedInstanceError` risk from `db.session.close()`

Tests updated and passing. Please take another look when you have a chance!

crazywoola:
Hi @EvanYao826, thanks for opening this pull request.

### Why this is being closed

The linked issue set has multiple open pull requests. Under the duplicate-PR rule, pull requests from the issue author are ranked first, then creation time is used as the FIFO tie-breaker. The retained result for this overlap is: issue #37884 -> PR #37918; issue #37886 -> PR #37918. This pull request is therefore being closed as a remaining duplicate.

### Next steps

PR #38086, opened 2026-06-27
fix(core): replace bare except with specific exceptions in CotAgentOutputParser

### LINKED ISSUE #38085: Bare except: in CotAgentOutputParser silently swallows all agent code-block parse failures
### Self Checks

### Dify version

1.15.0

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

1. Clone the Dify repository and set up the backend locally (`api/README.md`)
2. Configure a ReAct / CoT agent with any tool (e.g. web search)
3. Use a model that returns tool calls inside a markdown code fence , or simulate it by having the LLM return a truncated/malformed JSON block:

{"action": "search", "action_input":

4. Send a message that triggers tool use
5. Observe `docker-compose logs api` (or your local Flask logs)

**Result:** The agent silently produces no action. No exception is raised, no warning is logged, nothing appears in the logs. The failure is completely invisible.

**Root cause (code pointer):**
`api/core/agent/output_parser/cot_output_parser.py`, function `extra_json_from_code_block`, line 52 — bare `except:` swallows all exceptions including `KeyboardInterrupt` and `SystemExit`.

### ✔️ Expected Behavior

When the LLM returns a malformed JSON code block, the parser should:

1. Catch only `json.JSONDecodeError` and `ValueError` , the specific exceptions that `json.loads` and `re.sub` can raise
2. Emit a `WARNING` log entry containing the exception message, so the failure is visible in `docker-compose logs` and any connected log aggregator (Datadog, Loki, etc.)
3. Return `[]` and allow the agent stream to continue gracefully , same as today, but no longer silent

Critically, `KeyboardInterrupt` and `SystemExit` should **not** be caught here and must propagate normally to allow clean process shutdown.

### ❌ Actual Behavior

The agent silently returns no action and produces zero log output. The bare `except:` on line 52 of `api/core/agent/output_parser/cot_output_parser.py` catches every exception, including `KeyboardInterrupt` and `SystemExit` , and discards it, returning `[]` with no trace of the failure anywhere.

There is no way to distinguish between "the LLM returned a valid response with no tool call" and "the LLM returned a malformed code block that failed to parse" from logs alone.

ShauryaaSharma:
fixes #38085 

## What does this PR do?

Replaces a bare `except:` in `CotAgentOutputParser.extra_json_from_code_block` with `except (json.JSONDecodeError, ValueError)` and adds a `logger.warning` so malformed LLM code-block JSON surfaces in production logs instead of
being silently discarded.

## Why?

A bare `except:` has two concrete problems:

- It swallows `KeyboardInterrupt` and `SystemExit`, which can prevent clean process shutdown inside a streaming generator
- It produces zero log output, making it impossible to know when an LLM returns a malformed code block

## Changes

- `api/core/agent/output_parser/cot_output_parser.py` , replace bare `except:` with `except (json.JSONDecodeError, ValueError) as e` + `logger.warning`
- `api/tests/unit_tests/core/agent/output_parser/test_cot_fix_verification.py` 20 new tests across 4 suites

## Tests

- AST-level proof the bare except is gone
- `logger.warning` is called with exception details on bad JSON
- No spurious warnings on valid code blocks or plain text
- `KeyboardInterrupt` propagates correctly out of the generator
- 11 regression tests confirm all existing streaming behaviours are unaffected


ShauryaaSharma:
Hey @dosu Pushed an update to fix the CI failures.

The autofix/ruff errors were straightforward, renamed the ambiguous variable `l` to `line` (E741) and removed the `caplog=None` default argument (PT028).

The 4 test failures were more interesting. The original test file used `sys.modules.setdefault()` to stub out `graphon` and `AgentScratchpadUnit` at import time this worked locally where `graphon` isn't installed, but silently broke in CI because `graphon==0.5.3` is installed there and the real modules are already loaded before our stubs run. `setdefault()` skips 
existing entries, so the mock was pointing to an orphaned `MagicMock` that the parser never actually used.

Fixed by dropping the stub approach entirely and using `mocker.patch()` on `AgentScratchpadUnit.Action` directly, the same pattern used in the existing `test_cot_output_parser.py`. Works correctly whether or not `graphon` is installed.

PR #38089, opened 2026-06-27
fix(storage): replace bare except with specific ClientError handling in exists()

### LINKED ISSUE #38088: [Bug] exists() in AwsS3Storage and OracleOCIStorage silently treats all errors as file not found
### Self Checks

### Dify version

1.15.0

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

1. Configure Dify with S3 or Oracle OCI as the storage backend
2. Intentionally break the IAM/credentials so head_object returns a 403 Forbidden
   (or simulate S3 throttling by returning a 429, or take S3 down to get a 500)
3. Trigger any flow that calls storage.exists() — for example:
   - A Firecrawl website extraction (services/website_service.py:372)
   - A Jieba keyword index save (core/rag/datasource/keyword/jieba/jieba.py:156)
   - A storage migration command (commands/storage.py:677)
4. Check docker-compose logs api for any error output

### ✔️ Expected Behavior

exists() should return False only for genuine "file not found" responses (404 or NoSuchKey). Any other ClientError, 403 Forbidden, 429 throttling, 500 server error, should be re-raised so the caller can handle or surface it.

This is already how load_once() and load_stream() behave in both files.
exists() should be consistent with them.

### ❌ Actual Behavior

exists() silently returns False for every exception without any log output. 

A 403 (broken IAM), 429 (throttling), or 500 (S3 outage) is completely indistinguishable from "file not found" , the caller proceeds as if the file doesn't exist with no indication anything went wrong.

In the case of a 403, this can cause the app to overwrite or skip files that actually exist, creating a data loss risk.

ShauryaaSharma:
fixes #38088 

## What does this PR do?

Replaces bare `except:` in `AwsS3Storage.exists()` and `OracleOCIStorage.exists()`with `except ClientError` that re-raises on anything that is not a genuine 404 / NoSuchKey response.

## Why?

Both storage backends used a bare `except:` that silently returned `False` for every possible exception, including:

- 403 Forbidden (broken IAM policy) → app thinks file doesn't exist, may overwrite or skip it entirely → data loss risk
- 429 / SlowDown (S3 throttling) → app proceeds as if storage is empty during a traffic spike → silent degradation
- 500 Internal Server Error → S3 outage appears as missing files
- Network failures → same

Every other method in both files (load_once, load_stream, __init__) already uses specific ClientError handling correctly. exists() was the only outlier.

## Changes

- `api/extensions/storage/aws_s3_storage.py` — replace bare `except:` with `except ClientError`, return False only on 404 / NoSuchKey, raise otherwise
- `api/extensions/storage/oracle_oci_storage.py` — same fix, also tidied import order to stdlib → third-party → local
- `api/tests/unit_tests/extensions/storage/test_s3_oci_storage_exists.py` 16 new tests across both backends

## Tests

- Returns True when head_object succeeds
- Returns False on 404 and NoSuchKey
- Raises ClientError on 403 Forbidden
- Raises ClientError on 429 / SlowDown throttling
- Raises ClientError on 500 Internal Server Error
- Raises on non-ClientError exceptions (ConnectionError, etc.)
- KeyboardInterrupt propagates correctly
- All 8 cases verified for both AwsS3Storage and OracleOCIStorage


api/core/helper/position_helper.py
```diff
@@ -1,21 +1,20 @@
...
 from collections.abc import Callable
-from functools import lru_cache
+from functools import cache
 
 from configs import dify_config
 from core.tools.utils.yaml_utils import load_yaml_file_cached
 
 
-@lru_cache(maxsize=128)
+@cache
```
fatelei, line 10: why change this, cache decorator has no memory limit, memory can consume a lot

PR #38131, opened 2026-06-29
fix(web): prevent ssePost error handler from throwing

VectorPeak:
Re-submitted after fork recovery. Original PR #37799 was closed due to fork deletion. Fixes SSE error handler that was throwing instead of handling errors gracefully.

### DIFF
--- a/web/service/base.ts
+++ b/web/service/base.ts
@@ -611,7 +611,7 @@ export const ssePost = async (
       )
     })
     .catch((e) => {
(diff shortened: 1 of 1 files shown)


VectorPeak:
Closing as superseded: current main already normalizes the caught error with \\String(e)\\ before checking \\.includes('TypeError: Cannot assign to read only property')\\, which covers the issue this PR fixed and avoids the merge conflict in \\web/service/base.ts\\. Thanks for catching this.

PR #38166, opened 2026-06-29
feat: support dataset permission migrate to rbac

fatelei:
> [!IMPORTANT]
>

## Summary

support migrate dataset permission to rbac service

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/commands/rbac.py
+++ b/api/commands/rbac.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 from collections.abc import Iterator
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
@@ -8,8 +9,11 @@ from sqlalchemy import select
 
 from configs import dify_config
 from core.db.session_factory import session_factory
-from models import TenantAccountJoin, TenantAccountRole
-from services.enterprise.rbac_service import ListOption, RBACService
+from core.rbac import RBACResourceWhitelistScope
+from models import Dataset, DatasetPermission, DatasetPermissionEnum, TenantAccountJoin, TenantAccountRole
+from services.enterprise.rbac_service import ListOption, RBACService, ReplaceMemberBindings, ReplaceUserAccessPolicies
+
+_RBAC_DEFAULT_ACCESS_POLICY_ID = "default"
 
 _LEGACY_ROLE_TO_BUILTIN_TAG = {
     TenantAccountRole.OWNER.value: "owner",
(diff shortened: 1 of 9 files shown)


api/commands/rbac.py
```diff
@@ -110,3 +112,154 @@ def migrate_member_roles_to_rbac(tenant_id: str | None, dry_run: bool) -> None:
         click.echo(click.style("Dry run completed. No RBAC bindings were written.", fg="yellow"))
     else:
         click.echo(click.style(f"RBAC member-role migration completed. Migrated {migrated_count} members.", fg="green"))
+
+
+def _dataset_permission_value(permission: DatasetPermissionEnum | str | None) -> str:
```
QuantumGhost: Why not use enum values? Enum values are easier to read and maintain.

api/commands/rbac.py
```diff
@@ -110,3 +112,154 @@ def migrate_member_roles_to_rbac(tenant_id: str | None, dry_run: bool) -> None:
...
+        return permission
+    return permission.value
+
+
+def _rbac_dataset_scope_for_legacy_permission(permission: DatasetPermissionEnum | str | None) -> str:
+    permission_value = _dataset_permission_value(permission)
+    if permission_value == DatasetPermissionEnum.ALL_TEAM.value:
+        return "all"
+    if permission_value in {DatasetPermissionEnum.ONLY_ME.value, DatasetPermissionEnum.PARTIAL_TEAM.value}:
+        return "specific"
```
QuantumGhost: Is there a enumeration class for `scope`?

If there is, I would suggest using enum instead.

api/commands/rbac.py
```diff
@@ -110,3 +112,154 @@ def migrate_member_roles_to_rbac(tenant_id: str | None, dry_run: bool) -> None:
...
+
+@click.command(
+    "rbac-migrate-dataset-permissions",
+    help="Migrate legacy dataset permission scopes and partial members into RBAC dataset access bindings.",
+)
+@click.option("--tenant-id", help="Only migrate datasets in a single workspace.")
+@click.option("--dataset-id", help="Only migrate a single dataset.")
+@click.option(
+    "--operator-account-id",
+    help="Account id used as the RBAC operator. Defaults to omitted inner account id.",
```
QuantumGhost: What the meaning of the phrase "inner account id" here?

api/commands/rbac.py
```diff
@@ -110,3 +112,154 @@ def migrate_member_roles_to_rbac(tenant_id: str | None, dry_run: bool) -> None:
...
+)
+@click.option("--batch-size", default=500, show_default=True, type=click.IntRange(min=1))
+@click.option("--dry-run", is_flag=True, default=False, help="Preview the migration without writing RBAC bindings.")
```
QuantumGhost: I would propose changing the default value of  `dry-run` to true to avoid accidentally apply migration.

api/commands/rbac.py
```diff
@@ -110,3 +112,154 @@ def migrate_member_roles_to_rbac(tenant_id: str | None, dry_run: bool) -> None:
...
+    """Backfill RBAC dataset access config from legacy `Dataset.permission`.
+
+    Legacy mapping:
+    - all_team_members -> RBAC dataset whitelist scope "all"
+    - partial_members  -> RBAC dataset whitelist scope "specific" plus each partial member gets the
+      virtual default policy
+    - only_me          -> RBAC dataset whitelist scope "specific" with no member policy bindings
+
+    The command replaces each dataset's RBAC whitelist scope first. RBAC clears
+    existing per-user policy bindings during that replace, then this command
```
QuantumGhost, line 320: This  should be highlighted in the --help output. Otherwise the user of this data migration command may not aware the side effect.

api/commands/rbac.py
```diff
@@ -1,260 +1,267 @@
...
-                    futures = [
-                        executor.submit(
-                            _replace_member_role,
-                            workspace_id,
-                            owner_account_id,
-                            member_account_id,
-                            resolved_role_id,
+
+                scanned_count += 1
+                if not dry_run:
```
QuantumGhost: I would suggest print the proposed changes to stdout in dry-run mode, for example, the call arguments to `replace_whitelist` method. Besides, I suggest logging proposed changes in a structured format (I.E. JSON), so proposed changes could be verified / analyzed with a script.

Ideally we should log every state affected by the migrations, and their values before and after the migration. However, I examined the code and found that `RBACService.DatasetAccess` are a wrapper of external HTTP endpoints, which makes logging remote state hard to implement.

api/commands/rbac.py
```diff
@@ -1,260 +1,267 @@
...
+                        payload=ReplaceMemberBindings(scope=scope),
+                    )
+                    scope_migrated_count += 1
+
+                if should_bind_partial_members:
+                    partial_dataset_count += 1
+                    for member_account_id in partial_member_ids:
+                        if dry_run:
+                            continue
+                        RBACService.DatasetAccess.replace_user_access_policies(
```
QuantumGhost: The suggestion above about proposed change logging applies to here, too.

api/commands/rbac.py
```diff
@@ -1,260 +1,267 @@
...
+    scope_migrated_count = 0
+    user_policy_migrated_count = 0
+    partial_dataset_count = 0
 
-        replace_jobs: list[tuple[str, str]] = []
-        for member_account_id, legacy_role in batch:
-            resolved_role_id = role_ids_by_legacy_role.get(legacy_role)
-            if not resolved_role_id:
-                raise ValueError(f"Unsupported legacy workspace role: {legacy_role}")
+    with session_factory.create_session() as session:
```
QuantumGhost: I would against holding a session for the entire migration. Instead, the session should only be used to load `dataset_rows` and `permission_rows`. Those ORM records should be loaded with `expire_on_commit=False`. 

This avoiding long-running transaction in production environment.

api/tests/unit_tests/commands/test_legacy_model_type_migration.py
```diff
@@ -336,6 +336,13 @@ def _insert_load_balancing_model_config(
         )
 
 
+def test_data_migrate_group_registers_dataset_permission_rbac_migration(command_module) -> None:
```
QuantumGhost, line 339: I would suggest creating a integration test and verify the migration script works as expected.

fatelei: rbac service can not be deploy in integration env

QuantumGhost:
~~Besides, codex suggests the following problem:~~

@fatelei  and I had a private conversation and he confirmed that the access bindings are only meant for members, not owners. There are special handing for resource owners in RBAC service.

<details>
<summary>Outdated and incorrect</summary>

------
### Critical

**The migration replaces dataset RBAC bindings but never restores the resource owner’s access.**

The command calls `RBACService.DatasetAccess.replace_whitelist(...)`, and its own docstring says this clears existing per-user policy bindings. After that, it only recreates bindings for `partial_members`, but it never restores the owner/maintainer binding.

This breaks existing behavior:

- In the legacy model, both `only_me` and `partial_members` implicitly allow the dataset maintainer to access the dataset.
- Newly created datasets already call `try_sync_creator_access_policy_member_bindings(...)` to restore the creator binding in RBAC.

So this migration can remove the owner’s RBAC access to old datasets. Re-running the command can also delete an existing owner binding again.

**Suggested fix**

- After each `replace_whitelist(...)`, restore the owner/creator binding first, then recreate `partial_members` default-policy bindings.
- Add regression tests for:
  - `only_me`: the owner can still access the dataset after migration
  - `partial_members`: the owner still has access even if they are not listed in `dataset_permissions`

</details>

GareArc:
Need to resolve conflict first.

fatelei:
> Need to resolve conflict first.

done

GareArc (approved):
LGTM

QuantumGhost (approved):

Issue #37985, opened 2026-06-26
enhancement: object ownership binding consistency


WH-2099:
Parent issue: langgenius/dify#37983

## Overview

This issue groups flows where an outer route or token resolves one resource, but a nested object id is later looked up without consistently binding it back to that outer resource.

The descriptions below use concrete code locations instead of audit shorthand.

No implementation proposal is included here.

## Affected code points

- [X] Service API text-to-audio message lookup: `api/controllers/service_api/app/audio.py:168`, `api/controllers/service_api/app/audio.py:180`, `api/services/audio_service.py:132`.  <br>The Service API text-to-audio route can pass a message id into shared audio service logic that reads the message answer by id.
- [X] Dataset segment routes: `api/services/dataset_service.py:4059`, `api/controllers/service_api/dataset/segment.py:352`, `api/controllers/console/datasets/datasets_segments.py:486`.  <br>Segment read, update, delete, or child chunk flows can resolve the path dataset and document before later handling a segment id separately.
- [X] Console document batch deletion: `api/controllers/console/datasets/datasets_document.py:465`, `api/services/dataset_service.py:1924`, `api/services/dataset_service.py:1937`.  <br>Batch document deletion receives document ids under a dataset route and later deletes documents selected by those ids.
- [X] Annotation mutation: `api/controllers/console/app/annotation.py:318`, `api/controllers/service_api/app/annotation.py:281`, `api/services/annotation_service.py:303`, `api/services/annotation_service.py:394`.  <br>Annotation update and delete flows resolve an app at the route layer and later operate on annotation ids in the service layer.
- [X] App MCP server update: `api/controllers/console/app/mcp_server.py:146`, `api/models/model.py:2116`.  <br>The MCP server update route resolves a path app while the server row to update is selected from a body id.
- [X] Trial app read routes: `api/controllers/console/explore/trial.py:506`, `api/controllers/console/explore/trial.py:585`, `api/controllers/console/app/wraps.py:41`.  <br>Several trial app read routes load an app by id through trial-specific wrapping that does not consistently carry the expected trial app relationship.
- [X] Web, Console, Installed App, and Trial App text-to-audio message lookup: `api/controllers/web/audio.py:133`, `api/controllers/console/app/audio.py:139`, `api/controllers/console/explore/trial.py:422`, `api/services/audio_service.py:132`.  <br>Multiple text-to-audio entry points share the same message-id branch in the audio service.
[...]

Issue #37983, opened 2026-06-26
enhancement: track resource boundary consistency improvements


WH-2099:
## Overview

This parent issue tracks engineering follow-ups to make resource boundary checks more consistent across routes, serializers, async tasks, runtime services, and plugin or agent service boundaries.

The common pattern is that one layer has the expected workspace, app, dataset, user, credential, runtime, or package context, while a later layer handles a nested identifier without carrying that context all the way through.

The sub-issues below group the work by behavior area and include concrete code locations for review.

## Sub-issues

* langgenius/dify#37984 File reference and lifecycle boundaries
* langgenius/dify#37985 Object ownership checks for nested ids
* langgenius/dify#37986 Credential visibility and usage boundaries
* langgenius/dify#37987 Runtime task, HITL, and async boundaries
* langgenius/dify#37988 Resource-level permission checks
* langgenius/dify#37989 Temporary capability ownership checks
* langgenius/dify#37990 Plugin daemon, agent backend, and package boundaries

## Behavioral impact

These patterns can produce reads, writes, signed URLs, runtime actions, pending import consumption, package reuse, or service actions outside the scope that the caller-facing route appears to represent.

Detailed code locations are listed in the sub-issues.

No implementation proposal is included here.

WH-2099:
Conclusion:

A strongly typed Ref is used throughout the resource chain. After authentication is completed at the API/trust boundary, the raw ID is converted to `DatasetRef -> DocumentRef -> SegmentRef`, and the scattered IDs are no longer transmitted internally.

The first phase begins with explicit parameter passing:
- `DatasetRef(tenant_id, dataset_id, ctor_token)`
- `DocumentRef(dataset, document_id)`
- `SegmentRef(document, segment_id)`

Refs are implemented using `NamedTuple` and uniformly constructed through `DatasetRefService`, with a private constructor/ctor token controlling the trusted entry point. Subsequent reading and writing of documents/segments follows the complete attribution chain: tenant -> dataset -> document -> segment.

At the same time, it is necessary to clarify the logic for determining whether a document belongs to a dataset or a segment belongs to a document. The abstraction of the DAO layer will be discussed later and is not a prerequisite for the first stage.

WH-2099:
Reviewed the parent and all seven sub-issues against the latest code available at this audit:

- `dify/main@90d6046345a4d53ca3e1dcd9a419cca49a537858`
[...]

PR #38177, opened 2026-06-29
fix(api): scope nested resource lookups by owner refs

WH-2099:
## Summary

Refs #37983.
Refs #37985.

This PR tightens nested resource ownership checks by carrying trusted owner context through service lookups instead of resolving inner ids by themselves.

- Add typed `NamedTuple` refs for dataset, app, and workflow ownership chains.
- Bind dataset segment, child chunk, document batch delete, TTS message lookup, annotation mutation, App MCP server update, tag update/delete, workflow version mutation, instruction generation, and RAG pipeline document error updates to their outer resource context.
- Add focused tests that assert the generated SQLAlchemy statements include the owner columns such as tenant, app, dataset, document, pipeline, or workflow owner ids.

## Why

Several routes authenticate or resolve an outer resource first, then pass a nested id into lower-level service logic.
The lower layer could then read or mutate by the nested id without preserving the route's full ownership chain.

This makes the trusted boundary explicit: after the outer resource is resolved, internal code uses typed refs or scoped queries that include the relevant parent ids.

## Validation

- `git diff --check`
- `env DEBUG=false HTTP_PROXY= HTTPS_PROXY= http_proxy= https_proxy= ALL_PROXY= all_proxy= NO_PROXY= no_proxy= uv run --project api pytest -o addopts='' ...`

Result: `788 passed, 2 warnings`.

### DIFF
--- a/api/controllers/console/datasets/datasets_segments.py
+++ b/api/controllers/console/datasets/datasets_segments.py
@@ -58,8 +58,9 @@ from graphon.model_runtime.entities.model_entities import ModelType
 from libs.helper import dump_response, escape_like_pattern
 from libs.login import login_required
 from models import Account
-from models.dataset import ChildChunk, DocumentSegment
+from models.dataset import Dataset, Document, DocumentSegment
 from models.model import UploadFile
+from services.dataset_ref_service import DatasetRefService, SegmentRef
 from services.dataset_service import DatasetService, DocumentService, SegmentService
 from services.entities.knowledge_entities.knowledge_entities import ChildChunkUpdateArgs, SegmentUpdateArgs
 from services.errors.chunk import ChildChunkDeleteIndexError as ChildChunkDeleteIndexServiceError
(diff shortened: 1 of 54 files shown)


api/services/rag_pipeline/rag_pipeline.py
```diff
@@ -1005,9 +1026,14 @@ def update_workflow(
         :param tenant_id: Tenant ID
         :param account_id: Account ID (for permission check)
         :param data: Dictionary containing fields to update
+        :param workflow_ref: Optional trusted owner-bound workflow reference
         :return: Updated workflow or None if not found
         """
-        stmt = select(Workflow).where(Workflow.id == workflow_id, Workflow.tenant_id == tenant_id)
+        lookup_workflow_id = workflow_ref.workflow_id if workflow_ref else workflow_id
+        lookup_tenant_id = workflow_ref.tenant_id if workflow_ref else tenant_id
```
wylswz: Could we just simply remove `workflow_id` and `tenant_id` from the arg list?

api/services/audio_service.py
```diff
@@ -134,7 +147,14 @@ def invoke_tts(text_content: str, app_model: App, voice: str | None = None, is_d
                 uuid.UUID(message_id)
             except ValueError:
                 return None
-            message = session.get(Message, message_id)
+            app_ref = AppRefService.create_app_ref(app_model)
```
wylswz: The usage of `Ref` lives in the implementation instead of the contract, I think it's easier to enforce contracts (like type checking, structural testing) than implementations.

api/controllers/console/datasets/datasets_segments.py
QuantumGhost, line 324: This part still does not enforce nested scope lookup yet.

wylswz (approved):
LGTM

PR #38200, opened 2026-06-30
fix(workflow): guard on_tool_execution stdout traces behind DEBUG

### LINKED ISSUE #38199: on_tool_execution prints tool outputs to stdout unconditionally (missing DEBUG guard)
### Summary

In `api/core/callback_handler/workflow_tool_callback_handler.py`, `DifyWorkflowCallbackHandler.on_tool_execution()` calls `print_text(...)` for every tool output **unconditionally**.

Every other callback in the parent `DifyAgentCallbackHandler` (`on_tool_start`, `on_tool_end`, `on_tool_error`, `on_agent_start`, `on_agent_finish`, `on_datasource_start`) guards its `print_text` calls with `if dify_config.DEBUG:`, and `ignore_agent` returns `not dify_config.DEBUG`. The workflow subclass method is the only one missing this guard — it looks like an oversight.

```python
# api/core/callback_handler/workflow_tool_callback_handler.py
def on_tool_execution(self, ...):
    for tool_output in tool_outputs:
        print_text("\n[on_tool_execution]\n", color=self.color)
        print_text("Tool: " + tool_name + "\n", color=self.color)
        print_text("Outputs: " + tool_output.model_dump_json()[:1000] + "\n", color=self.color)
        print_text("\n")
        yield tool_output
```

### Impact

1. **Wasted serialization on a hot path.** `tool_output.model_dump_json()` fully serializes the *entire* tool output on every tool-node execution, only to print a 1000-char preview — `[:1000]` slices the already-serialized string. For tool nodes returning large payloads (HTTP / web-scraping / large JSON), this is meaningful, avoidable CPU work for every tool call.
2. **Unconditional stdout noise.** It is written via a raw `print()` to stdout, **not** routed through logging, so it is emitted regardless of `LOG_LEVEL` and inflates stdout / log-collection volume in every deployment.

### Steps to reproduce

Run any workflow containing a Tool node with `DEBUG` unset and `LOG_LEVEL=WARNING`. The worker stdout still shows:

```
[on_tool_execution]
Tool: <tool_name>
Outputs: {...}
```

### Expected behavior

Consistent with the sibling callbacks: emit these traces only when `dify_config.DEBUG` is enabled, and skip `model_dump_json()` otherwise.

### Proposed fix

Wrap the `print_text(...)` calls in `on_tool_execution` with `if dify_config.DEBUG:`, matching the rest of the callback family. The `yield tool_output` behavior is unchanged. (`agent_tool_callback_handler.py` already does this for its own methods.)

### Version

Observed on `main`; present in the 1.x line.

MRZHUH:
## Summary

`DifyWorkflowCallbackHandler.on_tool_execution` was the only callback in the `DifyAgentCallbackHandler` family that wrote to stdout **without** the `if dify_config.DEBUG:` guard that every sibling method uses (`on_tool_start`, `on_tool_end`, `on_tool_error`, `on_agent_start`, `on_agent_finish`, `on_datasource_start`; `ignore_agent` returns `not dify_config.DEBUG`). This wraps its `print_text(...)` calls in the same guard.

Two improvements when `DEBUG` is off (the default):

- No unconditional stdout output — the calls use a raw `print()` that bypasses `LOG_LEVEL`, so they were emitted in every deployment.
- `tool_output.model_dump_json()` is no longer invoked just to build a 1000-char preview, removing avoidable full serialization of large tool outputs on a hot path.

Functional behavior is unchanged — outputs are still `yield`ed; only the debug printing is gated.

Local checks: `ruff check` is clean on the changed files and `tests/unit_tests/core/callback_handler/test_workflow_tool_callback_handler.py` passes (10 passed), with a new case covering the `DEBUG`-off short-circuit.

Fixes #38199

## Screenshots

| Before | After |
|--------|-------|
| N/A — backend stdout/logging behavior | N/A |

## Checklist

- [ ] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods — ran `ruff check` (clean) and the affected unit tests (10 passed); full `make type-check` not run locally

### DIFF
--- a/api/core/callback_handler/workflow_tool_callback_handler.py
+++ b/api/core/callback_handler/workflow_tool_callback_handler.py
@@ -1,6 +1,7 @@
 from collections.abc import Generator, Iterable, Mapping
 from typing import Any
 
+from configs import dify_config
 from core.callback_handler.agent_tool_callback_handler import DifyAgentCallbackHandler, print_text
 from core.ops.ops_trace_manager import TraceQueueManager
 from core.tools.entities.tool_entities import ToolInvokeMessage
@@ -19,8 +20,9 @@ class DifyWorkflowCallbackHandler(DifyAgentCallbackHandler):
         trace_manager: TraceQueueManager | None = None,
     ) -> Generator[ToolInvokeMessage, None, None]:
         for tool_output in tool_outputs:
-            print_text("\n[on_tool_execution]\n", color=self.color)
-            print_text("Tool: " + tool_name + "\n", color=self.color)
-            print_text("Outputs: " + tool_output.model_dump_json()[:1000] + "\n", color=self.color)
-            print_text("\n")
+            if dify_config.DEBUG:
+                print_text("\n[on_tool_execution]\n", color=self.color)
+                print_text("Tool: " + tool_name + "\n", color=self.color)
+                print_text("Outputs: " + tool_output.model_dump_json()[:1000] + "\n", color=self.color)
+                print_text("\n")
             yield tool_output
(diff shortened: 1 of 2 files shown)


hjlarry (approved):
LGTM

PR #38220, opened 2026-06-30
feat(api): abort active workflow runs during Celery warm shutdown

linw1995:
> [!IMPORTANT]
>

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


api/extensions/workflow_warm_shutdown.py
```diff
@@ -0,0 +1,107 @@
...
+_WORKER_SHUTDOWN_DISPATCH_UID = "dify.workflow_warm_shutdown.shutdown"
+WORKFLOW_WARM_SHUTDOWN_ABORT_REASON = "Workflow stopped because the worker is shutting down."
+type _NativeThreadStarter = Callable[[Any, tuple[object, ...]], int]
+
+
+def _start_native_thread(function: Any, args: tuple[object, ...]) -> int:
+    try:
+        from gevent import monkey
+
+        start_new_thread = monkey.get_original("_thread", "start_new_thread")
```
QuantumGhost: This is incorrect. 

If you want to start a native thread as monitor, you could use https://www.gevent.org/api/gevent.threadpool.html

Generally speaking, you should not import the gevent module in application code. Accessing internal members of gevent should be avoided unless absolutely nesessary.

api/core/app/workflow/layers/persistence.py
```diff
@@ -411,6 +415,19 @@ def _fail_running_node_executions(self, *, error_message: str) -> None:
                 execution.elapsed_time = max((now - execution.created_at).total_seconds(), 0.0)
                 self._workflow_node_execution_repository.save(execution)
 
+    def _record_otel_abort_reason(self, *, reason: str) -> None:
```
QuantumGhost: IMO this should not belong to here. 

Consider move this to ObservabilityLayer

QuantumGhost:
The overall approach seems a bit weird, for the following reasons:

1. The cancellation logic (api/extensions/workflow_warm_shutdown.py) is separated from where the cancellation is registered. 
2. The original CommandChanenl is mainly used for cross-machine cancellation scenario. However, in the case of this PR, the cancelled workers and the cancellation does live inside a same machine. It seems like a waste to send Redis command to trigger cancellation.
3. The `CommandChanenl` is designed as an interface intentionally to provide other command source.

I would propose the following solutions:

1. Introduces a `CombinedCommandChannel`, to combine multiple CommandChannel into one:

```py
class CombinedCommandChannel(CommandChannel):
    def __init__(self, command_channels: Sequence[CommandChannel]): ...
```

2. Introduce `CelerySignalCommandChannel`, to convert Celery [worker_shutting_down](https://docs.celeryq.dev/en/stable/userguide/signals.html#id27) signals to Graphon commands
3. Combine classes introduced in 1 and 2, and rely on the command mechanism of Graphon.

QuantumGhost:
I think the shutdown modeling can be simplified further.

Instead of storing a process-global `AbortCommand`, I would suggest modeling Celery warm shutdown as a process-local `threading.Event` owned by a Celery adapter module, for example:

```python
_shutdown_started = threading.Event()

def _on_shutting_down(*args: object, **kwargs: object) -> None:
    _shutdown_started.set()

def shutdown_started() -> bool:
    return _shutdown_started.is_set()
```

Then `CelerySignalCommandChannel` can just depend on a shutdown-state getter and translate that state into a Graphon abort command, e.g.:

```python
class CelerySignalCommandChannel(CommandChannel):
    def __init__(
        self,
        shutdown_state_getter: Callable[[], bool],
        abort_reason: str = WORKFLOW_WARM_SHUTDOWN_ABORT_REASON,
    ) -> None:
        ...
```

This gives a cleaner boundary:

- the Celery adapter owns the worker shutdown lifecycle state
- `CelerySignalCommandChannel` only translates that state into Graphon commands
- workflow runners / `WorkflowEntry` continue to depend only on the `CommandChannel` abstraction

I would also suggest keeping an instance-level flag inside `CelerySignalCommandChannel` (for example `self._abort_emitted`) so each channel emits the abort command at most once. That avoids the module-level `_abort_command`, avoids runtime `reset`, and still preserves the late-created channel behavior after shutdown has already started.

With this design, `active_workflow_tasks.py` also looks unnecessary for the cancellation path, since the shutdown signal can be injected into `CelerySignalCommandChannel` directly instead of being split across multiple modules.

If the active task registry is only there for cancellation, I think it can be removed. If you still want to log the number of in-flight workflow runs during worker shutdown, then keeping the current counting mechanism makes sense as an observability / diagnostic concern, but it should stay separate from the command propagation mechanism.

QuantumGhost (approved):

PR #37388, opened 2026-06-12
fix(api): prevent plugin provider cache stampedes

VeraPyuyi:
## Summary
- add tenant/generation-scoped single-flight locking for plugin model provider cache misses
- wait briefly only when another worker owns the refresh lock, and fall back directly when the lock API fails
- cache empty provider lists as valid refresh results so waiters do not stampede the plugin daemon

Fixes #37334

## Validation
- `api/.venv/Scripts/python.exe -m pytest -o addopts='' api/tests/unit_tests/services/plugin/test_plugin_service.py -q` (`30 passed`)
- `api/.venv/Scripts/python.exe -m pytest -o addopts='' api/tests/unit_tests/core/plugin/test_model_runtime_adapter.py -q` (`29 passed`)
- `api/.venv/Scripts/ruff.exe check api/core/plugin/plugin_service.py api/tests/unit_tests/services/plugin/test_plugin_service.py api/tests/unit_tests/core/plugin/test_model_runtime_adapter.py`
- `git diff --check origin/main..codex/plugin-model-provider-cache-singleflight`

Issue #34264, opened 2026-03-30
The knowledge retrieval process node is too slow, causing significant workflow latency.


kgpp34:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] Please do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing? Tell me about your story.

Dify version: 1.11.4

# abstract 
In this version, I use workflow to run a knowledge retrieval process, but when i run this workflow, i found that the knowledge retrieval node consumes a lot of cpu time, the node latency even cost 1 min

# context
1. my knowledge retrieval node has 10+datasets which totally contains 1000+ documents
2. machine has 16 cpu cores
3. i used the py-spy to perf dify workflow, cmd is : py-spy record -o profile.svg --duration 60 --pid {PID}

<img width="1307" height="634" alt="Image" src="https://github.com/user-attachments/assets/2e9f721b-c7c2-4ac0-a7be-e4c6f5b001e8" />

# result
we found that in the workflow process, when the dify-api communicate with plugin_daemon, the function(fetch_model_providers) consumes lot of cpu time, and the pydantic slowed down the whole process and performance, so may be we should find a way to resolve this problem

### 3. Can you help us with this feature?

- [ ] I am interested in contributing to this feature.

fatelei:
Flame graph shows get_provider_model_bundle consuming 20.02% CPU. Root causes:
1. get_configurations() has no cache — every call triggers 6+ DB queries + 1 HTTP request
2. Each call site creates fresh ProviderManager/PluginModelRuntime instances, so the in-memory _provider_entities cache is never reused
3. In RAG retrieval, embedding_search and full_text_index_search both independently instantiate Vector → VectorFactory → ModelManager.for_tenant() → full config rebuild

WH-2099:
Update: the provider-list cache stampede portion is now covered by #37388, which has been merged.

[...]

PR #38226, opened 2026-06-30
fix(api): keep provider refresh waiters single-flight

### LINKED ISSUE #38277: test(api): stabilize ZIP archive fixture collection under xdist
## Summary

`Run API Tests / API Unit Tests (3.12)` can fail during pytest collection when xdist is enabled because one parametrized test uses ZIP bytes generated at import time.

Recent failing example:

- PR: https://github.com/langgenius/dify/pull/38226
- Run: https://github.com/langgenius/dify/actions/runs/28516727225
- Job: https://github.com/langgenius/dify/actions/runs/28516727225/job/84530474612

## Failure

The job fails before running test assertions:

```text
ERROR collecting gw0
Different tests were collected between gw1 and gw0.
```

The collected node id differs for this test:

```text
tests/unit_tests/services/test_agent_config_service.py::test_inspect_skill_maps_invalid_archives_to_service_errors
```

The diff is in the generated ZIP bytes shown in the parametrized test id:

```text
-...test_inspect_skill_maps_invalid_archives_to_service_errors[...Yb...]
+...test_inspect_skill_maps_invalid_archives_to_service_errors[...Xb...]
```

## Root Cause

`api/tests/unit_tests/services/test_agent_config_service.py` defines `_zip_bytes` like this:

```python
def _zip_bytes(members: dict[str, bytes]) -> bytes:
    buffer = io.BytesIO()
    with zipfile.ZipFile(buffer, "w") as archive:
        for name, payload in members.items():
            archive.writestr(name, payload)
    return buffer.getvalue()
```

The parametrized test calls `_zip_bytes(...)` at module collection time:

```python
@pytest.mark.parametrize(
    "archive_bytes",
    [
        b"not-a-zip-archive",
        _zip_bytes({"README.md": b"missing skill md"}),
    ],
)
```

`zipfile.writestr(name, payload)` stores the current timestamp in the ZIP entry metadata when `name` is a string.

With `pytest -n auto`, each xdist worker imports and collects the module independently.

If workers generate the ZIP bytes on different timestamp ticks, the bytes differ.

Pytest includes the bytes repr in the parametrized node id, so xdist sees different collected tests and aborts collection.

## Suggested Fix

Make the ZIP fixture deterministic at the source:

- Build entries with `zipfile.ZipInfo`.
- Set a fixed `ZipInfo.date_time`.
- Optionally add stable `ids` to the `parametrize` call so raw bytes never appear in node ids.

This should make collection stable across xdist workers and remove the flaky CI failure.

WH-2099:
## Summary

This is a follow-up to #37388 for the provider-list refresh path tracked under #34264.

The previous single-flight lock prevented ordinary cache-miss stampedes, but waiters could still fall through to the plugin daemon after the lock TTL deadline.
That made the Redis lock TTL act as both a crashed-owner lease and a waiter deadline.

This PR makes the coordination contract explicit:

- Redis lock TTL is only a lock lease.
- Waiters keep reading the current provider generation and cache until they either read cache or acquire the refresh lock.
- Redis cache or lock failures still fall back to direct daemon discovery to preserve availability.
- Generation changes naturally restart the cache/lock path on the new generation.

## CI Stability

This PR also fixes #38277.

`test_inspect_skill_maps_invalid_archives_to_service_errors` generated ZIP bytes during pytest collection.
Because `zipfile.writestr(name, payload)` stores the current timestamp in ZIP entry metadata, xdist workers could collect different parametrized node ids and fail before running tests.

The ZIP fixture now uses a fixed entry timestamp, and the parametrized cases have stable ids.

## Tests

- env DEBUG=false HTTP_PROXY= HTTPS_PROXY= ALL_PROXY= NO_PROXY= http_proxy= https_proxy= all_proxy= no_proxy= uv run --project api pytest -o addopts='' api/tests/unit_tests/services/plugin/test_plugin_service.py api/tests/unit_tests/core/plugin/test_model_runtime_adapter.py -q
- uv run --project api ruff check api/core/plugin/plugin_service.py api/tests/unit_tests/services/plugin/test_plugin_service.py api/tests/unit_tests/core/plugin/test_model_runtime_adapter.py
- env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u NO_PROXY -u http_proxy -u https_proxy -u all_proxy -u no_proxy DEBUG=false uv run --project api pytest -p no:benchmark -n 2 api/tests/unit_tests/services/test_agent_config_service.py::test_inspect_skill_maps_invalid_archives_to_service_errors
- git diff --check

Refs #34264
Fixes #38277

### DIFF
--- a/api/core/plugin/plugin_service.py
+++ b/api/core/plugin/plugin_service.py
@@ -17,7 +17,7 @@ import time
 from collections.abc import Iterator, Mapping, Sequence
 from contextlib import contextmanager
 from mimetypes import guess_type
-from typing import Literal, NoReturn, Protocol
+from typing import Literal, Protocol
 
 from pydantic import BaseModel, TypeAdapter, ValidationError
 from redis import RedisError
@@ -225,24 +225,10 @@ class PluginService:
     ) -> None:
         cache_key = cls._get_plugin_model_providers_cache_key(tenant_id, generation)
         try:
-            payload = _provider_entities_adapter.dump_json(list(providers)).decode("utf-8")
+            payload = _provider_entities_adapter.dump_json(list(providers))
             redis_client.setex(cache_key, dify_config.PLUGIN_MODEL_PROVIDERS_CACHE_TTL, payload)
         except (RedisError, RuntimeError):
             logger.warning("Failed to cache plugin model providers for tenant %s.", tenant_id, exc_info=True)
-
-    @classmethod
-    def _raise_plugin_model_providers_refresh_timeout(
-        cls, tenant_id: str, generation: int, cause: BaseException | None = None
-    ) -> NoReturn:
-        logger.warning(
-            "Timed out waiting for plugin model providers refresh lock for tenant %s generation %s.",
-            tenant_id,
-            generation,
-        )
-        error = RuntimeError(f"Timed out waiting for plugin model providers refresh lock for tenant {tenant_id}.")
-        if cause is not None:
-            raise error from cause
-        raise error
 
     @classmethod
     @contextmanager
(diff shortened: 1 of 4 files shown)


cqjjjzr:
I feel that we should not make this an infinite wait. A good idea would be introduce a hard timeout that once exceeded, fail the whole process.

crazywoola (dismissed):

cqjjjzr (approved):
LGTM

PR #38243, opened 2026-07-01
fix: stop create_document_index handler from swallowing all indexing errors

### LINKED ISSUE #38190: Bug: create_document_index handler swallows all indexing errors
### Self Checks

- [x] This is only for a bug report.

### Dify version

main (1.15.x)

### Cloud or Self Hosted

Self Hosted

### Steps to reproduce

1. Open `api/events/event_handlers/create_document_index.py`.
2. Note the `contextlib.suppress(Exception)` wrapper around `IndexingRunner.run`.
3. Compare with `api/tasks/document_indexing_task.py`, which catches `DocumentIsPausedError` and logs other failures explicitly.

### Expected behavior

When document indexing is paused, the handler should log the pause message. When indexing fails for other reasons, the failure should be logged instead of being discarded silently.

### Actual behavior

The outer `contextlib.suppress(Exception)` swallows every exception from `IndexingRunner.run`. The inner `except DocumentIsPausedError` block never handles non-pause failures, and indexing errors on this path produce no log output.

### Proposed fix

Remove `contextlib.suppress(Exception)` and align with the task handlers: catch `DocumentIsPausedError` for the info log, and use `logger.exception` for unexpected failures.

### Additional context

I found this while reviewing indexing error handling consistency across the API. I will open a PR shortly.

EvanYao826:
## Summary

The `create_document_index` handler wraps `IndexingRunner.run` with `contextlib.suppress(Exception)`, which silently discards every exception. When indexing fails for reasons other than `DocumentIsPausedError`, the error produces no log output and the failure is invisible.

### Before

```python
with contextlib.suppress(Exception):
    try:
        indexing_runner = IndexingRunner()
        indexing_runner.run(documents)
        ...
    except DocumentIsPausedError as ex:
        logger.info(click.style(str(ex), fg="yellow"))
```

### After

```python
try:
    indexing_runner = IndexingRunner()
    indexing_runner.run(documents)
    ...
except DocumentIsPausedError as ex:
    logger.info(click.style(str(ex), fg="yellow"))
except Exception:
    logger.exception("Failed to index documents")
```

### Changes

- Removed the outer `contextlib.suppress(Exception)` layer
- Added `except Exception` block with `logger.exception` so unexpected failures are logged with full traceback
- Removed unused `import contextlib`

Fixes #38190

### DIFF
--- a/api/events/event_handlers/create_document_index.py
+++ b/api/events/event_handlers/create_document_index.py
@@ -1,4 +1,3 @@
-import contextlib
 import logging
 import time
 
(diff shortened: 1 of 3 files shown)


wylswz:
There are style check failures.

PR #38274, opened 2026-07-01
refactor: manage rag pipeline sessions explicitly

41tair:
> [!IMPORTANT]
>

## Summary
manage rag pipeline sessions explicitly

https://github.com/langgenius/dify/issues/37403

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [X] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [X] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [X] I've updated the documentation accordingly.
- [X] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods


api/controllers/service_api/dataset/rag_pipeline/rag_pipeline_workflow.py
```diff
@@ -16,6 +17,7 @@
     register_schema_model,
     register_schema_models,
 )
+from controllers.console.app.wraps import with_session
```
QuantumGhost, line 20: I suggests move `with_session` to a shared module (E.G. `controllers/common`) to avoid import console wrapper from service api packages.

api/services/rag_pipeline/pipeline_template/built_in/built_in_retrieval.py
```diff
@@ -21,13 +22,15 @@ def get_type(self) -> str:
         return PipelineTemplateType.BUILTIN
 
     @override
-    def get_pipeline_templates(self, language: str, current_tenant_id: str | None = None) -> dict[str, Any]:
+    def get_pipeline_templates(
```
QuantumGhost, line 25: I would suggest inject session through the Retrieval constructor, instead of inject directly through method arguments.

api/services/rag_pipeline/rag_pipeline_transform_service.py
QuantumGhost: Here the transaction is committed, which may affect modifications happened outside of the `transform_dataset` function.

Replacing `session.commit()` with `session.flush()` seems a good solution.

api/services/rag_pipeline/rag_pipeline_transform_service.py
```diff
@@ -289,29 +289,29 @@ def _deal_dependencies(self, pipeline_yaml: dict[str, Any], tenant_id: str):
...
+        session.flush()
 
         dataset.pipeline_id = pipeline.id
         dataset.runtime_mode = DatasetRuntimeMode.RAG_PIPELINE
         dataset.updated_by = current_user.id
         dataset.updated_at = datetime.now(UTC).replace(tzinfo=None)
-        db.session.add(dataset)
-        db.session.commit()
+        session.add(dataset)
+        session.commit()
```
QuantumGhost: ditto

QuantumGhost: Same underlying concern here. I agree this is largely pre-existing debt, but this PR only pushes explicit session handling halfway through the call chain. Along the same use case we still mix request-injected sessions, `db.session`, and service-created sessions, so the session owner and transaction boundary remain hard to reason about.

I am not asking this PR to clean up all historical debt, only to keep the touched path internally consistent. If this path adopts caller-owned sessions, the adjacent loaders/helpers should not switch back to another session owner.

api/controllers/service_api/dataset/rag_pipeline/rag_pipeline_workflow.py
asukaminato0721: good idea.

api/services/rag_pipeline/pipeline_template/built_in/built_in_retrieval.py
asukaminato0721: yeah, this will reduce quite some signature changes.

Class(session).method(...)

instead of

Class().method(session, ...)

but many apis are classmethod so...

api/controllers/service_api/dataset/rag_pipeline/rag_pipeline_workflow.py
41tair: I will create a new PR to make this change

api/services/rag_pipeline/pipeline_template/built_in/built_in_retrieval.py
41tair: I agree with this direction. Passing `session` through method arguments is mostly a transitional step for removing implicit `db.session` access and making the DB boundary testable.

The longer-term shape I have in mind is not necessarily `Class(session).method(...)`, but rather injecting a repository/dependency object into the class, e.g. `Class(data_repo).method(...)`. The repo can be built from the session/sessionmaker at the composition layer, while the retrieval/service methods keep business-focused signature

api/services/rag_pipeline/rag_pipeline_transform_service.py
41tair: fix

api/controllers/service_api/dataset/rag_pipeline/rag_pipeline_workflow.py
41tair: https://github.com/langgenius/dify/pull/38379

QuantumGhost (approved):

PR #38367, opened 2026-07-03
refactor(openapi): resource-oriented paths for /openapi/v1 + difyctl version gate

lin-snow:
### ⚠️ Breaking Change

Aligns the user-scoped `/openapi/v1` surface (consumed by the `difyctl` CLI) with our API Design Guide: verb path suffixes are replaced by standard resource `GET`s and AIP-136 `:action` custom methods.

| Before | After |
|--------|-------|
| `GET /apps/{id}/describe` | `GET /apps/{id}` |
| `GET /permitted-external-apps/{id}/describe` | `GET /permitted-external-apps/{id}` |
| `POST /apps/{id}/run` | `POST /apps/{id}:run` |
| `POST /apps/{id}/tasks/{tid}/stop` | `POST /apps/{id}/tasks/{tid}:stop` |
| `POST /apps/{id}/files/upload` | `POST /apps/{id}/files` |
| `GET /apps/{id}/export` | `GET /apps/{id}/dsl` |
| `GET /apps/{id}/check-dependencies` | `GET /apps/{id}/dependencies:check` |
| `POST /workspaces/{ws}/apps/imports/{iid}/confirm` | `POST /workspaces/{ws}/apps/imports/{iid}:confirm` |
| `POST /workspaces/{ws}/switch` | `POST /workspaces/{ws}:switch` |
| `GET\|POST /apps/{id}/form/human_input/{tok}` | `/apps/{id}/human-input-forms/{tok}` (+ `:submit` for POST) |
| `PUT /workspaces/{ws}/members/{mid}/role` | `PATCH /workspaces/{ws}/members/{mid}` |

`difyctl` is the only consumer and ships in lockstep with the server, so this is a hard cutover — no `/v2`, no dual-mount. RFC 8628 device-flow routes and the `limit` pagination param are intentionally left unchanged.

The change is self-contained in this repo: server routes, the regenerated committed oRPC/Zod contract (`packages/contracts`) plus a codegen tweak that nests `:action` methods cleanly (`apps.byAppId.run` rather than `apps.appIdRun`), the regenerated Markdown docs, and the `difyctl` (`cli/`) call-sites + mock server.

### 🔀 Version compatibility gate

Because the path cutover is breaking, an outdated `difyctl` that still calls the old paths would otherwise get a bare `404/405`. This PR adds a **symmetric version gate** so each side refuses a too-old counterpart with a clear "upgrade" message — each side stores the oldest counterpart it supports and hard-fails **only on the lower bound** (so there are no false positives; "too new" stays a soft nudge):

| Side | Holds | Rejects | How |
|------|-------|---------|-----|
| server (`/openapi/v1`) | `MIN_DIFYCTL_VERSION` = `0.2.0-alpha` | difyctl `< 0.2.0-alpha` | **HTTP 426 Upgrade Required** + docs link, via an app-level `before_app_request` hook so it fires even for *removed* paths (which no longer match a route and would otherwise 404). Allowlists `_version`/`_health`; fails open for non-difyctl / unparseable User-Agents. |
| difyctl | `difyctl.compat.minDify` = `1.16.0` | Dify server `< 1.16.0` | **hard error (exit 6)** before running the command, cached per host for 1h (positive-only, so a just-upgraded server clears at once); fails open on probe error; checked fresh at `login` before persisting the session. |

A server 426 also maps to the CLI's version-compat exit code (6). `difyctl` is bumped to `0.2.0-alpha` — the first build that speaks the new paths — and the two floors move together at the cutover (dify `1.16.0` ↔ difyctl `0.2.0-alpha`).

> **Deploy note:** the server gate must ship in lockstep with a released `difyctl 0.2.0-alpha`, so outdated clients have an upgrade target. Deploying the path cutover already implies releasing a new `difyctl`.

**Verification:** backend openapi unit tests (449, incl. the new version-gate suite), swagger snapshot tests (37), and the CLI suite (1252) all pass, together with the contract and CLI type-checks, eslint, and the release-manifest validator. Integration/container tests are CI-only.

## Screenshots

N/A — backend/CLI API refactor, no UI change.

## Checklist

- [ ] This change requires a documentation update: N/A — `/openapi/v1` is not part of the public docs; the in-repo generated Markdown is regenerated in this PR.
- [x] I've updated the documentation accordingly (regenerated `api/openapi/markdown/openapi-openapi.md`).
- [x] I ran the relevant lint & type-checks (ruff via pre-commit, backend openapi tests, `tsgo` for the contract and CLI).

### DIFF
--- a/api/controllers/openapi/_version_gate.py
+++ b/api/controllers/openapi/_version_gate.py
@@ -4,8 +4,6 @@ difyctl and the ``/openapi/v1`` surface ship in lockstep. A breaking path change
 (resource-oriented paths) means an outdated difyctl would call removed paths and
 get a bare 404; this gate returns ``426 Upgrade Required`` with an upgrade hint
 instead.
-
-Design record: dify-todo/inprogress/difyctl/openapi/version.md
 """
 
 from __future__ import annotations
--- a/api/controllers/openapi/_version_gate.py
+++ b/api/controllers/openapi/_version_gate.py
@@ -1,0 +1,72 @@
+"""Version gate: reject outdated difyctl clients on /openapi/v1 with HTTP 426.
+
+difyctl and the ``/openapi/v1`` surface ship in lockstep. A breaking path change
+(resource-oriented paths) means an outdated difyctl calls removed paths and would
+otherwise get a bare 404. This gate reads the difyctl version from the User-Agent
+and returns ``426 Upgrade Required`` with an upgrade hint when the client is older
+than the minimum this server supports.
+
+Design record: dify-todo/inprogress/difyctl/openapi/version.md
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Final
+
+from flask import Blueprint, Response, request
+from packaging.version import InvalidVersion, Version
+
+from controllers.openapi._errors import ErrorBody, OpenApiErrorCode
+
+# Oldest difyctl this server serves. Bumped in lockstep with breaking
+# /openapi/v1 changes (paired with difyctl's own version + its MIN_DIFY_VERSION).
+MIN_DIFYCTL_VERSION: Final = "0.2.0"
+
+_UPGRADE_HINT: Final = "Upgrade difyctl: https://docs.dify.ai/en/cli/install"
+
+# difyctl sends `User-Agent: difyctl/<semver> (<os>; <arch>; <channel>)`.
+_DIFYCTL_UA_RE = re.compile(r"^difyctl/(\d+\.\d+\.\d+(?:-[\w.]+)?)")
+
+_PREFIX: Final = "/openapi/v1/"
+
+# Paths a too-old client must still reach to discover that it is outdated.
+_ALLOWLIST: Final = frozenset({"/openapi/v1/_version", "/openapi/v1/_health"})
+
+
+def _upgrade_required_response(client_version: str) -> Response:
+    body = ErrorBody(
+        code=OpenApiErrorCode.UPGRADE_REQUIRED,
+        message=f"difyctl {client_version} is no longer supported; upgrade to >= {MIN_DIFYCTL_VERSION}.",
+        status=426,
+        hint=_UPGRADE_HINT,
+    )
+    return Response(body.model_dump_json(exclude_none=True), status=426, mimetype="application/json")
+
+
+def attach_version_gate(bp: Blueprint) -> None:
+    """Reject difyctl clients older than ``MIN_DIFYCTL_VERSION`` with 426.
+
+    Registered app-wide (``before_app_request``) rather than blueprint-scoped so it
+    also fires for requests to *removed* paths — those no longer match an openapi
+    route and would 404 before a blueprint-scoped ``before_request`` ever runs. The
+    prefix guard scopes it back to ``/openapi/v1``. Fails open for non-difyctl or
+    unparseable User-Agents (only a confidently-too-old difyctl is blocked).
+    """
+
+    @bp.before_app_request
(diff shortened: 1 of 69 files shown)


api/controllers/openapi/app_dsl.py
```diff
@@ -153,7 +153,7 @@ def get(self, app_id: str, *, auth_data: AuthData, query: AppDslExportQuery):
         return AppDslExportResponse(data=data), 200
 
 
-@openapi_ns.route("/apps/<string:app_id>/check-dependencies")
+@openapi_ns.route("/apps/<string:app_id>/dependencies")
```
GareArc: The noun reads as "list this app's dependencies", but the handler only returns the bad ones (referenced in the DSL but not installed). I would suggest adding a `:check` action suffix.

lin-snow: Good catch, agreed. `/dependencies` over-promised — the handler only returns the leaked deps (referenced in the DSL but not installed), not the full set. Changed it to the read-only custom method `GET /apps/{app_id}/dependencies:check`, which also leaves the plain noun free for a future full listing. Contract + docs regenerated and difyctl/mock/tests updated in 0863f94. Thanks!

GareArc (dismissed):
LGTM

api/controllers/openapi/_version_gate.py
```diff
@@ -0,0 +1,71 @@
...
+from controllers.openapi._errors import ErrorBody, OpenApiErrorCode
+
+# Oldest difyctl this server serves. Bumped in lockstep with breaking
+# /openapi/v1 changes (paired with difyctl's own version + its MIN_DIFY_VERSION).
+MIN_DIFYCTL_VERSION: Final = "0.2.0-alpha"
+
+_UPGRADE_HINT: Final = "Upgrade difyctl: https://docs.dify.ai/en/cli/install"
+
+# difyctl sends `User-Agent: difyctl/<semver> (<os>; <arch>; <channel>)`.
+_DIFYCTL_UA_RE = re.compile(r"^difyctl/(\d+\.\d+\.\d+(?:-[\w.]+)?)")
```
GareArc, line 23: Version comparison on both sides is suffix-sensitive now. I would suggest limit the comparison to the numeric core (major.minor.patch) and exclude the channel/prerelease suffix from ordering.

api/controllers/openapi/_version_gate.py
```diff
@@ -0,0 +1,71 @@
...
+from typing import Final
+
+from flask import Blueprint, Response, request
+from packaging.version import InvalidVersion, Version
+
+from controllers.openapi._errors import ErrorBody, OpenApiErrorCode
+
+# Oldest difyctl this server serves. Bumped in lockstep with breaking
+# /openapi/v1 changes (paired with difyctl's own version + its MIN_DIFY_VERSION).
+MIN_DIFYCTL_VERSION: Final = "0.2.0-alpha"
```
GareArc: Hand-edited constant inside a code file. Suggest storing it the same way the server's own version is propagated(via pyproject.toml)

api/pyproject.toml
```diff
@@ -52,6 +52,12 @@ dependencies = [
 # Before adding new dependency, consider place it in
 # alphabet order (a-z) and suitable group.
 
+[tool.dify]
+# Oldest difyctl served on /openapi/v1. Bump in lockstep with breaking /openapi/v1
+# changes (paired with difyctl's own version + its compat.minDify). See
+# dify-todo/inprogress/difyctl/openapi/version.md.
```
GareArc: Seems like a local doc reference.

GareArc (approved):
lgtm

wylswz (approved):

PR #38373, opened 2026-07-03
fix(api): don't mutate workflow.graph_dict in single-node run

### LINKED ISSUE #38353: make _get_graph_and_variable_pool_for_single_node_run not mut input dict
how to fix it @dosu

manan-tech:
## Summary

Fixes #38353.

`WorkflowBasedAppRunner._get_graph_and_variable_pool_for_single_node_run` filters the workflow graph down to a single iteration/loop node. To do that it `cast(dict, …)`-ed the mapping returned by `workflow.graph_dict` and mutated it in place:

```python
graph_config = cast(dict[str, Any], graph_config)
...
graph_config["nodes"] = node_configs
...
graph_config["edges"] = edge_configs
```

`Workflow.graph_dict` is declared as a read-only `Mapping[str, Any]`, and there's an explicit TODO on it noting it can't be cached because *"some code in the codebase may modify the returned dict"* (`api/models/workflow.py`). This method is exactly that code. It's only harmless today because the property re-decodes JSON on every access — so the later full-graph read in the same method (`extract_variable_selector_to_variable_mapping(graph_config=workflow.graph_dict, …)`) gets a fresh, unmutated dict. If `graph_dict` were ever cached, the mutation would corrupt that read.

## Fix

Build a filtered **copy** and pass it to `DifyGraphInitContext` and `Graph.init`, instead of mutating the source mapping:

```python
filtered_graph_config = {**graph_config, "nodes": node_configs, "edges": edge_configs}
```

The now-unused `cast` import is removed. Behavior is unchanged (the filtered graph still reaches graph init exactly as before), but `workflow.graph_dict` is no longer mutated — honoring its read-only contract and unblocking the caching TODO.

## Tests

Added `test_get_graph_and_variable_pool_for_single_node_run_does_not_mutate_graph_dict`, which runs the method against a multi-node graph and asserts:
- the shared `workflow.graph_dict` still contains all original nodes/edges (not filtered in place), and
- the graph actually built (`Graph.init`) receives the filtered copy (only the target node).

`uv run python -m pytest tests/unit_tests/core/app/apps/test_workflow_app_runner_core.py` → 15 passed. `ruff` clean; `pyrefly` reports 0 errors on the changed source file.

## Checklist

- [x] I ran the affected unit tests, `ruff`, and `pyrefly` — all pass.

### DIFF
--- a/api/core/app/apps/workflow_app_runner.py
+++ b/api/core/app/apps/workflow_app_runner.py
@@ -1,7 +1,7 @@
 import logging
 import time
 from collections.abc import Mapping, Sequence
-from typing import Any, cast
+from typing import Any
 
 from pydantic import ValidationError
 
@@ -266,8 +266,6 @@ class WorkflowBasedAppRunner:
         if not graph_config:
             raise ValueError("workflow graph not found")
 
-        graph_config = cast(dict[str, Any], graph_config)
-
         if "nodes" not in graph_config or "edges" not in graph_config:
             raise ValueError("nodes or edges not found in workflow graph")
 
@@ -288,8 +286,6 @@ class WorkflowBasedAppRunner:
             or (start_node_id and node.get("id") == start_node_id)
         ]
 
-        graph_config["nodes"] = node_configs
-
         node_ids = [node.get("id") for node in node_configs]
 
         # filter edges only in the specified node type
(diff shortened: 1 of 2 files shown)


asukaminato0721:
cc @QuantumGhost this fix a todo

Issue #37604, opened 2026-06-18
ERROR Argument `SimpleNamespace` is not assignable to parameter


asukaminato0721:
run `uv run pyrefly check` under api/ directory, and fix them.

@dosu give me a full file list and tell me how to fix them.

asukaminato0721:
@dosu no, dont use Mock, use real class

PR #38392, opened 2026-07-03
refactor(test): replace SimpleNamespace with typed mocks in agent file request tests

ojasarora77:
#37604

## Summary

Replaces `SimpleNamespace` test doubles in `test_agent_file_request_service.py` with `MagicMock(spec=File)` for the download-request service tests.

From Cursor

## Test plan
- [x] `uv run pytest tests/unit_tests/services/test_agent_file_request_service.py -q -o addopts=` (6 passed)
- [x] `ruff check` on changed file

### DIFF
--- a/api/tests/unit_tests/services/test_agent_file_request_service.py
+++ b/api/tests/unit_tests/services/test_agent_file_request_service.py
@@ -2,18 +2,22 @@
 
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


ojasarora77:
Updated to use a real `graphon.file.File` instance instead of `MagicMock` — thanks for the feedback.

PR #38393, opened 2026-07-03
refactor(test): replace SimpleNamespace with typed mocks in schedule service tests

ojasarora77:
#37604

## Summary

Replaces the `_workflow` helper's `SimpleNamespace` cast with `Mock(spec=Workflow)` in `test_schedule_service.py`, matching the pattern already used elsewhere in the same file.

From Cursor

## Test plan
- [x] `uv run pytest tests/unit_tests/services/test_schedule_service.py -q -o addopts=` (50 passed)
- [x] `ruff check` on changed file

### DIFF
--- a/api/tests/unit_tests/services/test_schedule_service.py
+++ b/api/tests/unit_tests/services/test_schedule_service.py
@@ -1,3 +1,4 @@
+import json
 import unittest
 from datetime import UTC, datetime
 from typing import Any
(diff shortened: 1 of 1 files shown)


asukaminato0721:
no, don't use Mock

asukaminato0721 (changes requested):
.

ojasarora77:
Updated to use `Workflow.new()` instead of `Mock` — thanks for the review.

asukaminato0721 (approved):

PR #38470, opened 2026-07-06
fix(api): bind scoped lookups to nested owner refs

### LINKED ISSUE #37723: fix(api): bind pipeline document status updates to owner refs
## Problem

Pipeline document status paths passed document, dataset, and tenant IDs independently.

The `ERROR` path originally omitted tenant scope, while the `WAITING` path resolved a document only by ID.

PR #38177 established owner-bound lookups, but duplicated owner coordinates and one manual RAG failure query still expressed the same invariant in different ways.

## Decision

- Root refs remain flat and carry tenant plus root ID.
- Stable single-parent chains use nested immutable `NamedTuple` refs.
- Dataset children nest as `DatasetRef → DocumentRef → SegmentRef`.
- App children such as messages, annotations, and MCP servers nest their `AppRef`.
- `WorkflowRef` stays flat because its owner is polymorphic across App and Pipeline.
- Production code constructs refs through the relevant ref service.
- Every consuming query still applies the complete owner predicates because refs are value objects, not authorization tokens.
- Refs stay internal and are reconstructed from validated database state after payload or async boundaries.

## Scope

- Validate Pipeline tenant and Dataset ownership before creating refs.
- Resolve pipeline `ERROR` and `WAITING` updates through one document resolver.
- Align the RAG node-failure path from #38177 with the same DatasetRef and DocumentRef resolver.
- Migrate Dataset and App child consumers to nested owner access.
- Do not add `NewType`, constructor tokens, generic ref bases, compatibility fields, repositories, migrations, or indexes.

## Acceptance

- Cross-tenant and cross-dataset documents are not updated.
- Document resolution uses document, dataset, and tenant predicates.
- Missing or mismatched owners remain a no-op where existing behavior requires it.
- Ruff, mypy, pyrefly, and focused backend tests pass.

Implemented by #38470.

Related to #37983 and #38177.

WH-2099:
## Summary

- Nest Dataset, Document, Segment, and stable App child refs under their validated owner refs.
- Validate Pipeline tenant and Dataset ownership before constructing document refs.
- Route pipeline `ERROR`, `WAITING`, and RAG node-failure document updates through the same owner-bound resolver.
- Migrate document deletion, segment, child chunk, message, annotation, and MCP server consumers to the resulting owner chains.
- Record the reusable owner-ref rules in `api/AGENTS.md`.

## Design

- Root refs carry tenant plus root ID.
- Stable single-parent resources nest their parent ref.
- `WorkflowRef` remains flat because its owner can be either an App or a Pipeline.
- Refs are immutable value objects, not authorization tokens.
- Production factories establish the owner chain, and consuming SQL queries enforce the complete predicates.
- Refs do not cross Pydantic or Celery boundaries.

## Validation

- 482 targeted backend unit tests passed across the pipeline, Dataset, RAG, App audio, annotation, and controller paths.
- Ruff check and format check passed for all changed Python files.
- mypy and pyrefly passed for all changed production files.
- Container integration tests remain CI-only per repository convention.

Closes #37723

Follows up #38177

Refs #37983

### DIFF
--- a/api/services/annotation_service.py
+++ b/api/services/annotation_service.py
@@ -94,9 +94,11 @@ class AppAnnotationService:
     def _get_annotation_by_ref(annotation_ref: AnnotationRef, session: Session) -> MessageAnnotation | None:
         return session.scalar(
             select(MessageAnnotation)
+            .join(App, App.id == MessageAnnotation.app_id)
             .where(
                 MessageAnnotation.id == annotation_ref.annotation_id,
-                MessageAnnotation.app_id == annotation_ref.app_id,
+                App.id == annotation_ref.app.app_id,
+                App.tenant_id == annotation_ref.app.tenant_id,
             )
             .limit(1)
         )
@@ -343,15 +345,15 @@ class AppAnnotationService:
         session.commit()
         # if annotation reply is enabled , add annotation to index
         app_annotation_setting = session.scalar(
-            select(AppAnnotationSetting).where(AppAnnotationSetting.app_id == annotation_ref.app_id).limit(1)
+            select(AppAnnotationSetting).where(AppAnnotationSetting.app_id == annotation_ref.app.app_id).limit(1)
         )
 
         if app_annotation_setting:
             update_annotation_to_index_task.delay(
                 annotation.id,
                 annotation.question_text,
-                annotation_ref.tenant_id,
-                annotation_ref.app_id,
+                annotation_ref.app.tenant_id,
+                annotation_ref.app.app_id,
                 app_annotation_setting.collection_binding_id,
             )
 
@@ -368,7 +370,7 @@ class AppAnnotationService:
 
         annotation_hit_histories = session.scalars(
             select(AppAnnotationHitHistory).where(
-                AppAnnotationHitHistory.app_id == annotation_ref.app_id,
+                AppAnnotationHitHistory.app_id == annotation_ref.app.app_id,
                 AppAnnotationHitHistory.annotation_id == annotation_ref.annotation_id,
             )
         ).all()
@@ -379,14 +381,14 @@ class AppAnnotationService:
         session.commit()
         # if annotation reply is enabled , delete annotation index
         app_annotation_setting = session.scalar(
-            select(AppAnnotationSetting).where(AppAnnotationSetting.app_id == annotation_ref.app_id).limit(1)
+            select(AppAnnotationSetting).where(AppAnnotationSetting.app_id == annotation_ref.app.app_id).limit(1)
         )
 
         if app_annotation_setting:
             delete_annotation_index_task.delay(
                 annotation.id,
-                annotation_ref.app_id,
-                annotation_ref.tenant_id,
+                annotation_ref.app.app_id,
+                annotation_ref.app.tenant_id,
                 app_annotation_setting.collection_binding_id,
             )
 
@@ -395,8 +397,13 @@ class AppAnnotationService:
         # Fetch annotations and their settings in a single query
         annotations_to_delete = session.execute(
             select(MessageAnnotation, AppAnnotationSetting)
+            .join(App, App.id == MessageAnnotation.app_id)
             .outerjoin(AppAnnotationSetting, MessageAnnotation.app_id == AppAnnotationSetting.app_id)
-            .where(MessageAnnotation.id.in_(annotation_ids), MessageAnnotation.app_id == app_ref.app_id)
+            .where(
+                MessageAnnotation.id.in_(annotation_ids),
+                App.id == app_ref.app_id,
+                App.tenant_id == app_ref.tenant_id,
+            )
         ).all()
 
         if not annotations_to_delete:
@@ -576,7 +583,7 @@ class AppAnnotationService:
         stmt = (
             select(AppAnnotationHitHistory)
             .where(
-                AppAnnotationHitHistory.app_id == annotation_ref.app_id,
+                AppAnnotationHitHistory.app_id == annotation_ref.app.app_id,
                 AppAnnotationHitHistory.annotation_id == annotation_ref.annotation_id,
             )
             .order_by(AppAnnotationHitHistory.created_at.desc())
--- a/api/services/annotation_service.py
+++ b/api/services/annotation_service.py
@@ -94,11 +94,9 @@ class AppAnnotationService:
     def _get_annotation_by_ref(annotation_ref: AnnotationRef, session: Session) -> MessageAnnotation | None:
         return session.scalar(
             select(MessageAnnotation)
-            .join(App, App.id == MessageAnnotation.app_id)
             .where(
                 MessageAnnotation.id == annotation_ref.annotation_id,
-                App.id == annotation_ref.app.app_id,
-                App.tenant_id == annotation_ref.app.tenant_id,
+                MessageAnnotation.app_id == annotation_ref.app.app_id,
             )
             .limit(1)
         )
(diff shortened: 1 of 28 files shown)


api/services/dataset_ref_service.py
```diff
@@ -40,17 +40,23 @@ def create_dataset_ref(dataset: Dataset) -> DatasetRef:
...
             return None
-        return DocumentRef(
-            tenant_id=dataset_ref.tenant_id,
-            dataset_id=dataset_ref.dataset_id,
-            document_id=document.id,
-        )
+        return DatasetRefService.create_document_ref_from_id(dataset_ref, document.id)
+
+    @staticmethod
+    def create_document_ref_from_id(dataset_ref: DatasetRef, document_id: str) -> DocumentRef:
```
QuantumGhost, line 47: I'm wondering if we should allow direct construction without validating ownership. This seems a source of vulnerability.

At least we should add a short docstring and warn the caller that this function should only be called with trusted ids.

api/services/rag_pipeline/pipeline_generate_service.py
```diff
@@ -37,8 +38,12 @@ def generate(
         try:
             workflow = cls._get_workflow(pipeline, invoke_from, session)
             if original_document_id := args.get("original_document_id"):
-                # update document status to waiting
-                cls.update_document_status(original_document_id, session=session)
+                dataset = pipeline.retrieve_dataset(session)
+                if dataset is None or dataset.tenant_id != pipeline.tenant_id:
+                    raise ValueError("Pipeline dataset is required")
```
QuantumGhost, line 43: Could we use a dedicated error type here?

api/services/rag_pipeline/pipeline_generate_service.py
```diff
@@ -37,8 +38,12 @@ def generate(
         try:
             workflow = cls._get_workflow(pipeline, invoke_from, session)
             if original_document_id := args.get("original_document_id"):
```
QuantumGhost, line 40: The `original_document_id` may be untrusted.

api/services/annotation_service.py
```diff
@@ -390,8 +392,13 @@ def delete_app_annotations_in_batch(cls, app_ref: AppRef, annotation_ids: list[s
         # Fetch annotations and their settings in a single query
         annotations_to_delete = session.execute(
             select(MessageAnnotation, AppAnnotationSetting)
+            .join(App, App.id == MessageAnnotation.app_id)
```
QuantumGhost: Is it really necessary to join `App` table here? IMO we should avoid this since `AppRef.app_id` should come from trusted source.

api/services/annotation_service.py
```diff
@@ -94,9 +94,11 @@ class AppAnnotationService:
     def _get_annotation_by_ref(annotation_ref: AnnotationRef, session: Session) -> MessageAnnotation | None:
         return session.scalar(
             select(MessageAnnotation)
+            .join(App, App.id == MessageAnnotation.app_id)
```
QuantumGhost: Is it necessary to join the `App` table here?

api/services/app_ref_service.py
```diff
@@ -54,17 +51,16 @@ def create_message_ref(
...
+            app=app_ref,
             message_id=message_id,
             end_user_id=end_user_id,
             account_id=account_id,
         )
 
     @staticmethod
     def create_annotation_ref(app_ref: AppRef, annotation_id: str) -> AnnotationRef:
-        return AnnotationRef(tenant_id=app_ref.tenant_id, app_id=app_ref.app_id, annotation_id=annotation_id)
+        return AnnotationRef(app=app_ref, annotation_id=annotation_id)
```
QuantumGhost, line 63: The comment below about `DocumentRef` construction applies to here, too.

api/services/app_ref_service.py
```diff
@@ -54,17 +51,16 @@ def create_message_ref(
...
 
     @staticmethod
     def create_mcp_server_ref(app_ref: AppRef, server_id: str) -> AppMCPServerRef:
-        return AppMCPServerRef(tenant_id=app_ref.tenant_id, app_id=app_ref.app_id, server_id=server_id)
+        return AppMCPServerRef(app=app_ref, server_id=server_id)
```
QuantumGhost, line 68: ditto

api/services/audio_service.py
```diff
@@ -33,7 +33,15 @@
 class AudioService:
     @staticmethod
     def _get_message_by_ref(session: Session, message_ref: MessageRef) -> Message | None:
-        stmt = select(Message).where(Message.id == message_ref.message_id, Message.app_id == message_ref.app_id)
+        stmt = (
+            select(Message)
+            .join(App, App.id == Message.app_id)
```
QuantumGhost: Is it really necessary to join `App` table here?

api/services/dataset_ref_service.py
```diff
@@ -40,17 +40,23 @@ def create_dataset_ref(dataset: Dataset) -> DatasetRef:
...
+        return DocumentRef(dataset=dataset_ref, document_id=document_id)
 
     @staticmethod
     def create_segment_ref(document_ref: DocumentRef, segment_id: str) -> SegmentRef:
-        return SegmentRef(
-            tenant_id=document_ref.tenant_id,
-            dataset_id=document_ref.dataset_id,
-            document_id=document_ref.document_id,
-            segment_id=segment_id,
+        return SegmentRef(document=document_ref, segment_id=segment_id)
```
QuantumGhost, line 54: ditto

api/services/dataset_ref_service.py
WH-2099: Addressed. The DatasetRef is created from an already-authorized dataset, while the child ID remains a candidate locator. I documented that refs are not authorization proof and that ownership is enforced when the ref is resolved with tenant and dataset predicates.

api/services/rag_pipeline/pipeline_generate_service.py
WH-2099: Addressed. The service now fails before dispatch when the candidate document does not resolve in the pipeline dataset. The Celery runner revalidates target and original document refs independently, and IndexProcessor scopes document and segment reads and deletes by document, dataset, and tenant.

api/services/rag_pipeline/pipeline_generate_service.py
WH-2099: I considered this, but I am intentionally keeping ValueError here for now. No caller currently branches on this condition, and equivalent pipeline dataset checks use the same simple error handling. Since the RAG pipeline is expected to be refactored/replaced, introducing a dedicated error taxonomy now would add short-lived complexity without changing behavior.

api/services/annotation_service.py
WH-2099: Good catch. I removed the redundant App join. The lookup now binds MessageAnnotation.id directly to the trusted AppRef.app_id, preserving the child ownership predicate without re-reading the authorized parent.

api/services/audio_service.py
WH-2099: Good catch. I removed the redundant App join. The message lookup now binds Message.id directly to the trusted app_id and retains the end-user/account predicates.

api/services/dataset_ref_service.py
WH-2099: Addressed. create_segment_ref now documents that the segment ID is only a candidate locator and that ownership must be enforced when the ref is consumed.

api/services/app_ref_service.py
WH-2099: Addressed. create_annotation_ref now documents that the annotation ID is only a candidate locator and that ownership must be enforced when the ref is consumed.

api/services/annotation_service.py
WH-2099: Good catch. I removed the App join. AppRef is created from an already-authorized App, so the batch query now scopes MessageAnnotation directly by the candidate annotation IDs and the trusted app_id.

api/services/app_ref_service.py
WH-2099: Addressed. create_mcp_server_ref now carries the same warning: the server ID is only a candidate locator, and ownership is enforced by the consuming lookup.

QuantumGhost (approved):

Issue #35510, opened 2026-04-23
improve output node


hjlarry:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] Please do not modify this template :) and fill in all the required fields.

### 1. Is this request related to a challenge you're experiencing? Tell me about your story.

Now there are 4 issues for the output node of workflow:

1. you can add multi output nodes,  but you only can add second output node by clicking the button of an edge. 
2. the user require to add a variable after add the output node, maybe add a default variable is better
3. when two output nodes has a same name variable, one of it will lost, but there is no any hint
4. if there are two variables in a single output node,  they will display in different line in the workflow result panel. but if the two vars in different output nodes, they will display mixed together

### 2. Additional context or comments

1. 

<img width="1000" height="864" alt="Image" src="https://github.com/user-attachments/assets/17c576bc-cd53-4441-a490-98d4b86318f2" />

2.

<img width="1236" height="500" alt="Image" src="https://github.com/user-attachments/assets/b73155d2-fbd4-4429-b280-253c072d9a29" />

3.

<img width="1540" height="456" alt="Image" src="https://github.com/user-attachments/assets/bd6e6193-8642-4f42-8da1-95dbca6fd79e" />

4.

<img width="2024" height="504" alt="Image" src="https://github.com/user-attachments/assets/7d8a2242-e9e7-44ce-831c-12c9b29831f3" />

### 3. Can you help us with this feature?

- [x] I am interested in contributing to this feature.

PR #38488, opened 2026-07-07
fix: allow duplicate End node output variables across branches

### LINKED ISSUE #38440: How can different end nodes output the same key variable?
### Self Checks

### Dify version

1.15.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Set two end nodes with the same input key

### ✔️ Expected Behavior

the workflow able to publish

### ❌ Actual Behavior

Error: A variable cannot be duplicated, causing the workflow to be unable to publish

EvanYao826:
## Summary

Remove the duplicate End output variable check that was unnecessarily blocking workflow publishing. In a workflow with conditional branches, only one End node executes per run, so having two End nodes in different branches with the same output variable name is not a conflict.

Fixes #38440

## Changes

- **`web/app/components/workflow/hooks/use-checklist.ts`**: Removed `getDuplicateEndOutputMessages()` function and all its call sites
- **`web/app/components/workflow/hooks/__tests__/use-checklist.spec.ts`**: Updated test to verify that duplicate output variable names across End nodes are now allowed (instead of detected as errors)

### DIFF
--- a/web/app/components/workflow/hooks/__tests__/use-checklist.spec.ts
+++ b/web/app/components/workflow/hooks/__tests__/use-checklist.spec.ts
@@ -423,7 +423,7 @@ describe('useChecklist', () => {
     ])
   })
 
-  it('should detect duplicate output variables across end nodes', () => {
+  it('should allow duplicate output variables across end nodes (different branches)', () => {
     const startNode = createNode({ id: 'start', data: { type: BlockEnum.Start, title: 'Start' } })
     const firstEndNode = createNode({
       id: 'end-1',
@@ -454,8 +454,9 @@ describe('useChecklist', () => {
     const firstWarning = result.current.find((item: ChecklistItem) => item.id === 'end-1')
     const secondWarning = result.current.find((item: ChecklistItem) => item.id === 'end-2')
 
-    expect(firstWarning?.errorMessages.some(message => message.includes('duplicateOutputVariable'))).toBe(true)
-    expect(secondWarning?.errorMessages.some(message => message.includes('duplicateOutputVariable'))).toBe(true)
+    // Different branches → only one End node executes, so duplicate output variables are allowed
+    expect(firstWarning?.errorMessages.some(message => message.includes('duplicateOutputVariable'))).toBe(false)
+    expect(secondWarning?.errorMessages.some(message => message.includes('duplicateOutputVariable'))).toBe(false)
   })
 
   it('should sync checklist items to the workflow store without render phase update warnings', async () => {
--- a/web/app/components/workflow/hooks/use-checklist.ts
+++ b/web/app/components/workflow/hooks/use-checklist.ts
@@ -108,43 +108,6 @@ const START_NODE_TYPES: BlockEnum[] = [
   BlockEnum.TriggerPlugin,
 ]
 
-const getDuplicateEndOutputMessages = (
-  nodes: Node[],
-  t: ReturnType<typeof useTranslation>['t'],
-) => {
-  const variableOccurrences = new Map<string, string[]>()
-
-  nodes.forEach((node) => {
-    if (node.type !== CUSTOM_NODE || node.data.type !== BlockEnum.End)
-      return
-
-    const outputs = ((node.data as { outputs?: Array<{ variable?: string }> }).outputs) || []
-    outputs.forEach((output) => {
-      const variable = output.variable?.trim()
-      if (!variable)
-        return
-
-      const occurrences = variableOccurrences.get(variable) || []
-      occurrences.push(node.id)
-      variableOccurrences.set(variable, occurrences)
-    })
-  })
-
-  const nodeMessages = new Map<string, string[]>()
-  variableOccurrences.forEach((nodeIds, variable) => {
-    if (nodeIds.length <= 1)
-      return
-
-    Array.from(new Set(nodeIds)).forEach((nodeId) => {
-      const messages = nodeMessages.get(nodeId) || []
-      messages.push(t('errorMsg.duplicateOutputVariable', { ns: 'workflow', variable }))
-      nodeMessages.set(nodeId, messages)
-    })
-  })
-
-  return nodeMessages
-}
-
 export const useChecklist = (nodes: Node[], edges: Edge[], options?: { flowType?: FlowType }) => {
   const { t } = useTranslation()
   const language = useGetLanguage()
@@ -233,7 +196,6 @@ export const useChecklist = (nodes: Node[], edges: Edge[], options?: { flowType?
   const needWarningNodes = useMemo<ChecklistItem[]>(() => {
     const list: ChecklistItem[] = []
     const filteredNodes = nodes.filter(node => node.type === CUSTOM_NODE)
-    const duplicateEndOutputMessages = getDuplicateEndOutputMessages(filteredNodes, t)
     const { validNodes } = getValidTreeNodes(filteredNodes, edges)
     const installedPluginIds = new Set(modelProviders.map(p => extractPluginId(p.provider)))
 
@@ -308,8 +270,6 @@ export const useChecklist = (nodes: Node[], edges: Edge[], options?: { flowType?
           }
           if (hasInvalidVar)
             errorMessages.push(t('errorMsg.invalidVariable', { ns: 'workflow' }))
-
-          errorMessages.push(...(duplicateEndOutputMessages.get(node!.id) || []))
         }
 
         const isStartNodeMeta = nodesExtraData?.[node!.data.type as BlockEnum]?.metaData.isStart ?? false
@@ -446,7 +406,6 @@ export const useChecklistBeforePublish = () => {
     } = workflowStore.getState()
     const nodes = getNodes()
     const filteredNodes = nodes.filter(node => node.type === CUSTOM_NODE)
-    const duplicateEndOutputMessages = getDuplicateEndOutputMessages(filteredNodes, t)
     const { validNodes, maxDepth } = getValidTreeNodes(filteredNodes, edges)
 
     if (maxDepth > MAX_TREE_DEPTH) {
@@ -558,12 +517,6 @@ export const useChecklistBeforePublish = () => {
 
       if (errorMessage) {
         toast.error(`[${node!.data.title}] ${errorMessage}`)
-        return false
-      }
-
-      const duplicateOutputMessages = duplicateEndOutputMessages.get(node!.id) || []
-      if (duplicateOutputMessages.length > 0) {
-        toast.error(`[${node!.data.title}] ${duplicateOutputMessages[0]}`)
         return false
       }
 


lyfuci:
Thanks for picking this up — the false positive in #38440 is real and worth fixing. I think removing
the check outright trades one bug for another though, and there are a few concrete problems with the
PR as it stands.

### 1. The new test doesn't exercise the case the PR claims to fix, and its assertion cannot pass

The rewritten test builds this graph:

```
start --> end-1
start --> end-2
```

Both edges leave `start` through the same handle, so this is a **parallel fan-out**, not a conditional
branch: both Output nodes run in the same execution. The test therefore now asserts that the genuine
data-loss case is acceptable, and there is no test anywhere for the if/else topology from #38440.

The assertion itself also cannot pass. Once the check is removed, `end-1` produces no checklist item at
all, so `firstWarning` is `undefined` and `firstWarning?.errorMessages.some(...)` short-circuits to
`undefined` rather than `false` — `expect(undefined).toBe(false)` fails. That is what the last CI run on
this branch reported:

```
FAIL  app/components/workflow/hooks/__tests__/use-checklist.spec.ts
  > useChecklist > should allow duplicate output variables across end nodes (different branches)
AssertionError: expected undefined to be false
```

`expect(firstWarning).toBeUndefined()` would be the assertion that matches the intent.

### 2. It re-opens the bug the check was added for

The check came in with #35511, fixing #35510, whose item 3 is *"when two output nodes has a same name
variable, one of it will lost, but there is no any hint"* — and item 4 (*"if the two vars in different
output nodes, they will display mixed together"*) confirms two Output nodes can emit in the same run.
Removing the check restores that silent overwrite for parallel branches, again with no hint.

### 3. It also drops detection of a variable duplicated inside one Output node

`getDuplicateEndOutputMessages` counts occurrences per variable name, so a single Output node
declaring `result` twice yielded `nodeIds = [n, n]` and was flagged. That case is a real conflict
regardless of topology, and it is lost too.

### 4. The branch is `CONFLICTING` with `main`

It was written against an older revision — it still uses `t('errorMsg.duplicateOutputVariable', …)`
whereas `main` now uses the typed accessor form `t(($) => $['errorMsg.duplicateOutputVariable'], …)`.

---

### Suggested direction: make the check branch-aware rather than removing it

Your diagnosis is right — the rule is too broad, not wrong. Flag a duplicate only when the two Output
nodes can actually run in the same execution:

- Exclusive branching is already visible in the graph as **distinct `edge.sourceHandle` values leaving
  one node** — `true`/case ids/`false` for If/Else, class ids for Question Classifier, `source` vs
  `fail-branch` for error handling, action ids vs `__timeout` for Human Input. Several edges leaving the
  *same* handle are a parallel fan-out, where all targets run.
- So: walking from the entry nodes, record for each node the sets of branch decisions that lead to it.
  Two Output nodes are exclusive when every way of reaching one contradicts every way of reaching the
  other; anything else may co-execute and is still reported. Tracking whole decision sets rather than
  one branch at a time matters, because branches often merge into a shared error Output node.

That keeps #38440 publishable, keeps #35510's protection for parallel Output nodes, and keeps the
intra-node case. Notably the existing parallel test then passes **unchanged** — no need to flip it.

Two things that argue this really is an editor-side over-restriction rather than a rule the rest of the
stack depends on:

- `WorkflowToolConfigurationUtils.get_workflow_graph_output`
  (`api/core/tools/utils/workflow_configuration_sync.py`) already collapses same-named outputs across
  End nodes into a single output variable — *"Later end nodes override duplicated variable definitions."*
  For mutually exclusive branches that merge is exactly the schema a user wants: one `result` parameter
  regardless of which branch ran. There is no backend validation rejecting the duplicate, so the editor
  is the only thing blocking it.
- The same check also gates the run/debug path, not just publishing: `useWorkflowRunValidation` calls
  `useChecklist` and bails on `needWarningNodes.length > 0`. So today a workflow like #38440's cannot be
  test-run either, which makes the false positive more costly than it first looks.

One implementation trap worth knowing about if you take this route: the `edges` array can contain
temporary `source_tmp` edges (`data._isTemp`, added by `handleNodeSelect` while highlighting variable
dependencies). They have to be filtered out, or they invent paths that bypass real branches and the
warning starts flickering as nodes are selected.

I've opened #39785 implementing exactly this — you're credited in the description for the diagnosis, and it closes #38440. Happy to fold in anything you prefer from your approach, or to close mine if you'd rather finish this one; you got here first, so it's your call.

lyfuci:
Hi @EvanYao826 — following up on my earlier comment.

I had opened #39785 with an alternative implementation for #38440. A maintainer closed it under the author-first/FIFO rule (yours was opened first) and asked me to coordinate with you here instead. So the work is yours to use however you like — this PR stays yours.

## What's in it

Instead of removing `getDuplicateEndOutputMessages`, it makes the check branch-aware, so the #38440 false positive goes away while the original guard stays intact:

- New `web/app/components/workflow/utils/end-output-conflicts.ts` — walks the graph from the entry node and records, for each node, the set of branch decisions (branch node → `sourceHandle` taken) that can reach it. Two Output nodes are treated as mutually exclusive only when *every* way of reaching one contradicts *every* way of reaching the other.
- Mutual exclusivity is inferred from the data (different `sourceHandle` off the same node), so there's no node-type allow-list to keep in sync — if-else, question-classifier, `fail-branch` and human-input `__timeout` are all covered by the same rule.
- It filters `_isTemp` edges (the ones injected for variable-dependency highlighting). Without that, selecting a node invents paths that bypass branches and the warning flickers as you click around.
- Per @iamjoel's review on #39785, it also compares `value_type`, splitting the result into `duplicateName` (nodes can co-execute → one value gets silently dropped) and `conflictingTypes` (mutually exclusive but incompatible types → the published output schema doesn't match one of the branches, since `get_workflow_graph_output` merges by name and the later End wins). Type compatibility follows the existing `filterVar` / `filterVarByType` convention: `any` is a wildcard, otherwise types must match, and a missing `value_type` is not flagged.

Two things that pushed me away from removing the check outright:

- It was added in #35511 for #35510 item 3 — parallel End nodes (same `sourceHandle`, so both execute) sharing a variable name silently drop one value. Removing the check brings that back with no warning at all.
- It's a hard publish gate rather than an advisory: `features-trigger.tsx` throws when `needWarningNodes.length > 0`. So downgrading it to a soft warning isn't an option either — the check itself has to get smarter.

The pre-existing parallel-fan-out test passes **unchanged**, which is the main signal that the original guard survived. 25 new unit tests plus 2 hook-level ones. CI was green when the PR was closed (29 passing / 7 skipped).

## Where to get it

- Diff: #39785
- Branch: `lyfuci/dify:fix/end-output-duplicate-branch-aware` (2 commits, `36e2563` + `0044aaf`)
- Patch: `https://github.com/langgenius/dify/pull/39785.patch`

## Happy to do the work

Whichever is easiest for you:

1. **Add me (@lyfuci) as a collaborator on your fork `EvanYao826/dify`.** Then I can push straight to `fix/allow-duplicate-end-output-variables`, rebase it onto current `main` (it's showing conflicts at the moment), and apply the change here. The PR and the credit stay yours.
2. I open a PR against your branch, and you merge it once you're happy with it.
3. You cherry-pick the two commits yourself — completely fine too, no need to involve me.

One note on option 1: "Allow edits from maintainers" is already enabled on this PR, but that only grants push access to dify maintainers, not to me — it needs a collaborator invite on your fork.

Either path needs a rebase first; glad to handle that part as well. Just let me know which you prefer.

lyfuci:
@crazywoola @iamjoel — a status note on this PR, since #39785 was closed in its favour under the author-first/FIFO rule.

That decision assumed this PR is the active implementation for #38440. As of today it has not moved since it was opened on 7 July (37 days):

- **One commit, and none since.** `4bb67c2` (7 Jul) is still the head.
- **CI has been red from the start.** The only run on this branch (`28833434132`) has `Web Tests` failing — specifically `Run Web Tests / Web Tests (1/4)`. The rewritten test asserts `expect(firstWarning?.errorMessages.some(...)).toBe(false)`, but once the check is removed there is no checklist item for `end-1` at all, so the expression evaluates to `undefined`, not `false`. `expect(firstWarning).toBeUndefined()` would be the assertion matching the intent.
- **The branch no longer merges** — GitHub reports it as conflicting. It predates the typed `t(($) => $['errorMsg.duplicateOutputVariable'], …)` accessor now used on `main`.
- **No review yet**, and the author has not responded to my two comments here (30 Jul, 6 Aug) or posted anywhere in the repo since 7 Jul.

Beyond the mechanics, the approach removes `getDuplicateEndOutputMessages` outright. That reinstates #35510 item 3 — two Output nodes on a *parallel* fan-out silently dropping one value, with no hint — and also drops detection of a variable name declared twice inside a single Output node.

Meanwhile the false positive is still live: I re-read `use-checklist.ts` on `main` at `c9c7fb9` and the topology-blind check is unchanged, with both call sites intact. Two releases have shipped since the issue was filed (1.16.0 on 17 Jul, 1.16.1 on 28 Jul), neither carrying a fix. Note the check gates the run/debug path as well as publishing (`useWorkflowRunValidation`), so an affected workflow cannot be test-run either.

I'm not asking to override the FIFO rule — just to apply whatever staleness fallback it has. If the original author is unavailable, could #39785 be reopened? I'm happy to rebase it onto current `main` first; CI was green when it was closed (29 passing / 7 skipped), and it already addresses @iamjoel's `value_type` review.

@EvanYao826 — still glad to route the work through your branch instead if you'd prefer to keep this PR. Any of the three options in my previous comment works for me; just say which.

chz3344:
This issue is blocking my workflow updates. I have to refactor the output structure across all my workflows, even though they don't run in parallel.

The implementation looks solid, and I'd really appreciate if this could land in the upcoming branch — I'd prefer not to rebuild the container just for this change.

lyfuci:
> This issue is blocking my workflow updates. I have to refactor the output structure across all my workflows, even though they don't run in parallel.
> 
> The implementation looks solid, and I'd really appreciate if this could land in the upcoming branch — I'd prefer not to rebuild the container just for this change.

I've been struggling with this issue for a long time too.

lyfuci:
Two findings from digging into why CI is red here. Both are small — sharing them in case they save you time.

**1. The failing test is this PR's own assertion shape, not the production change.**

`Run Web Tests (1/4)` fails on exactly one test — `should allow duplicate output variables across end nodes (different branches)`, at `use-checklist.spec.ts:458`:

```
AssertionError: expected undefined to be false
```

With the check removed, `end-1` produces no checklist item at all, so `result.current.find(item => item.id === 'end-1')` returns `undefined` and the optional chain yields `undefined`, which is not `false`. `toBeFalsy()` — or simply asserting the item is absent — would pass. Everything else is green (`1 failed | 615 passed`), so this one line is what stands between the PR and a green run.

**2. The topology in that test is a parallel fan-out, not two branches.**

Its edges are `createEdge({ source: 'start', target: 'end-1' })` and `createEdge({ source: 'start', target: 'end-2' })`. `createEdge` (`web/app/components/workflow/__tests__/fixtures.ts:97`) doesn't set `sourceHandle`, so both edges leave the Start node through the same handle — the graph fans out and **both End nodes execute**.

That is precisely the case the check was added for in #35511 (per #35510 item 3): two End nodes emitting the same variable name within one execution, where one value is silently dropped.

So flipping this particular test to expect no warning doesn't remove the #38440 false positive — it removes the true positive, and the new test name asserts a topology the fixture doesn't actually build. A test that reproduces #38440 would need an if-else (or question-classifier) node in between, with the two edges leaving *different* `sourceHandle`s.

I'm not raising this to relitigate the approach — the second point is just something that would be easy to miss, and the first is the thing blocking CI.

The branch-aware variant I offered earlier is still at #39785 / `lyfuci/dify:fix/end-output-duplicate-branch-aware`, and the offer to hand it over in whatever form is easiest for you still stands.

Issue #38518, opened 2026-07-07
Deleting a knowledge base while its documents are still indexing leaves orphaned segments, child_chunks and pgvector tables


shaba:
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report.
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.x (self-hosted, Community Edition), pgvector vector store, parent-child chunker

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Create a Knowledge base backed by a RAG pipeline and start ingesting a **large** source (tens of thousands of documents), so indexing runs for a while.
2. While the documents are **still indexing**, delete that Knowledge base from the UI (`DELETE /console/api/datasets/{id}`).

### ✔️ Expected Behavior

Deleting a dataset should first **cancel/await the in-flight indexing** for that dataset, then clean up completely, leaving **no orphaned rows or vector tables**: `documents`, `document_segments`, `child_chunks`, dataset_* rows and the `embedding_vector_index_<uuid>_nod` pgvector table should all be gone.

### ❌ Actual Behavior

The dataset row is removed (`DELETE` returns 204), but the cleanup does **not** complete and large amounts of data are **orphaned**:

- The still-running indexing pipeline keeps executing against the deleted dataset and repeatedly fails with `KnowledgeIndexNodeError: Dataset <id> not found.` (`core/rag/index_processor/index_processor.py` `index_and_clean`), producing a storm of `Node knowledgeBase failed with ABORT strategy` errors.
- `clean_dataset_task` does not effectively remove the data. In our case, after deleting 3 knowledge bases mid-indexing we were left with:
  - `document_segments`: ~2.4M orphaned rows (dataset_id no longer in `datasets`)
  - `child_chunks`: ~830k orphaned rows
  - 3 orphaned `embedding_vector_index_<uuid>_nod` pgvector tables (~34 GB)
  - `documents` / `dataset_process_rules`: ~27k / ~25k orphaned rows

  These were never cleaned and had to be removed manually.

### Root-cause notes

Two compounding problems:

[...]

PR #37773, opened 2026-06-22
fix: clean orphan segments during dataset deletion

xiaobao-k8s:
## Summary

Fixes #37066.

When document deletion and dataset deletion race in separate async tasks, the document row can be gone before `clean_dataset_task` runs while dataset-scoped `document_segments` still remain. `clean_dataset_task` previously skipped segment cleanup when `documents` was empty, leaving orphan segments behind. This patch always cleans dataset-scoped segments, even if the document rows are already gone.

## Changes

- Move `DocumentSegment` cleanup outside the `documents` existence branch in `clean_dataset_task`.
- Add a regression unit test for orphan segment cleanup when no document rows are found.

## Test Plan

```bash
cd api
uv run pytest -o addopts='' tests/unit_tests/tasks/test_clean_dataset_task.py::TestEdgeCases::test_clean_dataset_task_deletes_orphan_segments_without_documents -q -s
uv run pytest -o addopts='' tests/unit_tests/tasks/test_clean_dataset_task.py -q
cd ..
git diff --check
cd api && uv run ruff check tasks/clean_dataset_task.py tests/unit_tests/tasks/test_clean_dataset_task.py
```

## Real behavior proof

behavior: Dataset deletion cleanup now deletes dataset-scoped `document_segments` even when the corresponding `documents` rows have already been deleted by a racing async document-delete task.
environment: Repository `langgenius/dify`, worktree `/datad/github/dify-issue37066`, branch `fix/issue-37066-knowledge-storage`, commit `870bc5f89`, Python 3.12 via `uv`.
steps: Ran the targeted regression test and full clean_dataset_task unit test file after the patch:
```bash
cd /datad/github/dify-issue37066/api
uv run pytest -o addopts='' tests/unit_tests/tasks/test_clean_dataset_task.py::TestEdgeCases::test_clean_dataset_task_deletes_orphan_segments_without_documents -q -s
uv run pytest -o addopts='' tests/unit_tests/tasks/test_clean_dataset_task.py -q
```
evidence: Targeted test output:
```text
.
1 passed, 2 warnings in 1.83s
```
Full file output:
```text
........                                                                 [100%]
8 passed, 2 warnings in 2.07s
```
observedResult: The new test constructs a dataset cleanup run with zero documents but one remaining segment, and verifies that `DELETE FROM document_segments` is issued. Existing clean_dataset_task unit tests still pass.
[...]

PR #38519, opened 2026-07-07
fix(dataset): clean deleted dataset in committed steps and bounded batches

shaba:
Rebased onto current `main` and rewritten. The child-chunk leak that used to be bundled here is now a separate, much smaller PR — #41261 — so this one is purely about how the cleanup transacts. Refs #38518.

### The problem

`clean_dataset_task` runs after the dataset row is already gone. If the cleanup fails there is no owner left to trigger a retry, so a failure is permanent by construction.

Today the whole cleanup is one transaction, and the child tables are addressed through `IN (...)` lists built from every matching row:

```python
segment_delete_stmt = delete(DocumentSegment).where(DocumentSegment.id.in_(segment_ids))
```

One timeout or deadlock — a large dataset, or a lock still held by indexing that was running when the delete was issued — rolls back everything and orphans the documents, segments, files and the vector index for good. The single `except Exception` at the end also flattens every failure into one message, so the log does not say which part gave up.

### The change

The cleanup becomes a sequence of independent steps, each committed as it completes:

1. index (vector store + keyword table)
2. segments
3. documents
4. attachments
5. the remaining dataset-scoped rows (process rules, queries, app joins, metadata, pipeline/workflow)

`_run_step` runs one step, and on failure logs it with the step's name, rolls back that step's uncommitted work, and lets the remaining steps proceed. Work that was already committed stays committed regardless of what happens later. Partial cleanup beats none, and the log now names the step that failed.

**The index cleanup is deliberately step one with its own commit.** The index processors flush their relational side effects — child chunks, summaries, the keyword table — via `session.flush()` without committing. Under the current single transaction, a rollback in any later step silently undoes those deletions while the vector store has already been cleaned, leaving the relational rows pointing at an index that no longer exists. Committing right after the processor returns closes that window.

**Segments and documents are deleted in bounded batches** (`_DELETE_BATCH_SIZE = 1000`) instead of one statement per table, so each transaction and each parameter list stays small. The constant is read through the module global rather than a default argument so tests can patch it.

This also matches two rules from `api/AGENTS.md`: *"Keep write transactions explicit and bounded"* and *"Celery tasks that may be retried or redelivered must keep side effects idempotent."*

### On the size of this diff

Most of it is one 200-line function being split into named steps; the batching itself is a `while` loop and a constant. I looked at whether the transactional restructuring and the batching could ship as two PRs, and they cannot usefully: the restructuring alone is the same size as this whole diff, because the movement of code *is* the change. What was genuinely separable is the child-chunk fix, now in #41261 at +9 lines.

### Tests

Three cases added to `tests/unit_tests/tasks/test_clean_dataset_task.py`, all failing on `main`:

- `test_cleanup_commits_more_than_once` — counts `after_commit` events; one commit is what strands data
- `test_failing_step_keeps_earlier_work_and_still_runs_later_steps` — makes the documents step raise, then asserts the committed segment deletion survived, the later metadata step still ran, and the failing step's own rows were left alone
- `test_documents_are_deleted_across_multiple_batches` — batch size patched below the row count, asserts the loop drains the table

The file passes: 11 tests. `ruff check` and `ruff format --check` are clean.

### DIFF
--- a/api/tasks/clean_dataset_task.py
+++ b/api/tasks/clean_dataset_task.py
@@ -1,9 +1,12 @@
 import logging
 import time
+from collections.abc import Callable, Sequence
 
 import click
 from celery import shared_task
 from sqlalchemy import delete, select
+from sqlalchemy.orm import Session
+from sqlalchemy.sql.elements import ColumnElement
 
 from core.db.session_factory import session_factory
 from core.rag.index_processor.index_processor_factory import IndexProcessorFactory
@@ -12,6 +15,7 @@ from extensions.ext_storage import storage
 from models import WorkflowType
 from models.dataset import (
     AppDatasetJoin,
+    ChildChunk,
     Dataset,
     DatasetMetadata,
     DatasetMetadataBinding,
@@ -27,8 +31,17 @@ from models.workflow import Workflow
 
 logger = logging.getLogger(__name__)
 
+# The dataset row is already gone by the time this task runs, so if the cleanup
+# fails there is nothing left to trigger a retry. Deleting everything in a single
+# transaction meant one timeout/deadlock (e.g. on a multi-million-row dataset, or
+# a lock held by indexing that was still running at delete time) rolled the whole
+# task back and orphaned the documents, segments, child chunks and the vector
+# index forever. Instead: clean each concern independently and commit as we go, so
+# a failure in one step neither rolls back finished work nor skips the later steps,
+# and delete the large child tables in bounded batches.
+_DELETE_BATCH_SIZE = 1000
 
-# Add import statement for ValueError
+
 @shared_task(queue="dataset")
 def clean_dataset_task(
     dataset_id: str,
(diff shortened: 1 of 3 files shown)


shaba:
Scope / relationship to other work (see #38518):

This PR fixes the **cleanup-robustness** half — `clean_dataset_task` no longer deletes everything in one transaction that rolls back (and orphans documents, segments, child chunks and the pgvector index table) on a timeout or a lock held by indexing that is still running at delete time. Each concern is deleted in bounded, per-step-committed batches, each step is isolated so one failure doesn't skip the rest, and `child_chunks` is cleaned explicitly as a fallback.

It is **complementary to #37773** (document-vs-dataset segment deletion race) — a different code path, no overlap/conflict. The remaining part called out in #38518 — adding a guard so a dataset can't be deleted while its documents are still indexing (cancel/await in-flight indexing in `DatasetService.delete_dataset`) — is intentionally left as a separate follow-up.

Existing unit tests (`api/tests/unit_tests/tasks/test_clean_dataset_task.py`) pass unchanged (7 passed); the service-level `delete_dataset` tests mock `clean_dataset_task.delay`, so they are unaffected.

shaba:
Gentle ping — this has been open since 7 July without a review.

Context for anyone picking it up: the problem it fixes is still present on `main`. `clean_dataset_task` deletes documents, segments, child chunks, image files, attachments and bindings inside one transaction and commits once at the end (L191), rolling the whole thing back on any failure (L206). In practice a single failure mid-cleanup leaves the dataset row deleted and everything else orphaned, including the `embedding_vector_index_<uuid>_node` pgvector table, which then has to be cleaned up by hand.

This PR changes the cleanup to commit per step, so a later failure no longer undoes the work already done. One file, no API or schema change.

Happy to rebase, split it up, or adjust the batching if maintainers prefer a different shape — just let me know what would make it easier to review.

Related: #38518 (this issue) and #38522 (the missing guard against deleting a dataset while its documents are still indexing, which is the underlying race).

shaba:
Pushed a fix for the integration test failure; `Run API Tests / API Integration Tests (3.12)` and `API Unit Tests (3.12)` now both pass.

The failure was real and worth explaining. `test_clean_dataset_task_storage_exception_handling` makes every `storage.delete` raise, then reads `upload_file.id` after the task has run. On `main` that read is safe only because the storage exception propagates out of the document-file loop and rolls the entire cleanup back, so the row survives. With the cleanup no longer discarding committed work, the row is correctly gone and the expired instance raises `ObjectDeletedError` on reload.

The test was not asserting on this — it had `if len(remaining_files) > 0: print(...)` with a comment saying the behaviour depends on the implementation. It now captures the id before the task runs and states the expectation: the dataset row is already gone, so nothing will retry the cleanup, and a blob that cannot be deleted must not keep its row — the row would be orphaned forever, while a stray blob is still reclaimable by storage GC. That is already how the segment-image and attachment paths behave; this makes the document-file path agree with them.

The remaining `Web Full-Stack E2E` failure is `features/agent-v2/access-point.feature:69`, which is failing intermittently on unrelated PRs as well (for example run 32860893991 on `agent/sqlite-orm-app-pipelines`, same step and locator). It passed on an earlier run of this branch.

shaba:
Filed the E2E flake as #41263, with the sampling across recent CI runs and an analysis of where the race appears to be.

PR #38531, opened 2026-07-08
fix(cli): --insecure also skips TLS certificate verification

GareArc:
Resolves ENG-708

## Summary

`difyctl auth login --insecure` only allowed `http://` hosts. Against a self-signed `https://` host (e.g. an internal Dify instance) it still failed with `self signed certificate`. `--insecure` now also skips TLS certificate verification, so it covers both "no TLS" and "untrusted cert" as one blanket flag — consistent with how `--insecure`/`-k` works in most CLIs.

- `http/proxy.ts`: builds an undici `Agent`/`EnvHttpProxyAgent` with `rejectUnauthorized: false` when `insecure` is set
- Persisted per-host (`insecure_tls`, same pattern as the existing `scheme` field) so authenticated commands after login don't need `--insecure` again — and a later secure login to the same host clears it
- Threaded through login, logout, authed commands, and the version/compat probe via a shared `activeHostInfo()` helper

## Test plan
- [x] `pnpm type-check`, `pnpm lint:fix`, `pnpm test` (1217 passed, 0 failed)
- [x] Manual smoke test against a real self-signed HTTPS server: rejected without `--insecure`, accepted with it


cli/test/fixtures/tls/self-signed-key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
wylswz: I suggest generating the key pair in test setup instead of maintaining it in codebase, since its source is unknown (could be taken from arbitrary environment).

GareArc: Now generated on the fly

wylswz (approved):

PR #38548, opened 2026-07-08
fix: remove overly strict guard in dataset deletion handler to prevent orphaned vector collections

### LINKED ISSUE #38537: Deleting a dataset leaves an orphaned vector collection when `doc_form` or `indexing_technique` is empty (vector DB never cleaned up)
### Self Checks

### Dify version

v1.15.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Deploy Dify with `VECTOR_STORE=qdrant`.
2. Use a dataset whose `datasets.doc_form` and/or `datasets.indexing_technique` column is empty/null. This is common for knowledge bases created in older Dify versions and carried forward via upgrade. Index some documents so a Qdrant collection such as `Vector_index_<uuid>_Node` exists on the vector store.
3. Delete the dataset from the UI (or via `DELETE /datasets/{id}`).
4. Inspect Qdrant afterwards:

   ```
   GET /collections
   ```

   The `Vector_index_<uuid>_Node` collection is **still present**. Checking the Qdrant HTTP access log shows that **no `DELETE /collections/...` request was ever issued** during the deletion.

5. The orphaned collection persists permanently. Repeated deletes keep accumulating more orphans; they are never reconciled.

### ✔️ Expected Behavior

When a dataset is deleted, the vector-store collection that holds its embeddings must be reliably deleted (or scheduled for deletion that is guaranteed to run), so that no orphaned collection is left behind — regardless of whether `doc_form` / `indexing_technique` are populated.

### ❌ Actual Behavior

The `datasets` row is removed, but the corresponding Qdrant collection is **never deleted** and becomes an orphan. The deletion handler short-circuits before dispatching the cleanup task, and there is no retry/reconcile path, so the orphan is permanent.

### Root cause (code-level)

The signal handler that owns vector cleanup guards too aggressively and bails out for exactly the case the cleanup task is prepared to handle.

`api/events/event_handlers/clean_when_dataset_deleted.py`:

```python
@dataset_was_deleted.connect
def handle(sender: Dataset, **kwargs):
    dataset = sender
    if not dataset.doc_form or not dataset.indexing_technique:
        return                       # ← skips vector cleanup entirely
    clean_dataset_task.delay(
        dataset.id,
        dataset.tenant_id,
        dataset.indexing_technique,
        dataset.index_struct,
        dataset.collection_binding_id,
        dataset.doc_form,
        dataset.pipeline_id,
    )
```

The early `return` short-circuits cleanup for any dataset whose `doc_form` **or** `indexing_technique` is empty. Datasets created in older Dify versions (and migrated forward) frequently have one of these fields empty, so deleting them never dispatches `clean_dataset_task`, and the Qdrant collection is never removed.

This is also internally inconsistent. The very task it guards already defends against an empty `doc_form`:

```python
# api/tasks/clean_dataset_task.py
if doc_form is None or (isinstance(doc_form, str) and not doc_form.strip()):
    # Use default paragraph index type for empty/invalid datasets
    # to enable vector database cleanup
    doc_form = IndexStructureType.PARAGRAPH_INDEX
```

So the handler's guard is **stricter than the task it protects** — it blocks the precise case the task already handles. `indexing_technique` is likewise not required by the cleanup logic (it is only stored on the reconstructed `Dataset` object and not used to select the cleanup path), so guarding on it serves no purpose other than skipping cleanup.

A second, related weakness compounds the problem. Inside `clean_dataset_task`, the vector cleanup is wrapped in `try/except`; on failure it only logs and then proceeds to commit the rest of the DB cleanup:

```python
# api/tasks/clean_dataset_task.py
try:
    index_processor = IndexProcessorFactory(doc_form).init_index_processor()
    index_processor.clean(dataset, None, with_keywords=True, delete_child_chunks=True)
except Exception:
    logger.exception(...)
```

Because `delete_dataset` has already committed the deletion of the `datasets` row (`db.session.delete(dataset)` + `commit`) before the async task runs, a failed cleanup can never be retried — the orphan is permanent and silent.

### Impact

Orphans accumulate silently and invisibly (the API never reports them). On a qdrant container restart/recreate, **all** collections — including orphans — are loaded at once, spiking open file handles. If the container's `nofile` ulimit is modest, this triggers RocksDB `IO error: Too many open files` → Qdrant panic → `restart: always` loop → all knowledge-base retrieval fails. From the API side this surfaces as:

```
InactiveRpcError ... StatusCode.UNAVAILABLE
details: "DNS resolution failed for qdrant:6334: ... Domain name not found"
```

(The DNS error is a symptom: the qdrant container is crash-looping and intermittently unresolvable, not an actual DNS problem.)

We hit exactly this in production after a version upgrade that recreated the qdrant container: the container restart-looped ~100 times and knowledge-base retrieval was down.

AndrianBalanescu:
> [!IMPORTANT]
>

## Summary

Fixes #38537

When a dataset with empty `doc_form` or `indexing_technique` (common in datasets created by older Dify versions and migrated forward) was deleted, the signal handler in `clean_when_dataset_deleted.py` returned early and never dispatched `clean_dataset_task`. This left the corresponding vector store collection (e.g. Qdrant `Vector_index_<uuid>_Node`) permanently orphaned — the cleanup task was never run, and there was no retry/reconcile path.

The cleanup task already handles empty `doc_form` by falling back to `PARAGRAPH_INDEX` (see `clean_dataset_task.py` lines 79-89), and `indexing_technique` is not used to select the cleanup path — it is only stored on the reconstructed `Dataset` object. So the guard was stricter than the task it protected, blocking exactly the case the task already handles.

**Changes:**
- Removed the `if not dataset.doc_form or not dataset.indexing_technique: return` early-return guard in `api/events/event_handlers/clean_when_dataset_deleted.py` so cleanup is always dispatched.
- Updated the three integration tests in `test_dataset_service_delete_dataset.py` that codified the old (buggy) behavior — they now assert `clean_dataset_task.delay` is called with the correct arguments instead of asserting it was never called.

## Screenshots

N/A — backend-only change.

## Checklist

### DIFF
--- a/api/events/event_handlers/clean_when_dataset_deleted.py
+++ b/api/events/event_handlers/clean_when_dataset_deleted.py
@@ -6,8 +6,12 @@ from tasks.clean_dataset_task import clean_dataset_task
 @dataset_was_deleted.connect
 def handle(sender: Dataset, **kwargs):
     dataset = sender
-    if not dataset.doc_form or not dataset.indexing_technique:
-        return
+    # Always dispatch the cleanup task, even when doc_form or indexing_technique
+    # is empty. Datasets created in older Dify versions frequently have these
+    # fields null, and the cleanup task already handles empty doc_form by
+    # falling back to PARAGRAPH_INDEX (see clean_dataset_task.py).
+    # The previous early-return skipped vector cleanup for exactly these
+    # datasets, leaving orphaned collections in the vector store.
     clean_dataset_task.delay(
         dataset.id,
         dataset.tenant_id,
--- a/api/tasks/clean_dataset_task.py
+++ b/api/tasks/clean_dataset_task.py
@@ -33,10 +33,10 @@ logger = logging.getLogger(__name__)
 def clean_dataset_task(
     dataset_id: str,
     tenant_id: str,
-    indexing_technique: str,
-    index_struct: str,
-    collection_binding_id: str,
-    doc_form: str,
+    indexing_technique: str | None,
+    index_struct: str | None,
+    collection_binding_id: str | None,
+    doc_form: str | None,
     pipeline_id: str | None = None,
 ):
     """
--- a/api/tests/test_containers_integration_tests/services/test_dataset_service_delete_dataset.py
+++ b/api/tests/test_containers_integration_tests/services/test_dataset_service_delete_dataset.py
@@ -147,7 +147,7 @@ class TestDatasetServiceDeleteDataset:
         )
 
     def test_delete_empty_dataset_success(self, db_session_with_containers: Session):
-        """Delete an empty dataset without scheduling cleanup when both gating fields are absent."""
+        """Delete an empty dataset and still dispatch cleanup so vector store collections are removed."""
         # Arrange
         owner, tenant = DatasetDeleteIntegrationDataFactory.create_account_with_tenant(db_session_with_containers)
         dataset = DatasetDeleteIntegrationDataFactory.create_dataset(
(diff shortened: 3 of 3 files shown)


AndrianBalanescu:
Closing — no activity for 7+ days. Will reopen if still needed.

PR #38573, opened 2026-07-08
fix(web): re-add handleChangeConversation to handleNewConversation deps (#38403)

### LINKED ISSUE #38403: Possible regression since v1.13.0: embedded chatbot reset keeps using old conversation_id
## Description

I found a possible regression in the embedded chatbot reset behavior.

In Dify v1.6.0, the built-in reset / new conversation button seemed to correctly start a new conversation. However, since around Dify v1.13.0, and still on v1.15.0, after clicking the built-in reset button in the embedded Web App chatbot, subsequent messages may still be sent with the old `conversation_id`.

The chat UI appears to be reset, but the actual conversation seems to remain the previous one.

I am not a frontend expert, so my analysis may be wrong. However, I wanted to report the behavior because it seems reproducible and may be related to the reset conversation logic.

## Environment

- Dify version: observed since around v1.13.0, still reproducible on v1.15.0
- Deployment: Self-hosted
- App type: Embedded Web App chatbot

## Steps to reproduce

1. Open an embedded Web App chatbot.
2. Send a message and confirm that a `conversation_id` is created.
3. Click the built-in reset / new conversation button.
4. Send another message.
5. Check the network request or localStorage.

## Expected behavior

After clicking the reset / new conversation button, the embedded chatbot should clear the current conversation state and start a new conversation. The next message should not be sent to the old `conversation_id`.

## Actual behavior

After clicking the reset button, the UI looks reset, but the old `conversation_id` may still be used for subsequent messages.

## Related issues / PRs

This looks related to previous conversation reset issues:

- #33362
- #34731
- #35489
- #35519

However, I can still reproduce a similar behavior in later versions.

## Possible cause

I am not sure if this is the root cause, but I found a possible dependency regression.

In v1.6.0, `handleNewConversation` included `handleChangeConversation` in its dependency array:

```ts
}, [handleChangeConversation, setShowNewConversationItemInList, handleNewConversationInputsChange, setClearChatList])
```

In v1.15.0, `handleNewConversation` still calls `handleChangeConversation('')`, but `handleChangeConversation` is no longer included in the dependency array:

```ts
}, [isTryApp, setShowNewConversationItemInList, handleNewConversationInputsChange, setClearChatList])
```

This may cause a stale closure and prevent the latest `handleChangeConversation` / `conversationIdInfo` update logic from being used when resetting the conversation.

Adding `handleChangeConversation` back to the dependency array may fix the issue:

```diff
- }, [isTryApp, setShowNewConversationItemInList, handleNewConversationInputsChange, setClearChatList])
+ }, [isTryApp, setShowNewConversationItemInList, handleChangeConversation, handleNewConversationInputsChange, setClearChatList])
```

I may be missing some context, but this dependency was present in v1.6.0 and seems to have been removed when the `isTryApp` branch was introduced. This may not be a new regression in v1.15.0 specifically, but rather a regression introduced around the Try Apps changes and still reproducible in later versions.

Taranum01:
Fixes #38403

### Problem

Since PR #31287 ("feat: frontend part of support try apps"), the embedded chatbot's `handleNewConversation` (`web/app/components/base/chat/embedded-chatbot/hooks.tsx`) was missing `handleChangeConversation` from its `useCallback` deps array.

```diff
-  }, [handleChangeConversation, setShowNewConversationItemInList, handleNewConversationInputsChange, setClearChatList])
+  }, [isTryApp, setShowNewConversationItemInList, handleNewConversationInputsChange, setClearChatList])
```

After URL→localStorage resolution or any state change that touched `handleConversationIdInfoChange`'s deps (`appId`, `conversationIdInfo`, `userId`), `handleChangeConversation` regenerates but `handleNewConversation` keeps the previous render's closure. Clicking reset then calls `handleChangeConversation('')` against a stale `handleConversationIdInfoChange` that closes over the previous `conversationIdInfo`. The `setConversationIdInfo({ ...conversationIdInfo, ... })` write hits the discarded snapshot rather than the live state, so `conversationIdInfo[appId][userId]` is never reset. The next message re-uses the old `conversation_id`.

### Fix

Re-add `handleChangeConversation` to `handleNewConversation`'s deps array. This restores the v1.6.0 deps (pre-#31287) and keeps the two callbacks in lockstep:

```diff
-  }, [isTryApp, setShowNewConversationItemInList, handleNewConversationInputsChange, setClearChatList])
+  }, [isTryApp, setShowNewConversationItemInList, handleChangeConversation, handleNewConversationInputsChange, setClearChatList])
```

`react/exhaustive-deps` is now satisfied on line 371. `useCallback` only re-emits when a dep reference changes; at steady state (no state mutation across renders) both callbacks short-circuit and preserve identity, so the rerender cost is unchanged.

### Test

Added a regression test in `web/app/components/base/chat/embedded-chatbot/__tests__/hooks.spec.tsx` that:
1. Stores a `conversation_id` in localStorage and disables URL/system-variable resolution so the path under test is the localStorage branch.
2. Captures the initial `handleNewConversation` and `handleChangeConversation` references.
3. A no-prop `rerender` preserves both references (validates `useCallback` memoization at steady state).
4. Flips `mockStoreState.embeddedUserId` to force the line-75 effect → `userId` state → `handleConversationIdInfoChange` deps → `handleChangeConversation` to regenerate.
5. Asserts `handleNewConversation` regenerates alongside `handleChangeConversation` — without the fix this assertion fails (`Object.is` equality on the stale closure).

Run with:
```
pnpm exec vitest run app/components/base/chat/embedded-chatbot/__tests__/hooks.spec.tsx
```

### Verification

- `pnpm exec vitest run` on the touched spec: **36/36 pass**
- `pnpm exec vitest run` on `app/components/base/chat/` (1355 tests across 73 files): **all pass**
- `pnpm exec eslint --fix --pass-on-unpruned-suppressions` on the two changed files: 0 errors, 4 pre-existing `exhaustive-deps` warnings on unrelated `useEffect` blocks (lines 74, 105, 250, 311) — none introduced by this change
- `pnpm exec tsgo --noEmit`: clean
- Confirmed the new test fails on the unpatched code (reverted the 1-line deps change → test failed with `expected [AsyncFunction] not to be [AsyncFunction]`), then restored the fix and re-ran → test passes

### Risk

Minimal. Pure deps-array change. No public-API change, no schema/migration/controller/frontend logic outside the touched file. The change only affects callers of `handleNewConversation` (the embedded chatbot reset button) — and only positively: they now reach the freshest `handleChangeConversation` instead of a stale one.

Reverts the same `useCallback` deps edit as PR #31287 did back when it introduced the Try Apps branch. If PR #31287 needed a smaller-deps form for performance reasons, please call that out in review and we can revisit (the call to `handleChangeConversation('')[']` only happens once per reset click so the cost change is negligible).

### DIFF
--- a/web/app/components/base/chat/embedded-chatbot/__tests__/hooks.spec.tsx
+++ b/web/app/components/base/chat/embedded-chatbot/__tests__/hooks.spec.tsx
@@ -132,6 +132,26 @@ const renderWithClient = async <T,>(hook: () => T) => {
   }
 }
 
+const renderWithClientProps = async <T, P>(
+  hook: (props: P) => T,
+  initialProps: P,
+) => {
+  const queryClient = createQueryClient()
+  const wrapper = createWrapper(queryClient)
+  let result: ReturnType<typeof renderHook<T, P>> | undefined
+  act(() => {
+    result = renderHook(hook, { wrapper, initialProps })
+  })
+  await waitFor(() => {
+    if (queryClient.isFetching() > 0)
+      throw new Error('Queries are still fetching')
+  }, { timeout: 2000 })
+  return {
+    queryClient,
+    ...result!,
+  }
+}
+
 const createConversationItem = (overrides: Partial<ConversationItem> = {}): ConversationItem => ({
   id: 'conversation-1',
   name: 'Conversation 1',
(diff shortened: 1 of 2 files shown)


Taranum01:
Friendly bump — this PR has been sitting without review attention. Rebase + CI fixes applied, ready for review.

If anything needs to be addressed (additional tests, naming, scope, etc.), happy to iterate.

Taranum01:
Closing as superseded. Will re-derive against current `main` — `handleNewConversation` callback wiring may have shifted. Fresh PR to follow.

PR #38578, opened 2026-07-09
perf: select one message per run in SQL with DISTINCT ON

weijun-xia:
## Summary

Follow-up to #38359, addressing the reviewer suggestion:

> One possible follow-up: can we push the "one message per run" selection into SQL as well, so we don't load extra Message rows when multiple rows share the same workflow_run_id? If we do that, we should first make the aggregation rule explicit (for example latest or lowest-id message), since the current Python setdefault preserves whichever row comes back first.

### Before
`get_paginate_advanced_chat_workflow_runs` loaded every Message matching the page's run ids (`workflow_run_id IN (...)`) and de-duplicated to one per run in Python via `setdefault`, so:
- extra Message rows were fetched when a run had more than one message, and
- the kept row depended on the database row return order.

### After
Select a single message per run directly in SQL:

```
SELECT DISTINCT ON (workflow_run_id) ...
WHERE app_id = :app_id AND workflow_run_id IN (:run_ids)
ORDER BY workflow_run_id, created_at DESC, id DESC
```

- One row per run, so no extra Message rows are loaded.
- The aggregation rule is explicit and stable: the most recent message per run (`created_at` desc, `id` desc as a tie-break), rather than order-dependent.

`Message.id` is a UUID, so "lowest-id" is not time-ordered; `created_at` is used instead (there is a composite `(created_at, id)` index that supports this ordering).

### Notes
- Still one query for the whole page (no N+1), same as #38359.
- `DISTINCT ON` is PostgreSQL-specific, matching dify's supported database.

## Testing
- Added `test_get_paginate_advanced_chat_workflow_runs_dedups_one_message_per_run_in_sql`: compiles the query for the PostgreSQL dialect and asserts it uses `DISTINCT ON` with the explicit `created_at DESC` ordering (de-duplication happens in SQL, not Python).
- Existing workflow-run-service tests remain valid; `ruff check` and `ruff format` are clean.

### DIFF
--- a/api/services/workflow_run_service.py
+++ b/api/services/workflow_run_service.py
@@ -80,18 +80,31 @@ class WorkflowRunService:
         run_ids = [workflow_run.id for workflow_run in workflow_runs]
         messages_by_run_id: dict[str, Message] = {}
         if run_ids:
+            # Pick a single message per run in SQL via DISTINCT ON, instead of
+            # loading every matching row and de-duplicating in Python. This
+            # avoids fetching extra Message rows when a run has more than one
+            # message, and makes the selection rule explicit: the most recent
+            # message per run (created_at desc, id desc as a stable tie-break),
+            # rather than depending on the row return order.
             messages = db.session.scalars(
-                select(Message).where(
+                select(Message)
+                .where(
                     Message.app_id == app_model.id,
                     Message.workflow_run_id.in_(run_ids),
+                )
+                .distinct(Message.workflow_run_id)
+                .order_by(
+                    Message.workflow_run_id,
+                    Message.created_at.desc(),
+                    Message.id.desc(),
                 )
             ).all()
             for loaded_message in messages:
                 run_id = loaded_message.workflow_run_id
                 if run_id is None:
                     continue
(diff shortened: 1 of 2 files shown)


weijun-xia:
@wylswz following up on your review suggestion from #38359.

Before finalizing, I'd like to confirm one design decision with you: this PR makes the "one message per run" selection explicit as **the latest message per run**:

```sql
SELECT DISTINCT ON (workflow_run_id) ...
WHERE app_id = :app_id AND workflow_run_id IN (:run_ids)
ORDER BY workflow_run_id, created_at DESC, id DESC
```

I went with `created_at DESC` rather than the "lowest-id" option you mentioned, because `Message.id` is a UUID (not sequential / not time-ordered), so "lowest id" wouldn't actually map to the first/earliest message. Using `created_at` (with `id` as a stable tie-break) gives a well-defined "most recent message per run", and it matches how `workflow_event_snapshot_service` already resolves a run's message (`order_by(desc(created_at))`).

Could you confirm that **"latest message per run" is the aggregation rule you'd want here**? If you'd rather use "earliest message" (`created_at ASC`) or some other rule, I'm happy to switch — it's a one-line change in the `ORDER BY`.

PR #38586, opened 2026-07-09
feat: configurable per-node first-token timeout for LLM nodes (v1)

lin-snow:
## Summary

Adds an opt-in **first-token timeout** for workflow LLM-compatible nodes (LLM, question classifier, parameter extractor — in both workflow and chatflow). If the model does not stream its first token within the budget, the node fails with `FirstTokenTimeoutError` and is handled by the node's existing error strategy (retry / fail-branch / default). Off by default — no behavior change unless enabled. Works on the pinned `graphon==0.6.0`; no graphon changes required.

```mermaid
flowchart LR
  UI["web · model parameter panel<br/>first_token_timeout_ms<br/>(milliseconds, opt-in)"] --> CP["model.completion_params"]
  CP -->|"popped at the Dify boundary<br/>_normalize_completion_params"| MI["ModelInstance<br/>.first_token_timeout"]
  MI -->|"DifyPreparedLLM"| CV["api · ContextVar"]
  CV --> HX["httpx per-request<br/>read timeout"]
  HX -->|"first token too slow"| ERR["FirstTokenTimeoutError"]
  ERR --> STRAT["node error strategy<br/>retry / fail-branch / default"]
```

## How it works

- The config rides the same channel as `stop`: a synthetic frontend rule (`FIRST_TOKEN_TIMEOUT_PARAMETER_RULE` — int, milliseconds, default 2000, range 100–600000) writes `completion_params.first_token_timeout_ms`. It renders only in panels that pass `supportFirstTokenTimeout` (LLM / question classifier / parameter extractor), so surfaces whose model config is not consumed by this path never show a dead setting.
- Dify pops the key at the workflow model-config boundary (`_normalize_completion_params` in `core/app/llm/model_access.py`) into `ModelInstance.first_token_timeout` — the single ms→s conversion on the whole path. graphon and providers never see the key; invalid values (non-numeric, bool, non-positive) disable the gate.
- `DifyPreparedLLM` carries the value to the plugin-daemon transport through a `ContextVar`; enforcement is httpx's per-request `read` timeout in `BasePluginClient._stream_request`. The daemon withholds the SSE response headers until the model's first token, so the read timeout measures time-to-first-token directly.
- `FirstTokenTimeoutError` is Dify-local (`core/plugin/impl/first_token_timeout.py`, subclassing graphon's `InvokeError`); graphon's retry handler treats it like any node failure. A pin test asserts the type survives graphon's `_transform_invoke_error` unchanged, so `error_type` stays `FirstTokenTimeoutError` in run logs.

**Semantics** — the budget bounds **any single silent gap on the stream**; time-to-first-token is simply the first such gap (the daemon sends neither headers nor keep-alives before the first token, so the first gap ≡ TTFT). The `read` window resets on every received chunk, which means:

- it is a **stall detector, not a total-duration SLA** — a stream that keeps emitting within the window can run indefinitely;
- healthy inter-chunk gaps (tens of ms) sit an order of magnitude below any sane budget (≥ hundreds of ms), so normal streams never trip it mid-way;
- restricting the window to the first read only would require re-arming the connection's timeout mid-stream; keeping it for the whole stream is a deliberate simplicity trade-off, and also reaps streams that die mid-way instead of hanging until the daemon default.

| Situation | Surfaced as |
|---|---|
| No first token within the budget | `FirstTokenTimeoutError` → node error strategy |
| Inter-token gap exceeds the budget (after the first token) | transport error whose message names the configured window |
| Parameter not enabled / non-positive | gate disabled — no behavior change |

The budget deliberately **replaces** the operator-level `PLUGIN_DAEMON_TIMEOUT` read component for that request, so slow reasoning models can be granted more headroom than the daemon default.

**Coverage:** workflow / chatflow LLM, question classifier and parameter extractor nodes (all `LLMCompatibleNodeData`). Not rendered and not enforced elsewhere — agent nodes (models invoked by strategy plugins via backwards invocation), other workflow surfaces embedding the shared model panel (knowledge-retrieval, tool/trigger model params), and easy-UI apps (whose model-config converter also drops the key defensively).

## Screenshots

_The setting appears in the model parameter panel (next to Stop sequences) for the three supported node types. Screenshot to follow._

## Checklist

### DIFF
--- a/api/core/plugin/impl/base.py
+++ b/api/core/plugin/impl/base.py
@@ -57,13 +57,15 @@ match _plugin_daemon_timeout_config:
 
 
 def _read_timeout_for(first_token_timeout: float | None) -> httpx.Timeout | None:
-    """Narrow the daemon request timeout's ``read`` component to the first-token budget.
+    """Replace the daemon request timeout's ``read`` component with the first-token budget.
 
     The daemon withholds the response headers until the model's first token and sends no
     heartbeat before it, so httpx's ``read`` timeout (which covers both the header wait and
-    each body read) measures time-to-first-token. A non-positive budget disables the gate
-    and keeps the default timeout. ``httpx.Timeout(base, read=x)`` is rejected when ``base``
-    is a ``Timeout``, so the other components are copied explicitly.
+    each body read) measures time-to-first-token. The budget replaces the operator-level
+    read timeout outright rather than narrowing it — deliberately, so slow reasoning models
+    can be granted more headroom than ``PLUGIN_DAEMON_TIMEOUT``. A non-positive budget
+    disables the gate and keeps the default timeout. ``httpx.Timeout(base, read=x)`` is
+    rejected when ``base`` is a ``Timeout``, so the other components are copied explicitly.
     """
     base = plugin_daemon_request_timeout
     if not first_token_timeout or first_token_timeout <= 0:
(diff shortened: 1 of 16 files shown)


api/core/plugin/impl/base.py
```diff
@@ -181,30 +202,55 @@ def _stream_request(
...
         stream_kwargs: dict[str, Any] = {
             "method": method,
             "url": url,
             "headers": headers,
             "params": params,
             "files": files,
-            "timeout": plugin_daemon_request_timeout,
+            # The daemon holds the response headers until the first token, so a per-request
+            # read timeout equal to the first-token budget gates time-to-first-token.
+            "timeout": _read_timeout_for(first_token_timeout),
```
wylswz, line 216: This timeout could affect every event read in stream mode, which does not match its purpose.
```py
    def _receive_event(
        self, timeout: float | None = None
    ) -> h11.Event | type[h11.PAUSED]:
        while True:
            with map_exceptions({h11.RemoteProtocolError: RemoteProtocolError}):
                event = self._h11_state.next_event()

            if event is h11.NEED_DATA:
                data = self._network_stream.read(
                    self.READ_NUM_BYTES, timeout=timeout
                )
```

lin-snow:
consider using another design

PR #38598, opened 2026-07-09
fix: chunk workflow failure tracking data

CodingOnStar:
> [!IMPORTANT]
>

## Summary

- Change `workflow_run_failed` tracking to send serialized workflow data as `data.workflow_data_chunks`
- Split serialized workflow data into 900-character chunks to stay under Amplitude’s 1024-character string value limit
- Remove raw `workflow_data` object and `workflow_data_json` from the event payload to avoid polluting Amplitude event properties
- Skip workflow data chunks when serialization fails or workflow data is empty


MRZHUH (approved):
LGTM

PR #38620, opened 2026-07-09
fix(web): save agent configure draft on route leave

lyzno1:
## Summary
- Save the latest dirty Agent Configure draft when the Configure surface unmounts during client-side navigation.
- Reuse the existing page-close dirty-draft save path instead of only flushing the debounced autosave callback.
- Add coverage for leaving Configure before the debounce window fires.

## Root cause
The failing E2E scenario changed the Agent Configure prompt and immediately navigated away. The backend log showed no second `PUT /console/api/agent/:id/composer` after the UI edit, so the saved prompt remained the initial value. This was a product-side route-leave persistence gap, not a backend save issue and not a case where the E2E should wait longer.

## Product/test boundary
- The Configure persistence E2E is valid because it covers real user data loss on immediate SPA navigation.
- The external-runtime CI links failed before the external runtime step: the shared `e2e:full` preflight hit a stale docs URL assertion on an older commit. Current `main` already contains the docs URL update.
- A local true external-runtime chain passed after prepare/run, so external runtime itself was not the blocker reproduced by those CI links.

## Verification
- `pnpm -C web test features/agent-v2/agent-detail/configure/__tests__/use-agent-configure-sync.spec.tsx`
- `E2E_FORCE_WEB_BUILD=1 pnpm -C e2e e2e:full -- --name "Leaving Configure immediately after editing preserves prompt changes"`
- `pnpm -C e2e e2e:external:prepare`
- `pnpm -C e2e e2e:external`

### DIFF
--- a/web/features/agent-v2/agent-detail/configure/__tests__/use-agent-configure-sync.spec.tsx
+++ b/web/features/agent-v2/agent-detail/configure/__tests__/use-agent-configure-sync.spec.tsx
@@ -315,6 +315,40 @@ describe('useAgentConfigureSync', () => {
     })
   })
 
+  it('should save the latest dirty draft when Configure unmounts before autosave runs', async () => {
+    const { store, unmount } = renderUseAgentConfigureSync()
+
+    act(() => {
+      store.set(agentComposerDraftAtom, {
+        ...defaultAgentSoulConfigFormState,
+        prompt: 'Route leave prompt',
+      })
+    })
+
+    expect(composerPutMutationFn).not.toHaveBeenCalled()
+
+    await act(async () => {
+      unmount()
+      await Promise.resolve()
+    })
+
+    expect(composerPutMutationFn).toHaveBeenCalledTimes(1)
+    expect(composerPutMutationFn).toHaveBeenCalledWith(expect.objectContaining({
+      params: {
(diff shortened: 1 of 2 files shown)


BeautyyuYanli (approved):

lyzno1:
Local E2E verification completed on `codex/fix-agent-configure-leave-save` after the PR was opened.

- `E2E_FORCE_WEB_BUILD=1 pnpm -C e2e e2e:full`
  - `110 scenarios (60 passed, 50 skipped)`
  - `992 steps (750 passed, 242 skipped)`
  - no failed scenarios
- `pnpm -C e2e e2e:middleware:up`
- `pnpm -C e2e e2e:external:prepare`
  - generated `e2e/seed-report/agent-v2-external-runtime.json`
- `pnpm -C e2e e2e:external`
  - `10 scenarios (10 passed)`
  - `159 steps (159 passed)`
- `pnpm -C e2e e2e:middleware:down`

Notes:
- The first full run attempt did not reach Cucumber because local `web` dev-proxy was occupying port `5001`; I stopped that local process and reran from a clean E2E reset.
- After teardown, no `docker-` containers remained and ports `3000`/`5001` were free.

PR #38660, opened 2026-07-10
fix: stale credential cache in get_cached_provider_credentials for dynamic-select

### LINKED ISSUE #38633: Tool dynamic-select uses a credential_id-agnostic singleton cache that is never invalidated on credential update (stale credentials)
# Tool `dynamic-select` uses a credential_id-agnostic singleton cache that is never invalidated on credential update (stale credentials)

## Summary

When fetching options for a tool's `dynamic-select` parameter, Dify decrypts the provider
credentials through a **provider-level singleton cache** (`SingletonProviderCredentialsCache`)
that (a) ignores `credential_id` and (b) is **not invalidated** when a credential is created,
updated, or deleted. As a result:

- After editing a tool provider credential (e.g. changing a non-secret field such as
  `base_url`), the `dynamic-select` dropdown keeps using the **old** value for up to the cache
  TTL (24h), even though the DB row and the per-credential cache are already updated.
- With **multiple credentials** on the same provider (different `base_url` / API key), the
  singleton cache **collides** — whichever credential was decrypted first "wins", so the
  dropdown can silently use the wrong credential.

The regular tool invocation path is not affected because it uses the per-`credential_id`
cache. The inconsistency is specific to the `dynamic-select` (parameter options) path.

## Affected version

Dify 1.15.0 (code paths unchanged on `main` at time of writing). Plugin daemon 0.6.3.

## Steps to reproduce

1. Install a tool plugin with a `dynamic-select` parameter whose options are fetched from an
   external API using a `base_url` credential (any tool plugin implementing
   `_fetch_parameter_options`).
2. Configure a credential with `base_url = A`. Open the dropdown once (works; value `A` is now
   cached in the singleton cache).
3. Edit the same credential, set `base_url = B`, and save.
4. Open the dropdown again → it still calls `A` (stale) until the 24h TTL expires.
   - Variant: add a *second* credential with `base_url = B`; the dropdown may still use `A`
     because the singleton cache key does not include `credential_id`.

## Root cause (code references)

`get_dynamic_select_options` (tool branch) decrypts via `create_tool_provider_encrypter`:

- `api/services/plugin/plugin_parameter_service.py` — `PluginParameterService.get_dynamic_select_options`
  → `encrypter.decrypt(db_record.credentials)`
- `api/core/tools/utils/encryption.py` — `create_tool_provider_encrypter()` builds a
  `SingletonProviderCredentialsCache(tenant_id, provider_type, provider_identity)` — **no
  `credential_id`**.
- `api/core/helper/provider_cache.py` — `SingletonProviderCredentialsCache._generate_cache_key`
  → `"{provider_type}_credentials:tenant_id:{tenant_id}:id:{provider_type}.{identity_name}"`
  (one entry per provider, TTL 86400s).

Meanwhile the credential create/update/delete flow only invalidates the **per-credential**
cache:

- `api/services/tools/builtin_tools_manage_service.py` uses
  `ToolProviderCredentialsCache(tenant_id, provider, credential_id=...)` and calls
  `cache.delete()` on save/update/delete — this never touches the singleton cache above.
- Regular tool invocation also uses the per-credential cache
  (`api/core/tools/tool_manager.py`, `ToolProviderCredentialsCache`), which is why only
  `dynamic-select` is affected.

## Evidence

Confirmed against a running 1.15.0 instance: the Redis key

```
plugin_credentials:tenant_id:<tenant>:id:plugin.<provider>
```

held the **old** `base_url`, while the DB row and the per-credential
`tool_credentials:...:credential_id:<id>` key held the **new** value. Deleting the
singleton key made the dropdown pick up the correct value immediately (no restart needed).

## Suggested fix (any of)

1. Invalidate the singleton cache in the credential create/update/delete flow
   (`builtin_tools_manage_service`), in addition to the per-credential cache; **or**
2. Make the `dynamic-select` path use the per-`credential_id` `ToolProviderCredentialsCache`
   (consistent with `tool_manager`); **or**
3. Have `create_tool_provider_encrypter` take a `credential_id` and key the cache by it.

## Related

- #37986 (parent #37983) — credential visibility / usage-boundary consistency (mentions
  dynamic-option paths carrying `credential_id` inconsistently).
- #30160 / #30161 — frontend cache for *trigger* dynamic-select edit modal (different layer /
  provider type; not the same as this backend tool-credential cache).

gautamkishore:
Replace create_tool_provider_encrypter (uses SingletonProviderCredentialsCache) with create_provider_encrypter using NoOpProviderCredentialCache so every dynamic-select call decrypts fresh credentials from the database. The trigger path was unaffected since it reads credentials directly from the subscription record, not through the cached encrypter.

Closes #38633

### DIFF
--- a/api/services/plugin/plugin_parameter_service.py
+++ b/api/services/plugin/plugin_parameter_service.py
@@ -4,11 +4,12 @@ from typing import Any, Literal
 from sqlalchemy import select
 from sqlalchemy.orm import Session
 
+from core.helper.provider_cache import NoOpProviderCredentialCache
+from core.helper.provider_encryption import create_provider_encrypter
 from core.plugin.entities.parameters import PluginParameterOption
 from core.plugin.entities.plugin_daemon import CredentialType
 from core.plugin.impl.dynamic_select import DynamicSelectClient
 from core.tools.tool_manager import ToolManager
-from core.tools.utils.encryption import create_tool_provider_encrypter
 from core.trigger.entities.api_entities import TriggerProviderSubscriptionApiEntity
 from core.trigger.entities.entities import SubscriptionBuilder
 from extensions.ext_database import db
@@ -44,10 +45,11 @@ class PluginParameterService:
         match provider_type:
             case "tool":
                 provider_controller = ToolManager.get_builtin_provider(provider, tenant_id)
-                # init tool configuration
-                encrypter, _ = create_tool_provider_encrypter(
+                # init tool configuration (no caching — dynamic-select needs fresh credentials)
+                encrypter, _ = create_provider_encrypter(
                     tenant_id=tenant_id,
-                    controller=provider_controller,
+                    config=[x.to_basic_provider_config() for x in provider_controller.get_credentials_schema()],
+                    cache=NoOpProviderCredentialCache(),
                 )
 
                 # check if credentials are required
--- a/api/tests/test_containers_integration_tests/services/plugin/test_plugin_parameter_service.py
+++ b/api/tests/test_containers_integration_tests/services/plugin/test_plugin_parameter_service.py
@@ -43,7 +43,7 @@ class TestGetDynamicSelectOptionsTool:
         assert call_kwargs[0][5] == {}  # empty credentials
 
     @patch("services.plugin.plugin_parameter_service.DynamicSelectClient")
-    @patch("services.plugin.plugin_parameter_service.create_tool_provider_encrypter")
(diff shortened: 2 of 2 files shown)


fatelei:
test failed

gautamkishore:
Thanks @fatelei! I've pushed a fix — the test mocks still referenced the old `create_tool_provider_encrypter` name. Updated to `create_provider_encrypter`. CI should pass now.

gautamkishore:
CI only ran the labeler check — could you approve the workflow run? The test fix should be ready now.

PR #38688, opened 2026-07-10
test: use sqlite3 session in test_workflow_pause_events

asukaminato0721:
## Summary

## Validation

Part of #32454.


api/tests/unit_tests/core/app/apps/test_workflow_pause_events.py
```diff
@@ -63,6 +46,47 @@ def get_paused_nodes(self):
         return ["node-pause-1"]
 
 
+@pytest.fixture
```
41tair, line 47: Use sqlite_session and bind.

api/tests/unit_tests/core/app/apps/test_workflow_pause_events.py
```diff
@@ -216,8 +236,11 @@ def test_queue_workflow_paused_event_to_stream_responses(monkeypatch: pytest.Mon
     assert hi_resp.data.expiration_time == int(expiration_time.timestamp())
```
41tair, line 238: Use UTC timestamps to avoid time zone issues.

41tair (changes requested):

41tair:
UTC conversion is only applied to the live event path.Snapshot/reconnect still call timestamp() directly on the naive persisted expiration_time, causing timezone-dependent values and expiration-time changes after reconnect.

41tair (approved):

PR #38691, opened 2026-07-10
test: use sqlite3 session in test_snippet

asukaminato0721:
## Summary

## Validation

Part of #32454.


api/tests/unit_tests/models/test_snippet.py
```diff
@@ -1,10 +1,35 @@
...
+TENANT_ID = "11111111-1111-1111-1111-111111111111"
+WORKFLOW_ID = "22222222-2222-2222-2222-222222222222"
+APP_ID = "33333333-3333-3333-3333-333333333333"
+SNIPPET_ID = "44444444-4444-4444-4444-444444444444"
+ACCOUNT_1_ID = "55555555-5555-5555-5555-555555555555"
+ACCOUNT_2_ID = "55555555-5555-5555-5555-555555555556"
+
+
+@pytest.fixture
+def snippet_session(sqlite_engine: Engine, monkeypatch: pytest.MonkeyPatch) -> Iterator[Session]:
```
41tair: Use sqlite_session and keep the monkeypatch for db.session. When we fix the model @property problem, we can address the remaining part.

41tair (changes requested):

41tair (approved):

PR #38696, opened 2026-07-10
test: use sqlite3 session in test_external_dataset_service

asukaminato0721:
## Summary

## Validation

Part of #32454.


41tair:
Test Failed

41tair (changes requested):
Test Failed

41tair (approved):

PR #38698, opened 2026-07-10
test: use sqlite3 session in test_message_service

asukaminato0721:
## Summary

## Validation

Part of #32454.

### DIFF
--- a/api/tests/unit_tests/services/test_message_service.py
+++ b/api/tests/unit_tests/services/test_message_service.py
@@ -1,13 +1,35 @@
 import json
+from collections.abc import Iterator
 from datetime import datetime
-from unittest.mock import MagicMock, patch
+from decimal import Decimal
+from unittest.mock import MagicMock
 
 import pytest
+from sqlalchemy.engine import Engine
+from sqlalchemy.orm import Session, scoped_session
 
+import models.model as model_module
+import services.message_service as service_module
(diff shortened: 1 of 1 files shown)


api/tests/unit_tests/services/test_message_service.py
```diff
@@ -16,1247 +38,818 @@
...
-            conversation_id="conv-001",
-            session=mock_db.session,
+            conversation_id=conversation.id,
+            session=sqlite_session,
         )
 
-        # Assert
-        assert len(result.data) == 5
-        assert result.has_more is False
-        mock_conversation_service.get_conversation.assert_called_once()
```
41tair, line 539: we need to keep it

api/tests/unit_tests/services/test_message_service.py
```diff
@@ -16,1247 +38,818 @@
...
-            limit=10,
-            session=MagicMock(),
-        )
 
-        # Assert
-        assert isinstance(result, InfiniteScrollPagination)
-        assert result.data == []
-        assert result.limit == 10
-        assert result.has_more is False
+@pytest.fixture(autouse=True)
```
41tair, line 174: use sqlite_session.

41tair (changes requested):

41tair (approved):

PR #38740, opened 2026-07-10
test: use sqlite3 session in test_base_app_runner

asukaminato0721:
## Summary

- remove the fake ORM session from the multimodal fallback path
- bind the real SQLite engine and create only the MessageFile table
- verify the early return persists no file row while retaining tool/queue mocks

## Validation

- targeted pytest suite passed (17 tests)

Part of #32454.


41tair:
Conflicts

41tair (approved):

PR #38742, opened 2026-07-10
test: use sqlite3 session in test_llm_generator_missing

asukaminato0721:
## Summary

- replace suggestion-context query doubles with real SQLite Dataset rows
- cover tenant isolation and persisted creation ordering
- inject the deterministic database failure through a SQLAlchemy event hook while retaining tool-catalogue/model mocks

## Validation

- targeted pytest suite passed (18 tests)

Part of #32454.


api/tests/unit_tests/core/llm_generator/test_llm_generator_missing.py
```diff
@@ -1,8 +1,47 @@
...
+    """Expose the real SQLite session used by suggestion-context queries."""
+
+    session: Session
+
+    def __init__(self, session: Session) -> None:
+        self.session = session
+
+
+@pytest.fixture
+def dataset_session(sqlite_engine: Engine, monkeypatch: pytest.MonkeyPatch) -> Iterator[Session]:
```
41tair: use sqlite_session.

api/tests/unit_tests/core/llm_generator/test_llm_generator_missing.py
```diff
@@ -1,7 +1,111 @@
...
+
+    return ProviderEntity(
+        provider="test-provider",
+        label=I18nObject(en_US="Test Provider"),
+        supported_model_types=[ModelType.LLM],
+        configurate_methods=[],
+    )
+
+
+def _model_manager(sqlite_session: Session, *, has_default_model: bool = True) -> ModelManager:
```
41tair: Could we keep ModelManager mocked here? Only the Dataset query needs sqlite_session.

41tair (changes requested):

41tair (approved):

PR #38751, opened 2026-07-10
test: use sqlite3 session in test_account_service

asukaminato0721:
## Summary

- replace residual account-service ORM doubles with real SQLite sessions
- persist TenantAccountJoin rows and record real offset/limit query parameters
- use a real empty Session for RBAC early-return coverage while retaining external RBAC mocks

## Validation

- targeted pytest suite passed (100 tests)

Part of #32454.


41tair (commented):
Conflicts.Others is good.

41tair (dismissed):

41tair (approved):

PR #38753, opened 2026-07-10
test: use sqlite3 session in test_if_else

asukaminato0721:
## Summary

- remove mocked scoped-session close methods from if/else node tests
- bind a real SQLite scoped_session and establish caller-owned transactions
- verify session identity and transaction state survive node evaluation while preserving workflow/domain mocks

## Validation

- targeted pytest suite passed (12 tests)
- structural ORM-session scan passed

Part of #32454.

### DIFF
--- a/api/tests/unit_tests/core/workflow/nodes/test_if_else.py
+++ b/api/tests/unit_tests/core/workflow/nodes/test_if_else.py
@@ -1,13 +1,12 @@
 import time
 import uuid
-from unittest.mock import MagicMock, Mock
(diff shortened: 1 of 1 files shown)


41tair:
It seems we may not need a database session in these test cases, since the IfElse execution path doesn’t access the database.

41tair (changes requested):
It seems we may not need a database session in these test cases, since the IfElse execution path doesn’t access the database.

41tair (approved):

PR #38759, opened 2026-07-10
test: use sqlite3 session in test_agent_app_parameters

asukaminato0721:
Replace scalar-query session doubles with a real in-memory SQLite-backed session and persisted Agent and AgentConfigSnapshot rows. Verify missing records, publication state, tenant isolation, and projected app features. Part of #32454.


41tair (dismissed):

41tair:
Conflicts.

41tair (approved):

PR #38760, opened 2026-07-10
test: use sqlite3 session in test_auth_wraps

asukaminato0721:
Replace mocked session and context-manager objects with a real SQLite-bound sessionmaker. Persist and load an EndUser through the valid HMAC path while leaving authentication boundaries isolated. Part of #32454.

### DIFF
--- a/api/tests/unit_tests/controllers/inner_api/test_auth_wraps.py
+++ b/api/tests/unit_tests/controllers/inner_api/test_auth_wraps.py
@@ -2,10 +2,12 @@
 Unit tests for inner_api auth decorators
 """
 
-from unittest.mock import MagicMock, patch
+from unittest.mock import patch
+from uuid import NAMESPACE_URL, uuid5
 
 import pytest
 from flask import Flask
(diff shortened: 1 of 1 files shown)


api/tests/unit_tests/controllers/inner_api/test_auth_wraps.py
```diff
@@ -270,14 +277,13 @@ def protected_view(**kwargs):
             headers={"Authorization": "Bearer user123:wrong_signature", "X-Inner-Api-Key": "valid_key"}
         ):
             with patch.object(dify_config, "INNER_API", True):
-                with patch("controllers.inner_api.wraps.session_factory.create_session") as mock_create_session:
-                    result = protected_view()
+                result = protected_view()
```
41tair: This test no longer guarantees that invalid HMAC requests return before touching the database.

41tair (changes requested):

41tair (approved):

PR #38761, opened 2026-07-10
test: use sqlite3 session in test_sqlalchemy_workflow_execution_repository

asukaminato0721:
Replace SQLAlchemy session, factory, engine, user, and workflow-run doubles with a real in-memory SQLite database and persisted models. Verify inserts, JSON round trips, cache behavior, timestamps, updates, and cross-tenant overwrite rejection. Part of #32454.


api/tests/unit_tests/core/repositories/test_sqlalchemy_workflow_execution_repository.py
```diff
@@ -71,125 +74,147 @@ def sample_workflow_execution():
...
+        sqlite_session_factory: sessionmaker[Session],
+        account: Account,
+        sample_workflow_execution: WorkflowExecution,
+    ):
         repo = SQLAlchemyWorkflowExecutionRepository(
-            session_factory=mock_session_factory,
-            tenant_id=RESOURCE_TENANT_ID,
-            user=mock_account,
+            session_factory=sqlite_session_factory,
+            tenant_id=account.current_tenant_id,
```
41tair: The old conversion and save tests used RESOURCE_TENANT_ID, intentionally different from Account.current_tenant_id. The migrated tests use account.current_tenant_id for both.

41tair (changes requested):

41tair (approved):

PR #38762, opened 2026-07-10
test: use sqlite3 session in test_tool_label_manager

asukaminato0721:
Replace mocked SQLAlchemy sessionmaker chains and fabricated query rows with real in-memory SQLite sessions and persisted ToolLabelBinding models. Verify replacement, filtering, retrieval, batch mapping, and provider-type isolation through database state. Part of #32454.


41tair (approved):

WH-2099:
This change is included in #39156 through the latest main branch, which consolidates the related SQLite-backed unit-test migrations.

PR #38763, opened 2026-07-10
test: use sqlite3 session in test_passport

asukaminato0721:
Replace fabricated scalar sequences with a real in-memory SQLite session and persisted App, Site, and EndUser rows. Verify application and site lookup, empty end-user state, and session-ID collision queries while retaining external passport mocks. Part of #32454.


41tair (approved):

WH-2099:
This change is included in #39156 through the latest main branch, which consolidates the related SQLite-backed unit-test migrations.

PR #38764, opened 2026-07-10
test: use sqlite3 session in test_agent_config_service

asukaminato0721:
Replace SQLAlchemy session doubles in the agent configuration service tests with real in-memory SQLite sessions and persisted model state. Preserve external service boundaries and verify database-backed behavior. Part of #32454.

### DIFF
--- a/api/tests/unit_tests/services/test_agent_config_service.py
+++ b/api/tests/unit_tests/services/test_agent_config_service.py
@@ -4,11 +4,22 @@ from __future__ import annotations
 
 import io
 import zipfile
+from datetime import datetime
 from types import SimpleNamespace
-from unittest.mock import MagicMock, patch
+from unittest.mock import patch
 
 import pytest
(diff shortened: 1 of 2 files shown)


api/tests/unit_tests/services/test_agent_config_service.py
```diff
@@ -549,63 +642,6 @@ def test_manifest_uses_items_shape_without_download_urls() -> None:
...
-        AgentConfigSkillRefConfig(name="alpha")
-    with pytest.raises(ValueError, match="must not retain"):
-        AgentConfigFileRefConfig(
-            name="guide.txt",
-            file_kind="upload_file",
-            file_id="workspace-file-id",
-            is_missing=True,
-        )
-
-
```
41tair, line 608: why we delete those cases?

41tair (changes requested):

41tair (approved):

PR #38765, opened 2026-07-10
test: use sqlite3 session in test_qa_index_processor

asukaminato0721:
Replace the mocked SQLAlchemy session and scalar chain with a real SQLite-bound session factory and persisted DocumentSegment rows. Verify matching, wrong-node, and wrong-dataset filtering while preserving genuine external boundaries. Part of #32454.

### DIFF
--- a/api/tests/unit_tests/core/rag/indexing/processor/test_qa_index_processor.py
+++ b/api/tests/unit_tests/core/rag/indexing/processor/test_qa_index_processor.py
@@ -5,12 +5,17 @@ from unittest.mock import MagicMock, Mock, patch
 
 import pandas as pd
 import pytest
+from sqlalchemy.orm import Session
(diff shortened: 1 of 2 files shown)


api/tests/unit_tests/core/rag/indexing/processor/test_qa_index_processor.py
```diff
@@ -235,14 +250,49 @@ def test_load_skips_vector_for_non_high_quality(self, processor: QAIndexProcesso
 
         mock_vector_cls.assert_not_called()
 
+    @pytest.mark.usefixtures("_bind_sqlite_session_factory")
```
41tair: seems don't need this fixture.

41tair (changes requested):

41tair (approved):

PR #38559, opened 2026-07-08
refactor: explicit DB session propagation across backend paths

41tair:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

#37403

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [X] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [X] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [X] I've updated the documentation accordingly.
- [X] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods

PR #38786, opened 2026-07-10
refactor: pass db session from parameter in WorkflowRunService and controllers

### LINKED ISSUE #37403: make db.session pass from parameter.

saitejabandaru-in:
## Summary

Resolves #37403.

This PR refactors the `WorkflowRunService` method `get_paginate_advanced_chat_workflow_runs` to accept an explicit `session` parameter rather than relying on the implicit global `db.session` object. This makes the code more traceable, readable, and easier to test without global mock patching.

### Changes
- Updated `get_paginate_advanced_chat_workflow_runs` method in `WorkflowRunService` to accept `session: Session = None`. If not provided, it falls back to `db.session` for backwards compatibility.
- Updated the workflow run controller `WorkflowRunApi` to pass `db.session` to the service method call.
- Refactored `test_get_paginate_advanced_chat_workflow_runs_should_attach_message_fields_when_message_exists` in `test_workflow_run_service.py` to pass the mock/fake session directly, removing the need to monkeypatch `db.session`.
- Kept the other test using the monkeypatch to verify that the fallback/backwards-compatibility path functions as expected.


41tair:
Hi, thanks for your contribution. There are some style check failures. Could you address them?

saitejabandaru-in:
"Updated the branch with type check adjustments to resolve the mypy style check failure. scoped_session is now properly supported in the method signature. All tests and type checks are green!"

api/services/workflow_run_service.py
```diff
@@ -52,6 +52,7 @@ def get_paginate_advanced_chat_workflow_runs(
         app_model: App,
         args: WorkflowRunListArgs,
         triggered_from: WorkflowRunTriggeredFrom = WorkflowRunTriggeredFrom.DEBUGGING,
+        session: Session | scoped_session | None = None,
```
41tair: Could we make session required instead of retaining the None default and > db.session fallback? The goal of this refactor is to make the database dependency > explicit, but the fallback preserves the implicit global-session path. Please > update all call sites—including the unit and integration tests—to pass a session > directly. The controller can pass db.session(), while the integration test can > pass db_session_with_containers.

41tair (changes requested):

saitejabandaru-in:
Hi @41tair, I have updated the PR to address all comments:
1. Replaced the `db.session` global fallback with the concrete `db.session()` instance inside the controllers/views.
2. Made `session` fully required in the service signature.
3. Updated the type annotations to `Union[Session, scoped_session, Any]` to cleanly pass static analysis (Pyrefly diff check).

41tair:
Perhaps my description was unclear. Please see the modifications in https://github.com/langgenius/dify/pull/38274/changes.

saitejabandaru-in:
Closing this PR as #38559 has been merged to upstream main, which implements explicit DB session propagation across backend paths. Thanks!

PR #38869, opened 2026-07-13
fix(firecrawl): add bounded timeout to scrape/crawl requests

he-yufeng:
# Summary

`FirecrawlApp._post_request` and `_get_request` issue `httpx.post` / `httpx.get` with no timeout, so a stalled Firecrawl endpoint blocks the RAG extractor thread indefinitely, and the surrounding retry loop repeats that hang up to three times. This mirrors the bounded-timeout sweep already merged for the watercrawl client (#37515, #37685) and for Firecrawl credential validation (#37638): both helpers now pass a module-level `FIRECRAWL_REQUEST_TIMEOUT = httpx.Timeout(60.0, connect=10.0)`. The read budget stays generous because a single scrape can legitimately take a while, while the short connect budget and finite read mean an unresponsive endpoint fails fast instead of hanging the ingestion worker.

# Screenshots

Backend-only change, no UI.

# Checklist

- [x] I have added tests that prove my fix is effective.
- [x] `ruff format` and `ruff check` pass on the changed files.
- [x] This is a bug fix; no documentation update is required.

# Test

Added `test_post_and_get_request_pass_bounded_timeout`, which asserts both helpers send `FIRECRAWL_REQUEST_TIMEOUT` with a non-None read and connect budget. The firecrawl unit tests pass (32 passed), and `ruff format` / `ruff check` are clean on both files.

From Claude Code

### DIFF
--- a/api/core/rag/extractor/firecrawl/firecrawl_app.py
+++ b/api/core/rag/extractor/firecrawl/firecrawl_app.py
@@ -6,6 +6,12 @@ import httpx
 
 from extensions.ext_storage import storage
 
+# Bounded timeout for Firecrawl scrape/crawl requests. Without it httpx blocks
+# forever, so a stalled endpoint hangs the extractor thread (and, via the retry
+# loop, does it three times). The read budget is generous because a single
+# scrape can legitimately take a while; the connect budget stays short.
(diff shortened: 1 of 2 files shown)


he-yufeng:
Covered upstream: #39831 landed the same bounded timeout for the Firecrawl extractor, so this is redundant now. Closing.

PR #38923, opened 2026-07-14
test: move console OAuth coverage to unit tests

asukaminato0721:
## What changed

## Why

## Impact

## Validation


41tair:
Conflict

41tair (approved):

PR #38932, opened 2026-07-14
test: move message cleanup coverage to unit tests

asukaminato0721:
## What changed

## Why

## Impact

## Validation

### DIFF
--- a/api/tests/test_containers_integration_tests/services/test_messages_clean_service.py
+++ b/api/tests/test_containers_integration_tests/services/test_messages_clean_service.py
@@ -4,7 +4,7 @@ import datetime
 import json
 import uuid
 from decimal import Decimal
-from unittest.mock import MagicMock, patch
+from unittest.mock import patch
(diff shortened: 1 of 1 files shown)


api/tests/unit_tests/services/test_messages_clean_service.py
```diff
@@ -685,3 +685,56 @@ def test_run_records_completion_metrics_on_failure(self):
             service.run()
         assert len(completion_calls) == 1
         assert completion_calls[0]["status"] == "failed"
+
+
+class TestMessagesCleanServiceConstruction:
+    """Verify constructor helpers without exercising database cleanup."""
+
+    def test_from_time_range_validation(self) -> None:
```
41tair: This seems to be a duplicate of test_start_from_end_before_raises_value_error(. Other functions in this file seem to have the same issue

41tair (approved):

PR #38936, opened 2026-07-14
test: move password reset coverage to unit tests

asukaminato0721:
## What changed

## Why

## Impact

## Validation

### DIFF
--- a/api/tests/unit_tests/controllers/console/auth/test_password_reset.py
+++ b/api/tests/unit_tests/controllers/console/auth/test_password_reset.py
@@ -2,14 +2,12 @@
 
 from __future__ import annotations
 
-from types import SimpleNamespace
+from collections.abc import Iterator
 from unittest.mock import MagicMock, patch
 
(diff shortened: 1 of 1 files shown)


api/tests/unit_tests/controllers/console/auth/test_password_reset.py
```diff
@@ -325,11 +339,6 @@ def test_verify_code_wrong_code(self, mock_add_rate_limit, mock_get_data, mock_i
 class TestForgotPasswordResetApi:
     """Test cases for resetting password with verified token."""
 
-    @pytest.fixture
-    def app(self, flask_app_with_containers: Flask, db_session_with_containers: Session):
-        ensure_dify_setup(db_session_with_containers)
-        return flask_app_with_containers
-
     @pytest.fixture
     def mock_account(self):
```
41tair: Seems we don't need this fixture.

41tair (approved):

PR #38950, opened 2026-07-14
fix: log error and nullify credential instead of silently suppressing decryption failures

xiaoming589:
## Problem

When the RSA key pair becomes mismatched (e.g. after a container rebuild that regenerates the private key), credential decryption raises \ValueError\. The code in \provider_manager.py\ silently suppressed this error using:

- \contextlib.suppress(ValueError)\ in \_get_and_decrypt_credentials()\
- \except ValueError: pass\ in \_to_system_configuration()\
- \except ValueError: pass\ in load balancing model credentials decryption

This caused the **encrypted ciphertext** to be returned as the credential value (e.g. the API key). The ciphertext was then sent to the model provider as the actual API key, producing cryptic errors like:

\\\
Authentication Fails, Your api key: ****DOup is invalid
\\\

The garbled key suffix changes on every restart because a new key pair is generated each time, making the issue very hard to diagnose.

## Fix

Replace all three silent suppressions with explicit error handling:

1. **Log the decryption failure** with \logger.error()\, including the variable name, tenant ID, and guidance to run \lask reset-encrypt-key-pair\
2. **Set the credential variable to \None\** so downstream code gets a clear \None\ instead of encrypted garbage

This way, when the key pair is mismatched:
- The logs show a clear error message explaining the root cause
- The model provider receives \None\ as the API key, producing a clear \credential not configured\ error instead of sending garbage to the provider
- The user can fix it by running \lask reset-encrypt-key-pair\

## Root Cause Context

This was discovered on a Docker Desktop (Windows) deployment where \STORAGE_TYPE=opendal\ with \OPENDAL_FS_ROOT=storage\ (relative path). After container restarts, the RSA private key file was regenerated, but the \encrypt_public_key\ in the \	enants\ table still referenced the old key pair. This mismatch made every stored credential undecryptable.

## Changes

\pi/core/provider_manager.py\

| Location | Before | After |
|----------|--------|-------|
| \_get_and_decrypt_credentials()\ | \with contextlib.suppress(ValueError):\ | \	ry/except\ with \logger.error()\ + set to \None\ |
| \_to_system_configuration()\ | \except ValueError: pass\ | \logger.error()\ + set to \None\ |
| Load balancing credentials | \except ValueError: pass\ | \logger.error()\ + set to \None\ |

## Testing

- Verified on Dify v1.14.2 deployment
- Before fix: model invocation sent garbage API key to DeepSeek, got \401 Authentication Fails\
- After fix: model invocation gets \None\ credential, produces clear error; logs show \Failed to decrypt credential variable api_key for tenant ...; the RSA key pair may be out of sync. Run \lask reset-encrypt-key-pair\ to fix.\

## Type of Change


crazywoola:
Hi @xiaoming589! Thanks so much for taking the time to contribute to Dify — we really appreciate it.

PR #38959, opened 2026-07-14
feat: knowledge add more trace

fatelei:
> [!IMPORTANT]
>

## Summary

knowledge add more opentelemetry trace #39417

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

before

image

after

image

## Checklist

### DIFF
--- a/api/core/rag/retrieval/dataset_retrieval.py
+++ b/api/core/rag/retrieval/dataset_retrieval.py
@@ -9,7 +9,6 @@ from collections.abc import Generator, Mapping
 from typing import Any, Union, cast
 
 from flask import Flask, current_app
-from opentelemetry.trace import Status, StatusCode, get_current_span
 from sqlalchemy import and_, func, literal, or_, select, update
 from sqlalchemy.orm import Session, sessionmaker
 
@@ -803,7 +802,7 @@ class DatasetRetrieval:
 
             if query:
                 query_thread = threading.Thread(
-                    target=propagate_context(self._multiple_retrieve_thread),
+                    target=propagate_context(self._multiple_retrieve_thread_safely),
                     kwargs={
                         "flask_app": current_app._get_current_object(),  # type: ignore
                         "available_datasets": available_datasets,
@@ -829,7 +828,7 @@ class DatasetRetrieval:
             if attachment_ids:
                 for attachment_id in attachment_ids:
                     attachment_thread = threading.Thread(
-                        target=propagate_context(self._multiple_retrieve_thread),
+                        target=propagate_context(self._multiple_retrieve_thread_safely),
                         kwargs={
                             "flask_app": current_app._get_current_object(),  # type: ignore
                             "available_datasets": available_datasets,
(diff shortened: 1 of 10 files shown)


api/core/rag/retrieval/dataset_retrieval.py
```diff
@@ -1187,6 +1208,9 @@ def _run_retriever_thread(
                     attachment_ids=attachment_ids,
                 )
         except Exception as e:
+            span = get_current_span()
+            span.record_exception(e)
+            span.set_status(Status(StatusCode.ERROR, str(e)))
```
QuantumGhost: **Blocking:** the new `@trace_span()` conflicts with this catch-and-return path. This branch records the exception and sets the current span to `ERROR`, but the exception is swallowed and the decorated function returns normally. `SpanHandler.wrapper()` then unconditionally calls `set_status(OK)`, so the exported worker span ends up as `OK` with an exception event. The same ownership conflict exists in `_multiple_retrieve_thread` at lines 1916-1919.

Please keep exception transfer outside the traced operation: let the traced inner function raise so `SpanHandler` exclusively records/finalizes the failure, and let an untraced thread adapter update `cancel_event` and `thread_exceptions`. Please also add a regression assertion that the finished span status is `ERROR`.

Sources:
- Dify handler behavior: https://github.com/langgenius/dify/blob/b6bfc3e28cb49a54e07552d76f3446a4ca938c3b/api/extensions/otel/decorators/handler.py#L86-L95
- OpenTelemetry status precedence: https://opentelemetry.io/docs/specs/otel/trace/api/#set-status

**AI disclosure**: This review comment was drafted and analyzed with Codex. I have reviewed the analysis, and I am responsible for the content.

api/core/rag/retrieval/dataset_retrieval.py
```diff
@@ -99,6 +102,20 @@
 logger = logging.getLogger(__name__)
 
 
+def _propagate_otel_context[**P, R](func: Callable[P, R]) -> Callable[P, R]:
```
QuantumGhost: This duplicates the OpenTelemetry context capture/attach/detach helper already owned by `retrieval_service.py`. Keeping separate copies makes RAG business modules responsible for tracing-runtime lifecycle and requires every new thread entry point to remember the same wrapping policy.

Please centralize this in `extensions.otel` or a shared context-aware thread adapter, then reuse it from both retrieval modules.

Source: the existing implementation is at https://github.com/langgenius/dify/blob/b6bfc3e28cb49a54e07552d76f3446a4ca938c3b/api/core/rag/datasource/retrieval_service.py#L96-L107

**AI disclosure**: This review comment was drafted and analyzed with Codex. I have reviewed the analysis, and I am responsible for the content.

api/tests/unit_tests/extensions/otel/test_retrieval_tracing.py
```diff
@@ -0,0 +1,75 @@
...
+
+    with (
+        patch("extensions.otel.decorators.base.dify_config.ENABLE_OTEL", True),
+        patch.object(retrieval, "_check_knowledge_rate_limit"),
+        patch.object(retrieval, "_get_available_datasets", return_value=[]),
+    ):
+        assert retrieval.knowledge_retrieval(MagicMock(), request) == []
+
+    span_names = [span.name for span in memory_span_exporter.get_finished_spans()]
+    assert "core.rag.retrieval.dataset_retrieval.DatasetRetrieval.knowledge_retrieval" in span_names
```
QuantumGhost: The test name and docstring promise a child span, but this assertion only proves that a span with the expected name exists. No parent span is started, so the test would still pass if `knowledge_retrieval` created an unrelated root span.

Please start a workflow-node parent span around the call, locate the retrieval span, and assert both the shared `trace_id` and `child.parent.span_id == parent.context.span_id`.

**AI disclosure**: This review comment was drafted and analyzed with Codex. I have reviewed the analysis, and I am responsible for the content.

QuantumGhost (commented):
**AI disclosure**: This review was drafted and analyzed with Codex. I have reviewed the analysis, and I am responsible for the content.

I found one blocking observability correctness issue and two non-blocking design/test issues. The blocking issue causes caught retrieval-worker failures to be exported as successful spans.

QuantumGhost (approved):

PR #38962, opened 2026-07-14
fix(web): gate console atom consumers on bootstrap

### LINKED ISSUE #38840: `TypeError: Suspense query must be resolved before reading` thrown repeatedly on cold console login when server-side profile prefetch is skipped
## Self Checks

## Dify version

main (any build containing #38572 / `cee88de684`)

## Cloud or Self Hosted

Self Hosted (Docker)

## Steps to reproduce

1. Deploy web with a **relative** `API_PREFIX` (same-origin console API) and no `SERVER_CONSOLE_API_PREFIX`, so the server-side prefetch in `CommonLayoutHydrationBoundary` is skipped (`resolveServerConsoleApiUrl` returns `null`).
2. Log in with a fresh session (no warm React Query cache for `console.account.profile.get`). Incognito + throttled network makes it easy to hit.
3. Load any `(commonLayout)` console route.

## ✔️ Expected Behavior

Console loads without errors; external-service sync effects (Amplitude / Zendesk) run once the account profile query resolves.

## ❌ Actual Behavior

The console throws repeatedly until the profile query resolves:

```
TypeError: Suspense query must be resolved before reading: [["console","account","profile","get"],{"type":"query"}]
```

The stack points at the jotai-effect scheduler (`postMessage` / `unstable_scheduleCallback` frames), not a React render frame. Non-blocking: a page refresh clears it because the query is cached by then.

## Root cause

`web/app/(commonLayout)/external-service-sync.tsx` (mounted in `ConsoleRuntimeProviders`, added in #38570) mounts two `jotai-effect` atoms:

- `web/context/amplitude-identity-sync.ts` — `amplitudeIdentitySyncAtom`
- `web/context/zendesk-conversation-sync.ts` — `zendeskConversationSyncAtom`

Both call `get(userProfileAtom)` unguarded. Since #38572, `userProfileAtom` (`web/context/account-state.ts`) reads `accountProfileQueryAtom`, which is built with `atomWithResolvedSuspenseQuery` (`web/utils/query-atoms.ts`). That helper intentionally throws when the underlying suspense query is still pending:

```ts
if (result instanceof Promise) {
  throw new TypeError(
    `Suspense query must be resolved before reading: ${JSON.stringify(getOptions(get).queryKey)}`,
  )
}
```

`atomEffect` callbacks run in jotai-effect's own scheduler, outside React's render/commit cycle, so there is no Suspense boundary to absorb the pending state. The "already resolved" invariant the helper relies on is normally guaranteed by the server-side `fetchQuery(serverUserProfileQueryOptions())` in `web/app/(commonLayout)/hydration-boundary.tsx` — but that prefetch is skipped entirely when `resolveServerConsoleApiUrl` cannot resolve an absolute console API URL (relative `API_PREFIX`, no `SERVER_CONSOLE_API_PREFIX`), which is a common self-hosted/Docker configuration. The client then fetches cold, the effect fires while the query is pending, and the helper throws. The effect re-evaluates on every query state change, so the error repeats until the query resolves.

## Suggested fix

Don't let a `jotai-effect` read a resolved-suspense-only atom. Give the two sync effects a non-suspense read path (e.g. a plain `atomWithQuery`-backed atom exposing `{ data, isPending }`) and bail out early while pending / no profile id. Gating the `ExternalServiceSync` mount behind a Suspense boundary is not sufficient, because the effect scheduler runs outside the render/catch cycle and the cold-cache case has nothing resolved at mount time anyway.

Files:

- `web/context/account-state.ts`
- `web/utils/query-atoms.ts`
- `web/context/amplitude-identity-sync.ts`
- `web/context/zendesk-conversation-sync.ts`
- `web/app/(commonLayout)/external-service-sync.tsx`
- `web/app/(commonLayout)/hydration-boundary.tsx`

Jingyi-Dify:
## Summary

- Prevent the common Console shell from mounting until profile and system-feature queries are available in the shared query cache.
- Keep atom consumers unmounted after an unauthenticated profile response while the existing auth redirect flow runs.
- Add focused regression coverage for pending bootstrap queries and profile 401 responses.

## Screenshots

N/A — no visual change.

## Checklist

- [x] I've added focused regression tests and kept this as a single atomic change.
- [x] I ran focused frontend tests and the staged-file formatter/linter.

From Codex

### DIFF
--- a/web/app/(commonLayout)/__tests__/console-bootstrap-gate.spec.tsx
+++ b/web/app/(commonLayout)/__tests__/console-bootstrap-gate.spec.tsx
@@ -4,6 +4,9 @@ import { act, render, screen, waitFor } from '@testing-library/react'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { ConsoleBootstrapGate } from '../console-bootstrap-gate'
 
+const profileQueryKey = ['console', 'account', 'profile', 'get']
+const systemFeaturesQueryKey = ['console', 'system-features', 'get']
+
 type Deferred<T> = {
   promise: Promise<T>
   resolve: (value: T) => void
@@ -24,20 +27,18 @@ const createDeferred = <T,>(): Deferred<T> => {
 const mocks = vi.hoisted(() => ({
   profileQuery: undefined as Deferred<{ id: string }> | undefined,
   systemFeaturesQuery: undefined as Deferred<{ branding: { enabled: boolean } }> | undefined,
-  isLegacyBase401: vi.fn((error: unknown) => error instanceof Response && error.status === 401),
 }))
 
 vi.mock('@/features/account-profile/client', () => ({
-  isLegacyBase401: mocks.isLegacyBase401,
   userProfileQueryOptions: () => ({
-    queryKey: ['console', 'account', 'profile', 'get'],
+    queryKey: profileQueryKey,
     queryFn: () => mocks.profileQuery!.promise,
   }),
 }))
 
 vi.mock('@/features/system-features/client', () => ({
   systemFeaturesQueryOptions: () => ({
-    queryKey: ['console', 'system-features', 'get'],
+    queryKey: systemFeaturesQueryKey,
     queryFn: () => mocks.systemFeaturesQuery!.promise,
   }),
 }))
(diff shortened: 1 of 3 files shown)


Jingyi-Dify:
Related GitHub issue: #38840.

hyoban (approved):

PR #38964, opened 2026-07-14
fix(workflow): preserve falsy inputs in single-node runs

### LINKED ISSUE #38963: Workflow single-node execution drops explicit falsy inputs
## Dify version

main (`40df83de`)

## Cloud or Self Hosted

Self Hosted (Source)

## Steps to reproduce

1. Run a workflow node through the debugger or single-node execution path.
2. Map an input such as `node.enabled` to a downstream selector.
3. Provide `False`, `0`, or an empty string as the explicit value for that input.

## Expected Behavior

The explicit value is placed in the variable pool unchanged.

## Actual Behavior

The input is treated as absent. The mapper falls through to the short-key alias and, when that alias is absent, does not add the value to the variable pool.

## Root cause

`WorkflowEntry.mapping_user_inputs_to_variable_pool` uses a truthiness check before falling back from the full input key to its short-key alias. Valid falsy values therefore follow the missing-value path.

ShiroKSH:
Fixes #38963

## Summary

- Preserve explicit `False`, `0`, and empty-string inputs during workflow single-node execution.
- Retain the existing short-key fallback only when the full key has no value.
- Cover the regression with parameterized unit tests.

The input mapper used a truthiness check before looking up the short-key alias. Valid falsy values therefore followed the missing-input path and were not placed in the variable pool.

## Screenshots

Not applicable; backend workflow execution change.

## Checks

- `uv run --project api --no-default-groups --group dev pytest api/tests/unit_tests/core/workflow/test_workflow_entry.py api/tests/unit_tests/core/workflow/test_workflow_entry_helpers.py -q` (47 passed)
- `ruff check` and `ruff format --check` on changed files
- `pyrefly check api/core/workflow/workflow_entry.py` (0 errors)
- `mypy api/core/workflow/workflow_entry.py --check-untyped-defs --disable-error-code=import-untyped` (success)

## Checklist

### DIFF
--- a/api/core/workflow/workflow_entry.py
+++ b/api/core/workflow/workflow_entry.py
@@ -576,7 +576,7 @@ class WorkflowEntry:
 
             # get input value
             input_value = user_inputs.get(node_variable)
-            if not input_value:
+            if input_value is None:
                 input_value = user_inputs.get(node_variable_key)
             if input_value is None:
                 continue
--- a/api/tests/unit_tests/core/workflow/test_workflow_entry_helpers.py
+++ b/api/tests/unit_tests/core/workflow/test_workflow_entry_helpers.py
@@ -933,6 +933,20 @@ class TestMappingUserInputsBranches:
 
         variable_pool.add.assert_not_called()
 
+    @pytest.mark.parametrize("input_value", [False, 0, ""])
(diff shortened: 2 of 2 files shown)


ShiroKSH:
Hi maintainers, this PR has been open for over two weeks. When you have a moment, I would appreciate a review. Thank you.

PR #38975, opened 2026-07-15
perf(workflow-generator): parallelize node config generation

crazywoola:
## Summary

- replace the sequential full-graph builder with node-specific config builders that run concurrently, then assemble topology and canvas metadata deterministically in the runner; the legacy full-graph builder path is removed
- make builder concurrency operator-tunable via `WORKFLOW_GENERATOR_NODE_BUILDER_MAX_WORKERS` (default 6, so the planner's recommended 3–6-node plans build as a single wave), with fail-fast cancellation of queued builder calls when one fragment fails
- fold auto-mode classification into the planner: it echoes a `mode` output field (with terminal-node inference and a conversational default as lenient fallbacks), removing one serial LLM round-trip and a duplicate model resolution per `mode="auto"` request
- preserve refine-flow wiring under the parallel pipeline: the existing-graph summary surfaces branch `source_handle`s verbatim, kept children recover containment from their existing wrapper, kept nodes shed stale container markers when their container is removed, and a kept container's entry edge follows the existing draft instead of planner listing order
- give every node builder the app-mode contract (`sys.query`/`sys.files` guidance in advanced-chat, no-system-variables rule in workflow) and the declared `start_inputs` in the shared plan JSON
- raise the frontend generation timeout to 180s and make it runtime-configurable via `NEXT_PUBLIC_WORKFLOW_GENERATION_TIMEOUT_MS`, registered across both docker deployment flavors
- monkey-patch gevent in the `python -m app` path (docker `DEBUG=true`, IDE debugging), which previously served through gevent's pywsgi server unpatched — any blocking LLM invoke or `Future.result()` wait pinned the single-threaded hub and froze the whole API process during generation

## Why

This is a follow-up to #38838.

That PR made the workflow generator's API contracts explicit, but the build stage still asked one model call to generate the complete ReactFlow graph. Its long prompt and completion dominated generation time and limited practical usage to faster, smaller models.

Moving deterministic topology, wrappers, layout, and defaults into the runner keeps each model request focused on one node's semantic config and allows independent nodes to build concurrently. A multi-agent review of the branch then surfaced ten verified refine-flow and concurrency defects (branch-handle loss, container containment drift, no fail-fast), all fixed here with regression tests.

## Impact

- reduces the common `start → llm → end` critical-path system prompt from 10,913 characters to at most 2,559 characters (23.4% of the previous prompt)
- a typical 3–6-node plan builds in one concurrent wave instead of up to three sequential waves; `mode="auto"` saves an additional serial classifier round-trip
- reuses `keep` node configs during refine without an LLM call; refine no longer rewires if-else/question-classifier/human-input branches or breaks iteration/loop containment
- a failed fragment cancels queued builder calls instead of letting them run to completion before the error envelope is emitted
- preserves the existing plan/result event and error envelope contracts; `result.mode` always carries a concrete mode
- rollback knob: `WORKFLOW_GENERATOR_NODE_BUILDER_MAX_WORKERS=2` restores the previous concurrency without a deploy

## Local debugging (VS Code template)

`.vscode/launch.json.template`'s "Python: API (gevent)" config now launches the API as `python -m gevent.monkey --module app` with debugpy's `"gevent": true`, so the stdlib is monkey-patched before the app imports. Combined with the new `__main__` guard in `api/app.py`, this fixes the system-wide freeze where a debugged API served every request on one unpatched OS thread and each blocking LLM call stalled the entire process — previously most visible as severe lag during (especially repeated) cmd+k generations. The stale Next.js full-stack debug config was dropped. Existing local `launch.json` files should be regenerated from the template.

## Configuration

- `WORKFLOW_GENERATOR_NODE_BUILDER_MAX_WORKERS` (api, default 6) — max concurrent node-builder LLM calls per generation request
- `NEXT_PUBLIC_WORKFLOW_GENERATION_TIMEOUT_MS` / `WORKFLOW_GENERATION_TIMEOUT_MS` (web, default 180000) — frontend hard timeout for a generation request, wired through env.ts, the web docker entrypoint, both compose flavors, and all env examples

## Validation

- `282 passed` — workflow generator, generator service, and llm-generator unit tests
- `55 passed` — workflow-generator frontend specs (Vitest)
- diff coverage of the generator modules measured locally at 100% (codecov patch methodology)
- Ruff and mypy passed for the changed modules; `tsc --noEmit` passed for web


docker/.env.example
```diff
@@ -137,6 +137,7 @@ API_SENTRY_PROFILES_SAMPLE_RATE=1.0
 WEB_SENTRY_DSN=
 AMPLITUDE_API_KEY=
 TEXT_GENERATION_TIMEOUT_MS=60000
+WORKFLOW_GENERATION_TIMEOUT_MS=180000
```
laipz8200, line 140: Only include variables required for services to start.
Do not add optional variables to this file.

iamjoel (approved):

PR #38981, opened 2026-07-15
fix(web): guard tracing array writes in handleSend against unmatched findIndex

### LINKED ISSUE #38980: Workflow node/iteration/loop "finished" trace updates silently dropped in chat streaming (missing `-1` index guard in `handleSend`)
### Self Checks

### Dify version

1.16.0-rc1 (also present on main @ 15acaf1fa6)

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

This is a logic bug in `web/app/components/base/chat/chat/hooks.ts`, in the streaming (`handleSend`) SSE callbacks used to build the workflow trace panel shown during/after a chat run.

`onNodeFinished`, `onIterationFinish`, and `onLoopFinish` all locate the matching "started" entry in `responseItem.workflowProcess.tracing` via `Array.prototype.findIndex(...)`, then write the finished data back with `tracing[index] = {...}`:

- `onNodeFinished` — `web/app/components/base/chat/chat/hooks.ts:1354-1368`
- `onIterationFinish` — `web/app/components/base/chat/chat/hooks.ts:1303-1316`
- `onLoopFinish` — `web/app/components/base/chat/chat/hooks.ts:1401-1414`

None of these three check that `findIndex` actually found a match (i.e. that the index is `> -1`) before writing to `tracing[index]`.

By contrast, the structurally identical callbacks used for `handleResume` (same file, ~150 lines earlier: `onNodeFinished` at line 664-681, `onIterationFinish` at line 645-651, `onLoopFinish` at line 705-724) all correctly guard the write with `if (currentIndex > -1)` / `if (iterationIndex > -1)` / `if (loopIndex > -1)`.

`Array.prototype.findIndex` returns `-1` when no element matches. In JavaScript, `arr[-1] = value` does **not** grow or replace an array element — it silently defines a non-index string property (`"-1"`) on the array object. The assignment does not throw, so this fails silently: the finished node/iteration/loop's data is never written into the tracing array shown to the user.

This is reachable whenever a `node_finished` / `iteration_finished` / `loop_finished` SSE event arrives during `handleSend` for a node whose corresponding "started" tracing entry was never pushed — e.g. an out-of-order or dropped SSE event, or a node id that doesn't match anything currently in `tracing` (this can also legitimately happen for nodes inside iterations/loops depending on ordering, since `onNodeStarted` explicitly skips pushing entries when `iteration_id`/`loop_id` is set — see lines 1338-1340).

There is already a test that exercises exactly this "no matching entry" case (`web/app/components/base/chat/chat/__tests__/hooks.spec.tsx:2899`, `sendCallbacks.onNodeFinished({ data: { id: 'missing-idx' } })`), but it only asserts that an unrelated field (`chatList[1].message_files`) is still defined — i.e. that nothing crashes — not that the tracing array is left unmodified. So the regression currently passes CI undetected.

### ✔️ Expected Behavior

When a `node_finished` / `iteration_finished` / `loop_finished` event has no matching "started" entry in the tracing array, the write should be skipped (guarded by `index > -1`), matching the behavior already implemented in the equivalent `handleResume` callbacks. The tracing array should never be corrupted with data assigned at index `-1`.

### ❌ Actual Behavior

In `handleSend`'s SSE callbacks, `onNodeFinished`, `onIterationFinish`, and `onLoopFinish` unconditionally write to `tracing[index]` even when `index === -1`, so the update is silently lost instead of being safely ignored (or the array being corrupted with a stray `-1` property, depending on how it's later serialized/consumed). This can make the workflow trace panel in the chat UI appear incomplete or stuck, with no error or console warning to indicate why.

amogh-nagri-11:
## Summary

Fixes #38980.

Three SSE callbacks in `handleSend` (`web/app/components/base/chat/chat/hooks.ts`) locate a node's "started" entry in `responseItem.workflowProcess.tracing` with `findIndex(...)`, then write the finished payload straight back to `tracing[index]` — without ever checking that the entry was found:

| Callback | Site |
| --- | --- |
| `onIterationFinish` | `hooks.ts:1303-1316` |
| `onNodeFinished` | `hooks.ts:1354-1368` |
| `onLoopFinish` | `hooks.ts:1401-1414` |

`Array.prototype.findIndex` returns `-1` on no match, and in JavaScript `arr[-1] = value` neither throws nor updates the array — it silently defines a stray `"-1"` string property on the array object. So when a `node_finished` / `iteration_finished` / `loop_finished` event arrives without a matching "started" entry, the update is dropped and the array is quietly polluted, with no error or console warning.

This is reachable on out-of-order or dropped SSE events, and for nodes inside iterations/loops, where `onNodeStarted` deliberately skips pushing a tracing entry when `iteration_id` / `loop_id` is set (`hooks.ts:1338-1340`) — so the finished event has nothing to match. The visible result is a workflow trace panel that looks incomplete or stuck.

Two things kept this hidden:

- **The guard was already written — just not here.** The structurally identical callbacks in `handleResume`, ~150 lines earlier in the same file, all correctly guard with `if (index > -1)`. This PR brings `handleSend` to parity rather than introducing a new pattern.
- **A non-null assertion made the code look checked.** Two of the sites ended in `findIndex(...)!`. Since `findIndex` returns `number` and never a nullish value, that `!` was inert — it silenced nothing and asserted nothing, but it reads like the result had been validated.

## Changes

- Added the missing `> -1` guards to all three callbacks, matching `handleResume`.
- Dropped the inert `)!` non-null assertions on the two `findIndex` calls that carried them.
- Strengthened the existing test for this exact path. `hooks.spec.tsx:2899` already fired `onNodeFinished({ data: { id: 'missing-idx' } })`, but only asserted that an unrelated field (`chatList[1].message_files`) was still defined — i.e. that nothing crashed. It now asserts the tracing array is actually left intact: length unchanged, surviving entry untouched, and no stray `-1` property. That is why the bug shipped past a green CI.

No behavior change when the index resolves normally, which is every non-degenerate stream.

## Screenshots

Not applicable — no visual change; this restores trace data that was being dropped.

## Checklist

## Verification

- `pnpm vitest run app/components/base/chat/chat/__tests__/hooks.spec.tsx` — 83/83 passing.
- The strengthened assertion **fails against the pre-fix code** and passes with the fix, so it is a real regression guard rather than a restatement of current behavior.
- Lint findings in the touched file are pre-existing and identical on `main`.

### DIFF
--- a/web/app/components/base/chat/chat/__tests__/hooks.spec.tsx
+++ b/web/app/components/base/chat/chat/__tests__/hooks.spec.tsx
@@ -2916,6 +2916,14 @@ describe('useChat', () => {
     })
 
     expect(result.current.chatList[1]!.message_files).toBeDefined()
+
+    // Finished events with no matching started entry must not corrupt the tracing array.
+    // `findIndex` returns -1, and an unguarded `tracing[-1] = value` silently defines a
+    // stray non-index "-1" property on the array object instead of being a true no-op.
+    const tracing = result.current.chatList[1]!.workflowProcess!.tracing!
+    expect(tracing).toHaveLength(1)
+    expect(tracing[0]!.node_id).toBe('n-new')
+    expect(Object.prototype.hasOwnProperty.call(tracing, '-1')).toBe(false)
   })
 
   it('should cover handleSwitchSibling target message not found early returns', () => {
--- a/web/app/components/base/chat/chat/hooks.ts
+++ b/web/app/components/base/chat/chat/hooks.ts
@@ -1308,11 +1308,13 @@ export const useChat = (
               (item.execution_metadata?.parallel_id ===
                 iterationFinishedData.execution_metadata?.parallel_id ||
                 item.parallel_id === iterationFinishedData.execution_metadata?.parallel_id),
-          )!
-          tracing[iterationIndex] = {
-            ...tracing[iterationIndex],
-            ...iterationFinishedData,
-            status: WorkflowRunningStatus.Succeeded,
+          )
+          if (iterationIndex > -1) {
+            tracing[iterationIndex] = {
+              ...tracing[iterationIndex],
+              ...iterationFinishedData,
+              status: WorkflowRunningStatus.Succeeded,
+            }
           }
 
           updateCurrentQAOnTree({
@@ -1365,7 +1367,8 @@ export const useChat = (
                 nodeFinishedData.execution_metadata?.parallel_id
             )
           })
-          responseItem.workflowProcess!.tracing[currentIndex] = nodeFinishedData as any
+          if (currentIndex > -1)
+            responseItem.workflowProcess!.tracing[currentIndex] = nodeFinishedData as any
 
           updateCurrentQAOnTree({
             placeholderQuestionId,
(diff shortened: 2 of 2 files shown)


amogh-nagri-11:
Just checking in here. @asukaminato0721, would you mind taking a look here whenever possible? Happy to make any change required!

Issue #37088, opened 2026-06-05
Collaborate with Teammates have a bug, sometimes  it will lost changes or Overwriting others' changes


jikangding:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.14.2

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Collaborate with Teammates have a bug, sometimes  it will lost changes or Overwriting others' changes

### ✔️ Expected Behavior

it can save all changes

hjlarry:
@jikangding Thanks for reporting this. We want to narrow down the exact collaboration scenario, because “lost changes” can happen at different points in the flow.

Could you help us with a few details?

1. What were the exact steps before the change was lost?
   For example: User A edited node X, User B edited node Y, then one user saved/refreshed/reconnected, etc.

2. Were both users editing the same node/field, or different nodes/fields?

3. When did the change disappear?
   - immediately while editing
   - after autosave/manual save
   - after refresh/reopen
   - after reconnecting from a network interruption
   - after another teammate joined the workflow

4. After refreshing both clients, which version remained in the workflow draft?

These details will help us distinguish between a live collaboration merge issue and a draft-save conflict issue.

We are already reviewing the collaboration merge and save-conflict paths, but a reproducible sequence would help us verify the exact failure mode and fix it safely.

PR #30781, opened 2026-01-09
feat: collaboration

hjlarry:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

image

Collaboration is an experimental Labs feature for real-time workflow co-editing. 

It allows workspace members to edit the same workflow together, with synced graph updates, online presence, and shared visibility into who is working where. The feature also supports workflow comments, threaded replies, and mentions to help teams discuss changes directly in context. 

As a Labs feature, Collaboration is still evolving and may continue to improve over time.

### How to enable it?

set the environment variable:

```
ENABLE_COLLABORATION_MODE=true
SERVER_WORKER_CLASS=geventwebsocket.gunicorn.workers.GeventWebSocketWorker
```

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.
- [x] I ran `make lint` and `make type-check` (backend) and `cd web && npx lint-staged` (frontend) to appease the lint gods

PR #38997, opened 2026-07-15
fix: prevent hidden-tab collaboration leader from saving stale drafts

### LINKED ISSUE #38994: Workflow collaboration: a hidden-tab leader keeps saving its stale canvas, losing/overwriting other users' edits
### Dify version

1.14.2 (root cause also present on current main)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Collaboration mode enabled. Users A and B editing the same workflow.

```
1. User B opens the workflow first → B becomes the collaboration "leader".
2. User B switches to a different browser tab (the workflow tab stays
   connected but is now hidden/inactive).
3. User A adds/moves nodes and clicks Publish.
4. Reopen the workflow and check the draft / published version.
```

### ✔️ Expected Behavior

User A's edits are saved to the draft and published.

### ❌ Actual Behavior

User A's edits are silently lost. After A publishes and the workflow is reopened, both the draft and the published version show B's old state from the moment B switched tabs — none of A's node changes are there. A saw no error and no warning; the canvas looked correct on A's screen the whole time, so it appears as if the publish simply discarded the work. If A keeps editing, the stale state can also get pushed back onto other collaborators' canvases, overwriting their changes too.

---

Likely root cause: only the "leader" session persists the draft, saving from its own React Flow canvas. Leader selection already filters on a per-session `graph_active` flag (introduced in #30781, persistence half-wired in #38242), but the frontend never reports tab visibility, so the flag is always effectively `true`. A hidden tab keeps its leadership even though its canvas is frozen — CRDT→canvas application runs inside `requestAnimationFrame`, which browsers pause in background tabs — so it keeps saving stale data over everyone else's edits.

Possibly related to #37088.

Kevin9703:
> [!IMPORTANT]
>

## Summary

Fixes #38994 (root cause behind #37088)

In collaboration mode the workflow draft is persisted only by the "leader" session, which saves from its own React Flow canvas. Leader selection already filters candidates on a per-session `graph_active` flag, but nothing ever sets it: the frontend never reports tab visibility, and the client's `graphViewActive` field has been write-never since it was added in #30781. As a result a background tab — whose canvas is frozen because CRDT→canvas application runs inside `requestAnimationFrame` (paused by browsers in hidden tabs) — stays leader and keeps persisting its stale snapshot, silently overwriting other collaborators' edits and publishing outdated drafts.

This PR wires the visibility signal end to end:

- **Frontend reports tab visibility.** A new `graph_view_state` collaboration event is emitted on `visibilitychange` (listener attached in `connect()`, removed in `forceDisconnect()`). Since the server recreates a session as visible on every (re)join, a hidden tab re-reports itself right after the post-join `status` event.
- **Backend prefers visible leaders.** Session info now persists `graph_active` end to end (`get_session_info` / `update_session_graph_active`); `_select_graph_leader` becomes two-tier (visible sessions first, any active session as a fallback so a room is never left leaderless); a leader that reports hidden is demoted when a visible session exists; and `sync_request` re-routes away from a connected-but-hidden leader, preferring the requester — whose canvas always contains its own edits, since local edits apply synchronously.
- **Canvas restore on tab focus.** When a tab becomes visible again, the canvas is restored from the CRDT document (which stays live in background tabs) instead of re-importing the possibly-stale DB draft, which could otherwise broadcast a rollback to every client.

`graph_active` defaults to `true` everywhere, so the change is backward compatible — an old frontend against a new backend behaves exactly as before.

Covered by new backend unit tests (repository + service leader-election / `graph_view_state` paths) and frontend unit tests (visibility reporting, event payload, re-report on rejoin).

## Screenshots

| Before | After |
|--------|-------|
| | |

## Checklist


hjlarry:
@Kevin9703 Thank you for identifying the root cause of the hidden-tab leader issue and putting together this fix. I reproduced the original issue locally with two browsers and verified that this PR resolves it.

I added a follow-up commit to harden several related synchronization boundaries discovered during review:

- Added ordered visibility updates so a delayed `graph_view_state` event cannot overwrite a newer state or incorrectly change the selected saver.
- Added end-to-end acknowledgement for `sync_request`. The selected saver now imports the requester’s CRDT snapshot, persists the draft, and returns the saved `hash` and `updatedAt`. Publishing waits for this acknowledgement and aborts if synchronization fails.
- Rebuild CRDT state as untrusted after reconnect and recover it from an authoritative snapshot or server draft. Late snapshots and stale HTTP responses are ignored, and canvas editing is temporarily blocked with a visible syncing indicator until recovery completes.
- Made collaborative saves update only `graph` for existing drafts. Features, environment variables, and conversation variables already use dedicated APIs, so graph synchronization should not overwrite those fields using stale cached values. Initial draft creation and non-collaborative saves keep their existing full-save behavior.

These changes cover additional races around visibility event ordering, reconnect recovery, publishing before the draft reaches the database, and independently persisted workflow fields.

Thanks again for finding and fixing the original issue.

hjlarry (dismissed):
LGTM

hjlarry (dismissed):

hjlarry (dismissed):

hjlarry (dismissed):

hjlarry (approved):

PR #39001, opened 2026-07-15
refactor: use catalog for workflow archive maintenance

zhaohao1004:
## Summary

- Discover V2 delete/restore candidates from `workflow_run_archive_bundles` with month-scoped keyset pagination; no R2 LIST.
- Add deterministic manifest lookup and strict catalog/manifest/archive-object integrity validation.
- Treat FULL, content-consistent SUBSET, and ABSENT live data as successful delete outcomes; delete only verified remaining live rows or directly converge the bundle to `_DELETED`.
- Fail closed only when a candidate has archive-integrity failure, live/archive divergence, marker/state conflict, or an operational/concurrency error.
- Add optional exact-shard delete discovery plus a mixed-layout preflight, so future parallel delete jobs cannot silently skip historical `00-of-01` bundles.
- Add `(year, month, shard, id)` for sharded discovery while retaining `(year, month, id)` for serial discovery.
- Keep archive/backfill history unchanged and introduce no persisted checkpoint table.

## Delete candidate outcomes

| Outcome | Definition | Action |
| --- | --- | --- |
| FULL | Every live row exists unchanged in the validated archive | Delete all live rows, write `_DELETED`, and continue |
| SUBSET | For every table, live IDs are a subset of archive IDs and same-ID content matches | Delete only the verified remaining live rows, write `_DELETED`, and continue |
| ABSENT | No live rows remain for any archived table | Skip DB deletion, write `_DELETED`, and continue |
| BLOCKED | Archive integrity failure, extra live row, same-ID content mismatch, marker/state conflict, or operational/concurrency failure | Do not advance past the failed candidate; fail the current worker |

## Pagination and failure scope

- V2 delete accepts `--target-month YYYY-MM`, optional `--after-catalog-id`, and optional paired `--run-shard-index` / `--run-shard-total`.
- The default remains one bounded page. `--all-pages` advances page-by-page until an empty page and retains only the current page in memory.
- A page is only a bounded keyset-query batch, not a checkpoint or failure-isolation unit.
- FULL, SUBSET, and ABSENT candidates return success and processing continues with the next candidate in the current page.
- On a BLOCKED candidate, previously successful bundles remain committed, the failed candidate does not advance the cursor, and the current worker exits non-zero. The remainder of the current page and later pages for that shard are not processed.
- When invoked by independent Indexed Job workers, one failed shard does not cause the CLI to stop other shard workers; after all indexes finish, the overall Kubernetes Job is still failed if any index failed.
- Mixed-layout shard preflight is a month-level guard: a preflight failure happens before page processing and is not a single-shard data-isolation case.
- Failures report the failed catalog ID plus the safe resume cursor; dry-run uses a preview-only cursor.
- Historical or mixed-shard months remain available through the serial path without shard options.

## Validation

- `uv run --project api pytest api/tests/unit_tests/commands/test_archive_workflow_runs.py api/tests/unit_tests/services/retention/workflow_run/test_bundle_archive_maintenance.py -q` — 47 passed
- Focused Ruff format/lint — passed
- Focused Pyrefly and Mypy — passed
- Alembic heads — single head `3c9f8e2a1d7b`
- Offline PostgreSQL upgrade/downgrade SQL — creates both indexes concurrently and drops them in reverse order
- `git diff --check` — passed


api/migrations/versions/2026_07_20_1200-9b2d7e4f6a81_add_workflow_run_archive_bundle_shard_cursor_index.py
```diff
@@ -0,0 +1,43 @@
+"""add workflow-run archive bundle shard cursor index
+
+Revision ID: 9b2d7e4f6a81
```
hjlarry: Would it be better to combine the two migration files into one?

hjlarry (approved):
LGTM

snakevash (approved):

PR #39008, opened 2026-07-15
fix(agent): preserve Unicode separators in SSE streams

zyssyz123:
## Summary

- split Agent Backend SSE text using only the CR/LF line endings defined by the SSE specification
- preserve U+0085, U+2028, and U+2029 inside JSON payloads instead of treating them as frame boundaries
- escape those separators in Agent Backend output for compatibility with clients that use universal line splitting
- add sync, async, and server serialization regression coverage

## Root cause

`httpx.iter_lines()` / `aiter_lines()` uses universal Unicode line splitting. A tool result containing a raw U+2028 was therefore split in the middle of the SSE `data:` JSON. The API client then raised `DifyAgentStreamError: malformed SSE data frame` even though the Agent Backend emitted a valid SSE frame.

## Validation

- `uv run --project dify-agent pytest -q dify-agent/tests/local/dify_agent/client dify-agent/tests/local/dify_agent/server` (96 passed on `main`)
- Ruff check and format check passed for all changed files
- deployed the backend-only change to `deploy/agent` and recreated Agent Backend, API, and API worker services
- runtime env, command, mount, network, and restart configuration matched the pre-deploy state
- replayed the exact DIFY-2799 request on dev: HTTP 200, 376 valid JSON events, 14 raw U+2028 characters, final `message_end`, no error event, and no `malformed SSE data frame`
- history API persisted the result as `status=normal` with 16 agent thoughts, the final answer, and token usage

Linear: DIFY-2799

### DIFF
--- a/dify-agent/src/dify_agent/client/_client.py
+++ b/dify-agent/src/dify_agent/client/_client.py
@@ -94,6 +94,48 @@ class _ReconnectableStreamError(Exception):
         super().__init__(str(error))
 
 
+class _SSELineDecoder:
+    """Split SSE text using only the line endings defined by the SSE specification."""
+
+    _buffer: str
+
+    def __init__(self) -> None:
+        self._buffer = ""
+
+    def decode(self, text: str) -> list[str]:
+        data = self._buffer + text
+        self._buffer = ""
+        lines: list[str] = []
+        start = 0
+        index = 0
+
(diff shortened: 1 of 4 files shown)


lyzno1 (commented):
!

lyzno1 (approved):

PR #39026, opened 2026-07-15
feat: redact workflow secrets from run logs, live stream, and DSL export

CourTeous33:
Value-matching secret backstop that redacts environment-variable and conversation-variable secret values from workflow run-log surfaces. Defense-in-depth alongside the graphon serializer masking (companion PR).

## Changes
- **`secret_scrub` module** (`core/workflow/secret_scrub.py`): registry collector (`collect_workflow_secret_values`) + shape-preserving redactor (`redact_secret_values`). Strict no-op when the registry is empty; substring match for secrets ≥8 chars, exact match for shorter; documents its own limitations in the module docstring.
- **Persistence** (`layers/persistence.py`): redacts node inputs / process_data / outputs / metadata, workflow outputs, and all error / error_message fields. Registry wired through all three app-runners.
- **Live SSE stream** (`workflow_response_converter.py`): redacts every stream builder — node finish/retry, workflow finish, iteration/loop completed — across inputs/outputs/process_data/metadata/error, plus the secondary error channel. Registry wired through both task pipelines.
- **DSL export** (`app_dsl_service.py`, `controllers/console/app/app.py`): manager-only (owner/admin) guard on the secret-bearing `/export` endpoint.

## Notes
- Standalone: does not require the graphon version bump. The graphon serializer masking layers in additionally once released.
- Deferred as documented follow-ups: conversation-variable encryption at rest, and a tool/model/datasource credential registry.

### DIFF
--- a/api/core/workflow/secret_scrub.py
+++ b/api/core/workflow/secret_scrub.py
@@ -1,0 +1,74 @@
+from collections.abc import Iterable, Mapping
+
+from graphon.variables import SecretVariable, VariableBase
+
+SECRET_PLACEHOLDER = "[REDACTED_SECRET]"
+# Values shorter than this are redacted only on an exact whole-string match;
+# substring-redacting a short value would corrupt unrelated text containing it.
+_MIN_SUBSTRING_LENGTH = 8
+
+
+def collect_workflow_secret_values(
+    environment_variables: Iterable[VariableBase],
+    conversation_variables: Iterable[VariableBase],
+) -> tuple[str, ...]:
+    """Collect the plaintext of every secret env/conversation variable.
+
+    Returns:
+        A de-duplicated tuple of non-empty secret string values, in first-seen order.
+    """
+    values: list[str] = []
+    for variable in (*environment_variables, *conversation_variables):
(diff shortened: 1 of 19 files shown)


api/core/workflow/secret_scrub.py
laipz8200, line 1: I believe that all handling of the display content for secret variables should be implemented through methods within the secret variables themselves.

We should consider passing the secret variables more frequently instead of strings, while avoiding the introduction of a separate file to handle these types of issues.

laipz8200 (changes requested):

CourTeous33:
These chanegs are not necessary, since I am addressing light, avoidable secret leaks by heavy code changes. pay attenttion to these following:

- Don't read or concatenate secrets inside a Code node — the Code node's inputs/outputs are plaintext and won't be masked.
- Don't process and output secrets inside a Template Transform or Variable Assigner node — they flatten the secret into plaintext and write it to the logs.
- Don't concatenate a secret into a string and then output it (e.g., "Bearer " + secret as a node's output).
- Don't let an external API echo the secret back in its response body — the response body goes into the outputs log.

PR #38739, opened 2026-07-10
test: use sqlite3 session in test_extract_thread_messages

asukaminato0721:
## Summary

- replace fabricated conversation message query results with persisted SQLite Message rows
- exercise descending creation order, parent threading, empty-answer exclusion, and conversation isolation
- retain the pure in-memory thread extraction cases

## Validation

- targeted pytest suite passed (10 tests)
- Ruff formatting and lint checks passed
- structural session-double scan passed

Part of #32454.

PR #36727, opened 2026-05-27
feat: add answer regeneration branches

Jackmoyu001:
## Summary

  - Add Service API support for regenerating the latest active answer and switching answer variants in the latest answer group.
  - Track the active conversation branch with `conversation.active_message_id` so future chat turns follow the selected answer path.
  - Expose message branch metadata in `/messages`: `is_active`, `is_in_active_thread`, `can_switch`, and `can_regenerate`.
  - Add migration, unit coverage, OpenAPI markdown docs, console API docs, and Docker restart/rebuild notes.

  ## Behavior

  - Only the latest answer group can be switched or regenerated.
  - Once the user sends the next chat turn, previous answer groups are locked.
  - Future chat context follows the active message chain from `conversation.active_message_id`.

  ## Test Plan

  - [x] `git diff --check`
  - [x] `python -m py_compile ...` for changed backend Python files
  - [ ] `uv run --project api pytest ...`

  `pytest` did not execute locally because dependency installation failed while building `chroma-hnswlib==0.7.6`; the local Windows environment is missing Microsoft Visual C++ 14.0+ Build Tools. This is an
  environment/toolchain failure before test collection, not a test assertion failure.

PR #39032, opened 2026-07-15
fix(prompt): scope extract_thread_messages' no-parent check to the tracked message

### LINKED ISSUE #39031: Bug: extract_thread_messages can pull in an unrelated regeneration-root message and truncate the real thread early
### Self Checks

### Dify version

main (latest)

### Cloud or Self Hosted

Self-hosted / source-level bug in API.

### Steps to reproduce

`api/core/prompt/utils/extract_thread_messages.py`:

```python
def extract_thread_messages(messages: Sequence[Message]):
    thread_messages: list[Message] = []
    next_message = None

    for message in messages:
        if not message.parent_message_id:
            # If the message is regenerated and does not have a parent message, it is the start of a new thread
            thread_messages.append(message)
            break

        if not next_message:
            thread_messages.append(message)
            next_message = message.parent_message_id
        else:
            if next_message in {message.id, UUID_NIL}:
                thread_messages.append(message)
                next_message = message.parent_message_id

    return thread_messages
```

The `if not message.parent_message_id:` check runs unconditionally for **every** message in the loop, not just the one currently being tracked via `next_message`. Per the comment above it, a message with no `parent_message_id` (as opposed to the legacy `UUID_NIL` sentinel) marks "the start of a new thread" — i.e. a regeneration point. A single conversation can contain **multiple** such regeneration-root messages (one per regenerate action over the conversation's life), only one of which is actually the ancestor of the thread currently being walked.

If an *unrelated* regeneration-root message (not equal to the current `next_message` target) happens to sit between the message we're walking from and its real parent in the `created_at DESC`-ordered list, this check fires on it first — the walk incorrectly grabs that unrelated message and `break`s, silently dropping every real ancestor beyond it (and including a message that never belonged to this thread).

Minimal reproduction (pure function, no DB needed — mirrors the real `created_at DESC` message order from `get_thread_messages_length`'s query):

```python
from dataclasses import dataclass

UUID_NIL = "00000000-0000-0000-0000-000000000000"

@dataclass
class Message:
    id: str
    parent_message_id: str | None
    answer: str = "some answer"

def extract_thread_messages(messages):
    thread_messages = []
    next_message = None
    for message in messages:
        if not message.parent_message_id:
            thread_messages.append(message)
            break
        if not next_message:
            thread_messages.append(message)
            next_message = message.parent_message_id
        else:
            if next_message in {message.id, UUID_NIL}:
                thread_messages.append(message)
                next_message = message.parent_message_id
    return thread_messages

# Real thread: C -> B -> A (A is the true root).
# X is an unrelated, earlier regeneration-root message from the same conversation
# (parent_message_id=None), created chronologically between B and C but NOT an
# ancestor of C.
msgC = Message(id="C", parent_message_id="B")
msgX = Message(id="X", parent_message_id=None)
msgB = Message(id="B", parent_message_id="A")
msgA = Message(id="A", parent_message_id=None)

messages = [msgC, msgX, msgB, msgA]  # created_at DESC order
print([m.id for m in extract_thread_messages(messages)])
```

Output: `['C', 'X']`

### ✔️ Expected Behavior

`['C', 'B', 'A']` — `X` is unrelated to the thread being walked from `C` and should be skipped, not treated as the thread's root.

### ❌ Actual Behavior

`extract_thread_messages` returns `['C', 'X']`: it wrongly appends the unrelated regeneration-root message `X` and stops the walk there, losing the real ancestors `B` and `A`.

### Impact

`extract_thread_messages` is the shared thread-reconstruction primitive used by:
- `core/memory/token_buffer_memory.py` (conversation history sent to the LLM)
- `core/agent/base_agent_runner.py` (agent conversation history)
- `core/prompt/utils/get_thread_messages_length.py` → `core/app/apps/advanced_chat/app_generator.py`'s `dialogue_count`

A corrupted/truncated thread here can silently shrink the conversation history an LLM call actually sees, and undercounts `dialogue_count`, in any conversation that has had more than one regeneration.

### Proposed fix direction

Only treat "no `parent_message_id`" as the thread's true root when the message being examined is the one actually being tracked (i.e. scope the check to the branch that already matched via `next_message`), rather than checking it unconditionally on every message in the loop. Happy to submit a PR with a regression test (mirroring the existing `test_extract_thread_messages_mixed_with_legacy_messages`-style tests in `api/tests/unit_tests/core/prompt/test_extract_thread_messages.py`) once a maintainer confirms the intended semantics — in particular, whether the `None`-vs-`UUID_NIL` distinction for "start of new thread" is meant to always take precedence, or should only apply when it's actually the message we're chasing.

---
Disclosure: this bug was found and reproduced with AI assistance (Claude Code); I verified the standalone repro above executes exactly as shown before filing.

chuenchen309:
Fixes #39031

## Summary

`extract_thread_messages`' "start of a new thread" check (`if not message.parent_message_id`) ran unconditionally on every message in the loop, not just the one currently being tracked via `next_message`. A conversation can contain several `parent_message_id=None` messages — one per regeneration over its lifetime. If an unrelated, earlier regeneration root sits between the message being walked and its real parent in the `created_at DESC` ordering, the walk grabs that root and stops, losing the real ancestor chain.

This affects `core/memory/token_buffer_memory.py` (LLM conversation memory), `core/agent/base_agent_runner.py` (agent history) and `core/prompt/utils/get_thread_messages_length.py` (`advanced_chat` dialogue count).

The fix scopes the "no parent" check to the branch that already matched via `next_message`, applied on both sides of the loop. It is deliberately narrow — not the broader regeneration-branch-selection feature from the earlier (closed, unmerged) #36727, which added a separate `active_message_id` parameter and did not scope this check.

## Tests

- New regression test `test_extract_thread_messages_ignores_unrelated_regeneration_root`: fails on unfixed `main` (the walk stops at the unrelated regeneration root), passes with the fix.
- `tests/unit_tests/core/prompt/`, `core/memory/` and `core/agent/`: 356 passed.
- `ruff format --check` and `ruff check` clean on both changed files.

Rebased onto current `main`; the only conflict was the test file reorganisation in #38739, and the new test is placed alongside the existing `MockMessage`-based cases.

## Checklist

- [x] I ran `ruff format` / `ruff check` and the backend unit tests listed above.

AI-assisted with agentic coding tools; all changes were reviewed and tested by me.

From Claude Code

### DIFF
--- a/api/core/prompt/utils/extract_thread_messages.py
+++ b/api/core/prompt/utils/extract_thread_messages.py
@@ -9,17 +9,18 @@ def extract_thread_messages(messages: Sequence[Message]):
     next_message = None
 
     for message in messages:
-        if not message.parent_message_id:
-            # If the message is regenerated and does not have a parent message, it is the start of a new thread
-            thread_messages.append(message)
-            break
-
         if not next_message:
             thread_messages.append(message)
+            if not message.parent_message_id:
+                # If the message is regenerated and does not have a parent message, it is the start of a new thread
+                break
             next_message = message.parent_message_id
         else:
             if next_message in {message.id, UUID_NIL}:
                 thread_messages.append(message)
+                if not message.parent_message_id:
+                    # Same "start of a new thread" case, but for the message currently being tracked
+                    break
                 next_message = message.parent_message_id
 
     return thread_messages
--- a/api/tests/unit_tests/core/prompt/test_extract_thread_messages.py
+++ b/api/tests/unit_tests/core/prompt/test_extract_thread_messages.py
@@ -137,6 +137,23 @@ def test_extract_thread_messages_breaks_when_parent_is_none():
     assert result[0].id == id2
 
 
+def test_extract_thread_messages_ignores_unrelated_regeneration_root():
+    # An earlier regeneration can leave a parent_message_id=None message in the
+    # same conversation that isn't an ancestor of the thread being walked. The
+    # unconditional "no parent" check must not grab it and stop the walk early.
+    id_a, id_b, id_c, id_x = str(uuid4()), str(uuid4()), str(uuid4()), str(uuid4())
+    messages = [
+        MockMessage(id_c, id_b),
+        MockMessage(id_x, None),  # unrelated regeneration root, not an ancestor of C
+        MockMessage(id_b, id_a),
+        MockMessage(id_a, None),  # true root of C's thread
+    ]
+
+    result = extract_thread_messages(messages)
+
+    assert [msg.id for msg in result] == [id_c, id_b, id_a]
+
+
 @pytest.mark.parametrize("sqlite_session", [(Message,)], indirect=True)
 def test_get_thread_messages_length_excludes_newly_created_empty_answer(sqlite_session: Session):
     id1, id2 = str(uuid4()), str(uuid4())


chuenchen309:
Checking back on this one — it's been open since Jul 15 with no review, and I wanted to confirm it still applies rather than just bump it.

`api/core/prompt/utils/extract_thread_messages.py` has had no commits since this PR was opened, and the function on `main` is byte-identical to the version this fixes. Running it as it stands there:

```python
# messages in created_at DESC order, as the callers pass them.
# X is the root of an earlier regeneration in the same conversation,
# and is not an ancestor of C.
messages = [Message("C", "B"), Message("X", None), Message("B", "A"), Message("A", None)]

extract_thread_messages(messages)
# -> ['C', 'X']          walking C's thread
# -> ['C', 'B', 'A']     same conversation without the earlier regeneration
```

So the walk both stops early *and* pulls an unrelated message into the thread. Whatever consumes that thread — `token_buffer_memory`, `base_agent_runner`, `get_thread_messages_length` — sees `X` as part of the conversation and never sees `B` or `A`. `#39031` is still open, and Dosu confirmed the analysis there.

One thing worth knowing if you review these together: **#39413** (`keep in-progress messages in get_thread_messages_length`, open since Jul 22) touches the same area but not the same function — it changes a *caller*, filtering what `get_thread_messages_length` counts, while this one changes the walk inside `extract_thread_messages`. The two are independent fixes; they only overlap textually in `test_extract_thread_messages.py`, where each adds its own test at the end. Whichever lands second just needs the test file merged.

Happy to rebase, split it further, or close it if you'd rather solve this a different way.

*AI-assisted; the output above is from running the current `main` source, not inferred.*

PR #38841, opened 2026-07-13
feat(agent): shellctl rewritten in go 

wylswz:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

closes #38671 

The original `shellctl` is re-written as `dify-agent-runtime` in go for faster start up time and smaller memory footprint. 

Including
- a `shellctl` server
- utility cli tools
- a `dify-agent` cli which communicates to dify agent backend

A full example with file upload and skill use was tested locally.

image

Co-authored-by: Claude Opus 4.6 (Anthropic)

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [ ] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [ ] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods

PR #39033, opened 2026-07-15
fix(agent-runtime): validate shellctl job IDs

### LINKED ISSUE #38610: [Refactor/Chore] Validate shellctl job ids at API and service boundaries

## Description

The new agent shell feature is currently intended for trusted users. Under the normal product flow, shellctl job ids are generated by the shellctl server and then tracked by the Dify shell layer. The model/user should not be able to invent arbitrary job ids for `shell_wait`, `shell_input`, or `shell_interrupt` because the layer checks the current session's tracked job list first.

CodeQL currently reports `py/path-injection` alerts around shellctl job-id usage in `dify-agent/src/shellctl/server/tmux.py`, including temporary file creation and cleanup in `send_input()`.

These alerts are likely false positives or low-risk under the current design because:

- `run_job()` allocates job ids server-side with `generate_job_id()`.
- Existing product calls remember job ids only after shellctl returns them.
- The Dify shell layer rejects untracked job ids before calling shellctl commands.
- Shellctl service methods look up the job row before acting on a job id.

Still, shellctl should validate job id shape at its own API/service boundary as defense in depth.

## Motivation

Even trusted-user features benefit from local invariant checks. A strict job id validator would:

- encode the intended job id contract close to the shellctl boundary;
- make future direct shellctl API usage safer;
- reduce noise from CodeQL path-injection alerts;
- avoid relying on every caller to preserve the Dify shell layer's tracked-job invariant.

Recommended scope:

- Add a shared job id pattern matching the generated format, for example the timestamp prefix plus the configured alphabet suffix.
- Validate `job_id` path parameters in shellctl API routes or immediately inside `ShellctlService` / `TmuxController` entrypoints.
- Return a structured 400 error for malformed job ids before touching tmux, temporary files, or artifact paths.
- Add local tests for malformed job ids such as path separators, `..`, absolute-path-looking values, whitespace, and shell metacharacters.

## Original CodeQL Annotations

- Alert 346: `dify-agent/src/shellctl/server/tmux.py:235`
  - Code: https://github.com/langgenius/dify/blob/7fabc39e97e4d54787384afdc35f7c021ff8e929/dify-agent/src/shellctl/server/tmux.py#L235
  - CodeQL alert: https://github.com/langgenius/dify/security/code-scanning/346
  - Title: `Uncontrolled data used in path expression`
  - Message: `This path depends on a user-provided value.`
- Alert 344: `dify-agent/src/shellctl/server/tmux.py:283`
  - Code: https://github.com/langgenius/dify/blob/7fabc39e97e4d54787384afdc35f7c021ff8e929/dify-agent/src/shellctl/server/tmux.py#L283
  - CodeQL alert: https://github.com/langgenius/dify/security/code-scanning/344
  - Title: `Uncontrolled data used in path expression`
  - Message: `This path depends on a user-provided value.`

## Additional Context

Relevant files in the current PR branch:

- `dify-agent/src/shellctl/shared/runtime.py`
- `dify-agent/src/shellctl/server/api.py`
- `dify-agent/src/shellctl/server/service.py`
- `dify-agent/src/shellctl/server/tmux.py`
- `dify-agent/tests/local/shellctl/test_shellctl_service.py`

This is a hardening issue, not a statement that the current trusted-user new agent flow is exploitable through arbitrary job id path traversal.

natemiller23:
## Summary

- add a shared validator for the current shellctl job-id contract: exactly 16 lowercase hexadecimal characters, matching `GenerateJobID`
- reject malformed path job IDs at every shellctl HTTP route with a structured `400 invalid_job_id` response before request decoding or service work
- enforce the same invariant at every public service entrypoint before SQLite, tmux, temporary-file, or artifact-path access
- cover generated/valid IDs plus path separators, `..`, absolute-path-looking values, whitespace, uppercase, wrong lengths, and shell metacharacters

The original issue referenced the former Python shellctl server. That server was replaced by the Go `dify-agent-runtime` in #38841, so this applies the requested defense-in-depth validation to the current implementation.

Fixes #38610

Issue: https://github.com/langgenius/dify/issues/38610

## Test plan

- `go test ./internal/server -run 'TestValidateJobID|TestJobRoutesRejectMalformedJobIDBeforeCallingService|TestServiceMethodsRejectMalformedJobIDBeforeUsingDependencies' -count=1 -v`
- `go vet ./...`
- verified the broader Windows failures are unchanged from upstream baseline (`TestConfigPaths` assumes POSIX paths and `TestDumpCLIHelpMatchesCheckedIn` is already stale); the focused server tests pass and Linux CI will run the repository's canonical `go test -race -count=1 ./...`

## OpenJobs

Work completed for [openjobs.bot](https://openjobs.bot), listing `48bf71d4-afb3-4b52-8ecd-d0c8de84476e`.

From Hermes Agent.

### DIFF
--- a/dify-agent-runtime/internal/server/service.go
+++ b/dify-agent-runtime/internal/server/service.go
@@ -318,6 +318,9 @@ func (s *Service) waitForPipeReady(jobID, readyFile string) error {
 
 // WaitJob blocks until output, completion, truncation, or timeout.
 func (s *Service) WaitJob(jobID string, req *WaitJobRequest) (*JobResult, error) {
+	if err := ValidateJobID(jobID); err != nil {
+		return nil, err
+	}
 	row, err := s.db.GetJob(jobID)
 	if err != nil {
 		return nil, err
@@ -408,6 +411,9 @@ func (s *Service) WaitJob(jobID string, req *WaitJobRequest) (*JobResult, error)
 
 // TailJob returns the tail of a job's output.
 func (s *Service) TailJob(jobID string, outputLimit int) (*JobResult, error) {
+	if err := ValidateJobID(jobID); err != nil {
+		return nil, err
+	}
 	row, err := s.db.GetJob(jobID)
 	if err != nil {
 		return nil, err
@@ -425,6 +431,9 @@ func (s *Service) TailJob(jobID string, outputLimit int) (*JobResult, error) {
 
 // GetJobStatus materializes the current status from SQLite + live tmux state.
 func (s *Service) GetJobStatus(jobID string) (*JobStatusView, error) {
+	if err := ValidateJobID(jobID); err != nil {
+		return nil, err
+	}
 	sessionExists, pipeActive, err := s.liveRuntimeState(jobID)
 	if err != nil {
 		return nil, err
@@ -467,6 +476,9 @@ func (s *Service) ListJobs(status *JobStatusName, limit int) (*ListJobsResponse,
 
 // SendInput sends input to a running job and waits.
 func (s *Service) SendInput(jobID string, req *InputJobRequest) (*JobResult, error) {
+	if err := ValidateJobID(jobID); err != nil {
+		return nil, err
+	}
 	view, err := s.GetJobStatus(jobID)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


dify-agent-runtime/internal/server/service.go
```diff
@@ -496,6 +508,9 @@ func (s *Service) SendInput(jobID string, req *InputJobRequest) (*JobResult, err
 
 // TerminateJob terminates a running job.
 func (s *Service) TerminateJob(jobID string, graceSeconds float64) (*JobStatusView, error) {
+	if err := ValidateJobID(jobID); err != nil {
```
wylswz, line 511: This is a shotgun validation. You could define the `ValidateJobID` as 
```go
func ValidateJobID(jobID str) (JobID, error)
```

So that type representing a valid job id is propagated down to service layer, where repeated validations are not necessary.

PR #39185, opened 2026-07-17
fix(api): return 400 instead of 500 when 'file' is missing from installed-app audio transcription

### LINKED ISSUE #39184: Installed-app audio transcription returns 500 instead of 400 when 'file' is missing
## Installed-app audio transcription returns 500 instead of 400 when `file` is missing

**Where:** `api/controllers/console/explore/audio.py`, `ChatAudioApi.post`

**What happens**

POST `/installed-apps/<installed_app_id>/audio-to-text` reads the multipart `file` field with a direct dict access:

```python
file = request.files["file"]
```

If a caller sends the endpoint without a `file` form field, this raises `KeyError: 'file'`. Flask's default error handler turns that into a 500. The endpoint catches `NoAudioUploadedServiceError` and turns it into the right 400 with code `no_audio_uploaded`, but it never reaches that path because the `KeyError` happens before the service call.

**What I expected**

A 400 with `code: no_audio_uploaded`, the same response the sibling endpoint at `api/controllers/console/app/audio.py:200` returns. That endpoint already uses the safe pattern (`request.files.get("file")`) and lets the service layer raise `NoAudioUploadedServiceError` when `file is None`, which the controller maps to `NoAudioUploadedError`.

**Reproduction**

```bash
curl -X POST \
  -H "Authorization: Bearer <token>" \
  https://<host>/installed-apps/<id>/audio-to-text
# -> 500 Internal Server Error (with a generic message)
```

Send a multipart request with an empty body or with any field other than `file`. The same call against `POST /apps/<app_id>/audio-to-text` already returns 400.

**Proposed fix**

One-line change in `ChatAudioApi.post`:

```python
file = request.files.get("file")
```

The service layer (`AudioService.transcript_asr`) already raises `NoAudioUploadedServiceError` when `file is None`, and the existing `except NoAudioUploadedServiceError` handler in the controller maps it to `NoAudioUploadedError` (HTTP 400, code `no_audio_uploaded`). No new error class needed.

**Why this is worth fixing**

- The contract is already documented for the console-side endpoint (`api/controllers/console/app/audio.py:200`). The installed-app variant is the only audio-to-text path that diverges.
- A 500 here hides the real cause from the caller and shows up as a noisy crash in logs and error trackers.
- The fix is one character of behavior change (`[` → `.get(`), zero new dependencies, zero design questions.

**Test coverage**

`api/tests/unit_tests/controllers/console/explore/test_audio.py::TestChatAudioApi` already mocks the service layer for the success/error-mapping paths. The missing case is "no `file` field on the request", which currently raises an uncaught `KeyError` instead of `NoAudioUploadedError`. A new test that posts an empty multipart body should assert `NoAudioUploadedError` is raised.

**Environment**
- `langgenius/dify` main branch, current HEAD
- Python 3.12, Flask multipart parsing enabled by default

Harsh23Kashyap:
## Summary

The installed-app audio-to-text endpoint (`POST /installed-apps/<installed_app_id>/audio-to-text`) returned 500 when callers omitted the multipart `file` field. Direct `request.files["file"]` access raised `KeyError` and Flask's default handler turned it into an unhandled 500.

Switched to `request.files.get("file")`. The service layer (`AudioService.transcript_asr`) already raises `NoAudioUploadedServiceError` when `file is None`, which the controller's existing `except NoAudioUploadedServiceError` block maps to `NoAudioUploadedError` (HTTP 400, code `no_audio_uploaded`). The sibling console endpoint at `api/controllers/console/app/audio.py:200` already uses this pattern.

Fixes #39184.

## Changes

- `api/controllers/console/explore/audio.py` (1 line): `request.files["file"]` -> `request.files.get("file")`
- `api/tests/unit_tests/controllers/console/explore/test_audio.py` (1 new test): posts an empty multipart body and asserts `NoAudioUploadedError` is raised.

## Verification

```
# From api/
uv run ruff check controllers/console/explore/audio.py tests/unit_tests/controllers/console/explore/test_audio.py
# All checks passed!

.venv/bin/python -m pytest tests/unit_tests/controllers/console/explore/test_audio.py -v
# 24 passed, 3 warnings in 4.86s
```

The new test (`test_post_missing_file_returns_no_audio_uploaded`) covers the previously untested missing-file path. All 24 tests in the file pass.

## Risk

None expected. Behavior change is from 500 (uncaught `KeyError`) to 400 (`NoAudioUploadedError`), matching the existing console endpoint. No new dependencies, no API contract change beyond the status code.

### DIFF
--- a/api/controllers/console/explore/audio.py
+++ b/api/controllers/console/explore/audio.py
@@ -54,7 +54,7 @@ class ChatAudioApi(InstalledAppResource):
         if app_model is None:
             raise AppUnavailableError()
 
-        file = request.files["file"]
+        file = request.files.get("file")
 
         try:
             response = AudioService.transcript_asr(
--- a/api/tests/unit_tests/controllers/console/explore/test_audio.py
+++ b/api/tests/unit_tests/controllers/console/explore/test_audio.py
@@ -76,6 +76,15 @@ class TestChatAudioApi:
 
         assert resp == {"text": "ok"}
 
+    def test_post_missing_file_returns_no_audio_uploaded(self, app: Flask, installed_app):
+        with app.test_request_context(
+            "/",
+            data={},
+            content_type="multipart/form-data",
+        ):
+            with pytest.raises(NoAudioUploadedError):
+                self.method(installed_app)
+
     def test_app_unavailable(self, app: Flask, installed_app, audio_file):
         with (
             app.test_request_context(


Harsh23Kashyap:
First-time follow-up. 400 instead of 500 for missing 'file' in installed-app audio transcription. CI green, no review threads. Standing by.

PR #39200, opened 2026-07-17
refactor(web): remove legacy service wrappers

lyzno1:
## Summary

- remove `web/service/billing.ts`, `use-billing.ts`, `use-dataset-card.ts`, and `use-oauth.ts`
- move query and mutation ownership into the consuming components using generated oRPC TanStack Query options
- define the real `/billing/subscription` and `/features` response schemas, then regenerate the TypeScript contracts
- remove handwritten billing response DTOs and update affected tests to exercise request and user-visible behavior

## Why

These service files were thin compatibility wrappers around generated API contracts. They duplicated query keys, request DTOs, and endpoint types, which let frontend types drift from the backend schema. In particular, `/billing/subscription` was documented as an untyped dictionary and `/features` incorrectly documented `vector_space` even though that endpoint intentionally omits it.

## Impact

Consumers now use generated `queryOptions` or `mutationOptions` at their owning component or hook. Event-driven payment and usage checks use `queryClient.fetchQuery` so requests remain lazy. No compatibility aliases or wrapper shims remain.

## Validation

- `pnpm --dir web type-check`
- 10 focused frontend test files: 198 tests passed
- 68 focused backend schema, controller, feature, and billing service tests passed
- API contracts regenerated with `pnpm --filter @dify/contracts gen-api-contract`


lyzno1:
Closing this broad refactor so it can be replaced by smaller, independently verified PRs based on the latest main.

PR #39208, opened 2026-07-17
fix: stream answer nodes inside iteration/loop containers

### LINKED ISSUE #38867: Answer Node not work in iteration (Also missing "message" event in SSE response )
### Self Checks

### Dify version

1.16.0-rc1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Image

### ✔️ Expected Behavior

The answer node within the iteration outputs correctly.

in 1.14.2, it works just fine, and /chat-messages api received "message" event in streaming mode, after upgrade to 1.16.0-rc1, only "message_end" event received which related to message

### ❌ Actual Behavior

look like the flow just ignored the answer node

Image

and "message" event is missing(see my chatflow events which has answer node in iteration)

Image

MatthewSynthia:
## Summary

Fixes #38867. An answer node inside an iteration (or loop) executes but never streams: no SSE `message` events are produced for it, and the persisted message loses the iteration content. Worked in 1.14.2, broken since 1.15.0.

## Root Cause

Since graphon 0.5, answer-node streaming is synthesized by `ResponseStreamFilter` sessions rather than emitted by the node. Answer nodes inside a container run in the container's **child graph**, which breaks the wiring twice:

1. The parent run's filter registers the inner answer node (it appears in the top-level graph config) but can never activate its streaming session — `_build_paths_map` finds no traversal path from the parent root to it, so the session sits pending forever.
2. Child engines built by `_WorkflowChildEngineBuilder` had no response stream filter at all, so nothing synthesized the inner answer's chunks there either.

The node itself runs fine (correct `node_finished` output), which is why the workflow "succeeds" while the user sees nothing — only `message_end` arrives.

## Fix

Dify-side wiring only, no graphon changes (same approach as #38540):

- `DifyResponseStreamFilter` (new, `api/core/workflow/response_stream_filter.py`): skips registering response nodes that have no path from the graph root (they belong to a container's child graph), and passes stream chunks forwarded by containers (`in_iteration_id`/`in_loop_id` set) straight through.
- `_WorkflowChildEngineBuilder` now wraps child engines in `ResponseFilteredChildEngine`, so each iteration/loop pass synthesizes streaming for its own response nodes with a fresh filter (sessions are one-shot, and inner answers run once per pass).
- All existing construction sites (`iter_dify_graph_engine_events`, `WorkflowEntry`, both app generators, `PauseStatePersistenceLayer`) construct the subclass; pause/resume serialization is inherited unchanged.

## Before / After

Chatflow: `start → iteration([apple, banana] → inner answer "{{item}}\n") → answer "done"` (fixture `iteration_contains_answer.yml`):

|  | Streamed to client |
|--------|-------|
| Before | `'done'` |
| After | `'apple\nbanana\ndone'` |

Loop chatflow (existing, previously orphaned fixture `loop_contains_answer.yml`):

|  | Streamed to client |
|--------|-------|
| Before | `'hi + 2'` (inner loop answers lost) |
| After | `'1\n2\nhi + 2'` |

After-fix event stream shows the inner answer chunks forwarded live during each iteration pass:

```
NodeRunIterationNextEvent   node=iteration_node
NodeRunStreamChunkEvent     selector=['iteration_node', 'item'] chunk='apple' in_iteration_id=iteration_node
NodeRunStreamChunkEvent     selector=['answer_in_iteration', 'answer'] chunk='\n' in_iteration_id=iteration_node
...
NodeRunStreamChunkEvent     selector=['answer_final', 'answer'] chunk='done'
```

`is_final` on the per-pass chunks is not propagated into `QueueTextChunkEvent` (only text/selector/container ids), so the SSE message pipeline is unaffected by multiple per-pass final markers.

## Tests

- New `test_answer_in_container_streaming.py`: regression coverage for both the iteration case (#38867) and the loop case, using real (non-mock) nodes through the production child-engine builder path. Both fail without this fix (streamed answer collapses to the outer answer only).
- Updated `test_workflow_entry_helpers.py` construction assertions; `_TableTestChildEngineBuilder` mirrors production child-engine wrapping.
- `uv run pytest tests/unit_tests` — 15756 passed, 6 skipped.
- `make lint && make type-check` — clean.

## Checklist

From Claude Code

### DIFF
--- a/api/core/app/apps/advanced_chat/app_generator.py
+++ b/api/core/app/apps/advanced_chat/app_generator.py
@@ -46,6 +46,7 @@ from core.ops.ops_trace_manager import TraceQueueManager
 from core.prompt.utils.get_thread_messages_length import get_thread_messages_length
 from core.repositories import DifyCoreRepositoryFactory
 from core.repositories.factory import WorkflowExecutionRepository, WorkflowNodeExecutionRepository
+from core.workflow.response_stream_filter import DifyResponseStreamFilter
 from extensions.ext_database import db
 from factories import file_factory
 from graphon.filters import ResponseStreamFilter
@@ -575,7 +576,7 @@ class AdvancedChatAppGenerator(MessageBasedAppGenerator):
             )
 
             graph_layers: list[GraphEngineLayer] = list(graph_engine_layers)
-            resolved_response_stream_filter = response_stream_filter or ResponseStreamFilter()
+            resolved_response_stream_filter = response_stream_filter or DifyResponseStreamFilter()
             if pause_state_config is not None:
                 graph_layers.append(
                     PauseStatePersistenceLayer(
--- a/api/core/app/apps/workflow/app_generator.py
+++ b/api/core/app/apps/workflow/app_generator.py
@@ -41,6 +41,7 @@ from core.helper.trace_id_helper import (
 from core.ops.ops_trace_manager import TraceQueueManager
 from core.repositories import DifyCoreRepositoryFactory
 from core.repositories.factory import WorkflowExecutionRepository, WorkflowNodeExecutionRepository
+from core.workflow.response_stream_filter import DifyResponseStreamFilter
 from extensions.ext_database import db
 from factories import file_factory
 from graphon.filters import ResponseStreamFilter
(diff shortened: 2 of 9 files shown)


MatthewSynthia:
@laipz8200 @QuantumGhost ready when you have a moment. Streams answer nodes nested inside iteration/loop containers (fixes #38867), checks are green. Let me know if anything needs changing.

guopeng1994:
any updates?

PR #39216, opened 2026-07-18
refactor(web): avoid direct setState in effect in jina reader crawler (#25194)

pepepeboom:
part of #25194

## Summary

Removes the direct `setState` call inside the `useEffect` in `web/app/components/datasets/create/website/jina-reader/index.tsx`, following the pattern of #25212.

The effect bumped `controlFoldOptions` (the signal that tells `OptionsWrap` to fold the options panel) whenever `step` changed away from `init`. `step` only transitions inside the `handleRun` event handler (`setStep(Step.running)` when a crawl starts, `setStep(Step.finished)` in its `finally`), so the fold trigger is now set directly at those two transitions instead of in a follow-up effect — same behavior, one less render pass.

This was the only `useEffect` in the file, so the now-unused `useEffect` import is removed. The file's `eslint-react/set-state-in-effect` entry is also removed from `oxlint-suppressions.json`, so the rule is enforced for this file again.

Verified:

- `pnpm lint:oxlint` (`vp lint`) passes with the suppression removed — and fails with `eslint-react(set-state-in-effect)` on this file when the old code is restored, confirming the rule is active again.
- All 60 existing specs in `jina-reader/__tests__/index.spec.tsx` pass, including "should fold options when step changes from init" and "should update controlFoldOptions when step changes".

Logic refactor only, no behavior change.

## Checklist

- [x] No new tests added: the existing `jina-reader/__tests__/index.spec.tsx` suite already covers the options-fold behavior; this is a single atomic change.
- [x] I ran the frontend lint — the pre-commit `vp check` hook passed on the staged files, and the un-suppressed `eslint-react/set-state-in-effect` rule passes on the changed file with no new issues introduced.

From Claude Code


WTW0313 (approved):
LGTM

PR #39220, opened 2026-07-18
fix(api): honor array-element limit and byte budget for list[File] in VariableTruncator (#39218)

### LINKED ISSUE #39218: VariableTruncator._truncate_array: File-array items bypass both array_element_limit and the size budget
### Self Checks

- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones. This is a different root cause from #37192 (which is about the `truncated` flag being under-reported for non-`File` array elements); this issue is about `File` elements bypassing the size/count limits entirely.

### Dify version

main (tested against commit from 2026-07-18)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

`VariableTruncator._truncate_array` in `api/services/variable_truncator.py` (lines 267-305) is supposed to guarantee that a truncated array never exceeds `array_element_limit` items nor the caller's `target_size` byte budget. It does this correctly for arrays of primitives (str/int/float/bool/list/dict), but a special-case branch added for `File` elements (the "Dirty fix" comment on line 281, from #26133) bypasses **both** guarantees:

```python
def _truncate_array(self, value: list[object], target_size: int) -> _PartResult[list[object]]:
    truncated_value: list[object] = []
    truncated = False
    used_size = self.calculate_json_size([])

    target_length = self._array_element_limit

    for i, item in enumerate(value):
        # Dirty fix:
        # The output of `Start` node may contain list of `File` elements,
        # causing `AssertionError` while invoking `_truncate_json_primitives`.
        #
        # This check ensures that `list[File]` are handled separately
        if isinstance(item, File):
            truncated_value.append(item)
            continue
        if i >= target_length:
            return _PartResult(truncated_value, used_size, True)
        if i > 0:
            used_size += 1  # Account for comma
        ...
```

Because the `isinstance(item, File)` check and its `continue` happen **before** the `i >= target_length` count check and before `used_size` is ever incremented, every `File` item in the array is unconditionally kept, and its size is never added to `used_size`. So a `list[File]` value:

1. is never capped at `array_element_limit` (default 20) items, and
2. never counts against `max_size_bytes` (default 1000 KB), and
3. never gets `truncated=True` set, even though the actual serialized size can be arbitrarily larger than the budget.

This directly contradicts the class's own stated guarantee ("ensuring the final size doesn't exceed specified limits", `VariableTruncator` docstring) and the purpose of `truncate_variable_mapping`, which uses the returned `used_size` to compute the remaining byte budget for sibling keys in the same `inputs`/`outputs`/`process_data` mapping — an artificially low `used_size` for a `File` array means other, unrelated keys in the same mapping get an inflated budget calculation.

**Minimal reproduction** (unit level, against the unmodified `VariableTruncator`/`graphon.file.File`):

```python
from graphon.file.models import File, FileTransferMethod, FileType

def make_file(name):
    return File(
        id=name, type=FileType.DOCUMENT, transfer_method=FileTransferMethod.REMOTE_URL,
        remote_url=f"https://example.com/{name}.txt", filename=f"{name}.txt",
        extension=".txt", mime_type="text/plain", size=1024,
    )

truncator = VariableTruncator(array_element_limit=3, max_size_bytes=1000, string_length_limit=50)

# Baseline: a plain-int array IS correctly capped at 3 items.
int_result = truncator._truncate_array(list(range(10)), target_size=200)
print(len(int_result.value), int_result.truncated)   # 3 True

# Bug: 500 File items, with the SAME 3-item limit and 200-byte target.
files = [make_file(f"f{i}") for i in range(500)]
file_result = truncator._truncate_array(files, target_size=200)
print(len(file_result.value), file_result.truncated, file_result.value_size)
```

### ✔️ Expected Behavior

`file_result.value` should be capped at `array_element_limit` (3) `File` entries (or at minimum `used_size`/`truncated` should reflect the real size of the `File` payload, matching the guarantee `_truncate_array` provides for every other element type).

### ❌ Actual Behavior

```
3 True
500 False 2
```

All 500 `File` entries survive (`array_element_limit=3` is never applied), `truncated` is reported as `False`, and the reported `used_size` is `2` bytes even though the actual serialized size of the 500 returned `File` objects is over 120,000 bytes (verified with `VariableTruncator.calculate_json_size(file_result.value)` on the same run). A workflow node (e.g. an HTTP Request or Start node) that returns a large list of file references — a completely ordinary shape for RAG/document-processing flows — will have that array pass through truncation untouched and unaccounted for, regardless of how small `WORKFLOW_VARIABLE_TRUNCATION_MAX_SIZE`/`WORKFLOW_VARIABLE_TRUNCATION_ARRAY_LENGTH` are configured.

### Suggested fix

`_truncate_json_primitives` (same file) already has a correct, dedicated `File` branch that returns the file untouched but with its real computed size (`case File(): return _PartResult(val, self.calculate_json_size(val), False)`); it's just not reachable from `_truncate_array` because the `isinstance` gate on the line that calls `_truncate_json_primitives` doesn't include `File`:

```python
if item is None or isinstance(item, (str, list, dict, bool, int, float, UpdatedVariable)):
    part_result = self._truncate_json_primitives(item, remaining_budget)
else:
    raise UnknownTypeError(f"got unknown type {type(item)} in array truncation")
```

Removing the early `isinstance(item, File): ... continue` bypass and adding `File` to that tuple would let `File` items flow through the normal count-limit and size-budget accounting (still without truncating the `File` object itself, since `_truncate_json_primitives`'s `File` branch already handles that), which looks like the originally-intended fix for #26133 without the size/count side effect.

Happy to open a PR for this if useful, though I see `_truncate_array` already has several duplicate open/closed PRs in flight for a different bug (#37192) — let me know if you'd rather this be folded into that effort.

Taranum01:
## Summary

`VariableTruncator._truncate_array` had a "Dirty fix" branch added in
[#26133](https://github.com/langgenius/dify/pull/26133) that
**unconditionally appended every `File` element to `truncated_value`
*before* the count cap and the byte-budget check, and *before*
`used_size` was ever incremented**. As a result a `list[File]` was:

1. never capped at `array_element_limit` (default 20 items),
2. never counted against `max_size_bytes` (default 1000 KB), and
3. always reported `truncated=False`.

This directly contradicts the class's own stated guarantee
("ensuring the final size doesn't exceed specified limits",
`VariableTruncator` docstring) and silently inflated the per-key byte
budget reported back to `truncate_variable_mapping`, which uses
`used_size` to compute the remaining budget for sibling keys in the
same `inputs`/`outputs`/`process_data` mapping.

`_truncate_json_primitives` already has a correct, dedicated `File`
branch that returns the file untouched but reports its real computed
size (`case File(): return _PartResult(val, self.calculate_json_size(val), False)`).
The fix removes the early `isinstance(item, File): ... continue`
bypass and adds `File` to the tuple of element types handled by the
normal count-and-size accounting loop. `File` items still aren't
truncated (the dedicated `File` branch keeps that), but they now
flow through the same count cap and byte budget as every other
element type — which is what #26133 originally needed.

## Reproduction (per the issue)

```python
truncator = VariableTruncator(array_element_limit=3, max_size_bytes=1000, string_length_limit=50)

# Baseline: a plain-int array IS correctly capped at 3 items.
int_result = truncator._truncate_array(list(range(10)), target_size=200)
# -> (3, True)

# Bug: 500 File items, with the SAME 3-item limit and 200-byte target.
files = [make_file(f"f{i}") for i in range(500)]
file_result = truncator._truncate_array(files, target_size=200)
# Before fix: (500, False, 2)
# After  fix: (3, True, 713)
```

## Regression tests

Added `TestFileArrayTruncationRegression39218` in
`tests/unit_tests/services/test_variable_truncator.py` with 5 cases:

- count cap honored (500 files → 3 with cap=3),
- `used_size` reflects real serialized size (not 2),
- byte budget honored (50 files with small budget → 1 file returned),
- mixed arrays count Files toward the cap,
- a single File in an array stays intact.

Each new test was verified to fail against the un-fixed source (via
`git stash` of the implementation file) and pass with the fix. Full
file: **52 passed**.

## Verification

- `uv run pytest tests/unit_tests/services/test_variable_truncator.py` → **52 passed**.
- `ruff check` on changed files → clean.
- `ruff format --check` on changed files → clean.

Fixes #39218

### DIFF
--- a/api/services/variable_truncator.py
+++ b/api/services/variable_truncator.py
@@ -278,14 +278,14 @@ class VariableTruncator(BaseTruncator):
         target_length = self._array_element_limit
 
         for i, item in enumerate(value):
-            # Dirty fix:
-            # The output of `Start` node may contain list of `File` elements,
-            # causing `AssertionError` while invoking `_truncate_json_primitives`.
-            #
-            # This check ensures that `list[File]` are handled separately
-            if isinstance(item, File):
-                truncated_value.append(item)
-                continue
+            # ``File`` is routed through ``_truncate_json_primitives`` (whose
+            # dedicated ``File`` branch returns the file as-is with its real
+            # serialized size). That preserves the count cap
+            # (``array_element_limit``) and the byte budget (``target_size``)
+            # for ``list[File]`` — the original "Dirty fix" branch above this
+            # loop bypassed both guarantees and reported ``used_size=2`` even
+            # when the returned array serialized to well over the budget.
+            # See https://github.com/langgenius/dify/issues/39218.
             if i >= target_length:
                 return _PartResult(truncated_value, used_size, True)
             if i > 0:
@@ -295,7 +295,7 @@ class VariableTruncator(BaseTruncator):
                 break
 
             remaining_budget = target_size - used_size
-            if item is None or isinstance(item, (str, list, dict, bool, int, float, UpdatedVariable)):
+            if item is None or isinstance(item, (str, list, dict, bool, int, float, File, UpdatedVariable)):
                 part_result = self._truncate_json_primitives(item, remaining_budget)
             else:
                 raise UnknownTypeError(f"got unknown type {type(item)} in array truncation")
--- a/api/tests/unit_tests/services/test_variable_truncator.py
+++ b/api/tests/unit_tests/services/test_variable_truncator.py
@@ -673,3 +673,118 @@ def test_dummy_variable_truncator_methods():
     assert isinstance(result, TruncationResult)
     assert result.result == segment
     assert result.truncated is False
+
+
+# ---------------------------------------------------------------------------
+# Regression tests for langgenius/dify#39218.
+#
+# Before the fix, ``_truncate_array`` had a "Dirty fix" branch that
+# unconditionally appended every ``File`` element to ``truncated_value``
+# *before* the count cap and the byte-budget check, and *before*
+# ``used_size`` was ever incremented. That made ``list[File]`` arrays:
+#   1. uncapped by ``array_element_limit``,
+#   2. uncounted against ``max_size_bytes``, and
+#   3. always reported ``truncated=False``.
+# The fix routes ``File`` through ``_truncate_json_primitives``'s dedicated
+# ``File`` branch, which returns the file as-is with its real serialized
+# size, while preserving the count cap and the byte budget.
+# ---------------------------------------------------------------------------
+
+
+class TestFileArrayTruncationRegression39218:
+    """``list[File]`` must respect ``array_element_limit`` and the byte budget."""
+
+    @pytest.fixture
+    def truncator(self) -> VariableTruncator:
+        return VariableTruncator(
+            array_element_limit=3,
+            max_size_bytes=1000,
+            string_length_limit=50,
+        )
+
+    @staticmethod
+    def _make_file(name: str = "f") -> File:
+        return File(
+            id=name,
+            type=FileType.DOCUMENT,
+            transfer_method=FileTransferMethod.REMOTE_URL,
+            remote_url=f"https://example.com/{name}.txt",
+            filename=f"{name}.txt",
+            extension=".txt",
+            mime_type="text/plain",
+            size=1024,
+        )
+
+    def test_file_array_respects_element_count_cap(self, truncator: VariableTruncator) -> None:
+        # Use a target_size larger than ``count * file_size`` so the byte
+        # budget never binds — only the count cap should fire.
+        # Each File serializes to ~237 bytes; 3 files = ~713 bytes.
+        files = [self._make_file(f"f{i}") for i in range(500)]
+
+        result = truncator._truncate_array(files, target_size=10_000_000)
+
+        # Before the fix, all 500 File entries survived (``len(value)==500``,
+        # ``truncated==False``). After the fix, the array is capped at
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
Verified against current main: _truncate_json_primitives already has a dedicated File branch (case File(): returns the file as-is with its real calculate_json_size), it just never got reached for array elements because the old Dirty fix branch short-circuited before the type dispatch. Routing File through the same isinstance check as the other primitives is the right fix, not a new code path, just removing the thing that was bypassing the existing one.

Test coverage is thorough, count cap, byte budget, real used_size reporting, mixed File/primitive arrays, and the single-file no-truncation case are all exercised separately rather than folded into one big assertion. Confirms the fix without over-fitting to one scenario.

fatelei (approved):

ErenAta16:
Thanks for picking this up and landing it, @Taranum01.

The shape is right: dropping the early-append branch and adding `File` to the `isinstance` tuple so it flows through `_truncate_json_primitives` means the count cap and the byte budget both apply, and `used_size` now reflects the real serialized size instead of the empty-array baseline of 2. Replacing the "Dirty fix" comment with one that explains *why* `File` belongs in the normal path (and links the issue) is the part that will stop this regressing later.

Good call adding both directions in the regression tests, the count cap firing with the byte budget slack and the byte budget firing independently. Testing only one would have let a fix that satisfies the cap but still miscounts bytes slip through.

PR #39264, opened 2026-07-20
fix(dataset_service): widen create_segment lock to cover insert and commit

### LINKED ISSUE #39022: Race condition in SegmentService.create_segment lets concurrent requests commit duplicate segment position values
### Self Checks

- [ ] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [ ] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.16.0-rc1 (also present on main)

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

`SegmentService.create_segment` (`api/services/dataset_service.py:3334-3417`) is supposed to use a per-document Redis lock (`add_segment_lock_document_id_{document.id}`) to make "read current max segment position, then insert a new segment at `max_position + 1`" atomic across concurrent requests.

The lock only wraps the *read* of `max_position` and the construction of the in-memory `DocumentSegment` object:

```python
lock_name = f"add_segment_lock_document_id_{document.id}"
try:
    with redis_client.lock(lock_name, timeout=600):
        max_position = session.scalar(
            select(func.max(DocumentSegment.position)).where(DocumentSegment.document_id == document.id)
        )
        segment_document = DocumentSegment(
            ...
            position=max_position + 1 if max_position else 1,
            ...
        )
        if document.doc_form == IndexStructureType.QA_INDEX:
            segment_document.word_count += len(args["answer"])
            segment_document.answer = args["answer"]

    session.add(segment_document)      # <-- outside the lock
    # update document word count
    assert document.word_count is not None
    document.word_count += segment_document.word_count
    session.add(document)
    session.commit()                   # <-- outside the lock
```

`session.add(segment_document)` / `session.commit()` sit outside the `with redis_client.lock(...)` block, at the same indentation as the surrounding `try:`. The lock is released as soon as the `max_position` read and object construction finish — well before the row is persisted.

If two "add segment" requests for the same document arrive concurrently, both can acquire the lock sequentially, each read the same `max_position` (since neither has committed yet when the other reads), release the lock, and only then commit. Both end up inserting a segment with the **identical `position`** value. There is no unique constraint on `(document_id, position)` in `models/dataset.py` (only non-unique indexes), so both inserts succeed silently.

For comparison, the structurally identical `multi_create_segment` (`api/services/dataset_service.py:3420` onward) keeps `session.add(...)` for each segment **inside** the lock's `with` block — confirming the intended pattern, and that `create_segment`'s placement is a mistake rather than a deliberate design choice.

### ✔️ Expected Behavior

Concurrent calls to create a segment on the same document should always be assigned strictly increasing, unique `position` values — the read-check-insert-commit sequence should be fully covered by the lock, matching `multi_create_segment`'s behavior.

### ❌ Actual Behavior

Two segments created concurrently on the same document can end up with the same `position` value, since the DB insert/commit happens after the lock protecting the position read has already been released. This silently corrupts segment ordering, which is relied on for display order and for prev/next segment lookups (`DocumentSegment.position == self.position - 1` in `models/dataset.py`).

sergioperezcheco:
SegmentService.create_segment currently wraps only the max_position read and the DocumentSegment construction in its per-document Redis lock (add_segment_lock_document_id_{document.id}). The session.add(), the document word-count update, session.add(document) and session.commit() all run *after* the lock has been released. That leaves a classic check-then-act window: two concurrent create-segment requests for the same document can both read the same max_position before either commits, so both insert a segment at the identical position value. Since there's no unique constraint on (document_id, position), both inserts succeed silently and segment ordering gets corrupted.

This moves session.add()/commit() (plus the attachment-binding write and the vector-index write that depend on the committed segment id) inside the with redis_client.lock(...) block, so the read-position / insert / commit sequence is atomic under the lock. This mirrors the existing multi_create_segment method, which already keeps its session.add()/commit() within its own lock block — so create_segment now follows the same established pattern.

Verified against the existing unit tests (test_dataset_service_segment.py and test_dataset_service_lock_not_owned.py, 45 tests pass). No behaviour change in the non-contended path; the only difference is the lock is held until commit instead of being released early.

Fixes #39022

### Checklist

- [x] My code follows the style guidelines of this project (Ruff clean)
- [x] I have added tests that prove my fix is effective or that my feature works (existing unit tests cover the lock-held path)

### DIFF
--- a/api/services/dataset_service.py
+++ b/api/services/dataset_service.py
@@ -3393,41 +3393,47 @@ class SegmentService:
                     segment_document.word_count += len(args["answer"])
                     segment_document.answer = args["answer"]
 
-            session.add(segment_document)
-            # update document word count
-            assert document.word_count is not None
-            document.word_count += segment_document.word_count
-            session.add(document)
-            session.commit()
+                # Persist the segment and the document word-count update inside
+                # the lock so the read-max-position / insert / commit sequence is
+                # atomic across concurrent requests. Otherwise two callers can
+                # both read the same max_position before either commits, producing
+                # duplicate positions. This matches multi_create_segment, which
+                # keeps session.add()/commit() within its lock block.
+                session.add(segment_document)
+                # update document word count
+                assert document.word_count is not None
+                document.word_count += segment_document.word_count
+                session.add(document)
+                session.commit()
 
-            if args["attachment_ids"]:
-                for attachment_id in args["attachment_ids"]:
-                    binding = SegmentAttachmentBinding(
-                        tenant_id=current_user.current_tenant_id,
-                        dataset_id=document.dataset_id,
-                        document_id=document.id,
-                        segment_id=segment_document.id,
-                        attachment_id=attachment_id,
+                if args["attachment_ids"]:
+                    for attachment_id in args["attachment_ids"]:
+                        binding = SegmentAttachmentBinding(
+                            tenant_id=current_user.current_tenant_id,
+                            dataset_id=document.dataset_id,
+                            document_id=document.id,
+                            segment_id=segment_document.id,
+                            attachment_id=attachment_id,
+                        )
+                        session.add(binding)
+                    session.commit()
+
+                # save vector index
+                try:
+                    keywords = args.get("keywords")
+                    keywords_list = [keywords] if keywords is not None else None
+                    VectorService.create_segments_vector(
+                        keywords_list, [segment_document], dataset, document.doc_form, session=session
                     )
-                    session.add(binding)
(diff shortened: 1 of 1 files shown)


crazywoola:

### Why this is being closed

Issue #39022 has multiple open pull requests. Under the duplicate-PR rule, pull requests from the issue author are ranked first, then creation time is used as the FIFO tie-breaker. PR #39029 is the retained pull request, so this one is being closed.

### Next steps

Please coordinate on PR #39029 if you would like to contribute further. Thank you for understanding.

PR #39294, opened 2026-07-20
fix(account): resolve account.id/email before session detach in reset-password flow

### LINKED ISSUE #39287: Forgot password endpoint returns 500 DetachedInstanceError for specific accounts, while others work fine (Resend mail provider)
### Self Checks

### Dify version

1.13.3

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

When using the "Forgot Password" functionality, the endpoint /console/api/forgot-password returns a 500 Internal Server Error for my primary account (xxx@live.com), but the same endpoint works correctly for another account (cxxx@outlook.com) – the latter successfully receives the password reset email via Resend.

After inspecting the logs (docker compose logs), I found a sqlalchemy.orm.exc.DetachedInstanceError which indicates that the Account object is not bound to a database session when trying to access its email attribute during the password reset flow.

Deploy Dify with Resend as the mail provider (configured via .env).

Have two user accounts:
Account A: xxx@live.com (triggers error)
Account B:xxx@outlook.com (works)
Submit a password reset request for Account A via the login page.
Observe that the request fails with a 500 error and no email is sent.
Submit the same request for Account B – the email is delivered successfully.

### ✔️ Expected Behavior

The password reset email should be sent to both accounts without errors, regardless of the email address or account ID.

### ❌ Actual Behavior

For Account A, the server returns a 500 error, and the following exception is logged in the api container:

`sqlalchemy.orm.exc.DetachedInstanceError: Instance <Account at 0x...> is not bound to a Session; attribute refresh operation cannot proceed`
The full traceback points to services/account_service.py in send_reset_password_email, where the code attempts to access account.email on an object that has become detached from the session.

sergioperezcheco:
The forgot-password and reset-password endpoints load the Account from the request's scoped session (db.session()). Under gevent that session can be recycled before TokenManager.generate_token runs, leaving the Account instance detached. Reading account.email / account.id on the detached instance then raises a DetachedInstanceError and the endpoint returns 500 — this is the root cause of #39287.

The fix resolves account.id and the email string while the request session is still bound, then threads them through explicitly. send_reset_password_email, generate_reset_password_token, and TokenManager.generate_token all gain an optional account_id parameter and prefer the passed-in email over account.email. The three entry points (console forgot-password, console login reset-password, web forgot-password) each pre-resolve account.id before the call and pass it down, so no code path lazily loads ORM attributes off a possibly-detached instance.

The change is additive and backward compatible — callers that don't pass account_id fall back to the existing account.id read. I kept it minimal and confined to the reset-password token flow.

A regression test drives send_reset_password_email with a pre-resolved account_id and asserts the persisted token payload carries that id without ever touching account.email / account.id on the (detached) Account.

Closes #39287

### DIFF
--- a/api/controllers/console/auth/forgot_password.py
+++ b/api/controllers/console/auth/forgot_password.py
@@ -82,9 +82,14 @@ class ForgotPasswordSendEmailApi(Resource):
             language = "en-US"
 
         account = AccountService.get_account_by_email_with_case_fallback(args.email, session=db.session())
+        # Read account.id here, while the request session is still bound, so the
+        # downstream token flow does not lazy-load it from a possibly-detached
+        # instance (DetachedInstanceError, #39287).
+        account_id = account.id if account else None
 
         token = AccountService.send_reset_password_email(
             account=account,
+            account_id=account_id,
             email=normalized_email,
             language=language,
             is_allow_register=FeatureService.get_system_features().is_allow_register,
--- a/api/controllers/console/auth/forgot_password.py
+++ b/api/controllers/console/auth/forgot_password.py
@@ -82,9 +82,14 @@ class ForgotPasswordSendEmailApi(Resource):
             language = "en-US"
 
         account = AccountService.get_account_by_email_with_case_fallback(req_data.email, session=db.session())
+        # Read account.id here, while the request session is still bound, so the
+        # downstream token flow does not lazy-load it from a possibly-detached
+        # instance (DetachedInstanceError, #39287).
+        account_id = account.id if account else None
 
         token = AccountService.send_reset_password_email(
             account=account,
+            account_id=account_id,
             email=normalized_email,
             language=language,
             is_allow_register=FeatureService.get_system_features().is_allow_register,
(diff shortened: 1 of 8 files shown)


sergioperezcheco:
The failing test `test_contract_cli_updates_checks_and_detects_openapi_drift` (in `tests/unit_tests/dev/test_generate_knowledge_fs_contract.py`) is unrelated to this PR — it only touches the account/auth layer. The failure is a `KeyError: 'PATH'` raised inside a `monkeypatch` block, which is a known environment-dependent flake in the knowledge-fs contract test (the CI runner's environ occasionally lacks `PATH` in the xdist worker). Could a maintainer re-run the "API Unit Tests (3.12)" job? Happy to make changes if there's any real concern with the account/auth changes.

sergioperezcheco:
The only failing test is `test_generate_knowledge_fs_contract.py::test_contract_cli_updates_checks_and_detects_openapi_drift` in `api/tests/unit_tests/dev/`, which errors with `KeyError: 'PATH'` during env access — looks like a CI runner issue rather than anything in this PR's scope (account/auth/session). The other 13723 tests pass. Would appreciate a re-run, or happy to investigate further if you think it's related.

sergioperezcheco:
The API test failures were a collection error, not a test assertion: the test imported `DeploymentEdition` from `enums.deployment_edition`, a module that doesn't exist on main (the package exposes it via `__init__`). Reverted the import to `from enums import DeploymentEdition` in 90cecc1c86 — CI should be green again.

PR #39304, opened 2026-07-20
fix(api): return 400 instead of 500 when 'file' is missing from web-app audio-to-text

### LINKED ISSUE #39303: Audio-to-text endpoints return 500 instead of 400 when the 'file' field is missing
## Audio-to-text endpoints return 500 instead of 400 when the `file` field is missing

Three sibling audio-to-text controllers under `api/controllers/` raise `KeyError: 'file'` (HTTP 500) when a multipart POST omits the `file` field, instead of returning the existing 400 `no_audio_uploaded` error.

PR #39219 fixes one of them (`api/controllers/console/explore/audio.py:57`). The other two have the same bug and the same one-line fix.

### Affected endpoints

| Endpoint | File | Line |
| --- | --- | --- |
| Web-app audio-to-text (public) | `api/controllers/web/audio.py` (`AudioApi.post`) | 79 |
| Trial-app audio-to-text | `api/controllers/console/explore/trial.py` (`TrialChatAudioApi.post`) | 648 |
| Service API audio-to-text | `api/controllers/service_api/app/audio.py` (`AudioApi.post`) | 104 |

### Current behaviour

```
curl -X POST $APP_URL/audio-to-text
```

Returns `500 Internal Server Error` with a stack trace pointing at `request.files["file"]`.

### Expected behaviour

Returns `400 Bad Request` with the existing `no_audio_uploaded` error payload. The service layer already supports this path: `AudioService._invoke_speech_to_text` raises `NoAudioUploadedServiceError()` when `file is None`, and each controller already has an `except NoAudioUploadedServiceError: raise NoAudioUploadedError()` handler.

### Proposed fix

Replace `request.files["file"]` with `request.files.get("file")` at each of the three lines above. This matches the pattern already used in `api/controllers/console/app/audio.py:200`.

### Why it matters

- The web-app endpoint is public. Any client that omits the file — frontend bug, retry, partial form — hits a 500 with a server stack trace instead of the documented 400.
- The trial and service API endpoints power trial-app audio and programmatic integrations respectively.
- All three return the same wrong status for the same reason. Fixing them together keeps the API consistent across surfaces and removes a class of avoidable Sentry noise.

### Scope

Three one-line code changes, each with one test that asserts a missing file returns 400.

### Backward compatibility

None. Existing callers that send a valid `file` field are unaffected. Clients that previously got a 500 for missing file will now get a 400 with the same `no_audio_uploaded` error payload the service layer was designed to emit.

### Risks

None observed. The change is a strict improvement: malformed requests get the documented 400 instead of a stack trace. The service-layer `None` guard is already in place and exercised by other call sites.

### Acceptance criteria

- `curl -X POST <endpoint>` with no `file` returns 400 and `error_code = "no_audio_uploaded"`.
- A request with a valid `file` still returns 200 and the transcribed text.
- Each touched controller has a unit test for the missing-file case.

Harsh23Kashyap:
## Summary

The web-app audio-to-text endpoint (`POST /audio-to-text`) returns HTTP 500 with a `KeyError` stack trace when a client submits a multipart POST without the `file` field. It should return HTTP 400 with the existing `no_audio_uploaded` error payload.

Same fix as PR #39219 (sibling `explore/audio.py`). See issue #39303 for the family of three sibling endpoints.

## What changed

- `api/controllers/web/audio.py:79` — `request.files["file"]` → `request.files.get("file")`.
- `api/tests/unit_tests/controllers/web/test_audio.py` — new `test_missing_file_field_returns_400` asserts the controller raises `NoAudioUploadedError` (HTTP 400) when the multipart `file` field is absent.

## How it works

The service layer already handles `file=None` cleanly:

```python
# api/services/audio_service.py
def _invoke_speech_to_text(cls, app_model, file, end_user):
    if file is None:
        raise NoAudioUploadedServiceError()
```

The controller already catches that and maps it to the documented 400:

```python
except NoAudioUploadedServiceError:
    raise NoAudioUploadedError()
```

The previous `request.files["file"]` access crashed with `KeyError` before this path could fire, so the `except NoAudioUploadedServiceError` clause was unreachable for this case.

## Verification

- `ruff check` and `ruff format --check` pass.
- `pytest tests/unit_tests/controllers/web/test_audio.py` — 13 passed, including the new `test_missing_file_field_returns_400`.
- Existing tests for the happy path and all error paths remain green.

## Backward compatibility

None. Clients that send a valid `file` field see no change. Clients that previously got a 500 for missing `file` will now get a 400 with the same `no_audio_uploaded` error payload the service was already designed to emit.

## Related

- Fixes #39303 (first of three siblings: web-app, trial-app, service API)
- Mirrors PR #39219 for the `explore/audio.py` sibling

### DIFF
--- a/api/controllers/web/audio.py
+++ b/api/controllers/web/audio.py
@@ -76,7 +76,7 @@ class AudioApi(WebApiResource):
     @web_ns.response(200, "Success", web_ns.models[AudioToTextResponse.__name__])
     def post(self, app_model: App, end_user: EndUser):
         """Convert audio to text"""
-        file = request.files["file"]
+        file = request.files.get("file")
 
         try:
             response = AudioService.transcript_asr(
--- a/api/tests/unit_tests/controllers/web/test_audio.py
+++ b/api/tests/unit_tests/controllers/web/test_audio.py
@@ -131,9 +131,7 @@ class TestAudioApi:
             raise NoAudioUploadedServiceError()
 
         with patch("controllers.web.audio.AudioService.transcript_asr", side_effect=fake_asr):
-            with app.test_request_context(
-                "/audio-to-text", method="POST", data={}, content_type="multipart/form-data"
-            ):
+            with app.test_request_context("/audio-to-text", method="POST", data={}, content_type="multipart/form-data"):
                 with pytest.raises(NoAudioUploadedError) as exc_info:
                     AudioApi().post(_app_model(), _end_user())
 
--- a/api/tests/unit_tests/controllers/web/test_audio.py
+++ b/api/tests/unit_tests/controllers/web/test_audio.py
@@ -119,6 +119,26 @@ class TestAudioApi:
             with pytest.raises(ProviderModelCurrentlyNotSupportError):
                 AudioApi().post(_app_model(), _end_user())
 
+    def test_missing_file_field_returns_400(self, app: Flask) -> None:
+        """A multipart POST with no `file` field must surface as 400, not 500.
+
+        Verifies the controller passes file=None to AudioService.transcript_asr
+        instead of raising a KeyError that would yield HTTP 500.
+        """
+
+        def fake_asr(*args, **kwargs):
+            assert kwargs["file"] is None
+            raise NoAudioUploadedServiceError()
+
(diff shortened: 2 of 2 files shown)


crazywoola:

### Why this is being closed

### Next steps

PR #38149, opened 2026-06-29
feat: Require privileged role for workspace renames

gauravbsinghal:
## Summary
Require admin or owner privileges before accepting workspace rename requests through the console workspace settings endpoint. This prevents normal workspace members from reaching the tenant mutation path while preserving owner/admin behavior.

close https://github.com/langgenius/dify/issues/38364
## Consequence
A non-admin workspace member can rename the shared workspace, altering tenant-wide metadata visible to every member without admin or owner approval.

## Reproduction
Workspace rename authorization bypass — actor: a normal workspace member (substitute your own equivalent)

1. Authenticate the normal workspace member:
   ```bash
   curl -i -X POST "$TARGET/console/api/login" \
     -H 'Content-Type: application/json' \
     --data '{"email":"<normal-member-email>","password":"<base64-normal-member-password>","remember_me":true}'
   ```
   -> Observed login success with access, refresh, and CSRF cookies.

2. Attempt to rename the shared workspace as that normal member:
   ```bash
   curl -i -X POST "$TARGET/console/api/workspaces/info" \
     -H 'Content-Type: application/json' \
     -H 'Authorization: Bearer <normal-member-access-token>' \
     -H 'X-CSRF-Token: <normal-member-csrf-token>' \
     --cookie 'access_token=<normal-member-access-token>; refresh_token=<normal-member-refresh-token>; csrf_token=<normal-member-csrf-token>' \
     --data '{"name":"Niro Tenant Alpha POC1"}'
   ```
   -> Observed vulnerable output: HTTP 200 with `{"result":"success","tenant":{"name":"Niro Tenant Alpha POC1","role":"normal",...}}`.

3. Control — read the workspace as an owner:
   ```bash
   curl -i -X POST "$TARGET/console/api/workspaces/current" \
     -H 'Content-Type: application/json' \
     -H 'Authorization: Bearer <owner-access-token>' \
     -H 'X-CSRF-Token: <owner-csrf-token>' \
     --cookie 'access_token=<owner-access-token>; refresh_token=<owner-refresh-token>; csrf_token=<owner-csrf-token>' \
     --data '{}'
   ```
   -> Observed owner output showed the shared workspace name changed to `Niro Tenant Alpha POC1`.

Expected secure: the normal member receives 403/permission denial and the workspace name remains unchanged; an owner can still rename the workspace.

## What changed
- Added the existing `is_admin_or_owner_required` guard to `WorkspaceInfoApi.post` before tenant lookup and mutation.
[...]

PR #39306, opened 2026-07-20
fix: restrict workspace renames to admins and owners

### LINKED ISSUE #38364: Enhancement: restrict workspace renames to admins and owners
## Summary

Workspace names are tenant-wide settings. The rename action should align with workspace administration permissions and be available only to workspace owners and admins.

## Expected behavior

- Workspace owners and admins can update the workspace name from settings.
- Normal workspace members can view workspace information but cannot rename the workspace.
- The API returns a permission denial when the requester does not have the required role.

## Proposed implementation

Add the existing admin/owner role guard to the console workspace info update endpoint and include regression coverage for both normal-member denial and owner/admin success.

Related PR: #38149

ArslanRasheed60:
Fixes #38364

## Summary

`POST /workspaces/info` renames the current workspace, but its decorator stack had no role guard — only `setup_required`, `login_required`, and `account_initialization_required`. Any member of a workspace, including `normal` and `dataset_operator` roles, could rename it for the whole tenant.

This applies the existing `is_admin_or_owner_required` decorator to `WorkspaceInfoApi.post`, so renaming is limited to admins and owners, and documents the resulting 403 in the console OpenAPI contract.

This is the same approach as the earlier #38149, which was closed unmerged. That PR looked complete and its only failing check was `Style Check / Python Style`. Since then `main` has refactored this endpoint to use `@with_session` / `TenantService.get_tenant_by_id`, so this is re-implemented against current `main` rather than a revival of that branch, with the lint clean this time.

The decorator is placed after `account_initialization_required` so authentication and account initialization still resolve first, and only then authorization — matching the ordering in #38149.

Changes:
- `api/controllers/console/workspace/workspace.py` — import and apply `is_admin_or_owner_required`, declare the 403 response
- `api/openapi/markdown/console-openapi.md` — document the 403
- `packages/contracts/generated/api/console/workspaces/types.gen.ts` — generated contract for the 403
- `api/tests/unit_tests/controllers/console/workspace/test_workspace.py` — tests

## Screenshots

N/A — backend authorization change, no UI surface.

## Testing

Added parametrized tests covering both directions:
- `normal` and `dataset_operator` are rejected with `Forbidden`, and the workspace is left untouched (`get_tenant_by_id` never called, `session.commit` never called, name unchanged)
- `owner` and `admin` can still rename successfully, and the change is committed

The positive case matters as much as the denial here: a guard that silently broke legitimate renames would be worse than the original bug.

A note on test scope, to be upfront: these tests re-apply the guard over the unwrapped handler to isolate its allow/deny semantics from the rest of the decorator stack, which needs a database to exercise. They verify the guard behaviour, not the wiring; the wiring is the decorator addition visible in the diff. I could not run the `controllers` unit tests locally — pytest on `api/tests/unit_tests/controllers/**` crashes with a native access violation on my Windows setup, unrelated to this change — so I have relied on CI for the full suite. `ruff check` and `ruff format --check` both pass cleanly against the whole `api` tree.

## Checklist

- [x] Lint: `make` isn't available on my machine, so I ran the underlying backend commands directly — `ruff check ./api` (All checks passed) and `ruff format --check ./api` (3150 files already formatted). I did not run the frontend `vp staged` step, as this change is backend-only.

From Claude Code


rkfshakti (approved):
Clean fix. The `@is_admin_or_owner_required` decorator is exactly the right guard for workspace rename — it's already used elsewhere in the codebase for similar privileged operations, so this is consistent with existing patterns.

The test coverage is thorough:
- Parametrized over all four roles (NORMAL, DATASET_OPERATOR, ADMIN, OWNER)
- Verifies both the allow and deny paths
- Asserts the tenant name is unchanged on denial (no partial update leak)
- Uses `unwrapped` to isolate the decorator from the rest of the stack — clever approach

One small thing: the `mock_get_tenant.assert_not_called()` assertion is a nice safety check that the guard fires before any business logic runs. Good attention to detail.

LGTM.

rkfshakti (changes requested):
Please review and share feedback. I have fixed them.

rkfshakti (approved):
I have fixed them, please review to share feedback, if any,

ArslanRasheed60:
cc @rkfshakti — thanks again for the review and approval! Just checking in — is there anything else needed on my end, or is this ready to be merged whenever a maintainer with merge rights gets a chance?

PR #39310, opened 2026-07-20
fix(account): show validation error message in reset-password/reset-email CLI

sergioperezcheco:
The reset-password and reset-email click commands wrapped valid_password / email_validate in a bare except: that discarded the ValueError both helpers raise. valid_password raises "Password must contain letters and numbers, and the length must be at least 8 characters." and email_validate raises "<addr> is not a valid email.", but operators only ever saw a raw regex pattern or a generic "Invalid email" line, with no indication of which rule was actually violated.

This catches ValueError specifically and echoes str(e), so the actionable reason is shown to whoever is running the command. Narrowing away from the bare except also lets KeyboardInterrupt / SystemExit propagate normally instead of being silently swallowed.

password_pattern is removed from the import since nothing in the file references it anymore.

Verified with ruff check / ruff format (both clean) and the pre-commit hooks pass. These two commands had no existing tests and I did not add coverage for a two-line error-message change, but happy to add a CliRunner test if the maintainers prefer.

### DIFF
--- a/api/commands/account.py
+++ b/api/commands/account.py
@@ -7,7 +7,7 @@ from sqlalchemy.orm import Session
 from constants.languages import languages
 from extensions.ext_database import db
 from libs.helper import email as email_validate
-from libs.password import hash_password, password_pattern, valid_password
+from libs.password import hash_password, valid_password
 from services.account_service import AccountService, RegisterService, TenantService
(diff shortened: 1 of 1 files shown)


crazywoola:

PR #39330, opened 2026-07-21
feat: gate agent management behind the agent.manage permission

GareArc:
Enforces the new `agent.manage` workspace permission across the API and web console (ESQ1-170).

## Changes

**API**
- Agent roster management endpoints require `agent.manage`.
- Draft build endpoints stay ungated: every handler already scopes to `account_id=current_user.id`, so drafts are per-user and the workflow panel calls exactly these endpoints.

**Web**
- `AgentsAccessGuard` redirects to `/apps` when the workspace lacks `agent.manage`, holding a loading state while workspace or permission keys resolve.
- Main-nav roster entry hidden without the permission. Route visibility refactored from a string-union if-ladder to explicit predicates, which removes an unlabelled fallthrough on `canUseAppDeploy`.
- `EditInConsoleLink` renders a real `Button` from the design system, and a disabled button with an explanatory tooltip when the user lacks the permission (previously a hand-rolled anchor).

**Tests**
- New guard spec; nav specs cover both permission states; console-link spec asserts on accessible roles.


web/app/(commonLayout)/@detailSidebar/agents/layout.tsx
lyzno1, line 1: Unnecessary files need to be deleted.

web/app/(commonLayout)/agents/agents-access-guard.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const currentWorkspaceId = useAtomValue(currentWorkspaceIdAtom)
+  const isLoadingCurrentWorkspace = useAtomValue(currentWorkspaceLoadingAtom)
+  const isLoadingWorkspacePermissionKeys = useAtomValue(workspacePermissionKeysLoadingAtom)
+  const canManageAgents = useCanManageAgents()
+  const router = useRouter()
+  const isLoadingAccess = isLoadingCurrentWorkspace || !!isLoadingWorkspacePermissionKeys
+  const shouldRedirect = !isLoadingAccess && !!currentWorkspaceId && !canManageAgents
+
+  useEffect(() => {
+    if (shouldRedirect) router.replace('/apps')
```
lyzno1:
```suggestion
    if (shouldRedirect) router.replace('/')
```

web/app/components/workflow/nodes/agent-v2/components/edit-in-console-link.tsx
lyzno1, line 1: https://base-ui.com/react/components/button#rendering-links-as-buttons
This is an anti-pattern. Do not render the link as a button. Use `<a>` instead.

web/app/(commonLayout)/agents/agents-access-guard.tsx
lyzno1, line 1: use named export instead of default export.

lyzno1 (commented):
Could you clarify whether enterprise_enabled is only intended to distinguish CE from EE when rendering CommunityEditionTip?
The frontend currently derives the deployment type from NEXT_PUBLIC_EDITION:
IS_CE_EDITION = EDITION === 'SELF_HOSTED'
IS_CLOUD_EDITION = EDITION === 'CLOUD'
IS_CE_EDITION effectively means self-hosted, so it currently includes both CE and EE.
If the missing piece is simply the EE distinction, could we extend the existing frontend edition/env model with IS_ENTERPRISE_EDITION instead of adding enterprise_enabled to the shared SystemFeatureModel? The latter expands both the Console and Web API contracts. Additionally, Cloud uses the local cloudSystemFeatures, which currently inherits enterprise_enabled: false, so Cloud may incorrectly render a Community Edition warning.
I suggest distinguishing CE, EE, and Cloud at the deployment-env layer and letting CommunityEditionTip consume that edition information directly.

api/controllers/console/agent/roster.py
```diff
@@ -542,6 +542,7 @@ class AgentAppListApi(Resource):
     @setup_required
     @login_required
     @account_initialization_required
+    @rbac_permission_required(RBACResourceScope.WORKSPACE, RBACPermission.AGENT_MANAGE, resource_required=False)
```
zyssyz123, line 555: The workspace gate is only applied to handlers in this module. `PUT /agent/{agent_id}/composer` in `api/controllers/console/agent/composer.py` still checks only `app_edit`, although it writes the shared normal draft with `account_id=None`. The workflow and snippet `.../agent-composer/save-to-roster` endpoints also create roster Agents while checking only `app_edit` or `snippets.create_and_modify`. A caller without `agent.manage` can therefore perform these management operations directly through the API. These mutations need the same workspace-level gate.

web/app/components/workflow/nodes/agent-v2/components/agent-roster-field.tsx
```diff
@@ -122,6 +122,7 @@ function AgentRosterDrawer({
   onClose: () => void
 }) {
   const { t } = useTranslation()
+  const canManageAgents = useCanManageAgents()
```
zyssyz123, line 125: `canManageAgents` is only applied to `EditInConsoleLink`. The adjacent Make Copy control remains enabled, and `showInlineActions` still exposes Save to Roster whenever `onSaveInlineToRoster` is present. For a user without `agent.manage`, Make Copy now ends in a backend 403, while Save to Roster reaches the currently ungated API path. These controls need to use the same permission.

web/docker/entrypoint.sh
```diff
@@ -14,6 +14,7 @@ set -e
 
 export NEXT_PUBLIC_DEPLOY_ENV=${DEPLOY_ENV}
 export NEXT_PUBLIC_EDITION=${EDITION}
+export NEXT_PUBLIC_ENTERPRISE_ENABLED=${NEXT_PUBLIC_ENTERPRISE_ENABLED:-${ENTERPRISE_ENABLED}}
```
zyssyz123, line 17: The current `dify-enterprise-docker-compose` configuration does not pass either `NEXT_PUBLIC_ENTERPRISE_ENABLED` or `ENTERPRISE_ENABLED` to the web service, and its `envs/core-services/web.env.example` does not define the flag. Existing EE deployments therefore export an empty value here; `emptyStringAsUndefined` plus the schema default makes it `false`, so `IS_COMMUNITY_EDITION` remains `true`. The EE deployment configuration must provide this flag as part of the same rollout, or this value must be derived from an environment signal already available in the EE web container.

zyssyz123 (commented):
The current head has three remaining gaps in server-side enforcement, workflow-node controls, and Enterprise Edition runtime configuration. Details are inline.

web/docker/entrypoint.sh
GareArc: This config should be hidden to CE users. It will be defined in the EE docker compose release instead.

api/controllers/console/agent/composer.py
```diff
@@ -118,7 +118,7 @@ class WorkflowAgentComposerCopyFromRosterApi(Resource):
     @login_required
     @account_initialization_required
     @edit_permission_required
-    @rbac_permission_required(RBACResourceScope.APP, RBACPermission.APP_EDIT)
+    @rbac_permission_required(RBACResourceScope.WORKSPACE, RBACPermission.AGENT_MANAGE, resource_required=False)
```
zyssyz123: [P1] Keep the source-resource permission when adding this workspace gate. This handler mutates the workflow draft, but replacing `APP_EDIT` means that, with RBAC enabled, a role that has `agent.manage` and no edit access to this app can still write to it (`edit_permission_required` is a no-op in that mode). The same replacement was made on save-to-roster and on both snippet variants. Operations that require Agent management need `agent.manage` in addition to the existing `app_edit` / `snippets.create_and_modify` check, not instead of it.

api/controllers/console/agent/roster.py
zyssyz123, line 555: [P1] This does not yet establish a server-side boundary for Agent management because the same Agent App remains mutable through the generic App APIs. `GET /agent/{agent_id}` exposes `app_id`; `/apps/{app_id}`, `/name`, `/icon`, `/api-enable`, and `/api-keys` all accept `mode=agent`, and `AppService` synchronizes or archives the backing roster Agent. `/apps/{app_id}/copy` and `/apps/imports` can also create roster Agents through the Agent DSL path. Those routes only check App permissions, so a caller without `agent.manage` can bypass this gate. Please add Agent-aware enforcement to the generic management/import paths (while preserving workflow-only hidden backing-app behavior), or reject roster Agent Apps from those legacy paths.

api/services/agent/roster_service.py
```diff
@@ -893,6 +891,32 @@ def get_agent_runtime_app_model(self, *, tenant_id: str, agent_id: str) -> App:
...
+        stay read-only: unlike :meth:`get_agent_runtime_app_model`, this never
+        materializes the hidden backing App for a workflow-only Agent. Returns
+        ``None`` when the Agent does not resolve, leaving the caller to decide
+        how to treat an unresolvable Agent.
+        """
+
+        agent = self._get_runtime_resolvable_agent(tenant_id=tenant_id, agent_id=agent_id)
+        if agent is None:
+            return None
+        return self.runtime_backing_app_id(agent)
```
zyssyz123: [P1] A workflow-only Agent must be authorized against its parent workflow App, not its hidden runtime App. `create_hidden_backing_app_for_workflow_agent` inserts the backing App directly and never initializes a resource access policy, so the RBAC service evaluates workspace defaults for this ID instead of any per-resource override on `agent.app_id`. A user excluded from the workflow can therefore pass `app_test_and_run` through the default policy (and a user explicitly granted on the workflow can be denied). For `AgentScope.WORKFLOW_ONLY`, this resolver should return the parent `agent.app_id`; `backing_app_id` should remain runtime-only. The new test that prefers the hidden App currently codifies the same authorization mismatch.

lyzno1 (dismissed):
Frontend changes LGTM.

lyzno1 (approved):

zyssyz123 (approved):

PR #39334, opened 2026-07-21
feat: make enterprise inner API timeout configurable

GareArc:
## Summary

Requests to the enterprise inner API relied on httpx's implicit 5s default — no call site passed `timeout`, and no env or config could change it.

On tenants with enough installed apps, `POST /webapp/permission/batch` exceeds that budget and `/console/api/installed-apps` fails with `httpx.ReadTimeout`, so the Explore page cannot render.

Adds `ENTERPRISE_API_TIMEOUT` (`EnterpriseFeatureConfig`, `PositiveFloat`, default `90.0`) and passes it to `client.request` in `BaseRequest.send_request`.

## Why 90s

The enterprise server enforces its own request timeout (`SERVER_TIMEOUT`, default `60s`). Defaulting the client above that means the server terminates a runaway request first and returns a clean 504, instead of the two timers racing.

Note this raises the effective budget for **all** enterprise inner-API calls, not just the batch permission one — a stalled enterprise pod now holds a gunicorn worker longer than it used to.

Ref: CUS-1422

### DIFF
--- a/api/configs/enterprise/__init__.py
+++ b/api/configs/enterprise/__init__.py
@@ -1,4 +1,4 @@
-from pydantic import Field
+from pydantic import Field, PositiveFloat
 from pydantic_settings import BaseSettings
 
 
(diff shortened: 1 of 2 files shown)


GareArc:
Superseded by #39337. This branch sits under `hotfix/**`, which the `hotfix` ruleset protects behind required PRs, so it could not be updated with the env var rename or the CI fix. Reopened from `fix/cus-1422-enterprise-api-timeout-e-1114`.

PR #39338, opened 2026-07-21
fix(api): keep workflow log archives owner-admin only

lyzno1:
## Summary

- enforce fixed workspace owner/admin access for all workflow log archive endpoints
- remove the unrelated `workspace.role.manage` RBAC gate from this Cloud-only capability
- cover owner/admin and non-manager roles with RBAC enabled and disabled, plus verify every endpoint invokes the fixed-role guard

## Root cause

Workflow log archives were introduced with both `is_admin_or_owner_required` and `workspace.role.manage`. When enterprise RBAC is enabled, the legacy manager guard becomes a no-op, so a custom role with role-management permission can read and export workflow archives while an owner/admin without that permission can be rejected. This contradicts the Cloud UI contract, which exposes archives only to workspace owners and admins.

## Impact

Archive listing, task creation, task polling, and file download now use the persisted workspace role directly. Access remains owner/admin-only regardless of the enterprise RBAC flag. Cloud edition, billing-enabled, and paid-plan entitlement checks are unchanged.

## Validation

- `uv run --project api pytest -q api/tests/unit_tests/controllers/console/test_workflow_run_archive.py` — 18 passed
- `uv run --project api ruff format --check api/controllers/console/workflow_run_archive.py api/tests/unit_tests/controllers/console/test_workflow_run_archive.py`
- `uv run --project api ruff check api/controllers/console/workflow_run_archive.py api/tests/unit_tests/controllers/console/test_workflow_run_archive.py`
- `uv run --project api pyrefly check api/controllers/console/workflow_run_archive.py api/tests/unit_tests/controllers/console/test_workflow_run_archive.py`

### DIFF
--- a/api/controllers/console/workflow_run_archive.py
+++ b/api/controllers/console/workflow_run_archive.py
@@ -4,20 +4,16 @@ from http import HTTPStatus
 from flask import redirect
 from flask_restx import Resource
 from pydantic import BaseModel, Field
-from werkzeug.exceptions import Conflict, NotFound
+from werkzeug.exceptions import Conflict, Forbidden, NotFound
 
 from controllers.common.fields import RedirectResponse
 from controllers.common.schema import register_response_schema_models, register_schema_models
 from controllers.console import console_ns
 from controllers.console.wraps import (
-    RBACPermission,
-    RBACResourceScope,
     account_initialization_required,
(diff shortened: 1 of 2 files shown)


hjlarry (approved):
LGTM

PR #39364, opened 2026-07-21
fix(api): HTTP Node BUG

chogenfran:
## Summary

HTTP Request nodes can fail during workflow draft/debug runs when a large response body is offloaded to file storage. The failure is not in the HTTP call itself; it happens while persisting the oversized draft variable.

Postgres rejects the insert because `upload_files.tenant_id` is written as an empty string:

```text
invalid input syntax for type uuid: ""
INSERT INTO upload_files (id, tenant_id, ...)
parameters: tenant_id='', key='upload_files//....txt'
```

Root cause: `_try_offload_large_variable()` previously derived tenant context from `Account.current_tenant_id`. In several workflow execution paths (debugger / background worker contexts), that value is unset. `FileService.upload_file()` then fell back to `tenant_id=""`, producing invalid storage keys (`upload_files//...`) and an invalid UUID write.

## Fix

1. Resolve tenant from the app record (`App.tenant_id`) via `_resolve_app_tenant_id()` when offloading large draft variables.
2. Pass that tenant explicitly into `FileService.upload_file(...)`.
3. Persist the same tenant on `WorkflowDraftVariableFile`.
4. Make `FileService.upload_file()` reject empty/missing tenant IDs instead of writing `""` into a UUID column.

## Changed files

- `api/services/workflow_draft_variable_service.py`
  - large-variable offload now uses app tenant, not account current-tenant state
- `api/services/file_service.py`
  - upload path requires a non-empty `tenant_id`
- `api/tests/unit_tests/services/workflow/test_workflow_draft_variable_service.py`
  - covers offload when `current_tenant` is missing
- `api/tests/unit_tests/services/test_file_service.py`
  - covers explicit tenant usage and empty-tenant rejection
- `docs/http-request-node-empty-tenant-id-hotfix.md`
  - incident notes / deploy notes for the hotfix image

## Why this is safe

- Normal uploads that already pass a valid `tenant_id` are unchanged.
- Draft offload no longer depends on mutable account tenant state.
- Empty tenant now fails early with a clear application error, instead of producing corrupted storage keys and a database type error.

## Test plan

- [x] `test_upload_file_requires_tenant_id`
- [x] `test_upload_file_uses_explicit_resource_tenant`
- [x] `test_try_offload_large_variable_uses_app_tenant_when_current_tenant_missing`

## Deploy notes

Only the API image needs updating. In `docker/docker-compose.yaml`, the following services share `langgenius/dify-api`:

- `api`
- `worker`
- `worker_beat`
- `api_websocket` (if collaboration profile is enabled)

No changes are required for `web`, `sandbox`, `plugin_daemon`, or `agent_backend`.

### DIFF
--- a/api/services/file_service.py
+++ b/api/services/file_service.py
@@ -86,15 +86,17 @@ class FileService:
         file_uuid = str(uuid.uuid4())
 
         resource_tenant_id = tenant_id if tenant_id is not None else extract_tenant_id(user)
+        if not resource_tenant_id:
+            raise ValueError("tenant_id is required to upload a file")
 
-        file_key = "upload_files/" + (resource_tenant_id or "") + "/" + file_uuid + "." + extension
+        file_key = "upload_files/" + resource_tenant_id + "/" + file_uuid + "." + extension
 
         # save file to storage
         storage.save(file_key, content)
 
         # save file to db
         upload_file = UploadFile(
-            tenant_id=resource_tenant_id or "",
+            tenant_id=resource_tenant_id,
             storage_type=StorageType(dify_config.STORAGE_TYPE),
             key=file_key,
             name=filename,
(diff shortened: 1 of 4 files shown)


41tair:
This has been fixed in #39307.

PR #39365, opened 2026-07-21
fix(logging): prevent gevent handler-lock wedge that hangs requests

MRZHUH:
## Problem

Under gevent (`GeventWebSocketWorker`), a `logging.Handler`'s lock is a gevent `BoundedSemaphore`. `Handler.handle()` acquires it around `emit()`. When a request greenlet is parked in that `acquire()` and a gevent **threadpool worker** (a different hub — e.g. gevent's DNS resolver, hit on every redis/db/plugin call) holds and releases the same semaphore, the cross-hub wakeup can be lost and the greenlet **sleeps forever**.

The request then hangs with **HTTP 200 and an empty body** — no exception, no timeout, and invisible to py-spy (a parked greenlet holds no OS thread). It is intermittent (~30% under load) and shows up as a chat/workflow request that returns 200 but never produces an answer.

Captured stack of a wedged greenlet:

```
geventwebsocket/handler.py  log_request -> self.logger.info(...)
logging/__init__.py         handle -> callHandlers -> Handler.handle -> self.acquire()
                            -> self.lock.acquire() -> threading._RLock._block.acquire()
gevent/thread.py            BoundedSemaphore.acquire   <- parked here forever
```

## Fix

Route every record through a lock-free `queue.SimpleQueue` and emit from a single `QueueListener`:

- The root logger now holds one `_LockFreeQueueHandler` whose `handle()` does **not** take the handler lock — `SimpleQueue.put_nowait` is a C-atomic, so the lock is unnecessary. Request greenlets (and threadpool workers) can no longer wedge on a semaphore they never acquire.
- The real file/console handlers run only on the listener, which is started on a **real OS thread** (`SimpleQueue.get` is a blocking C call and would freeze the hub if it ran in a greenlet under monkey-patch).
- Context filters move onto the queue handler, so `IdentityContextFilter`'s `end_users` lookup now runs once per record instead of once per output handler.

## Validation

- The wedge is at `Handler.acquire()`; the request path no longer calls it — the fix removes the exact line by construction.
- Under a synthetic lock-contention harness, the old path lets greenlets emit **90** logs while the fixed path emits **2579** (28x) over the same window, because the fixed greenlets never touch the contended lock.
- The permanent wedge is a rare intermittent race and is hard to force synthetically. Suggest verifying on a canary via traces: `200` responses with no terminal workflow span should drop to 0.

## Notes

- Adds one daemon listener thread + a continuously-drained in-memory queue (the standard stdlib `QueueHandler`/`QueueListener` pattern).
- Behaviour is unchanged when not running under gevent (the listener falls back to a normal `threading.Thread`).

### DIFF
--- a/api/extensions/ext_logging.py
+++ b/api/extensions/ext_logging.py
@@ -1,25 +1,137 @@
 """Logging extension for Dify Flask application."""
 
+import atexit
+import contextlib
 import logging
 import os
+import queue
 import sys
-from logging.handlers import RotatingFileHandler
+import threading
+from logging import LogRecord
+from logging.handlers import QueueHandler, RotatingFileHandler
 from typing import override
 
 from configs import dify_config
 from dify_app import DifyApp
 
+# Keep the listener (and its thread) referenced for the process lifetime so it
+# is never garbage-collected while the root logger still feeds its queue.
+_log_listener: "_QueueLogListener | None" = None
+
+
+def _spawn_real_thread(target, name: str) -> threading.Thread:
+    """Start a **real OS thread**, even under gevent monkey-patching.
+
+    gevent's ``monkey.patch_all()`` turns ``threading.Thread`` into a
+    greenlet-backed thread. The logging queue listener below blocks on
+    ``queue.SimpleQueue.get()`` -- a C-level call that is NOT cooperative -- so
+    running it on a greenlet would freeze the whole gevent hub. We therefore
+    resolve the original (unpatched) ``Thread`` and run the listener there.
+    """
+    # If gevent is unavailable or threading is unpatched, threading.Thread is
+    # already a real OS thread and the suppressed block simply leaves it alone.
+    thread_cls = threading.Thread
+    with contextlib.suppress(Exception):
+        from gevent import monkey
+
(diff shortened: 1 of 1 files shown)


api/extensions/ext_logging.py
```diff
@@ -1,25 +1,137 @@
...
+    resolve the original (unpatched) ``Thread`` and run the listener there.
+    """
+    # If gevent is unavailable or threading is unpatched, threading.Thread is
+    # already a real OS thread and the suppressed block simply leaves it alone.
+    thread_cls = threading.Thread
+    with contextlib.suppress(Exception):
+        from gevent import monkey
+
+        if monkey.is_module_patched("threading"):
+            thread_cls = monkey.get_original("threading", "Thread")
```
QuantumGhost, line 38: This does not launch a native thread. You should use `gevent.threadpool.ThreadPool(1)` instead.

QuantumGhost:
Besides, SimpleQueue is indeed patched: 

- https://www.gevent.org/changelog.html#id18
- https://github.com/gevent/gevent/pull/1253

PR #39376, opened 2026-07-21
fix(rag): use floor division for vector embedding batch count

sergioperezcheco:
The total_batches counter in Vector.create and Vector.create_multimodal was computed as len(texts) + batch_size - 1, which is missing the floor-division step. With a batch size of 1000, embedding 1001 documents logged "Processing batch 1/2000" instead of "Processing batch 1/2", so the progress denominator was off by ~batch_size. This only affects the info-level progress logs (the actual batching loop uses the correct range step), but it makes the logs misleading when debugging long embedding runs.

Changed both the text and multimodal paths to use ceiling division: (len(items) + batch_size - 1) // batch_size. Added a regression test that embeds 1001 docs and asserts the logged batch denominator is 2, not 2000 (verified it fails on the old formula).

### DIFF
--- a/api/core/rag/datasource/vdb/vector_factory.py
+++ b/api/core/rag/datasource/vdb/vector_factory.py
@@ -169,7 +169,7 @@ class Vector:
             start = time.time()
             logger.info("start embedding %s texts %s", len(texts), start)
             batch_size = 1000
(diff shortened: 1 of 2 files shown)


crazywoola:

Issue #38855, opened 2026-07-13
Service API audit: six inconsistencies in error handling and event emission


RiskeyL:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

main (`b03385ffb0`); the same behavior was verified on 1.15.0 during the docs team's Service API audit

### Cloud or Self Hosted

Cloud, Self Hosted (Docker)

### Steps to reproduce

While rewriting the Service API reference, the docs team code-verified every endpoint and found six backend inconsistencies. Each item below lists the trigger, the code location, and expected vs. actual. They are independent; happy to split into separate issues if that helps triage.

**1. Renaming a missing metadata field returns 400 instead of 404**

`PATCH /v1/datasets/{dataset_id}/metadata/{metadata_id}` with a nonexistent `metadata_id`.

`MetadataService.update_metadata_name` raises a bare `ValueError("Metadata not found.")` (`api/services/metadata_service.py`), which the Service API renders as `400 invalid_param`. The same request with a nonexistent `dataset_id` correctly returns `404 not_found`.

Expected: `404 not_found`. Actual: `400 invalid_param` with message `Metadata not found.`

**2. Built-in metadata toggle can report success without doing anything**

`POST /v1/datasets/{dataset_id}/metadata/built-in/{action}` (`DatasetMetadataBuiltInFieldActionServiceApi.post`, `api/controllers/service_api/dataset/metadata.py`). Two facets:

- The `action` parameter is typed `Literal["enable", "disable"]`, but the route accepts any string and the `match` has no default case, so e.g. `POST .../built-in/Enable` returns `{"result": "success"}` having done nothing.
- `enable_built_in_field` / `disable_built_in_field` (`api/services/metadata_service.py`) wrap their whole body in `except Exception: logger.exception(...)`, so a failed toggle (for example lock contention) also returns success.

[...]

PR #39383, opened 2026-07-22
fix(api): map InvokeRateLimitError to 429 in CompletionApi

sergioperezcheco:
POST /v1/completion-messages returns 500 internal_server_error when the workspace rate limit is hit, while POST /v1/chat-messages correctly returns 429 too_many_requests for the same condition. The only difference is that ChatApi.post catches InvokeRateLimitError and maps it via InvokeRateLimitHttpError, but CompletionApi.post is missing that handler and the exception falls through to the generic 500 catch-all.

This adds the same except InvokeRateLimitError clause to CompletionApi.post so both endpoints behave consistently. Both imports (InvokeRateLimitError from services.errors.llm and InvokeRateLimitHttpError from controllers.web.error) are already present in the file.

Identified as item 3 in the Service API audit issue #38855.

### DIFF
--- a/api/controllers/service_api/app/completion.py
+++ b/api/controllers/service_api/app/completion.py
@@ -260,6 +260,8 @@ class CompletionApi(Resource):
             raise ProviderQuotaExceededError()
         except ModelCurrentlyNotSupportError:
             raise ProviderModelCurrentlyNotSupportError()
+        except InvokeRateLimitError as ex:
+            raise InvokeRateLimitHttpError(ex.description)
         except InvokeError as e:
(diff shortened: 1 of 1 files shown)


crazywoola:

### Why this is being closed

The linked issue set has multiple open pull requests. Under the duplicate-PR rule, pull requests from the issue author are ranked first, then creation time is used as the FIFO tie-breaker. The retained result for this overlap is: issue #38855 -> PR #38871. This pull request is therefore being closed as a remaining duplicate.

### Next steps

PR #39396, opened 2026-07-22
fix(service-api): map completion-messages rate limit to 429

sergioperezcheco:
POST /v1/completion-messages currently returns 500 internal_server_error when the workspace rate limit is hit, while the sibling POST /v1/chat-messages endpoint correctly returns 429 rate_limit_error. The root cause is a missing exception handler: ChatApi.post catches InvokeRateLimitError and maps it to the HTTP error, but CompletionApi.post does not, so the same condition falls through to the generic 500 path.

This adds the identical except clause to CompletionApi.post so both generate endpoints surface the rate-limit condition consistently as 429. The required imports (InvokeRateLimitError from services.errors.llm and InvokeRateLimitHttpError from controllers.web.error) were already in place.

I also added a regression test (test_completion_api_post_rate_limit_maps_to_429) that mocks AppGenerateService.generate to raise InvokeRateLimitError and asserts CompletionApi.post re-raises it as the 429 HTTP error, mirroring the existing test for the chat endpoint.

This corresponds to item 3 (Rate limit surfaces as 500 on completion-messages but 429 on chat-messages) of the Service API audit reported in #38855.

### DIFF
--- a/api/controllers/service_api/app/completion.py
+++ b/api/controllers/service_api/app/completion.py
@@ -260,6 +260,8 @@ class CompletionApi(Resource):
             raise ProviderQuotaExceededError()
         except ModelCurrentlyNotSupportError:
             raise ProviderModelCurrentlyNotSupportError()
+        except InvokeRateLimitError as ex:
+            raise InvokeRateLimitHttpError(ex.description)
         except InvokeError as e:
             raise CompletionRequestError(e.description)
         except ValueError as e:
--- a/api/tests/unit_tests/controllers/service_api/app/test_completion.py
+++ b/api/tests/unit_tests/controllers/service_api/app/test_completion.py
@@ -431,6 +431,29 @@ class TestCompletionControllerLogic:
 
     @patch("controllers.service_api.app.completion.service_api_ns")
     @patch("controllers.service_api.app.completion.AppGenerateService")
+    def test_completion_api_post_rate_limit_maps_to_429(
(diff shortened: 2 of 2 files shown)


sergioperezcheco:
The Web Full-Stack E2E failure looks unrelated to this change. The failing assertions are frontend ZodError validation errors ("Invalid option: expected one of \"chat\"|\"completion\"") coming from the agent-v2 e2e specs, and the E2E workflow has been failing consistently on its own feature branch (feat/agent-v2-e2e-test) for weeks. This PR only touches the Python service API error-mapping path and all backend tests pass (API unit, integration, and coverage checks are green). If a maintainer could re-run or the E2E suite stabilizes upstream, this should clear.

sergioperezcheco:
The Web Full-Stack E2E failure looks unrelated to this change — it's failing on an orchestration-status E2E case that also fails on the base branch and on the feat/agent-v2-e2e-test branch. The API integration and unit tests (which actually exercise the rate-limit path) both pass. Happy to rebase if that helps, but I don't think there's anything in this diff that touches the web E2E flow.

crazywoola:

### Why this is being closed

The linked issue set has multiple open pull requests. Under the duplicate-PR rule, pull requests from the issue author are ranked first, then creation time is used as the FIFO tie-breaker. The retained result for this overlap is: issue #38855 -> PR #38871. This pull request is therefore being closed as a remaining duplicate.

### Next steps

PR #39405, opened 2026-07-22
fix(agent): isolate build and preview chat conversations

zyssyz123:
## Summary

- isolate Agent Preview (`draft`) and Build (`debug_build`) chat conversations per editor
- scope persisted conversation mappings and runtime sessions through distinct conversation IDs
- keep Agent detail/list and no-body refresh backward-compatible with the Build conversation
- allow explicit Preview refresh with `{"draft_type":"draft"}`
- document the refresh request body as optional while registering its OpenAPI schema for generated clients

## Root cause

`agent_debug_conversations` was unique only by tenant, Agent, and account. Both chat surfaces therefore resolved to the same conversation even though they loaded different config drafts. Messages, history, and conversation-owned runtime sessions crossed between Preview and Build.

## Database

- add non-null `draft_type` to `agent_debug_conversations`
- backfill existing mappings as `debug_build`, matching the existing Agent detail contract
- replace the old unique constraint with `(tenant_id, agent_id, account_id, draft_type)`
- include a tested downgrade that consolidates scoped rows before restoring the legacy constraint

## Verification

- `372 passed` across Agent controllers/services, Agent App runtime, session store, cleanup, and request builder tests
- `78 passed` for the final Agent controller and optional-body OpenAPI regression suite
- Ruff, Ruff format, Mypy, Alembic single-head, diff checks, and `pnpm -w check` passed
- PostgreSQL migration validated with upgrade, scoped uniqueness, downgrade, and legacy-row re-upgrade/backfill
- deployed final backend commit `c3cc4efefd` to `deploy/agent`; API, workers, priority worker, and Agent backend all ran that revision
- dev E2E on `agent.dify.dev`: optional request body and resolved request schema in runtime OpenAPI, separate Preview/Build conversation IDs, Agent detail bound to Build, explicit Preview refresh, legacy no-body Build refresh, and two cross-scope conversation requests rejected with 404
- earlier full SSE regression also verified separate histories, correct draft/build-draft config routing, independent refresh behavior, and runtime cleanup task completion
- all GitHub checks passed, including Autofix, API unit/integration, PostgreSQL/MySQL migration, Web full-stack E2E, type/style, and CodeQL

Linear: DIFY-2825

### DIFF
--- a/api/controllers/console/agent/roster.py
+++ b/api/controllers/console/agent/roster.py
@@ -62,7 +62,7 @@ from libs.datetime_utils import parse_time_range
 from libs.helper import dump_response
 from libs.login import login_required
 from models import Account
-from models.agent import Agent, AgentStatus
+from models.agent import Agent, AgentConfigDraftType, AgentStatus
 from models.agent_config_entities import AgentSoulConfig
 from models.enums import ApiTokenType
 from models.model import ApiToken, App, IconType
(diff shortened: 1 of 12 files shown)


MRZHUH (approved):
LGTM

PR #39413, opened 2026-07-22
fix(core): keep in-progress messages in get_thread_messages_length

### LINKED ISSUE #37880: Bug: get_thread_messages_length excludes in-progress messages with answer_tokens > 0
### Self Checks

### Dify version

main (latest)

### Cloud or Self Hosted

Self-hosted / source-level bug in API.

### Steps to reproduce

1. In a conversation thread, the newest message is still generating: `answer` is empty but `answer_tokens > 0` (streaming/in-progress state).
2. Call `get_thread_messages_length(conversation_id)` from `api/core/prompt/utils/get_thread_messages_length.py`.
3. Compare with `TokenBufferMemory.get_history_prompt_messages()` behavior in `api/core/memory/token_buffer_memory.py`.

Relevant code paths:

```python
# get_thread_messages_length.py (current)
if thread_messages and not thread_messages[0].answer:
    thread_messages.pop(0)

# token_buffer_memory.py (intended pattern)
if thread_messages and not thread_messages[0].answer and thread_messages[0].answer_tokens == 0:
    thread_messages.pop(0)
```

`advanced_chat/app_generator.py` uses `get_thread_messages_length(conversation.id) + 1` for `dialogue_count`.

### Expected Behavior

Only exclude the newest message when it is a freshly created placeholder with **both** an empty `answer` and `answer_tokens == 0`, matching `token_buffer_memory`.

### Actual Behavior

`get_thread_messages_length` excludes any newest message with a falsy `answer`, even when `answer_tokens > 0`. This under-counts thread length and can skew `dialogue_count` during in-progress generations.

sergioperezcheco:
get_thread_messages_length excluded the newest thread message whenever its answer was empty, even during streaming when answer_tokens > 0. This under-counted the thread length and skewed the dialogue_count passed to the agent generator in advanced_chat/app_generator.py.

The guard now also checks answer_tokens == 0, so only a freshly created placeholder message (empty answer and no tokens produced yet) is excluded. This matches the existing guard in TokenBufferMemory.get_history_prompt_messages, which already uses the same two-condition check.

I added a regression test for the in-progress case (empty answer, answer_tokens > 0) alongside the existing placeholder test. Both pass locally along with the rest of the extract_thread_messages suite.

Fixes #37880

The concrete caller that benefits: `AdvancedPromptGenerator.resume()` re-enters `_generate` with an existing message instead of creating one, so `thread_messages[0]` there is a pre-existing in-progress message rather than a fresh placeholder — exactly the case the old one-condition guard dropped, under-counting `dialogue_count` by one on every resume of a run whose answer column was still empty.

### DIFF
--- a/api/core/prompt/utils/get_thread_messages_length.py
+++ b/api/core/prompt/utils/get_thread_messages_length.py
@@ -17,8 +17,10 @@ def get_thread_messages_length(conversation_id: str, *, session: Session) -> int
     # Extract thread messages
     thread_messages = extract_thread_messages(messages)
 
-    # Exclude the newly created message with an empty answer
-    if thread_messages and not thread_messages[0].answer:
+    # Exclude only a freshly created placeholder message: one with an empty
+    # answer AND no produced tokens yet. An in-progress message whose answer is
+    # still streaming (empty answer but answer_tokens > 0) must be kept.
+    if thread_messages and not thread_messages[0].answer and thread_messages[0].answer_tokens == 0:
         thread_messages.pop(0)
 
     return len(thread_messages)
--- a/api/tests/unit_tests/core/prompt/test_extract_thread_messages.py
+++ b/api/tests/unit_tests/core/prompt/test_extract_thread_messages.py
@@ -19,6 +19,7 @@ def _persisted_message(
     parent_message_id: str,
     answer: str,
     created_at: datetime,
+    answer_tokens: int = 0,
 ) -> Message:
     message = Message(
         id=message_id,
@@ -34,16 +35,19 @@ def _persisted_message(
         parent_message_id=parent_message_id,
         created_at=created_at,
         updated_at=created_at,
+        message_tokens=0,
+        answer_tokens=answer_tokens,
     )
     message._inputs = {}
     return message
 
 
 class MockMessage:
-    def __init__(self, id, parent_message_id, answer="answer"):
+    def __init__(self, id, parent_message_id, answer="answer", answer_tokens=1):
         self.id = id
         self.parent_message_id = parent_message_id
         self.answer = answer
+        self.answer_tokens = answer_tokens
 
     def __getitem__(self, item):
         return getattr(self, item)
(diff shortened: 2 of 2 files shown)


chuenchen309:
Disclosure first: I have an open PR (#39032) that touches `extract_thread_messages.py` and adds a case to the same test file, so I have a stake in this area. That's also why I went looking — the two changes are complementary (thread walk vs. length guard), but whichever lands second will probably need a trivial rebase of `test_extract_thread_messages.py`.

The inconsistency you're fixing is real and checkable in two lines on `main`:

```python
# core/prompt/utils/get_thread_messages_length.py
if thread_messages and not thread_messages[0].answer:

# core/memory/token_buffer_memory.py:152
if thread_messages and not thread_messages[0].answer and thread_messages[0].answer_tokens == 0:
```

Same job — drop the freshly created placeholder — applied to the output of the same `extract_thread_messages`, with different conditions. One of the two has to be wrong, and the two-condition form is clearly the intended one.

**Where I think it actually bites, which the description undersells.** On the ordinary path the two guards are indistinguishable: `advanced_chat/app_generator.py` calls `_init_generate_records` and *then* `get_thread_messages_length(...) + 1`, so `thread_messages[0]` is a placeholder this request just created — `answer` empty, `answer_tokens` still 0 — and both versions pop it.

The difference needs a caller that reaches the count without having just created the message, and there is one:

```python
if conversation is not None and message is not None:
    pass
else:
    conversation, message = self._init_generate_records(...)
```

`resume()` ("Resume a paused advanced chat execution") calls `_generate(..., message=message)` with an existing message, so that branch skips record creation and `thread_messages[0]` is a pre-existing message rather than a fresh placeholder. If that message has produced tokens while its `answer` column is still empty, the current one-condition guard drops it and `dialogue_count` comes out one short on every resume. That's a much more concrete story than "during streaming", and it's the scenario I'd put in the test name.

**What I did not verify:** the exact write ordering of `Message.answer` versus `Message.answer_tokens` during a paused/resumed run — they're separate columns (`answer` `nullable=False`, `answer_tokens` defaulting to 0), but I didn't trace which is persisted when, so I can't claim from here that the empty-answer/non-zero-tokens window is actually observed on resume rather than merely representable. If you've seen it in practice, saying so in the description would close the last gap.

*AI-assisted review; the call sites and guards quoted above were read on current `main` rather than inferred.*

sergioperezcheco:
Thanks for the thorough trace — the resume() path is a better story than the one in the description, and you're right that the two-condition form is the intended one (it mirrors TokenBufferMemory.get_history_prompt_messages exactly, which is the guard I aligned with).

On the write ordering: I haven't observed the empty-answer/non-zero-tokens window on resume in practice — the fix is defensive, driven by the inconsistency between the two guards rather than a captured repro. The streaming case (placeholder created, answer empty while tokens accumulate) is the one I could reproduce, but the same condition covers the resume window you describe, since both reduce to "exclude only a fresh placeholder with zero produced tokens."

Agreed that whichever of #39032 and this lands second will need a trivial rebase of test_extract_thread_messages.py — I'll keep an eye on it.

sergioperezcheco:
Thanks for the careful read — the resume() walk-through is exactly the caller I had in mind, and you're right that the description undersold it. I've folded that scenario into the PR description. The regression test already covers the shape you described: empty answer with answer_tokens > 0 is kept, fresh placeholder (answer_tokens == 0) is still dropped, and the suite passes locally.

PR #39195, opened 2026-07-17
fix: handle cleared optional file tool inputs

stevenup:
## Summary

- Reset cleared optional `file` and `files` form inputs to an unset null constant in the workflow editor.
- Ignore legacy empty optional variable selectors at runtime while preserving validation for required and malformed non-empty selectors.
- Add frontend and backend regression coverage for empty-string and empty-list selector states.

Fixes #39045

## Screenshots

Not applicable. This is a behavioral fix with no visual UI changes.

## Test plan

- `uv run --project api pytest api/tests/unit_tests/core/tools/test_tool_manager.py -q` (50 passed)
- `make lint && make type-check`
- `cd web && pnpm test app/components/workflow/nodes/_base/components/__tests__/form-input-item.branches.spec.tsx app/components/workflow/nodes/_base/components/__tests__/form-input-item.spec.tsx` (17 passed)
- `cd web && pnpm exec vp staged --cwd .. --diff upstream/main...HEAD --fail-on-changes`
- `git diff --check upstream/main...HEAD`

## Checklist

- [x] This change does not require a documentation update.
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] Documentation is unchanged because this fix does not alter user-facing configuration or APIs.
- [x] I ran `make lint && make type-check` (backend) and the frontend staged checks for the full PR diff.

PR #39414, opened 2026-07-22
fix(workflow): reset cleared variable selector to constant/null

### LINKED ISSUE #39045: Bug: Clearing an optional file input in a tool node saves an invalid variable value
### Self Checks

### Dify version

1.13.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Use a plugin tool that has an optional `files` form parameter. A minimal parameter definition is:

```yaml
parameters:
  - name: image
    type: files
    required: false
    label:
      en_US: Reference images
    human_description:
      en_US: Optional reference images.
    llm_description: Optional reference images.
    form: form

  - name: image_urls
    type: string
    required: false
    label:
      en_US: Reference image URLs
    human_description:
      en_US: Optional reference image URLs.
    llm_description: Optional reference image URLs.
    form: form
```

1. Create a workflow with a Start node containing an `Array[File]` input.
2. Add the plugin tool node.
3. Leave the optional `image` field untouched and configure only `image_urls`.
4. Run the workflow. It succeeds. The untouched field is stored as:

   ```json
   {"type": "constant", "value": null}
   ```

5. Select `Start / input` for the optional `image` field. It is stored as a valid variable selector and the workflow also succeeds:

   ```json
   {"type": "variable", "value": ["start-node-id", "input"]}
   ```

6. Clear the selected variable from the optional `image` field.
7. Run the workflow again.

After the field is cleared, the workflow draft contains:

```json
{"type": "variable", "value": ""}
```

The tool node fails before the plugin is invoked:

```text
1 validation error for ToolInput
type
  Value error, value must be a list [type=value_error, input_value='variable', input_type=str]
```

Relevant stack trace:

```text
File "/app/api/core/tools/tool_manager.py", line 1042, in _convert_tool_parameters_type
  tool_input = ToolNodeData.ToolInput.model_validate(tool_configurations.get(parameter.name, {}))

pydantic_core._pydantic_core.ValidationError: 1 validation error for ToolInput
type
  Value error, value must be a list [type=value_error, input_value='variable', input_type=str]
```

The frontend variable-clear path is in:

```text
web/app/components/workflow/nodes/_base/components/form-input-item.tsx
```

It keeps `type: variable` while using an empty string as the cleared value. The backend validator in `api/core/workflow/nodes/tool/entities.py` correctly requires a variable selector to be a list of strings, so the frontend-generated state is invalid.

This is not specific to one image-generation plugin. It can affect any tool with an optional `file` or `files` form parameter after a user selects and then clears the variable.

Related issue #23400 contains the same validator message in a different pipeline/parameter scenario, but does not cover this reproducible optional-file clearing behavior.

### ✔️ Expected Behavior

Clearing an optional `file` or `files` parameter should restore it to an unset state, for example by removing the configuration entry or storing:

```json
{"type": "constant", "value": null}
```

The workflow should then run without requiring users to bind an unrelated file variable when they only want to use another optional input such as `image_urls`.

As defense in depth, the backend could also ignore `type: variable` configurations with an empty selector for optional parameters instead of passing them to `ToolInput.model_validate()`.

### ❌ Actual Behavior

Clearing the optional file variable stores `{"type":"variable","value":""}`. Workflow execution then fails immediately with `ToolInput` validation error `value must be a list`, before the plugin invocation endpoint is called.

sergioperezcheco:
When a user selects a variable for an optional file or files form parameter in a tool node and then clears it, VarReferencePicker calls onChange([]) (or onChange('') in some paths). The previous handleVariableSelectorChange kept type=variable with an empty value, which the backend ToolInput validator rejects with 'value must be a list' — crashing the workflow run before the tool is even invoked.

This change makes handleVariableSelectorChange detect an empty selector (empty string or empty array) and reset to {type: constant, value: null}, matching the initial untouched state. A helper test covers the empty-selector detection.

Closes #39045

### DIFF
--- a/web/app/components/workflow/nodes/_base/components/__tests__/form-input-item.helpers.spec.ts
+++ b/web/app/components/workflow/nodes/_base/components/__tests__/form-input-item.helpers.spec.ts
@@ -157,6 +157,20 @@ describe('form-input-item helpers', () => {
     expect(normalizeVariableSelectorValue('')).toBe('')
   })
 
+  it('should detect empty selector values as cleared', () => {
+    // Both empty string and empty array are "cleared" selectors.
+    // The FormInputItem uses these checks to reset to constant/null
+    // instead of persisting type=variable with an empty value.
+    const empty = normalizeVariableSelectorValue('')
+    expect(!empty || (Array.isArray(empty) && empty.length === 0)).toBe(true)
+
+    const emptyArray = normalizeVariableSelectorValue([])
+    expect(!emptyArray || (Array.isArray(emptyArray) && emptyArray.length === 0)).toBe(true)
+
+    const valid = normalizeVariableSelectorValue(['start', 'input'])
+    expect(!valid || (Array.isArray(valid) && valid.length === 0)).toBe(false)
+  })
+
   it('should derive remaining target variable types and label states', () => {
     const objectState = getFormInputState(createSchema({ type: FormTypeEnum.object }), undefined)
     const arrayState = getFormInputState(createSchema({ type: FormTypeEnum.array }), undefined)
--- a/web/app/components/workflow/nodes/_base/components/form-input-item.tsx
+++ b/web/app/components/workflow/nodes/_base/components/form-input-item.tsx
@@ -247,12 +247,27 @@ const FormInputItem: FC<Props> = ({
   }
 
   const handleVariableSelectorChange = (newValue: ValueSelector | string, variable: string) => {
+    const normalized = normalizeVariableSelectorValue(newValue)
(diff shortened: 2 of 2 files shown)


crazywoola:

### Why this is being closed

Issue #39045 has multiple open pull requests. Under the duplicate-PR rule, pull requests from the issue author are ranked first, then creation time is used as the FIFO tie-breaker. PR #39195 is the retained pull request, so this one is being closed.

### Next steps

Please coordinate on PR #39195 if you would like to contribute further. Thank you for understanding.

PR #39415, opened 2026-07-22
fix(agent): refresh stale inline preview drafts

zyssyz123:
## Summary

- keep an existing workflow-only Agent Preview draft synchronized when `node_job_only` creates a new active snapshot
- self-heal historical workflow-only drafts whose `base_snapshot_id` no longer matches the active snapshot
- route default Preview draft resolution through the shared Composer lifecycle while preserving roster drafts and account-scoped Build drafts

## Root cause

Inline Agent saves advanced `Agent.active_config_snapshot_id` and the workflow binding, but left an existing normal `AgentConfigDraft` on its old base snapshot. Default Preview treated any existing draft as authoritative and never checked it against the active snapshot, so the stale config could be reused indefinitely.

## Validation

- Ruff check and format check passed for all changed files
- `411` Agent service and Agent App unit tests passed
- `77` Agent generator and Console controller unit tests passed
- deployed commit `aa38c42827` to `deploy/agent` and confirmed the dev API image revision
- dev Console API E2E verified historical self-healing, same-config `node_job_only` synchronization, roster draft preservation, Build/Preview isolation and cleanup, and a successful Preview SSE run ending in `message_end`

Linear: [DIFY-2826](https://linear.app/dify/issue/DIFY-2826/inline-agent-%E7%9A%84-preview-%E4%BC%9A%E9%95%BF%E6%9C%9F%E8%AF%BB%E5%8F%96%E6%97%A7%E7%9A%84-agentconfigdraft)

### DIFF
--- a/api/core/app/apps/agent_app/app_generator.py
+++ b/api/core/app/apps/agent_app/app_generator.py
@@ -682,42 +682,27 @@ class AgentAppGenerator(MessageBasedAppGenerator):
             if draft_type == AgentConfigDraftType.DEBUG_BUILD.value
             else AgentConfigDraftType.DRAFT
         )
+        if effective_draft_type == AgentConfigDraftType.DRAFT:
+            from services.agent.composer_service import AgentComposerService
+
+            return AgentComposerService.get_or_create_normal_agent_draft(
+                session=session,
+                tenant_id=tenant_id,
+                agent=agent,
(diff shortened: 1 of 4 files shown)


lyzno1 (approved):

zyssyz123 (commented):
Removed: this review was submitted to the wrong pull request.

PR #39419, opened 2026-07-22
fix(api): refresh billing vector-space cache after segment deletion

he-yufeng:
## Summary

Follow-up to #39404. That PR made the document/dataset cleanup paths refresh the billing vector-space cache once index cleanup completes (`clean_document_task`, `batch_clean_document_task`, `clean_dataset_task`). The segment-level delete path frees the same vector storage through `delete_segment_from_index_task` but never schedules a refresh, so a tenant's vector-space quota stays stale after deleting individual segments until some unrelated cleanup happens to run.

This schedules the same best-effort refresh at the end of `delete_segment_from_index_task`, gated on the cleanup actually succeeding. That mirrors `clean_document_task`'s `vector_cleanup_succeeded` guard, so the early-return paths (missing dataset/document, document not in a valid state) don't fire it.

## Screenshots

| Before | After |
|--------|-------|
| N/A (backend task behavior) | N/A |

## Checklist

Tests: extended `test_delete_segment_commits_index_cleanup_without_attachments` to assert the refresh is scheduled with the dataset's tenant, and added `test_delete_segment_skips_billing_refresh_when_document_not_actionable` for the early-return path. `pytest tests/unit_tests/tasks/` (segment cleanup, billing refresh, document and dataset clean tasks): 20 passed. Ruff check/format clean on the touched files.

From Kimi Code


he-yufeng:
First nudge on this one. It still applies cleanly: the task file is untouched on main since mid-July, and the billing vector-space cache still goes stale when a segment is deleted through this path. Happy to rebase or rework if a maintainer wants a different direction.

PR #39421, opened 2026-07-22
fix(api): make site/api status toggles atomic under concurrent toggles

### LINKED ISSUE #39420: Rapid disable→enable of a Web App can leave it publicly unavailable (lost update in update_app_site_status)
**Self Checks**
- [x] This is only for a bug report, not a question.

**Dify version**

Reproduced against `main` at commit `4da4fa72cd9d011559fad3746547c935941724b2` (Self Hosted / Source). The affected code path is unchanged in current `main`.

**Cloud or Self Hosted**

Self Hosted (Source)

**Steps to reproduce**

1. Create and publish a Web App with public (anonymous) access enabled.
2. Issue two site-status toggles back-to-back, concurrently, without waiting for the first to complete — e.g. `POST /console/api/apps/<app_id>/site-enable` with `{"enable_site": false}` immediately followed by `{"enable_site": true}` (this also happens if an admin rapidly clicks the site-enable toggle in the console overview).
3. After both requests return HTTP 200, read the authoritative app detail.
4. Reload the anonymous Web App URL.

Observed: both requests return HTTP 200, but the final `enable_site` is `false` (the earlier disable), not the later `true` (restore). The public page then returns "App is unavailable".

Root cause — `api/services/app_service.py` (lines 815-832 at HEAD): https://github.com/langgenius/dify/blob/4da4fa72cd9d011559fad3746547c935941724b2/api/services/app_service.py#L815-L832

```python
def update_app_site_status(self, app: App, enable_site: bool, *, session: Session) -> App:
    if enable_site == app.enable_site:
        return app                     # stale read -> concurrent enable no-ops
    assert current_user is not None
    app.enable_site = enable_site
    app.updated_by = current_user.id
    app.updated_at = naive_utc_now()
    session.commit()                   # last committer wins, no version check
    app_was_updated.send(app)
    return app
```

Each request runs in its own session with no sequencing or optimistic version/`updated_at` check. When the enable request reads the pre-disable state, the `enable_site == app.enable_site` guard makes it return without writing, while the concurrent disable commits `false` — so the older intent wins.

Effect on public access — `api/controllers/web/passport.py` (lines 76-83): https://github.com/langgenius/dify/blob/4da4fa72cd9d011559fad3746547c935941724b2/api/controllers/web/passport.py#L76-L83

The passport reads `app_model.enable_site` and raises `NotFound()` when it is false, so the stale `false` keeps the public app unreachable until an admin toggles again.

Suggested fix: guard the write with an optimistic concurrency check (e.g. conditional `UPDATE ... WHERE updated_at = <read value>` / a version column), or serialize site-status writes per app, so an older in-flight toggle cannot overwrite a newer one.

**✔️ Expected Behavior**

After two toggles complete, the persisted `enable_site` reflects the most recent request. A restore (`enable_site=true`) issued after a disable should leave the public Web App reachable.

**❌ Actual Behavior**

The final persisted state can reflect the earlier disable rather than the later restore; both requests return HTTP 200, but the public Web App stays unavailable ("App is unavailable") until an admin toggles it again.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/b2be8c90-bcb5-4e24-9735-0185a89c8c4d.

he-yufeng:
## Summary

Fixes #39420. Rapid disable→enable toggles on a Web App can lose the restore: `update_app_site_status` guards the write with an in-memory check (`enable_site == app.enable_site`) on a possibly stale object, so the enable request reads the pre-disable state, skips its write, and returns 200 anyway. `enable_site=false` stays committed and the public page keeps 404ing until an admin toggles again.

The guard plus attribute assignment becomes a single conditional `UPDATE ... WHERE id AND <column> != <value>`: the write only lands when the row is still in the other state, so the last intent wins at SQL level regardless of which snapshot each request read. `app_was_updated` now fires only on a real transition (rowcount-driven), and the returned object is synced to the committed state (`updated_by`/`updated_at` only move on a real change). `update_app_api_status` has the same race and gets the same treatment.

## Screenshots

| Before | After |
|--------|-------|
| N/A (backend write path) | N/A |

## Checklist

Tests: two new parametrized cases per method. A stale-snapshot toggle still issues the conditional UPDATE (the regression from the issue), and no signal fires when the row is already at target. The existing commit-before-signal test is adjusted for the new write path. `pytest tests/unit_tests/services/test_app_service.py`: 29 passed (one pre-existing app-context failure deselected; it fails identically on clean main). Ruff check/format clean on the touched files.

From Kimi Code


he-yufeng:
Merged main to pick up a fresh CI run. For the record on the old red: Style Check and both API test jobs from the July run predate the recent main breakage window, and the one unit test that can fail locally (`test_bound_agent_id_is_none_for_non_agent_app`) only does so when pytest runs without the repo conftest, which provides the Flask app context the property resolves against. With conftest in the loop it passes, along with the toggle tests added here.

he-yufeng:
The red API lane was mine, and it is fixed in 78977363b4. `test_update_app_site_status_no_change` runs the service in the container suite with no login context, and my first version read `current_user.id` unconditionally, so a no-op toggle crashed there. `_apply_status_toggle` now re-reads the row inside the transaction and returns early when the row is already in the target state, so the no-change path never touches `current_user`; the conditional `UPDATE` still arbitrates the race between that read and the write. Unit suite: 38 passed locally, including a new regression test that calls the toggle with a `current_user` whose `.id` raises and asserts no write happens.

he-yufeng:
app_service.py:1042/1063 still do the read-compare-write without a row lock, so the race this patches is still open. CI was fully green on the 9-05 head. Mind taking another look?

PR #39426, opened 2026-07-22
fix(splitter): preserve spaces when recursive splitting on space separator with keep_separator=True

Taranum01:
Fixes langgenius/dify#39403.

## Problem

`FixedRecursiveCharacterTextSplitter.recursive_split_text` was dropping whitespace between words when `keep_separator=True` selected the `" "` separator. After splitting with `re.split(r" +", text)` (which silently strips the spaces) and the base `_join_docs` then stripping outer whitespace from each merged chunk, concatenating the chunks no longer reproduced the original text.

## Repro (from the issue)

```python
text = "여름철에는 항상 기상상황에 주목하며 주변 사람들과 함께 정보를 공유합니다."
splitter = FixedRecursiveCharacterTextSplitter(
    fixed_separator="\n\n", chunk_size=20, chunk_overlap=0, keep_separator=True,
)
result = splitter.split_text(text)
assert "".join(result) == text  # FAILS — joins to text without spaces
```

## Fix

1. In the `separator == " "` branch, when `keep_separator=True`, walk the source string and emit one trailing space on every split except the last so the merged chunks carry the inter-word whitespace.
2. After `_merge_splits`, re-attach the trailing separator to all but the last chunk to compensate for `_join_docs`'s outer-whitespace strip.

`keep_separator=False` and other separator branches are untouched.

## Tests added

In `TestFixedRecursiveCharacterTextSplitter`:

- `test_keep_separator_space_preserves_spaces_korean` — exact repro from the issue.
- `test_keep_separator_space_preserves_spaces_english` — English paragraph reconstructs cleanly under small chunk_size.
- `test_keep_separator_false_unaffected_by_fix` — unchanged behavior when `keep_separator=False`.
- `test_keep_separator_space_single_chunk_unchanged` — short text (no recursive split) returns the text unchanged.
- `test_keep_separator_space_does_not_double_space` — last chunk must not carry a trailing space.

## Verification

- Full `tests/unit_tests/core/rag/splitter/` suite: 109 passed, 4 skipped, 0 regressions.
- Related `tests/unit_tests/core/rag/indexing/test_index_processor_base.py`: 15 passed.
- Sanity-checked the tests: 3 of the 5 new tests fail on pre-fix code (the 2 that exercise `keep_separator=False` and the single-chunk path correctly pass on both).

Made with [Cursor](https://cursor.com)

PR #39429, opened 2026-07-23
fix(rag): preserve spaces when recursive-splitting on the space separator

### LINKED ISSUE #39403: FixedRecursiveCharacterTextSplitter removes spaces during recursive splitting
### Self Checks

### Dify version

1.15.0 (also reproducible on current main)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Run Dify 1.15.0 as a self-hosted installation using Docker Compose.
2. Create a Knowledge dataset and upload a DOCX document containing a long Korean paragraph with normal spaces between words.
3. Set the delimiter to `\n\n` and the chunk size to approximately `1000` or `1024`, so that the paragraph exceeds the chunk size and requires recursive splitting.
4. Process the document and inspect the generated chunks.
5. The spaces inside the recursively split paragraph are removed.
6. Increase the chunk size to `1500`, so that the same paragraph is not recursively split, and process it again.
7. The spaces are preserved when recursive splitting is not triggered.

The issue can also be reproduced without DOCX parsing on the current `main` branch:

```python
text = "여름철에는 항상 기상상황에 주목하며 주변 사람들과 함께 정보를 공유합니다."
splitter = FixedRecursiveCharacterTextSplitter(
    fixed_separator="\n\n",
    chunk_size=20,
    chunk_overlap=0,
    keep_separator=True,
)

result = splitter.split_text(text)

assert len(result) > 1
assert " ".join(result) == text
```

The final assertion fails because spaces are removed when the text is recursively split using the space separator.

### ✔️ Expected Behavior

Spaces from the original text should be preserved when a long paragraph is recursively split. Combining the resulting chunks should reproduce the original text without removing word boundaries.

### ❌ Actual Behavior

When a paragraph exceeds the chunk size and recursive splitting selects the space separator, spaces between words are removed from the generated Knowledge chunks.

Original:
여름철에는 항상 기상상황에 주목하며 주변 사람들과 함께 정보를 공유합니다.

Result:
여름철에는항상기상상황에주목하며주변사람들과함께정보를공유합니다.

The DOCX source and its document.xml contain the spaces correctly. Increasing the chunk size so that recursive splitting is not triggered also preserves the spaces.

In `recursive_split_text`, `re.split(r" +", text)` removes the space separators. However, the `keep_separator` restoration logic currently runs only in the non-space separator branch. Because merging uses an empty separator when `keep_separator=True`, the split words are concatenated without spaces.

sergioperezcheco:
When a chunk larger than chunk_size is recursively split on the space separator, the spaces between words were being dropped, so adjacent words got concatenated ('word1word2' instead of 'word1 word2'). This affects CJK and Korean text with normal inter-word spaces, and any language once a paragraph exceeds the chunk size and the splitter falls through the separator hierarchy to ' '.

The root cause is in FixedRecursiveCharacterTextSplitter.recursive_split_text: the ' ' separator uses re.split(r' +', text), which discards the spaces. The keep_separator logic then passes an empty separator to _merge_splits, assuming the separator was already re-appended — but the re.split path never re-appends it, so _join_docs joins the words with no spaces.

The fix passes the real ' ' separator to _merge_splits for the space case, so spaces are reinserted when chunks are merged back together. Other separators are unaffected since they go through the append path that does preserve the separator. Added a regression test (test_recursive_space_split_preserves_spaces) that splits space-delimited text exceeding chunk_size and asserts every output chunk contains spaces.

Closes #39403

### DIFF
--- a/api/core/rag/splitter/fixed_text_splitter.py
+++ b/api/core/rag/splitter/fixed_text_splitter.py
@@ -101,7 +101,14 @@ class FixedRecursiveCharacterTextSplitter(EnhanceRecursiveCharacterTextSplitter)
             splits = [s for s in splits if (s not in {"", "\n"})]
         _good_splits = []
         _good_splits_lengths = []  # cache the lengths of the splits
-        _separator = "" if self._keep_separator else separator
+        # The " " separator uses `re.split(r" +", text)` above, which discards the
+        # spaces rather than re-appending them (unlike the keep_separator append
+        # path for other separators). So the spaces must be reinserted when merging
+        # the splits back together; pass the real separator to `_merge_splits`.
+        if self._keep_separator and separator == " ":
+            _separator = " "
+        else:
+            _separator = "" if self._keep_separator else separator
         s_lens = self._length_function(splits)
         if separator != "":
             for s, s_len in zip(splits, s_lens):
--- a/api/tests/unit_tests/core/rag/splitter/test_text_splitter.py
+++ b/api/tests/unit_tests/core/rag/splitter/test_text_splitter.py
@@ -952,6 +952,26 @@ class TestFixedRecursiveCharacterTextSplitter:
         assert "word1" in combined
         assert "word2" in combined
 
+    def test_recursive_space_split_preserves_spaces(self):
+        """Regression test: recursive splitting on " " must keep spaces between words.
+
+        Previously, when a chunk exceeded `chunk_size` and was recursively split on
+        the " " separator, the spaces were dropped and words were concatenated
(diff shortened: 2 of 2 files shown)


crazywoola:

### Why this is being closed

Issue #39403 has multiple open pull requests. Under the duplicate-PR rule, pull requests from the issue author are ranked first, then creation time is used as the FIFO tie-breaker. PR #39426 is the retained pull request, so this one is being closed.

### Next steps

Please coordinate on PR #39426 if you would like to contribute further. Thank you for understanding.

PR #39235, opened 2026-07-19
fix(api): reject trailing newline in email() validator

lntutor:
Closes https://github.com/langgenius/dify/issues/39234

### Summary

`libs/helper.py::email()` validated with `re.match` against a `$`-anchored pattern. In Python `$` matches at end-of-string **or just before a trailing newline**, so `email("user@example.com\n")` returned the value as valid instead of raising.

`email()` backs the `EmailStr` type used across the auth surface (login, registration, forgot-password, activate, workspace member invite). A CR/LF in an accepted address is a mail header-injection primitive and defeats exact-match uniqueness/normalization (`"a@b.com\n" != "a@b.com"`).

### Change

```diff
-    if re.match(pattern, email) is not None:
+    if re.fullmatch(pattern, email) is not None:
```

`re.fullmatch` requires the entire string to match, rejecting a trailing newline. Behavior changes **only** for the trailing-newline case — every currently-valid address still validates (verified: `re.match` and `re.fullmatch` differ only on strings with content after the `$` match, i.e. a trailing `\n`).

### Tests

Adds `TestEmailValidator` to `tests/unit_tests/libs/test_helper.py` (the validator had no unit tests): accepts valid / special-local-part addresses, rejects trailing newline, missing domain, and plain strings. All 22 `test_helper` tests pass; `ruff check` and `ruff format --check` clean.

PR #39442, opened 2026-07-23
fix(api): reject trailing newlines in email() validator (#39234)

### LINKED ISSUE #39234: email() validator accepts a trailing newline (mail header-injection vector on the auth surface)
### Self Checks

### Dify version

main

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

`libs/helper.py::email()` validates with `re.match` against a pattern anchored by `$`:

```python
pattern = r"^[\w\.!#$%&'*+\-/=?^_`{|}~]+@([\w-]+\.)+[\w-]{2,}$"
if re.match(pattern, email) is not None:
    return email
```

In Python, `$` matches at the end of the string **or just before a trailing newline**, so a trailing `\n` slips through:

```python
>>> from libs.helper import email
>>> email("user@example.com\n")
'user@example.com\n'          # accepted; expected ValueError
```

### Impact

`email()` backs `EmailStr = Annotated[str, AfterValidator(email)]`, used across the auth surface (login, registration, forgot-password, activate, workspace member invite). A newline-bearing address:

- is a classic **mail header-injection** primitive if any downstream mailer assembles headers from the raw value, and
- defeats exact-match uniqueness / normalization: `"user@example.com\n" != "user@example.com"`, enabling shadow/duplicate accounts.

### Fix

Use `re.fullmatch` instead of `re.match`; it requires the whole string to match and rejects the trailing newline. This changes behavior only for the trailing-newline case — all currently-valid addresses still validate. PR with a unit test follows.

Harsh23Kashyap:
## Summary

Fixes #39234

Python's `re.match` anchors `$` at end-of-string **or just before a trailing newline**, so a raw `\n` slipped through `libs.helper.email()` and into `EmailStr`. The validator backs the auth surface (login, registration, forgot-password, activate, workspace member invite), where a newline-bearing address is a classic mail header-injection vector and lets `"user@example.com\n"` coexist with `"user@example.com"` (shadow accounts).

Switch `re.match` to `re.fullmatch` so the whole string must match. All currently-valid addresses still match; only the trailing-newline case is rejected. Add a unit test pinning the new behavior.

```python
>>> from libs.helper import email
>>> email("user@example.com")
'user@example.com'
>>> email("user@example.com\n")
ValueError: user@example.com
 is not a valid email.
```

## Why this is a security issue, not just a style nit

1. **Mail header injection.** Any downstream SMTP path that assembles headers from the raw `EmailStr` value (bounce handlers, password-resend confirmations, workspace invites) can have a `\n` interpreted as a header separator. The standard mitigation is to validate-and-reject at the entry point — exactly what `email()` is for.
2. **Shadow / duplicate accounts.** Email lookup uses exact equality on the column. A user registering as `attacker@x.com\n` doesn't collide with `attacker@x.com` at the DB layer; the canary at the validator was the only thing keeping them in sync.

## Changes

| File | Change |
|---|---|
| `api/libs/helper.py` | `re.match` → `re.fullmatch` on the email validator (line 225); 2 lines of comment referencing the issue |
| `api/tests/unit_tests/libs/test_email.py` | new `test_email_with_trailing_newline_raises` pinning the new behavior |

10 lines added, 2 removed across 2 files.

## Verification

- `pytest tests/unit_tests/libs/test_email.py -v` → **3 passed** (2 existing + 1 new)
- `pytest tests/unit_tests/libs/` (full module) → **587 passed, 16 warnings, 138 subtests passed**
- `ruff check` → clean
- `ruff format --check` → already formatted
- Manual sanity: 4 positive + 3 existing negative + 1 new negative (trailing newline) assertions all pass

The 16 warnings are pre-existing `InsecureKeyLengthWarning`s from `test_passport.py`'s short HMAC keys — unrelated to this change.

## Out-of-scope observation

While reading the area I noticed `api/libs/password.py:13` has the same `re.match`-with-`$` pattern, so a password ending in `\n` would also slip through. Keeping this PR scoped to the email validator; a follow-up PR can cover the password case with the same one-line fix + test.

## Checklist

- [x] This change requires a documentation update, included: not needed — `email()` is an internal validator, no user-facing docs reference the trailing-newline case
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. — issue #39234 was filed and a claim was posted before the PR
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. — one commit, one test
- [ ] I've updated the documentation accordingly. — N/A
- [x] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods — `ruff check` + `ruff format --check` clean (this is a backend-only change; no frontend touch)

### DIFF
--- a/api/libs/helper.py
+++ b/api/libs/helper.py
@@ -221,8 +221,9 @@ def current_timestamp() -> int:
 def email(email):
     # Define a regex pattern for email addresses
     pattern = r"^[\w\.!#$%&'*+\-/=?^_`{|}~]+@([\w-]+\.)+[\w-]{2,}$"
-    # Check if the email matches the pattern
-    if re.match(pattern, email) is not None:
+    # Use re.fullmatch so a trailing newline (which re.match's $ accepts)
+    # cannot smuggle a raw "\n" into the auth surface — see #39234.
+    if re.fullmatch(pattern, email) is not None:
         return email
 
     error = f"{email} is not a valid email."
--- a/api/tests/unit_tests/libs/test_email.py
+++ b/api/tests/unit_tests/libs/test_email.py
@@ -19,3 +19,10 @@ def test_email_with_invalid_email():
 
     with pytest.raises(ValueError, match="()@example.com is not a valid email."):
         email("()@example.com")
+
+
+def test_email_with_trailing_newline_raises():
+    # re.match's $ accepts a trailing "\n", which would otherwise smuggle a raw newline
+    # into the auth surface (mail header-injection vector, shadow accounts). See #39234.
+    with pytest.raises(ValueError, match="user@example.com\n is not a valid email."):
+        email("user@example.com\n")


crazywoola:

### Why this is being closed

Issue #39234 has multiple open pull requests. Under the duplicate-PR rule, pull requests from the issue author are ranked first, then creation time is used as the FIFO tie-breaker. PR #39235 is the retained pull request, so this one is being closed.

### Next steps

Please coordinate on PR #39235 if you would like to contribute further. Thank you for understanding.

Issue #39434, opened 2026-07-23
TTS audio fails to play in browser due to Content-Type mismatch (WAV data served as audio/mpeg)


chenjinming2071:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.15.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1.Configure TTS in ChatFlow using Tongyi's qwen3-tts-flash model.

2.Generate a conversation and click the audio playback button.

3.Observe that no sound is played.

### ✔️ Expected Behavior

Audio plays normally in the browser.

### ❌ Actual Behavior

TTS audio fails to play in the browser. The browser console shows the following error:

**Chrome**:
```
Uncaught (in promise) NotSupportedError: Failed to load because no supported source was found.
```

**Edge**:
```
Uncaught (in promise) NotSupportedError: Failed to load because no supported source was found.
```

**Firefox**:
```
Uncaught DOMException: MediaSource.addSourceBuffer: Type not supported in MediaSource
```

Upon inspecting the API response, I found:
- The actual audio data returned is **WAV format** (starts with `RIFF....WAVE`)
- But the HTTP response `Content-Type` is set to **`audio/mpeg`**

This is exactly the same root cause identified in [#35880](https://github.com/langgenius/dify/issues/35880) by @euxx:

> "The Tongyi TTS implementation does not specify an output format and returns the downloaded audio bytes unchanged. The actual response is WAV data (RIFF....WAVE, PCM 24 kHz), while Dify responds with Content-Type: audio/mpeg."

Although [#35901](https://github.com/langgenius/dify/pull/35901) was merged to fix AudioContext issues, the **audio format mismatch** between the actual WAV data and the `audio/mpeg` Content-Type header persists in v1.15.0.

[...]

PR #39446, opened 2026-07-23
fix(api): sniff the TTS response content type from the audio bytes

he-yufeng:
The TTS streaming path in `AudioService` hardcoded `content_type="audio/mpeg"` on every generator response (`api/services/audio_service.py`, both call sites). Providers that return a different container get mislabeled: Tongyi `qwen3-tts-flash` sends WAV bytes, so the browser sees `audio/mpeg` over a WAV payload and throws `NotSupportedError` (#39434).

Since the provider only hands us a byte stream, the type has to come from the bytes. This sniffs the first chunk: `RIFF....WAVE` maps to `audio/wav`, `OggS` to `audio/ogg`, `fLaC` to `audio/flac`, and everything else keeps the previous `audio/mpeg` default. The stream passes through unchanged and in order, and an empty stream keeps the old default. The OpenAPI annotation for `/text-to-audio` and its swagger test now list the possible types instead of declaring mpeg only.

Verified with a standalone harness that loads the edited module with the heavy project imports stubbed and real Flask: the WAV/MP3/OGG/unknown cases sniff correctly, the first chunk is preserved in order, and an empty stream stays `audio/mpeg`. New unit tests cover the sniff table, stream pass-through, and the empty case. Note the project venv on my machine is iCloud-evicted, so I could not run the repo's own pytest session end to end here; CI should confirm.

Refs #39434.

From Kimi Code


he-yufeng:
autofix attribution, since the failure was in the codegen step rather than the fix itself: the four-type `produces` list I had on this endpoint landed in the generated spec as four schema-less media types, and `@hey-api/openapi-ts` dies planning those ("Symbol finalName has not been resolved yet", jobIndex 57). Reproduced locally against the workspace config: the exact same spec crashes, a single media type passes, and four types with an explicit binary schema also pass. Interesting wrinkle: `/files/{file_id}/preview` already ships sixteen schema-less types and is fine, so the crash is picky about where the empty entries appear.

Not worth fighting the flask-restx serializer over, so the annotation now advertises a single `application/octet-stream` (same shape the preview endpoint uses) and the sniffed types stay documented in the 200 description. Runtime behavior is unchanged, the swagger test asserts the octet-stream shape now, and both suites pass locally (29 tests).

he-yufeng:
Nudging this one: the autofix-attribution note above still holds (the red was in the codegen step, not the fix), and the branch is green against current main. The TTS content type is now sniffed from the audio instead of trusted from the request. Could use a review pass.

he-yufeng:
The Web Full-Stack E2E red on the latest run is infrastructure, not this change: the job timed out installing Playwright browsers (15-minute cap hit during the download step) before any test ran, and this PR only touches the API TTS response sniffing. A re-run should clear it.

he-yufeng:
Superseded by #41043 (merged 2026-08-24), which now propagates and validates the actual audio MIME end to end, including byte-level inspection. No longer needed.

PR #39451, opened 2026-07-23
feat(trace): add provider-neutral unified tracing

### LINKED ISSUE #39450: Introduce provider-neutral unified tracing for nested workflow hierarchy
## Self Checks

## 1. Is this request related to a challenge you're experiencing? Tell me about your story.

Dify's tracing integrations currently assemble traces independently in each provider implementation. This duplicates Dify-specific logic for workflow executions, sessions, span status, and hierarchy, and it allows the same execution to be represented differently across providers.

Nested workflow correlation is the clearest gap. Phoenix has been enhanced to connect nested workflow spans to their actual parent tool spans, but the other tracing providers do not currently provide equivalent hierarchy assembly. As a result, users can receive disconnected root traces instead of one navigable parent-child tree.

The Phoenix implementation has now been validated in real customer workloads. That validation confirmed that complete parent-child hierarchy is necessary for debugging production workflows, especially when workflows invoke other workflows or asynchronous operations. Customers also need the same capability in LangSmith and, eventually, the other tracing providers rather than a Phoenix-only solution.

Maintaining separate hierarchy implementations for every provider will not scale and is likely to produce inconsistent behavior. Dify needs one provider-neutral place to assemble the logical trace before translating it to each provider's protocol.

## 2. Additional context or comments

Proposed direction:

- Build a canonical, provider-neutral trace tree in Dify Core.
- Load workflow executions and reconstruct predecessor, loop, iteration, nested-workflow, and asynchronous parent-child relationships once.
- Centralize session resolution, status/error normalization, and common trace metadata.
- Coordinate cross-task provider parent contexts through a shared Core mechanism.
- Keep provider adapters lightweight: map canonical spans to the provider SDK, emit them, and export or restore only the minimal provider-specific parent context.
- Preserve existing provider behavior behind an opt-in feature flag during migration.
- Support Phoenix and LangSmith first, then migrate other providers incrementally.

This boundary gives downstream providers consistent Dify semantics without forcing Core to depend on Phoenix, LangSmith, OpenTelemetry, or another provider SDK.

## 3. Can you help us with this feature?

Blackoutta:
> [!IMPORTANT]
>

## Summary

Fixes #39450

This PR introduces an opt-in, provider-neutral unified tracing path that assembles Dify trace semantics once in Core and delegates only provider-specific translation and transport to lightweight adapters.

The implementation:

- builds a canonical parent-first span tree for workflows, chatflows, messages, nodes, loops, and iterations;
- reconstructs predecessor and container hierarchy once instead of independently in each provider;
- coordinates nested-workflow and asynchronous child parent contexts through a versioned Core envelope;
- adds unified Phoenix and LangSmith adapters;
- preserves message, session, error, token usage, retry summary, and linked-root metadata;
- records the stable architecture and compatibility decisions in ADRs.

### Backwards compatibility

The change is fully backwards compatible:

- `OPS_TRACE_UNIFIED_ENABLED` defaults to `false`.
- When disabled, every provider continues to use its existing implementation.
- When enabled, only providers explicitly registered for unified tracing use the new path.
- Providers that have not migrated continue to use their existing implementations.
- Unified dispatch never falls back to legacy dispatch after it starts, preventing duplicate or split traces.
- Legacy and unified provider instances, registries, caches, and Redis namespaces remain isolated.

With the flag enabled, Phoenix and LangSmith currently receive the complete parent-child hierarchy, including nested workflows. Other tracing providers can be migrated incrementally by implementing the lightweight unified adapter contract.

### Phoenix

https://github.com/user-attachments/assets/d2feb51a-0672-429d-9066-9e650412d0ad

### LangSmith

https://github.com/user-attachments/assets/b01e47b9-c9ea-4ba4-98b0-19d7b9c096d9

## Test plan

Validated against a local merge of `origin/main` and this branch:

- UV lockfile validation and dependency sync
- Ruff formatting and lint checks
- Import Linter and response contract linter
- No-new-`getattr` and controller SQLAlchemy guards
- Dotenv Linter
- Pyrefly and Mypy
- API unit tests: 14,119 passed, 4 skipped
- Controller unit tests: 3,493 passed
- Offline database upgrade and downgrade SQL generation

CI-only middleware integration, live PostgreSQL/MySQL migration, and E2E jobs remain covered by the PR workflow.

## Checklist

- [x] I ran the backend lint, type-check, and unit-test commands required by CI.

From pi


docs/architecture/adr/unified-tracing/0001-canonical-trace-model.md
```diff
@@ -0,0 +1,64 @@
+# ADR-0001: Use a Provider-Neutral Canonical Trace Model
```
QuantumGhost: The three ADRs should be consolidated into a single ADR for the unified tracing runtime.

The canonical model, dispatch routing, and cross-task parent coordination are not independently selectable or reversible decisions. They define one architectural contract and must be evaluated together. The current split results in repeated motivation, supported semantics, and consequences, while several important contracts remain implicit.

QuantumGhost: **AI disclosure**: This comment was primarily drafted with Codex. I am responsible for the content.

## Architecture documentation

The three ADRs should be consolidated into one ADR for the unified tracing runtime, accompanied by a separate normative specification.

The canonical trace model, dispatch routing, and cross-task parent coordination are not independently selectable or reversible architectural decisions. They define one runtime contract and must be evaluated together. The current split repeats motivation, supported semantics, and consequences while leaving several hard-to-change decisions implicit.

### Consolidated ADR

The ADR should focus on decisions that will be difficult to change once providers, persisted task payloads, workers, and operational processes depend on them.

It should explicitly decide the following.

#### 1. Logical trace unit and topology

- Whether an adapter receives a complete logical trace or a fragment with an explicit external parent.
- Whether every local `parent_id` must reference a span in the same fragment.
- Whether only the fragment root may have an external parent.
- Whether adapters may only consume parent relationships resolved by Core and must never infer missing parents.
- Whether parent-first ordering is guaranteed.
- Whether sibling ordering is deterministic and whether it carries any causal meaning.
- How predecessor relationships, structured containment, parallel branches, repeated executions, synthetic wrappers, ambiguous relationships, and cycles are interpreted.

Loop/Iteration nesting must be an explicit support decision.

The initial contract version may intentionally support only non-nested Loop/Iteration structures. It does not need to be recursively compatible with future engine behavior. However, the ADR must explicitly define:

- the containment forms supported by the current contract version;
- how unsupported nested Loop/Iteration topology is detected;
- whether unsupported topology causes trace rejection, conservative degradation, or another explicit fallback;
- which component owns that decision;
- how the unsupported condition is reported and observed.

Nested containment must not be silently flattened or interpreted independently by each adapter unless that behavior is explicitly selected as part of the contract.

#### 2. Snapshot and identity

- Which execution snapshot is used to construct a trace: an immutable completion snapshot, current persistence state at task execution time, task payload data, or a defined combination.
- The distinction between canonical trace/span identity, delivery identity, and provider trace/run/span identity.
- Which identities or observable effects remain stable across retries, duplicate task delivery, worker restarts, and delayed execution.
- How synthetic and repeated executions receive deterministic identities.
- The distinction between immutable dispatch identity and provider destination identity.
- When mode, provider, destination, configuration revision, and contract version are frozen for one logical dispatch.

The contract should define stable observable effects rather than assume that every provider allows Dify to control provider-native identifiers.

#### 3. Emission protocol

- What “provider accepted” means for each supported adapter capability.
- Whether local SDK enqueueing, exporter completion, and remote acknowledgement represent different acceptance levels.
- The state transitions between pending, partially emitted, emitted, and parent-context published.
- Which provider effects have occurred at each state.
- Checkpoint granularity and ownership.
- Whether an entire fragment, only remaining spans, or only context publication may be retried.
- The required delivery and idempotency guarantees when a task is delivered more than once.

Provider emission and parent-context publication are non-transactional external effects. Their recovery semantics must be part of the architecture rather than left to individual adapters.

#### 4. Cross-task coordination

- How provider, normalized endpoint, project/resource, and provider account or workspace identity determine destination compatibility.
- The distinction between destination identity and configuration revision. Credential rotation or operational configuration changes do not necessarily represent a different destination.
- How parent-context ownership and tenant/application authorization are validated.
- Whether parent-context data is strictly coordination data and never an authorization token.
- The lifecycle of parent context, including publication, duplicate publication, lookup, expiry, and cleanup.
- How missing, expired, never-published, malformed, incompatible, and unsupported contexts are distinguished.
- Which conditions are retryable, terminal, or produce a linked root.
- What happens when retry exhaustion occurs after a compatible parent was expected.

Destination compatibility and authorization should be separate decisions.

#### 5. Evolution and ownership

- The responsibilities of Core, adapters, and the composition root.
- Which layer discovers and constructs adapters.
- Which layer loads workflow execution data.
- Which layer owns topology decisions, provider emission, checkpoint persistence, parent-context publication, and retry policy.
- Versioning rules for serialized boundaries such as task payloads and parent-context envelopes.
- Mixed-version behavior when retries are processed by different worker versions.
- Adapter capability and conformance requirements.
- How unsupported fields, span kinds, context versions, and provider features are handled.
- Whether adding nested Loop/Iteration support requires a new canonical contract version, a new topology capability, or both.
- How tasks, stored contexts, and older adapters behave during that upgrade.

### Replaceable implementation details

Replaceable implementation mechanisms should not dominate the ADR. Examples include:

- concrete Python model types;
- class and helper decomposition;
- registry container implementation;
- lazy-import mechanics;
- cache implementation and capacity;
- Redis key prefixes;
- exact TTL, retry-count, and delay values;
- the initial set of migrated providers;
- provider SDK calls and provider-native attribute mappings.

The ADR should state the invariants these mechanisms must satisfy—for example, runtime isolation or the required relationship between context retention and retry windows—without unnecessarily fixing their implementation.

### Normative specification

The referenced specification should make the runtime contract deterministic. It should define:

- trace-fragment and parent invariants;
- topology and containment semantics;
- the support matrix for the current topology contract version;
- deterministic handling of topology unsupported by that version;
- snapshot and identity semantics;
- dispatch and destination compatibility;
- provider acceptance and parent-publication lifecycle;
- partial-emission recovery, checkpoint, replay, and idempotency requirements;
- the complete retryable, terminal, linked-root, and unsupported failure matrix;
- authorization and data-boundary requirements;
- version compatibility and mixed-version rollout behavior;
- adapter capabilities and conformance criteria;
- the compatibility path for introducing nested Loop/Iteration support in a later version.

Where prose alone leaves meaningful room for interpretation, the specification should include a small number of normative examples that establish the intended semantics. These examples should be used only where the valid behavior is complex or difficult to infer.

Counterexamples should be included only where a plausible alternative interpretation must be explicitly excluded, together with the invariant it violates.

For example, the current specification should show the supported Loop/Iteration semantics for this contract version. If nested Loop/Iteration is intentionally unsupported, it should also show how that topology is classified and which tracing outcome is required. The purpose is not to make the initial version forward-compatible, but to ensure that the current limitation and the future compatibility boundary are deliberate, observable, and versioned.

The examples should clarify semantic behavior and conformance requirements, not prescribe a particular implementation, in-memory representation, or serialization format.

Until the consolidated ADR and normative specification establish this deterministic contract, the decision should remain in `Proposed` status. The implementation can then be reviewed against that contract.

Blackoutta: I consolidated the three ADRs into one proposed runtime ADR in `e4a8a821fe` and changed its status from **Accepted** to **Proposed**. The consolidated ADR now records the behavior already implemented and tested: Core/adapter ownership, parent-first deterministic ordering, non-causal sibling ordering, conservative predecessor/ambiguity/cycle handling, the currently supported non-nested Loop/Iteration wrappers, unified/legacy isolation, and the existing parent-context compatibility/failure matrix.

I have not added the normative specification yet. Before encoding the remaining semantics or changing the runtime, I would like to confirm the following proposed decisions.

### 1. Topology contract v1

- v1 covers the product topology currently supported by Dify. The editor does not allow Loop or Iteration nodes inside another container.
- Imported, historical, or future persisted data containing nested Loop/Iteration topology should be detected, conservatively parent affected spans to the workflow root, and emit an observable warning. Adapters must not interpret it independently.
- Parallel metadata does not add a hierarchy edge; persisted predecessor and supported container relationships remain authoritative.
- I propose making external parentage explicit and validating that every local `parent_id` references a span in the same fragment, only the fragment root can have an external parent, and spans are parent-first. This removes the current LangSmith-only inference from an out-of-fragment root `parent_id`.

Is this the intended v1 topology boundary and unsupported-topology behavior?

### 2. Snapshot, identity, and configuration changes

- The current workflow task payload contains the workflow/root completion data, while node executions are loaded from tenant-scoped persistence when the trace attempt runs.
- Persisted workflow, message, and node execution IDs and synthetic wrapper IDs are deterministic. I propose persisting a generated ID for standalone operations instead of generating a new UUID on every attempt.
- I do **not** propose freezing tracing configuration at enqueue time. Each delivery attempt should use the latest enabled mode/provider/destination. A user configuration change should affect pending and retried traces; a parent now targeting an incompatible destination becomes a linked root. One attempt still uses exactly one selected runtime and destination.
- The specification would explicitly state that configuration may change between attempts and that provider-native IDs or effects are not universally stable unless the adapter capability guarantees it.

Is resolving the latest configuration per attempt acceptable, or should configuration/destination be immutable for the complete logical dispatch?

### 3. Emission and recovery

For contract v1, I propose **at-least-once whole-fragment emission with no span checkpoint**:

- parent resolution occurs before any child fragment emission;
- adapters emit parent-first;
- a retry after partial external effects may replay the complete fragment and may duplicate provider effects;
- adapters must document any stronger provider-native idempotency they supply;
- there is no exactly-once guarantee and no cross-provider checkpoint protocol in v1.

Two current correctness defects would be fixed:

1. Phoenix uses `SimpleSpanProcessor`, which swallows exporter exceptions and ignores a failed export result. Parent context can therefore be published even when the parent was not exported. The adapter should publish context only after a strict synchronous successful export; failure must not publish context.
2. Parent-context TTL is currently 300 seconds while the default retry window is about 1,500 seconds. TTL should be configurable and validated to cover the retry and scheduling window.

Is whole-fragment at-least-once replay without checkpoints acceptable for v1?

### 4. Parent-context lifecycle and trust boundary

I propose documenting the current lifecycle explicitly:

- duplicate publication is last-write-wins and refreshes retention;
- lookup is non-consuming so multiple children can restore one parent;
- missing, expired, and never-published compatible contexts are intentionally represented by the same unavailable state and remain retryable until exhaustion;
- malformed, unsupported-version, and stored provider/destination mismatch are terminal;
- absent, legacy, cross-provider, and cross-destination parents become linked roots;
- retry exhaustion never silently creates a detached root.

The supported workflow path generates `ParentTraceContext` internally, workflow execution loading is tenant-scoped, execution IDs are globally unique, and provider context contains no credentials or authorization token. There is currently no second tenant/app ownership field in the Redis envelope. I propose documenting this internal trust boundary rather than adding a separate authorization mechanism unless explicit defense against corrupted internal task payloads is required.

Do you require tenant/app ownership to be redundantly encoded and checked in the coordination envelope?

### 5. Contract evolution

I propose minimal explicit versioning rather than a capability framework:

- task payload, canonical topology, and parent-context envelope use contract version 1;
- a registered unified adapter means complete v1 support;
- unknown versions and unsupported canonical kinds fail closed;
- older payloads without a unified contract version retain legacy interpretation;
- unified mode is enabled only after all `ops_trace` Celery workers are upgraded;
- a future v2 rollout first deploys readers for v1 and v2, then starts writing v2.

Is this sufficient for the initial mixed-version and adapter-conformance contract?

After these decisions are confirmed, I will add the normative specification, implement the agreed correctness changes, and add contract/conformance tests before changing the ADR status.

Blackoutta:
## Update: runtime reliability, Agent v2 tracing, and ADR alignment

I rebased the branch and completed the runtime and documentation work discussed in the review. The main changes are summarized below.

### New capabilities

#### Agent v2 unified tracing

- Added unified tracing for both Agent Apps and Agent v2 workflow nodes.
- Normalized Dify Agent / Pydantic AI stream events into provider-neutral Agent, LLM, and Tool operations, including stable run identity, inputs/outputs, usage, errors, and retry/resume roles.
- Preserved Agent trace fragments in workflow tracing state so they survive workflow Human Input pauses and are emitted under the exact owning node execution.
- Propagated Agent run and workflow-tool call identity across the Dify Agent → Core tool boundary, allowing nested workflow-tool traces to restore the correct logical parent.
- Added fail-open collection: event normalization or trace finalization failures are logged but do not change Agent execution results.
- Added the tracing configuration panel to the Agent v2 monitoring surface.
- Added Agent App Human Input tracing. The requesting and resuming Messages remain separate traces in the same conversation session, with stable wait identity and logical-link metadata connecting the two phases.

unified tracing now works for agent v2:


https://github.com/user-attachments/assets/063e12b8-0fcc-44f0-85ef-fbed5ec2f643



#### Human Input inside Loop/Iteration

- Workflow Human Input waits now prefer the persisted node execution ID as their owner, so repeated executions inside Loop or Iteration containers attach to the correct execution rather than relying on a static node ID.
- Older persisted records retain the static-node/time-proximity compatibility fallback.
- Workflow and Chatflow executions remain one logical trace across any number of pauses and resumptions; the final trace is emitted only after a terminal workflow outcome.

unified tracing now works for human input in loop/iteration:


https://github.com/user-attachments/assets/e8459b63-f68d-4580-9625-b487c1d5cc41

### Reliability and behavior changes

#### At-least-once final trace delivery

- Added durable final-trace handoff state to workflow pause snapshots.
- A global-timeout snapshot remains authoritative until the deterministic final payload has been persisted and accepted by Celery.
- Failed handoffs are retried by the existing Human Input timeout scan with a separate bounded budget configured by `OPS_TRACE_FINAL_TRACE_HANDOFF_MAX_RETRIES`.
- Pending final handoffs are protected from pause pruning; exhausted handoffs are marked failed, logged, and can later be pruned with their payload.
- Provider dispatch and final-payload handoff use separate retry budgets.
- Recoverable provider failures retain the stored payload only after Celery accepts the retry request. Terminal failure, retry scheduling failure, retry exhaustion, and success clean it up.
- Phoenix now requires synchronous exporter success before parent context can be published.
- LangSmith transport/API failures are mapped to the retryable Core error while authentication and user/configuration errors remain terminal.

#### Core-owned parent resolution and identity

- Standalone Message children (Moderation, Suggested Question, Dataset Retrieval, Tool Trace, and Generate Name) are now local-root fragments with `required_parent_context_id`; Core waits for and resolves the Message parent before adapter emission.
- Removed LangSmith-specific inference from an out-of-fragment local `parent_id`.
- Added a persisted optional `operation_id` so standalone operations retain canonical identity across retries while old payloads remain readable.
- Parent-context retention is configurable and validated to cover the configured retry window.
- Added structural validation for canonical fragments: unique IDs, exactly one local root, root without a local parent, parent-first local references, and mutually exclusive external-parent modes.

#### Adapter conformance

- Registered unified adapters must cover every current `CanonicalSpanKind`.
- Added missing Human Wait support:
  - Phoenix: OpenInference `CHAIN`
  - LangSmith: `run_type="chain"`
- Every adapter preserves the canonical kind as `dify.span.kind`.
- Logical Dify links are preserved as `dify.span.links`. v1 does not fabricate provider-native links because the canonical link currently contains a logical Dify ID rather than provider trace/span context.
- Reserved canonical metadata overrides conflicting caller metadata.
- Delivery remains whole-fragment at-least-once with no per-span checkpoint; replay may duplicate provider effects.

### ADR and normative contract alignment

The previous ADRs are consolidated into:

- `docs/architecture/adr/unified-tracing/0001-unified-tracing-runtime.md`
- `docs/architecture/adr/unified-tracing/0001-unified-tracing-runtime-spec.md`

The ADR status is now **Revised**. The ADR/spec now explicitly define:

- Core, adapter, and composition-root ownership;
- canonical fragment and parent-first topology invariants;
- the v1 producer boundary: supported Dify paths produce non-nested Loop/Iteration topology, and adapters never infer or flatten nested containment;
- workflow snapshot, canonical identity, delivery identity, and provider identity semantics;
- latest-configuration-per-attempt behavior and destination compatibility;
- synchronous provider acceptance boundaries;
- whole-fragment replay and at-least-once guarantees;
- parent-context lifecycle, retry/terminal/linked-root outcomes, retention, and trust boundary;
- serialized-boundary versioning and mixed-worker rollout requirements;
- Phoenix and LangSmith conformance behavior for Human Wait, links, identity, replay, and acceptance.

Intermediate design/plan documents were removed from the final diff; only the consolidated ADR and its normative specification remain.

### Verification

- Ruff checks and formatting checks pass.
- Latest focused regression suite: **117 tests passed**.
- Broader unified tracing / Human Input regression suite: **241 tests passed**.
- Alembic reports a single migration head.

Blackoutta:
## Update: unified tracing v1 scope narrowed

Following review feedback about potential conflicts with the upcoming subgraph expansion work, this branch has been rebased and the v1 scope has been reduced.

### Deferred from v1

- **Agent v2 execution tracing**
  - Agent run / tool-call / tool-result sub-spans
  - Agent App tracing callbacks and two-phase Human Input correlation
  - Agent tool-inner parent-context propagation on both the Dify API and `dify-agent` sides
  - The Agent v2 Monitoring tracing configuration panel and related translations
- **Human Input lifecycle tracing**
  - Workflow/Agent App human-wait spans
  - Pause/resume private tracing state (`WorkflowTraceState`)
  - Human Input lifecycle timestamp normalization and wait reconstruction
  - Global-timeout final-trace handoff and recovery
- **Database changes for final trace handoff**
  - Removed the `workflow_pauses.final_trace_status` / `final_trace_attempts` migration and model fields
  - Removed the dedicated final-handoff retry configuration and task/repository handling

The deferred designs are still documented in ADR-0001 under **“Out of scope (v1): deferred designs”**, but they are no longer normative or wired into v1 producer paths.

### Temporary/unrelated changes removed

- Reverted the temporary frontend change that allowed Human Input nodes inside Loop/Iteration containers.
- Reverted temporary `Makefile`, `dev/start-agent`, Docker local-agent setup, SSRF proxy, and `dify-agent/.example.env` changes.
- Dropped the branch-local embedding/model-type workaround during rebase because upstream already contains the equivalent (and cleaner) `_model_type_db_literals()` implementation.

### Still in scope

- Provider-neutral canonical workflow/Chatflow traces
- Deterministic parent-first hierarchy reconstruction
- Existing non-nested Loop/Iteration synthetic wrapper topology
- Nested-workflow and standalone Message-child parent-context coordination
- Unified/legacy runtime isolation and provider routing
- Durable whole-fragment, at-least-once provider export retry
- Conversation-name trace association via `message_id`

### Core/DB boundary

ADR-0001 now states explicitly that v1 introduces:

- **no Agent/Human Input tracing changes to `api/core/` outside `api/core/ops/`**, and
- **no database migration**.

The retained conversation-name `message_id` propagation is documented as an independent, general trace-correlation improvement rather than Agent/Human Input tracing work.

The updated branch has been force-pushed to this PR after rebase, and the follow-up revert/documentation commits are included in the PR head.

api/configs/feature/__init__.py
```diff
@@ -1336,16 +1337,35 @@ class NewAgentBetaConfig(BaseSettings):
...
         default=5,
     )
 
+    OPS_TRACE_PARENT_CONTEXT_TTL_SECONDS: PositiveInt = Field(
+        description="Retention in seconds for unified tracing parent contexts.",
+        default=1800,
+    )
+
+    @model_validator(mode="after")
+    def validate_parent_context_retention(self) -> "OpsTraceConfig":
```
laipz8200, line 1363: The validation still blocks startup due to incorrect configuration even when the flag is set to false.

api/core/ops/unified_trace/trace_builder.py
```diff
@@ -0,0 +1,450 @@
...
+            case _:
+                return None
+
+    def _build_workflow(self, trace_info: WorkflowTraceInfo) -> CanonicalTrace:
+        executions = self._load_workflow_executions(trace_info)
+        hierarchy = build_workflow_hierarchy(executions)
+        workflow_data = trace_info.workflow_data
+        workflow_start = _started_at(_read_attribute(workflow_data, "created_at") or trace_info.start_time)
+        workflow_end = _read_attribute(workflow_data, "finished_at") or trace_info.end_time
+        root_id = trace_info.message_id or trace_info.workflow_run_id
```
laipz8200: We should not confuse the message ID and the workflow run ID.

api/tests/unit_tests/providers/trace/unified_trace/test_arize_phoenix.py
laipz8200, line 1: Test code should not be located in the current directory.

api/tests/unit_tests/providers/trace/unified_trace/test_langsmith.py
laipz8200, line 1: Test code should not be located in the current directory.

api/providers/trace/trace-langsmith/src/dify_trace_langsmith/unified_trace.py
```diff
@@ -0,0 +1,176 @@
...
+                    end_time=canonical_span.end_time,
+                    error=canonical_span.error if canonical_span.status is CanonicalSpanStatus.ERROR else None,
+                    extra={"metadata": metadata},
+                    tags=["dify", "synthetic" if canonical_span.synthetic else "execution"],
+                    parent_run_id=parent_run_id,
+                    trace_id=trace_id,
+                    dotted_order=dotted_order,
+                    session_name=self._project_name,
+                )
+            except (
```
laipz8200, line 148: LangSmithError has not been catched.

laipz8200 (changes requested):

Blackoutta:
# Spec: Address unified tracing review feedback

## Problem Statement

The unified tracing pull request is intended to be opt-in and backwards compatible, but its current implementation introduces four review concerns that weaken that contract: disabled unified tracing can still block application startup through a feature-specific cross-field validation; Message identity and workflow-run identity are expressed through one fallback variable even though they represent different canonical entities; the new unified provider adapter tests do not follow the agreed centralized ownership layout; and the LangSmith adapter does not explicitly translate the complete LangSmith SDK exception boundary into retryable and terminal Core outcomes.

These issues should be resolved without changing the canonical topology recorded in the unified tracing ADR, broadening the feature scope, or making permanent configuration failures consume the transient retry budget.

## Solution

Polish the unified tracing implementation at its existing boundaries:

- Run the retry-window versus parent-context-retention validation only when unified tracing is enabled.
- Preserve separate Message and workflow-run identities in the canonical builder while retaining the ADR-defined topology: a Chatflow fragment uses its Message span as the fragment root and places its workflow span below it; a pure Workflow fragment uses the workflow-run span as its root.
- Centralize the Phoenix and LangSmith unified adapter tests under the API unified tracing provider test area.
- Explicitly classify LangSmith SDK failures. Known transient transport or service failures become the Core retryable dispatch signal. Authentication and user/configuration failures become terminal adapter failures. Any remaining `LangSmithError` is caught at the adapter boundary and conservatively treated as terminal.

## User Stories

1. As a self-hosted Dify operator with unified tracing disabled, I want unrelated unified tracing retention settings to be ignored by the cross-field validator, so that an opt-in feature cannot prevent my existing deployment from starting.
2. As a self-hosted Dify operator enabling unified tracing, I want inconsistent retry and parent-context retention settings to fail during configuration loading, so that nested trace parent context cannot expire before the configured retry window ends.
3. As a Dify maintainer, I want field-level configuration parsing to remain intact, so that malformed configuration values are still reported consistently with other Dify settings.
4. As a Dify maintainer, I want the feature flag to gate only the feature-specific relationship constraint, so that the fix remains minimal and backwards compatible.
5. As a Chatflow user, I want the Message span to remain the canonical fragment root, so that Message-owned child traces and provider parent-context publication retain their designed hierarchy.
6. As a Workflow user, I want a workflow without a Message to retain the workflow-run span as its canonical root, so that pure Workflow traces remain stable.
7. As a nested-workflow user, I want workflow-run identity to remain distinct from Message identity, so that external parent resolution continues to use the correct workflow and node execution concepts.
8. As a tracing provider adapter author, I want canonical Message and workflow-run identifiers to be expressed separately, so that provider translation cannot accidentally treat one domain identity as the other.
9. As a Dify architect, I want the implementation to remain conformant with the unified tracing ADR, so that review polishing does not silently redefine canonical topology.
10. As a Dify contributor, I want unified Phoenix and LangSmith adapter tests in one centralized API test area, so that the ownership of the cross-Core/provider contract is obvious.
11. As a Dify contributor, I want the existing unit-test runner to continue discovering the moved adapter tests, so that reorganizing them cannot reduce coverage.
12. As a Dify reviewer, I want known transient LangSmith failures to be mapped to `RetryableTraceDispatchError`, so that Celery retries only failures that may succeed with the same payload and configuration.
13. As a Dify operator, I want LangSmith connection failures, request timeouts, rate limits, and server-side API failures to remain retryable, so that temporary provider outages do not silently discard durable trace payloads.
14. As a Dify operator, I want LangSmith authentication failures to terminate immediately, so that an invalid or expired credential does not generate hundreds of identical requests.
15. As a Dify operator, I want LangSmith user or configuration errors to terminate immediately, so that requests which require a configuration or input change do not consume the transient retry budget.
16. As a Dify maintainer, I want authentication and user errors to be caught explicitly, so that the retry policy is visible in the adapter instead of being an accidental consequence of exception inheritance.
17. As a Dify maintainer, I want unforeseen `LangSmithError` subclasses to be caught at the provider boundary, so that provider SDK exceptions do not leak into Core as an undocumented contract.
18. As a Dify maintainer, I want unknown LangSmith SDK failures to default to terminal handling, so that a future SDK exception cannot create an unbounded or noisy retry pattern.
19. As a Dify debugger, I want translated terminal and retryable failures to preserve their original exception as the cause, so that logs retain the provider-level diagnostic information.
20. As a Dify user, I want tracing export failures to remain isolated from the Workflow or Message business outcome, so that observability failures do not change application execution results.

## Implementation Decisions

- The unified tracing enable flag gates the cross-field relationship between retry count, retry delay, and parent-context TTL. When the feature is disabled, that relationship is not enforced.
- Existing positive-integer field types remain unchanged. This work does not make arbitrary malformed values valid merely because the feature is disabled.
- When unified tracing is enabled, the parent-context TTL must still cover the complete configured retry window.
- Message identity and workflow-run identity remain separate named concepts throughout canonical Workflow/Chatflow construction.
- The canonical topology remains unchanged and follows the accepted ADR: Chatflow Message root, workflow child, then workflow node executions; pure Workflow workflow-run root, then node executions.
- The canonical trace schema, provider parent-context envelope, session resolution, deterministic provider identity mapping, and external-parent behavior remain unchanged.
- The Phoenix and LangSmith unified adapter unit tests move to the centralized API unified tracing provider test area. Existing legacy provider tests are not reorganized as part of this work.
- The LangSmith adapter explicitly catches known transient exceptions—server API failure, connection failure, rate limiting, and request timeout—and translates them to the existing Core retryable dispatch exception.
- The LangSmith adapter explicitly catches authentication and user/configuration exceptions and translates them to a terminal runtime failure without using the retryable Core exception.
- A final `LangSmithError` boundary catches remaining SDK-specific failures and treats them as terminal by default.
- Exception chaining is preserved for every translation.
- No new terminal Core exception hierarchy is introduced. The existing generic terminal-failure handling in the trace task is sufficient.
- Authentication and user/configuration errors are not retried merely to allow a possible operator configuration change during the retry window. Configuration recovery is external intervention, not transient request recovery.

## Testing Decisions

- Tests assert externally observable contracts rather than local variable names or exception implementation details.
- The configuration-model seam verifies that an inconsistent retry window is accepted when unified tracing is disabled, rejected when it is enabled, and accepted when enabled with sufficient retention.
- The canonical-builder seam verifies both supported root topologies: Chatflow uses the Message span as root with the workflow-run span as its child; pure Workflow uses the workflow-run span as root.
- Existing canonical identity and parent-first ordering tests remain the prior art for the builder assertions.
- The provider-adapter `emit` seam verifies classification at the highest boundary that directly observes SDK exceptions and Core outcomes.
- LangSmith connection, timeout, rate-limit, and API/server errors must produce `RetryableTraceDispatchError`.
- LangSmith authentication and user/configuration errors must produce a terminal translated runtime failure and must not be instances of `RetryableTraceDispatchError`.
- A direct base `LangSmithError` must also produce a terminal translated runtime failure, proving the SDK boundary is exhaustive.
- Provider adapter tests continue to verify that parent context is published only after provider acceptance succeeds.
- The moved Phoenix and LangSmith tests must run through the repository's existing API unit-test command and require no new test framework, fixture layer, or abstraction.

## Out of Scope

- Changing the ADR-defined Chatflow or Workflow canonical topology.
- Making the workflow-run span the root of every Chatflow fragment.
- Changing trace IDs, session IDs, deterministic provider IDs, parent-context envelopes, Redis namespaces, or durable payload semantics.
- Migrating additional tracing providers to the unified runtime.
- Retrying LangSmith authentication, user/configuration, not-found, conflict, or unknown base errors.
- Designing operator-driven credential reload or delayed configuration-recovery workflows.
- Reorganizing existing legacy provider test suites beyond the two new unified adapter suites.
- Adding a new Core terminal trace-dispatch exception type.
- Changing retry counts, retry delays, or default parent-context TTL values.
- Database migrations, public API changes, or environment variable renames.

## Further Notes

- This specification addresses the unresolved review threads from `laipz8200` on PR #39451.
- The identity clarification is an implementation-expression change, not a topology change. Any request to make the workflow-run span the Chatflow root would contradict the current unified tracing ADR and requires an explicit architecture decision.
- Retryability is defined by whether the same operation, with the same payload and configuration, has a reasonable chance of succeeding later. Authentication and user/configuration failures do not meet that criterion.

docs/architecture/adr/unified-tracing/0001-unified-tracing-runtime-spec.md
laipz8200, line 1: We haven't reached a consensus yet on introducing the ADR mechanism. I think we should hold off on adding that directory for now.

laipz8200 (approved):

PR #39491, opened 2026-07-24
fix: don't close the active comment panel when a different comment is resolved

### LINKED ISSUE #39475: Workflow comment panel snaps back to older comment after resolve/edit/reply fetch
## Workflow comment panel snaps back to older comment after resolve/edit/reply fetch

When a user has selected a comment in the workflow panel and then clicks a different comment while a resolve/edit/reply fetch for the first comment is still in flight, the panel briefly snaps back to the first comment when the slower fetch resolves.

### Steps to reproduce

1. Open a workflow with at least two unresolved comments (A and B).
2. Click comment A — its detail panel opens.
3. Quickly click comment B — its detail panel opens.
4. From B's panel, trigger any action that calls `refreshActiveComment` for A (e.g. resolve A via the API call that the panel makes, or simply have another collaborator resolve A via the collaboration listener).
5. When the refresh for A resolves, the panel jumps back to A even though B is the user's current selection.

### Expected

The active-comment panel should remain on B, the user's latest selection.

### Actual

A slower in-flight refresh for A overwrites the active comment, snapping the panel back to A.

### Cause

`refreshActiveComment` in `web/app/components/workflow/hooks/use-workflow-comment.ts` unconditionally calls `setActiveComment(detail)` once the fetch resolves. Every other place in the same file that calls `setActiveComment` (`handleCommentIconClick` line 309, the user reloader line 432, and the cache fallback line 458) guards with `activeCommentIdRef.current === commentId` first. `refreshActiveComment` is the only path missing that guard.

### Suggested fix

```diff
-      setActiveComment(detail)
+      // Only apply the fetched detail if this comment is still the active one;
+      // otherwise a slower in-flight refresh can clobber a newer selection.
+      if (activeCommentIdRef.current === commentId) setActiveComment(detail)
```

The guard uses the `activeCommentIdRef` already declared at line 78, so no new state/refs are needed and the deps array on the `useCallback` stays as-is (refs are stable across renders).

### Notes

- Reproduction requires concurrent user interaction; trivially reachable in any multi-comment workflow with collaboration enabled.
- I have only read the code; I have not reproduced this in the running app.

dparkmit24:
> [!IMPORTANT]
>

## Summary

Fixes #39475.

The diagnosis and the fix in this PR are @AlexMultiAgent's, from #39475: `refreshActiveComment` in `web/app/components/workflow/hooks/use-workflow-comment.ts` calls `setActiveComment(detail)` unconditionally, whereas every other `setActiveComment` site in the file first guards on `activeCommentIdRef.current === commentId`. This PR applies their proposed guard and adds a regression test; my contribution is reproducing the bug in the running app and verifying the fix.

**Reproduction** (self-hosted 1.16.0, collaboration enabled; deterministic, 3× consecutive, no network throttling):
1. Add two comments A and B on a workflow canvas.
2. Select B — B's detail panel opens.
3. Resolve A from the comment list.
4. B's detail panel disappears; A's panel does not open.

(Control: resolving A while A's own panel is open closes A's panel — correct.)

**Why it presents as the panel disappearing, rather than the snap-back described in #39475:** `handleCommentResolve` calls `refreshActiveComment(commentId)` with the resolved comment's id, so `setActiveComment` flips the active detail to A. The detail popup only renders where `activeComment?.id === comment.id` over `visibleComments` (`web/app/components/workflow/index.tsx`), and `visibleComments` filters out resolved comments — so A's marker is removed once it is resolved (A's panel can't render) and B's marker no longer matches the now-A active detail (B's panel unmounts). Same unguarded `setActiveComment`; the trigger comment is simply resolved rather than still visible.

**The fix** applies the fetched detail only when its comment is still the active one:
```diff
-      setActiveComment(detail)
+      // Only apply the fetched detail if this comment is still the active one;
+      // otherwise a refresh for another comment can clobber the current selection.
+      if (activeCommentIdRef.current === commentId) setActiveComment(detail)
```
Refreshes of the active comment are unaffected (covered by a test).

**Verified** on from-source builds of current `main`: the before/after recordings below are the same app and sequence, with and without this change.

**Not established:** this was traced from source and reproduced in the UI, but the trace is static — I did not instrument the running bundle. The `collaborationManager.emitCommentsUpdate` local-loopback path and the `showResolvedComments` toggle were not separately exercised; with "show resolved" on, A's marker would remain and the symptom would instead match #39475's snap-back, further indicating the same root cause.

Frontend-only change, so I ran `cd web && pnpm exec vp lint` and `tsc --noEmit` rather than the backend `make lint && make type-check`; the changed files are clean.

From Claude Code

## Videos
Click the links below to see videos of the bug and its fix.

**Before**
[Comment B's detail panel is open; resolving comment A from the list closes B's panel, and A's panel does not open.](https://github.com/user-attachments/assets/d7d58eb2-ee27-477f-b75d-94f803faf829)

**After**
[Same app and sequence, built from this branch: resolving A leaves B's detail panel open.](https://github.com/user-attachments/assets/1ec0644a-eb23-44b4-a19e-e321078a4ce6)

## Checklist

### DIFF
--- a/web/app/components/workflow/hooks/__tests__/use-workflow-comment.spec.ts
+++ b/web/app/components/workflow/hooks/__tests__/use-workflow-comment.spec.ts
@@ -669,4 +669,66 @@ describe('useWorkflowComment', () => {
 
     expect(store.getState().pendingComment).toBeNull()
   })
+
+  it('does not overwrite the active comment when a different comment is resolved/refreshed', async () => {
+    const commentA = baseComment()
+    const commentB: WorkflowCommentList = {
+      ...baseComment(),
+      id: 'comment-2',
+      content: 'second',
+      position_x: 50,
+      position_y: 80,
+    }
+    const activeBDetail = { ...baseCommentDetail(), id: commentB.id, content: 'B detail' }
+    // refreshActiveComment(A) — reached via resolving A — fetches A's detail:
+    mockFetchWorkflowComment.mockResolvedValue({
+      ...baseCommentDetail(),
+      id: commentA.id,
+      content: 'A detail (must not clobber B)',
+    })
+    mockResolveWorkflowComment.mockResolvedValue({})
+    mockFetchWorkflowComments.mockResolvedValue({ data: [{ ...commentA, resolved: true }, commentB] })
+
+    const { result, store } = renderWorkflowHook(() => useWorkflowComment(), {
+      queryClient: createSeededQueryClient(),
+      initialStoreState: {
+        comments: [commentA, commentB],
+        activeCommentId: commentB.id,
+        activeCommentDetail: activeBDetail,
+      },
+    })
+
+    await act(async () => {
+      await result.current.handleCommentResolve(commentA.id)
+    })
+
+    // B is still the selected comment; A's fetched detail must not replace it.
+    expect(store.getState().activeCommentId).toBe(commentB.id)
+    expect(store.getState().activeCommentDetail?.id).toBe(commentB.id)
+  })
+
+  it('still refreshes the active comment', async () => {
+    const commentA = baseComment()
+    mockFetchWorkflowComment.mockResolvedValue({
+      ...baseCommentDetail(),
+      id: commentA.id,
+      content: 'refreshed content',
+    })
+
+    const { result, store } = renderWorkflowHook(() => useWorkflowComment(), {
+      queryClient: createSeededQueryClient(),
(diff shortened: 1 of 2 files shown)


dparkmit24:
Closing in favour of #39486, which came first and is the author's own fix for their issue. Repro and before/after recordings are in this PR's description if useful.

crazywoola (approved):

PR #39506, opened 2026-07-24
fix(workflow): guard refreshActiveComment against stale active-comment overwrite

### LINKED ISSUE #39475: Workflow comment panel snaps back to older comment after resolve/edit/reply fetch

AlexMultiAgent:
## Summary

Fixes the workflow comment panel snapping back to a stale selection when an in-flight refresh for an older comment resolves after the user has clicked a newer one. Adds the `activeCommentIdRef.current === commentId` guard that every other `setActiveComment` call in the same hook already uses.

Fixes #39475

## Changes

- `web/app/components/workflow/hooks/use-workflow-comment.ts`: guard `refreshActiveComment`'s `setActiveComment(detail)` call so it only applies if the comment is still the active one.

## Why

Every other call site in this hook — `handleCommentIconClick` (line 309), the user reloader (line 432), and the cache fallback (line 458) — already guards with `activeCommentIdRef.current === commentId` to prevent stale fetches from clobbering newer selections. `refreshActiveComment` was the only path missing that guard, which is why a slower resolve/edit/reply fetch for a previously-active comment could snap the panel back to it.

## Checklist

## Notes for reviewers

- **No live-app testing.** I verified the fix statically against the surrounding code; reproduction in the running app is straightforward but I did not exercise it.
- **Lint/type-check status.** `pnpm check`/`vp staged` are not configured in a way I could run end-to-end in the sandbox (this repo pins Node 22 via `devEngines` and the sandbox has Node 25; `vp check` and `tsc` therefore either error on the toolchain or exceed the sandbox timeout on the full web project). The change is a single-line guard that mirrors verbatim CI-passing code at lines 309/432/458, so type-safety is preserved by construction. Happy to follow up with a test if the maintainers want one added.

From an automated review agent

### DIFF
--- a/web/app/components/workflow/hooks/use-workflow-comment.ts
+++ b/web/app/components/workflow/hooks/use-workflow-comment.ts
@@ -98,7 +98,9 @@ export const useWorkflowComment = () => {
         [commentId]: detail,
       }
       setCommentDetailCache(commentDetailCacheRef.current)
-      setActiveComment(detail)
+      // Only apply the fetched detail if this comment is still the active one;
+      // otherwise a slower in-flight refresh can clobber a newer selection.
+      if (activeCommentIdRef.current === commentId) setActiveComment(detail)
     },
     [appId, setActiveComment, setCommentDetailCache],
   )


AlexMultiAgent:
Closing as superseded by #39491 (merged). The same guard is now in upstream main. Thanks to dparkmit24 for the fix.

PR #39519, opened 2026-07-24
fix(web): re-fetch images on prop change without dropping their blob URLs

### LINKED ISSUE #39518: fix(web): re-fetch images on prop change without dropping their blob URLs
## ImagePreviewer never loads new images when the prop changes while the component stays mounted

`web/app/components/datasets/common/image-previewer/index.tsx` lines 97-112 — the loading `useEffect` has an empty dependency array but reads the `images` prop. So if the previewer instance is reused for a different document or a different set of images, the new images are never fetched. The user can land on a stuck loading state or a 'no image' placeholder while the actual images exist.

A naive fix that simply adds `images` to the dep array is also broken: the same `useEffect` returns a cleanup that revokes every cached blob URL and clears the entire cache, so on every change the cleanup would tear down URLs still needed by the new image list.

### Steps to reproduce

1. Open the image previewer on document A.
2. Without unmounting the previewer, switch to document B.
3. Observe: document B's images never load, and the cache for document A's URLs is wiped if the effect re-runs.

### Expected

When the `images` prop changes, the previewer fetches the new set of images. Blob URLs of still-present images are reused, not revoked. Cleanup runs on unmount only.

### Actual

The loading effect runs only once on mount. New images are not fetched. Any naive dep-array fix would also wipe cached blob URLs on every change, breaking subsequent previews.

### Suggested fix

Split into two effects:
- Loading effect with `[images]` deps, no cleanup.
- Unmount-only effect with `[]` deps that handles `isMounted.current` and the cache cleanup.

AlexMultiAgent:
## Summary

- `web/app/components/datasets/common/image-previewer/index.tsx`: split the loading + cleanup effect into two — a `[images]`-deps fetch effect that re-runs on prop changes, and an unmount-only effect that revokes blob URLs and clears the cache.
- `web/app/components/datasets/common/image-previewer/__tests__/index.spec.tsx`: new test `refetches when the images prop changes without releasing cached blob URLs` — renders with one set, waits for them to load, re-renders with a different set, asserts the new fetches fire AND existing blob URLs are not released until unmount.

Fixes #39518.

## Why

The original effect was a single block with `[]` deps that both fetched and (on cleanup) revoked every cached blob URL. Reusing the previewer for a different image list never fetched the new images; naively adding `images` to the deps would wipe URLs still needed by the new list.

## Test Plan

- `pnpm --dir web test run app/components/datasets/common/image-previewer/__tests__/index.spec.tsx`

### DIFF
--- a/web/app/components/datasets/common/image-previewer/__tests__/index.spec.tsx
+++ b/web/app/components/datasets/common/image-previewer/__tests__/index.spec.tsx
@@ -155,4 +155,30 @@ describe('ImagePreviewer', () => {
 
     await waitFor(() => expect(mockRevokeObjectURL).toHaveBeenCalledTimes(3))
   })
+
+  it('refetches when the images prop changes without releasing cached blob URLs', async () => {
+    const initialImages = [
+      { url: 'https://example.com/image1.png', name: 'image1.png', size: 1024 },
+      { url: 'https://example.com/image2.png', name: 'image2.png', size: 2048 },
+    ]
+    const nextImages = [
+      { url: 'https://example.com/image3.png', name: 'image3.png', size: 3072 },
+      { url: 'https://example.com/image4.png', name: 'image4.png', size: 4096 },
+    ]
+
+    const { rerender } = render(<ImagePreviewer images={initialImages} onClose={vi.fn()} />)
+    expect(await screen.findByRole('img', { name: 'image1.png' })).toBeInTheDocument()
+    expect(mockFetch).toHaveBeenCalledTimes(2)
+    expect(mockRevokeObjectURL).not.toHaveBeenCalled()
+
+    rerender(<ImagePreviewer images={nextImages} onClose={vi.fn()} />)
+    await waitFor(() => expect(mockFetch).toHaveBeenCalledTimes(4))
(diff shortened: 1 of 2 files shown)


AlexMultiAgent:
> CI note for reviewers: this PR (and every other open PR from this fork) is blocked at GitHub's first-time-contributor workflow approval gate.

Latest Main CI Pipeline run: **[30418111076](https://github.com/langgenius/dify/actions/runs/30418111076)** — `conclusion: action_required`, **zero jobs executed**. Three earlier runs on this PR (and analogous runs on the other open PRs from `AlexMultiAgent`) hit the same gate; `mergeable_state` reports `blocked` regardless of labeler / pyrefly / semantic-PR passing.

A maintainer needs to click **"Approve & run"** on any one queued workflow from `AlexMultiAgent` (or check "Allow all future runs from this contributor") to unblock CI for the entire fork — once approved, all subsequent secret-using runs auto-approve per GitHub's first-time-contributor model. I pushed an empty commit to refresh the run timestamp; the previous 3 runs were 4 days old and had scrolled out of the Actions tab.

Happy to address review feedback as soon as CI is green. Apologies for the noise.

PR #39523, opened 2026-07-24
fix(security): harden AgentDriveArchiveMemberApi signed-URL verification

### LINKED ISSUE #39522: fix(security): harden AgentDriveArchiveMemberApi signed-URL verification
## AgentDriveArchiveMemberApi signed-URL verification is non-constant-time, accepts future timestamps, and has no replay protection

`api/services/agent_drive_service.py` `verify_archive_member_signature` is the sole authorization on the unauthenticated `GET /files/agent-drive/archive-member` endpoint. Three weaknesses in the verification together let an attacker who intercepts a signed URL (logs, browser history, referer leaks, accidental sharing) read the underlying archive bytes indefinitely and also brute-force the signature itself.

### Issues

1. **Non-constant-time signature comparison** (`agent_drive_service.py:1206`). The verification uses `sign != cls._sign_archive_member_payload(payload)`, which short-circuits on the first mismatching byte and leaks signature bytes via timing. The signature can be brute-forced one byte at a time from a network-adjacent position.

2. **No lower-bound timestamp check** (`agent_drive_service.py:1209`). Only the upper bound is enforced (`current_time - int(timestamp) <= FILES_ACCESS_TIMEOUT`). A captured URL with a forged far-future timestamp stays valid indefinitely. There is no check that the timestamp is not in the future, even with a small clock-skew allowance.

3. **No replay protection.** The `nonce` field is part of the signed payload but is never recorded server-side, so a captured URL is fully replayable for the entire access window.

### Attack scenarios

- Intercept any signed archive-member URL — from server access logs, browser history, accidental sharing, or referer leakage — and reuse it indefinitely. Default `FILES_ACCESS_TIMEOUT` is 300 seconds; the lack of nonce tracking means a single interception is enough.
- Brute-force the signature one byte at a time by measuring response time. The endpoint is unauthenticated, so the attacker can probe from any IP.

### Fix

1. Replace `sign != ...` with `hmac.compare_digest(sign, ...)`.
2. Reject timestamps that are more than a small clock-skew window in the future, in addition to the existing upper-bound check.
3. Record the nonce in Redis with the same TTL as the access window using `SET ... NX` for atomic first-use semantics; repeat attempts find the key already present and are rejected.

AlexMultiAgent:
## Summary

- `api/services/agent_drive_service.py`:
  - import `redis_client` from `extensions.ext_redis`
  - module-level constants `_SIGNATURE_CLOCK_SKEW_SECONDS` and `_NONCE_KEY_PREFIX`
  - in `verify_archive_member_signature`: use `hmac.compare_digest` for the signature comparison, reject timestamps more than 30 s in the future, and record the nonce in Redis with TTL `FILES_ACCESS_TIMEOUT + 30` via `SET ... NX` for atomic first-use semantics.
- `api/tests/unit_tests/services/test_agent_drive_service.py`: four tests covering tampered signature, far-future timestamp, signature within the skew window, and replayed nonce.

Fixes #39522.

## Why

`GET /files/agent-drive/archive-member` has no auth other than the signed URL. The prior signature check used `!=` (timing-leaking), accepted any non-negative timestamp (forged far-future timestamps stayed valid indefinitely), and never recorded the nonce (any captured URL was replayable for the entire access window).

## Test Plan

- `uv run --project api pytest "api/tests/unit_tests/services/test_agent_drive_service.py::test_verify_archive_member_signature_*" -q`

### DIFF
--- a/api/services/agent_drive_service.py
+++ b/api/services/agent_drive_service.py
@@ -41,6 +41,7 @@ from sqlalchemy.orm import Session
 
 from configs import dify_config
 from core.app.file_access.controller import DatabaseFileAccessController
+from extensions.ext_redis import redis_client
 from extensions.ext_storage import storage
 from factories import file_factory
 from libs.uuid_utils import uuidv7
@@ -53,6 +54,8 @@ logger = logging.getLogger(__name__)
 _MAX_KEY_LENGTH = 512
 _DRIVE_REF_PREFIX = "agent-"
 _SKILL_MD_SUFFIX = "/SKILL.md"
+_SIGNATURE_CLOCK_SKEW_SECONDS = 30
+_NONCE_KEY_PREFIX = "agent_drive_archive_nonce:"
 _SKILL_ARCHIVE_NAME = ".DIFY-SKILL-FULL.zip"
 _ARCHIVE_MEMBER_DOWNLOAD_PURPOSE = "agent-drive-archive-member"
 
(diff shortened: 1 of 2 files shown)


AlexMultiAgent:
Brief re-submission note: the long-form context moved into the PR description. (Edited by author.)

AlexMultiAgent:
Quick bump — was this PR missed in the security-review queue?

`GET /files/agent-drive/archive-member` has no auth other than the signed URL. This PR closes three independent bypasses on `verify_archive_member_signature` (in `api/services/agent_drive_service.py`), each with its own focused regression test:

- **Timing-leaking compare** — `sign != cls._sign_archive_member_payload(...)` → `hmac.compare_digest`
- **Future-timestamp accepted** — only the past direction is constrained by `FILES_ACCESS_TIMEOUT`, so any forged far-future timestamp passes indefinitely
- **No nonce replay protection** — captured URLs are valid for the full `FILES_ACCESS_TIMEOUT` access window, since the nonce is never recorded as used

Each issue lives in its own test (`test_verify_archive_member_signature_rejects_tampered_signature`, `..._rejects_far_future_timestamp`, `..._accepts_signature_within_skew_window`, `..._rejects_replay_of_captured_nonce`) so review can be scoped to one at a time if useful.

Production diff is intentionally minimal: `redis_client` import, two module constants (`_SIGNATURE_CLOCK_SKEW_SECONDS`, `_NONCE_KEY_PREFIX`), and the hardened `verify_archive_member_signature`. Happy to split this into three smaller PRs (one per issue) if that's easier to review, or address any feedback inline. Let me know what works.

PR #39532, opened 2026-07-24
fix(agent): block unpublished agents in workflows

zyssyz123:
## Summary

- Treat an app-backed roster Agent as workflow-callable only when its active snapshot is backed by a publish/save revision.
- Preserve existing behavior for a published active snapshot when a newer unpublished draft exists.
- Enforce the rule in invite candidates, draft binding sync, publish validation, runtime binding resolution, and copy-from-roster.

## Root cause

`active_config_is_published` compares the active snapshot with the mutable draft, so `false` means either never published or published with later draft edits. Workflow paths only checked that an active snapshot existed and therefore allowed never-published imported Agent snapshots to be selected and executed. The fix derives callability from revision provenance of the active snapshot instead.

## Validation

- 706 related Agent and Workflow unit tests passed.
- Ruff format/check passed for all changed files.
- Pyrefly reported 0 errors; targeted Mypy passed for all 6 production files.
- `deploy/agent` commit: ea88aa8b0945f228a0ceff6d06413fc3eb129356.
- Dev build: https://github.com/langgenius/dify/actions/runs/30081615142
- Latest containing build: https://github.com/langgenius/dify/actions/runs/30081899235
- Latest dev deploy: https://github.com/langgenius/dify/actions/runs/30082136597
- On dev, a real never-published imported Agent is absent from invite candidates and rejected by runtime resolution, publish validation, draft binding sync, and copy-from-roster; the rejected copy transaction produced no writes.
- A published Agent with later draft edits remains selectable and both existing runtime bindings resolve its published active snapshot.
- Public root/system-features/version and internal API health return 200; no post-deploy 5xx, restart, OOM, or unexpected application error was observed.
- PR-only E2E synchronization waits for the apps query to resolve before the WebKit keyboard menu assertion; this addresses the pre-hydration CI race and does not require another dev deployment.

Linear: https://linear.app/dify/issue/DIFY-2837/未发布的-agent-依然可以被-workflow-调用

### DIFF
--- a/e2e/features/step-definitions/accessibility/keyboard-navigation.steps.ts
+++ b/e2e/features/step-definitions/accessibility/keyboard-navigation.steps.ts
@@ -72,6 +72,8 @@ When('I open and close the create app menu using the keyboard', async function (
   const page = this.getPage()
   const trigger = getCreateAppMenuTrigger(this)
 
+  await expect(page.getByRole('heading', { name: 'Build your first App' })).toBeVisible()
+  await expect(trigger).toHaveAttribute('aria-expanded', 'false')
   await trigger.press('Enter')
   await expect(trigger).toHaveAttribute('aria-expanded', 'true')
   await page.keyboard.press('Escape')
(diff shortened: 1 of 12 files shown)


e2e/features/step-definitions/accessibility/keyboard-navigation.steps.ts
```diff
@@ -72,6 +72,8 @@ When('I open and close the create app menu using the keyboard', async function (
   const page = this.getPage()
   const trigger = getCreateAppMenuTrigger(this)
 
+  await expect(page.getByRole('heading', { name: 'Build your first App' })).toBeVisible()
+  await expect(trigger).toHaveAttribute('aria-expanded', 'false')
```
lyzno1:
```suggestion
```

lyzno1 (approved):

PR #39549, opened 2026-07-24
fix(api): reject trailing newlines in valid_password() validator (#39548)

### LINKED ISSUE #39548: valid_password() accepts trailing newline/CRLF (auth surface bypass, sibling of #39234)
## Bug

`libs.password.valid_password()` accepts a password ending in a newline (LF or CRLF).

```python
>>> from libs.password import valid_password
>>> valid_password("Password123")
'Password123'
>>> valid_password("Password123\n")
'Password123\n'              # accepted
>>> valid_password("Password123\r\n")
'Password123\r\n'           # accepted (Windows line ending)
```

## Root cause

Same shape as the now-fixed #39234 (`email()` validator). `valid_password` uses `re.match` against a pattern anchored by `$`:

```python
# api/libs/password.py
password_pattern = r"^(?=.*[a-zA-Z])(?=.*\d).{8,}$"

def valid_password(password):
    pattern = password_pattern
    if re.match(pattern, password) is not None:
        return password
    raise ValueError(...)
```

In Python, `re.match`'s `$` matches at end-of-string or just before a trailing newline, so a password ending in `\n` (or `\r\n`) is accepted unchanged.

## Impact

`valid_password` backs every password-setting path on the auth surface (registration, password reset, password change, admin-set password). A value with a trailing newline is stored as-typed, which creates:

- **Silent password mismatch.** A user who copy-pastes a password with a trailing whitespace or line ending from a terminal may save `Password123\n` and be unable to log in with the same value from a normal form field. Login form inputs usually strip trailing whitespace, so the user is stuck.
- **Downstream injection risk.** Any code path that later assembles headers, logs, or shell commands from the raw stored value treats the trailing newline as a separator. This is the same primitive the email fix closed for the mail surface.
- **Cross-platform inconsistency.** `\r\n` is accepted; pure `\n` is accepted; the input is stored verbatim. Whichever client sanitized its own copy is the one that wins.

## Proposed fix

Switch `re.match` to `re.fullmatch` so the entire string must match. One-character change, mirroring the fix applied in #39234 for the email validator:

```python
if re.fullmatch(pattern, password) is not None:
    return password
```

`re.fullmatch` is strictly more restrictive than `re.match` and only rejects the trailing-newline case. All currently-valid passwords still validate.

## Alternatives considered

- **Strip the value first** (`.rstrip("\r\n")`): silently mutates the user input and breaks the security invariant that the value the user typed is the value that gets stored. Avoid.
- **Reject explicitly in the controller**: scatters the validation across callers, hides the bug class, and leaves any non-controller path (Celery tasks, console scripts) unguarded. The single fix in the validator is the right layer.
- **Sweep all `re.match`-with-`$` patterns across the repo** in one PR (helper.py alphanumeric, provider_ids, vdb-oceanbase, code_executor template_transformer, trace_id_helper): a real concern, but a separate, wider PR. Out of scope here; worth a follow-up issue.

## Scope

| File | Change |
|---|---|
| `api/libs/password.py` | `re.match` → `re.fullmatch` (line 13); 1-2 lines of comment referencing this issue |
| `api/tests/unit_tests/libs/test_password.py` | 3 new test cases: trailing LF, trailing CRLF, and one positive that a valid password is still accepted |

Estimated diff: ~6 lines added, ~1 line modified.

## Acceptance criteria

- `valid_password("Password123\n")` raises `ValueError`
- `valid_password("Password123\r\n")` raises `ValueError`
- All existing positive test cases in `test_password.py` still pass (no regression)
- `pytest api/tests/unit_tests/libs/test_password.py` is green
- `ruff check` and `ruff format --check` on the two touched files are clean

## Backward compatibility

Strictly more restrictive. Only passwords that include a trailing newline (or CRLF) change behavior, from "accepted" to "rejected". No currently-valid password is affected. No database migration; existing stored passwords with embedded trailing newlines remain valid for login (login is a hash-compare, not a re-validate), but new password-set/change attempts cannot create one.

## Risks

Effectively none. `re.fullmatch` is the documented Python equivalent of "match the whole string" and is already used elsewhere in the repo (e.g. the email fix from #39234). Behavior change is narrow and the impact of the existing behavior is negative.

## Related

- #39234 — same root cause for the `email()` validator (closed, fixed in #39320 / retained in #39235)
- `api/libs/helper.py:289` (`alphanumeric()`) — same `re.match`-with-`$` pattern, lower-impact (tool provider name validation, not auth), out of scope here

## Next step

Happy to open a PR with the one-line `re.match` → `re.fullmatch` change in `api/libs/password.py:13` and the test cases above. Same shape as the email fix already merged.

Harsh23Kashyap:
## Summary

Fixes #39548

`libs.password.valid_password()` validates with `re.match` against a pattern anchored by `$`. In Python `$` matches at end-of-string or just before a trailing newline, so a password ending in `\n` (or `\r\n`) slips through unchanged. The validator backs the auth surface (registration, password reset, password change, admin-set password), where a newline-bearing password creates a silent mismatch with form-stripped login input and a header/log-injection primitive for any downstream code that handles the raw value.

Switch `re.match` to `re.fullmatch` so the whole string must match. Mirrors the now-merged fix for the email validator (#39320) on the same family of `re.match`-with-`$` patterns.

```python
>>> from libs.password import valid_password
>>> valid_password("Password123")
'Password123'
>>> valid_password("Password123\n")
ValueError: Password must contain letters and numbers, and the length must be at least 8 characters.
>>> valid_password("Password123\r\n")
ValueError: Password must contain letters and numbers, and the length must be at least 8 characters.
```

## Why this is more than a style nit

1. **Silent password mismatch.** A user who copy-pastes a password with a trailing whitespace or line ending from a terminal may save `Password123\n` and be unable to log in with the same value from a normal form field. Login form inputs usually strip trailing whitespace, so the user is stuck.
2. **Header / log / shell injection primitive.** Any downstream code path that later assembles headers, logs, or shell commands from the raw stored value treats the trailing newline as a separator. Same primitive the email fix closed for the mail surface.
3. **Cross-platform inconsistency.** `\r\n` is accepted; pure `\n` is accepted; the input is stored verbatim. Whichever client sanitized its own copy is the one that wins.

## Changes

| File | Change |
|---|---|
| `api/libs/password.py` | `re.match` → `re.fullmatch` (line 13); 2 lines of comment referencing the issue |
| `api/tests/unit_tests/libs/test_password.py` | 2 new test cases (trailing LF, trailing CRLF) inside the existing `TestValidPassword` class |

14 lines added, 2 removed across 2 files.

## Verification

- `pytest tests/unit_tests/libs/test_password.py -v` → **9 passed** (7 existing + 2 new)
- `pytest tests/unit_tests/libs/` (full module) → **595 passed**, 16 warnings, 138 subtests
- `ruff check` → All checks passed
- `ruff format --check` → 2 files already formatted
- Manual sanity: 3 positive + 2 new negative (trailing LF, trailing CRLF) assertions all pass
- The 16 warnings are pre-existing `InsecureKeyLengthWarning`s from `test_passport.py`'s short HMAC keys, unrelated to this change.

## Design decisions

- **One-character behavior change.** `re.fullmatch` is strictly more restrictive than `re.match` and only rejects the trailing-newline case. All currently-valid passwords still match; no API change, no migration.
- **No `.rstrip()` shim.** Considered stripping the value first, but that silently mutates the user input and breaks the security invariant that what the user typed is what gets stored.
- **No sweep across other `re.match`-with-`$` patterns.** Considered including `helper.py:289` (`alphanumeric`), `provider_ids`, `vdb-oceanbase`, `code_executor/template_transformer`, `trace_id_helper` in one PR, but kept this one focused. Out of scope here; a follow-up issue is a better place.

## Backward compatibility

Strictly more restrictive. Only passwords that include a trailing newline (or CRLF) change behavior, from accepted to rejected. No currently-valid password is affected. No database migration; existing stored passwords with embedded trailing newlines remain valid for login (login is a hash-compare, not a re-validate), but new password-set attempts cannot create one.

## Risks

Effectively none. `re.fullmatch` is the documented Python equivalent of "match the whole string" and is already used elsewhere in the repo (the email fix in #39320). Behavior change is narrow and the impact of the existing behavior is negative.

## Out-of-scope observation

`api/libs/helper.py:289` (`alphanumeric()`) has the same `re.match`-with-`$` pattern. Lower-impact (tool provider name validation, not auth), so I kept it out of this PR. Worth a follow-up.

## Checklist

- [x] This change requires a documentation update, included: not needed; `valid_password` is an internal validator, no user-facing docs reference the trailing-newline case
- [x] I understand that this PR may be closed in case there was no previous discussion or issues — issue #39548 was filed this session
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change — one commit, two test cases for the one behavior change
- [ ] I've updated the documentation accordingly — N/A
- [x] I ran `make lint && make type-check` (backend) — `ruff check` + `ruff format --check` clean (this is a backend-only change; no frontend touch)

### DIFF
--- a/api/libs/password.py
+++ b/api/libs/password.py
@@ -9,8 +9,9 @@ password_pattern = r"^(?=.*[a-zA-Z])(?=.*\d).{8,}$"
 def valid_password(password):
     # Define a regex pattern for password rules
     pattern = password_pattern
-    # Check if the password matches the pattern
-    if re.match(pattern, password) is not None:
+    # Use re.fullmatch so a trailing newline (which re.match's $ accepts)
+    # cannot smuggle a raw "\n" or "\r\n" past the auth surface — see #39548.
+    if re.fullmatch(pattern, password) is not None:
         return password
 
     raise ValueError("Password must contain letters and numbers, and the length must be at least 8 characters.")
--- a/api/tests/unit_tests/libs/test_password.py
+++ b/api/tests/unit_tests/libs/test_password.py
@@ -42,6 +42,17 @@ class TestValidPassword:
 
         assert "at least 8" in str(exc_info.value)
 
+    def test_should_reject_password_with_trailing_newline(self):
+        """re.match anchors $ before a trailing newline, so a value ending in \\n slipped through.
+        re.fullmatch (the fix in #39548) rejects it like any other malformed password."""
+        with pytest.raises(ValueError, match="at least 8"):
+            valid_password("Password123\n")
+
+    def test_should_reject_password_with_trailing_crlf(self):
+        """Same root cause for Windows-style line endings."""
+        with pytest.raises(ValueError, match="at least 8"):
+            valid_password("Password123\r\n")
+
 
 class TestPasswordHashing:
     """Test password hashing and comparison"""


crazywoola:

### Why this is being closed

### Next steps

PR #39550, opened 2026-07-24
fix: use Dify's own doc_id instead of content hash for Weaviate object UUIDs

### LINKED ISSUE #39175: [Bug] Disabled Weaviate document vectors are not deleted by segment ID
### Self Checks

### Dify version

main (`61b07ab17d`)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Configure a knowledge base to use Weaviate.
2. Index a document with at least one segment.
3. Disable or delete the document in Dify.
4. Inspect the corresponding Weaviate collection or run retrieval without an active-document filter.

The Weaviate adapter stores each object with a UUID generated from `page_content`, while Dify passes the segment's logical `doc_id` to `delete_by_ids()`. The adapter then calls `delete_by_id(doc_id)`, so the object UUID does not match and the vector remains in Weaviate.

The content-derived UUID also makes segments with identical text collide across documents. This part was previously reported in #36137 and addressed by the still-open #36140, but that proposal does not address deletion by logical segment ID or existing content-UUID objects.

### ✔️ Expected Behavior

Disabling or deleting a document should remove every matching Weaviate object by its Dify segment `doc_id`. New objects should use the same `doc_id` identity contract as `BaseVector`, so equal text in different segments cannot overwrite another segment.

### ❌ Actual Behavior

`WeaviateVector._get_uuids()` derives object UUIDs only from content, but `delete_by_ids()` treats logical segment IDs as those object UUIDs. Missing-object responses are ignored, leaving disabled or deleted vectors in the collection. Identical segment text can also overwrite another segment because it produces the same object UUID.

FuHarrison:
Fixes #39175

## Problem

`WeaviateVector._get_uuids()` generates the UUID used as each object's
primary key in Weaviate by hashing the segment's `page_content` with
`uuid5` and a fixed namespace:

```python
def _get_uuids(self, documents: list[Document]) -> list[str]:
    URL_NAMESPACE = _uuid.UUID("6ba7b811-9dad-11d1-80b4-00c04fd430c8")
    uuids = []
    for doc in documents:
        uuid_val = _uuid.uuid5(URL_NAMESPACE, doc.page_content)
        uuids.append(str(uuid_val))
    return uuids
```

Meanwhile, `delete_by_ids()` in the same file looks segments up by
Dify's own `doc_id` (`DocumentSegment.index_node_id`, passed in via
`document.metadata["doc_id"]`). These two IDs are unrelated, causing:

1. **Disabling/deleting a document never actually removes its
   vectors from Weaviate** — `delete_by_ids()` gets 404s (silently
   swallowed), so vectors for disabled/deleted documents accumulate
   forever.
2. **Documents with identical segment content silently overwrite
   each other** — Weaviate's write is an upsert by UUID, so a
   byte-identical chunk in a different document replaces the
   original object.

## Fix

Use `doc.metadata["doc_id"]` as the Weaviate object UUID instead of
hashing content, falling back to a fresh random UUID only if `doc_id`
is missing or invalid. `delete_by_ids()` is unchanged — it was
already correct. This also brings `WeaviateVector` in line with the
default `_get_uuids()` already defined on the abstract base class
(`core/rag/datasource/vdb/vector_base.py`):

```python
def _get_uuids(self, texts: list[Document]) -> list[str]:
    return [text.metadata["doc_id"] for text in texts if text.metadata and "doc_id" in text.metadata]
```

## Relationship to existing PRs

- **#36140** combines `doc_id` into the content hash, which prevents
  collisions but leaves `delete_by_ids()` mismatched against the new
  hash, so deletion is still broken.
- **#37068** switches `delete_by_ids()` calls to
  `delete_by_metadata_field("doc_id", node_id)`, but only for the
  child-chunk update/delete call sites; the general document
  disable/delete path (`remove_document_from_index_task` /
  `clean_dataset_task`) still uses `delete_by_ids()` and remains
  affected. It also doesn't address write-time collisions.

This PR fixes both symptoms for the general document path in a
single, self-contained change to the Weaviate adapter, with no
changes needed to `vector_service.py` or other callers.

## Testing

Reproduced end-to-end locally with a clean Weaviate-backed
deployment and two documents that share one identical segment:

| Step | Before fix | After fix |
|---|---|---|
| Upload 2 docs, 5 segments each | 9 total objects (1 lost to collision) | 10 total objects (5 + 5) |
| Disable one document | Vector count unchanged | Drops to 0 |
| Re-enable it | N/A (never reached 0) | Back to 5 |
| Delete the other document | Orphaned vectors remain | Drops to 0 |

Added unit tests in `test_weaviate_vector.py` covering:
- `_get_uuids()` returns the given `doc_id` when present and valid
- Two documents with identical `page_content` but different `doc_id`
  produce different UUIDs (regression test for the collision bug)
- Falls back to a valid random UUID when `doc_id` is missing

## Screenshots

N/A — this is a backend logic fix with no UI changes. See the
before/after vector-count table above for the observable behavior
change.

## Checklist

### DIFF
--- a/api/providers/vdb/vdb-weaviate/src/dify_vdb_weaviate/weaviate_vector.py
+++ b/api/providers/vdb/vdb-weaviate/src/dify_vdb_weaviate/weaviate_vector.py
@@ -280,17 +280,20 @@ class WeaviateVector(BaseVector):
     @override
     def _get_uuids(self, documents: list[Document]) -> list[str]:
         """
-        Generates deterministic UUIDs for documents based on their content.
-
-        Uses UUID5 with URL namespace to ensure consistent IDs for identical content.
+        Use Dify's own doc_id (index_node_id, set in each Document's metadata by
+        the caller) as the Weaviate object UUID instead of hashing the page
+        content. This keeps write-time IDs consistent with what delete_by_ids()
+        looks up, and stops segments with identical text across different
+        documents from colliding on the same Weaviate object.
+        Falls back to a fresh random UUID if a document has no usable doc_id.
         """
-        URL_NAMESPACE = _uuid.UUID("6ba7b811-9dad-11d1-80b4-00c04fd430c8")
-
         uuids = []
         for doc in documents:
-            uuid_val = _uuid.uuid5(URL_NAMESPACE, doc.page_content)
-            uuids.append(str(uuid_val))
-
+            doc_id = (doc.metadata or {}).get("doc_id")
+            if doc_id and self._is_uuid(str(doc_id)):
+                uuids.append(str(doc_id))
+            else:
+                uuids.append(str(_uuid.uuid4()))
         return uuids
 
     @override
--- a/api/providers/vdb/vdb-weaviate/tests/unit_tests/test_weaviate_vector.py
+++ b/api/providers/vdb/vdb-weaviate/tests/unit_tests/test_weaviate_vector.py
@@ -759,6 +759,37 @@ class TestWeaviateVector(unittest.TestCase):
         assert wv._is_uuid("123e4567-e89b-12d3-a456-426614174000") is True
         assert wv._is_uuid("not-a-uuid") is False
 
+    def test_get_uuids_returns_doc_id_from_metadata(self):
+        """_get_uuids should reuse Dify's own doc_id instead of hashing content."""
+        wv = WeaviateVector.__new__(WeaviateVector)
+        doc_id = "123e4567-e89b-12d3-a456-426614174000"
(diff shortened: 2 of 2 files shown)


FFXN:
Thank you for identifying the root cause and proposing the correct direction for fixing the Weaviate UUID mismatch. Your investigation and PR were very helpful in validating both the content-collision and deletion issues.
The fix has now been superseded by #41915, which follows the same doc_id-based UUID approach and additionally handles backward-compatible cleanup for legacy UUID5 objects and positional alignment when doc_id is missing. Therefore, we’re closing this PR in favor of #41915.
Thanks again for your contribution!

PR #39552, opened 2026-07-24
fix: typing into the trailing Params/Headers row in the HTTP node no longer destroys the row

### LINKED ISSUE #39551: HTTP Request node: typing into the trailing Params/Headers row destroys the row
### Self Checks

### Dify version

1.16.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

On an **HTTP Request** node in the workflow editor, in the **Params** key/value table:

1. Add a param row and type a key (e.g. `state`).
2. Click into that row's **Value** field.
3. Type any single character **other than a space** (e.g. `a`, `1`, or `/`).

**Result:** the entire row is destroyed — the value you just typed *and* the key you had already entered both disappear.

Further reproductions of the same underlying behaviour:

- **Only the trailing (last) row is affected.** Typing into the value of a committed non-last row works normally. It is specifically the uncommitted trailing row that is destroyed.
- **A space is the only first character that survives.** Every other character destroys the row. Typing `30` enters `3`, destroys the row, and moves the cursor into a newly-created row — so multi-character values cannot be typed at all.
- **Pasting works; typing does not.** Pasting a value into the trailing row's value field succeeds; typing the same value character-by-character destroys the row. Pasting is currently the only way to enter a value by hand.

This reproduces on **every attempt (5/5)** for each case above, on a stock self-hosted 1.16.0 install. The **Headers** table on the same node uses the same component and shows the identical behaviour.

**Root cause (for reference):** params/headers are stored as a newline-joined string, and every change round-trips list → string → list. On each keystroke a microtask re-derives the list and reassigns every row's `id`, which remounts the Lexical editor mid-keystroke and drops the input; the add-row path also closes over a stale list. Fix and regression test in the linked PR.

**Related but separate:** pressing **Enter** in the trailing row's key field creates a new row and migrates the existing value into it. This shares the same newline-joined-string storage model but is not addressed by the linked PR (it would require changing the storage model); noting it here for visibility as a follow-up.

**Related Issue**: #38860 (form-data values truncated at the first colon) stems from the same newline-joined-string storage model, on a different surface.

Screen recording attached below.

https://github.com/user-attachments/assets/2c34d659-6220-44fa-a31b-0116b0518abe

### ✔️ Expected Behavior

Typing a value into the trailing param/header row should enter that value and preserve the key already typed — exactly as pasting does. No keystroke should destroy an existing row, and the value should never be lost.

### ❌ Actual Behavior

Typing any non-space character into the trailing row's value field destroys the whole row, including a key that was already entered. Values can only be entered by pasting. The same defect affects both the Params and Headers tables on the HTTP Request node.

dparkmit24:
## Summary

Fixes #39551.

In the HTTP Request node's **Params** and **Headers** key/value tables, typing any non-space character into the trailing (uncommitted) row destroyed the entire row — both the just-typed value and any key already entered. Values could only be entered by pasting.

**Root cause:** params/headers are stored as a newline-joined string, and every change round-trips list → string → list. Two mechanisms in the shared `use-key-value-list.ts` hook combined to drop input:

1. A microtask `useEffect` re-derived the list from the string on each change and reassigned *every* row's `id`, remounting the Lexical editor mid-keystroke and dropping the character being typed.
2. The add-row handler closed over a stale list, so an `onChange` immediately followed by an `onAdd` let the add overwrite the value that was just typed.

**Fix:** hold a `listRef` mirror so the add path no longer closes over a stale list, and preserve `id`s for rows that already exist — only assigning a new `id` to a genuinely new row — so the editor is no longer remounted on every keystroke. Both tables share the hook, so this fixes Params and Headers together. The newline-joined-string storage model is left unchanged (deliberately out of scope for this fix).

**Test:** added a regression test in the existing Vitest suite that mounts the key-value editor, types into the trailing row's value, and asserts the row survives with its key and value intact. It fails on current `main` and passes with this change; the full workflow-nodes suite and `tsc --noEmit` stay green.

**Not addressed here:** pressing Enter in the trailing row's key field still migrates the existing value into the newly-created row. That stems from the same newline-joined-string storage model and would require changing it, so it's left as a documented follow-up in #39551.

## Screenshots

**Before**
*Typing `open` into the trailing Params value row on `main` — the row is destroyed on the first character, losing both the key and the value.*

https://github.com/user-attachments/assets/73cd5d37-0fea-4cee-b1fd-7ec56a369654

---

**After**
*Same sequence with the fix — the row survives, key and value are preserved, and a new empty row appears as expected.*

https://github.com/user-attachments/assets/aaf7d220-8c03-4233-b23a-b6450b32b795

## Checklist

From Claude Code

### DIFF
--- a/web/app/components/workflow/nodes/http/components/key-value/__tests__/key-value-row-lifecycle.spec.tsx
+++ b/web/app/components/workflow/nodes/http/components/key-value/__tests__/key-value-row-lifecycle.spec.tsx
@@ -87,9 +87,7 @@ describe('HTTP key/value table — trailing-row lifecycle (F-HTTP-PARAMS-TABLE)'
 
     // And its editor must NOT have been remounted (stable instance id) — a remount
     // is what wipes the in-progress keystroke in the real Lexical editor.
-    const sinceIdx = getKeyInputs().findIndex(
-      (el) => (el as HTMLInputElement).value === 'since',
-    )
+    const sinceIdx = getKeyInputs().findIndex((el) => (el as HTMLInputElement).value === 'since')
     const valueIdAfter = getValueInputs()[sinceIdx]!.getAttribute('data-instanceid')
     expect(valueIdAfter).toBe(valueIdBefore)
   })
--- a/web/app/components/workflow/nodes/http/components/key-value/__tests__/key-value-row-lifecycle.spec.tsx
+++ b/web/app/components/workflow/nodes/http/components/key-value/__tests__/key-value-row-lifecycle.spec.tsx
@@ -1,0 +1,96 @@
+import { render, screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import * as React from 'react'
+import { useState } from 'react'
+import useKeyValueList from '../../../hooks/use-key-value-list'
+import KeyValue from '../index'
+
+// The HTTP node's key/value editor renders a Lexical `PromptEditor` at its leaf
+// (`input-item` -> `input-support-select-var` -> `PromptEditor`). Lexical's real
+// contenteditable behaviour is not modelled by happy-dom, and the existing
+// PromptEditor tests mock it out for the same reason. The row-lifecycle bug under
+// test lives *above* the editor — in `use-key-value-list` + `item.tsx`'s
+// onChange/onAdd orchestration — so we mock only the leaf editor with a plain
+// controlled input that honours the same `{ value, onChange }` contract.
+vi.mock('../key-value-edit/input-item', () => ({
+  __esModule: true,
+  default: ({
+    value,
+    onChange,
+    instanceId,
+  }: {
+    value: string
+    onChange: (v: string) => void
+    instanceId?: string
+  }) => (
+    <input
+      data-instanceid={instanceId}
+      data-field={instanceId?.startsWith('http-key-') ? 'key' : 'value'}
+      value={value}
+      onChange={(e) => onChange(e.target.value)}
+    />
+  ),
+}))
+
+// Mirrors the production wiring in `use-config.ts`:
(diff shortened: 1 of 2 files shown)


dparkmit24:
Scope note: `use-key-value-list.ts` and the key-value editor it feeds are imported only by the HTTP Request node — driving both the Params and Headers tables, which is why this one change fixes both. No other panel uses this hook or editor. The other key/value tables in the workflow editor use a different, unaffected model (the Webhook trigger's tables are array-based; the variable list is an array of `Variable[]`; paragraph input is a plain textarea). The HTTP node's form-data body is also string-based but is a separate, pre-existing defect (#38860), reload-time colon truncation), untouched by this change.

dparkmit24:
@crazywoola: The linked issue #39551 was erroneously closed when #38861 merged. That PR changes `transformToBodyPayload` so colons are kept in form-data body values, which doesn't fix the defect in #39551. Could you reopen #39551 and take a look at this PR?

The defect in #39551 is a different one on the Params/Headers tables: typing any non-space character into the trailing row destroys the row, because each keystroke re-derives the list and reassigns every row's `id`, remounting the Lexical editor mid-keystroke. It's a remount-on-keystroke problem rather than a parsing one, and #38861's change doesn't touch it. Re-checked against current main — still reproduces, 5/5, and pasting still works where typing doesn't.

This PR carries that fix with its regression test. Happy to rebase if useful.

dparkmit24:
Re-verified on 1.16.1 (self-hosted, Docker): typing a non-space character into the trailing Params row still destroys the row, key and value both. The earlier confirmation in this thread was against 1.16.0, so this is the same behaviour on the current release rather than a repeat of the same check.

dparkmit24:
@crazywoola — would you be able to take a look, or point me at the right reviewer? This one silently discards input a user has already typed, with no indication anything went wrong.

Issue #38351, opened 2026-07-03
[Refactor/Chore] Migrate handwritten web API helpers to generated contracts


hyoban:
## Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

## Current status — 2026-09-22

**Still open; migration is partially complete.** This update is based on `main` at [`f7dbed5d8613`](https://github.com/langgenius/dify/tree/f7dbed5d8613470c7a6e929ad19b65e6d7a1eeda). Open PRs below are not included in the main-branch counts.

### Remaining handwritten calls

A fresh TypeScript AST scan of production `.ts` / `.tsx` files under `web/`, counting calls to named or namespace imports resolved directly to `web/service/base.ts` (including import aliases), found:

| Category | Original baseline | Current main |
| --- | --- | --- |
| All base helper calls | 460 / 62 files | **404 / 56 files** |
| `get`, `post`, `put`, `del`, `patch` | 411 / 43 files | **357 / 37 files** |
| SSE, upload, public API, marketplace helpers | 47 / 23 files | **45 / 24 files** |
| `request` / `handleStream` | 2 calls | **2 calls / 2 files** |

Tests, mocks, stories, and declaration files are excluded. These are syntactic call-site counts, not unique endpoints or a count of safely migratable Console JSON requests. The ordinary-helper bucket can also contain dynamic URLs and special-response calls. The difference from the original baseline is a net change, not proof that every removed call was migrated.

Largest remaining ordinary-helper modules:

| Module under `web/service/` | Calls |
| --- | ---: |
| `common.ts` | 39 |
| `datasets.ts` | 30 |
| `use-plugins.ts` | 25 |
| `use-pipeline.ts` | 21 |
| `tools.ts` | 20 |
| `use-tools.ts` | 19 |
| `use-common.ts` | 17 |
| `use-workflow.ts` | 17 |
[...]

PR #39558, opened 2026-07-25
refactor(web): migrate vector space billing query to contract client

EvanYao826:
## Summary

- migrate vector-space billing requests from the deprecated handwritten REST helper to generated console contract query options
- use `consoleQuery.features.vectorSpace.get.queryOptions(...)` directly at all three call sites
- preserve conditional query enablement for dataset creation flows
- update existing feature tests to seed the generated query cache
- remove the obsolete handwritten helper and thin query wrapper

## Test Plan

- `pnpm --dir web test run __tests__/billing/billing-integration.test.tsx __tests__/billing/education-verification-flow.test.tsx app/components/datasets/create/step-one/__tests__/index.spec.tsx app/components/datasets/documents/create-from-pipeline/__tests__/index.spec.tsx` (94 passed)
- focused `vp check` on the changed service and test files
- `git diff --check`

Refs #38351


web/service/use-billing.spec.ts
lyzno1, line 1: Unnecessary test need to delete.

web/service/use-billing.ts
```diff
@@ -28,8 +25,8 @@ export const useBillingUrl = (enabled: boolean) => {
 
 export const useCurrentPlanVectorSpace = (enabled = true) => {
   return useQuery({
-    queryKey: currentPlanVectorSpaceQueryKey,
-    queryFn: () => fetchCurrentPlanVectorSpace(),
+    queryKey: consoleQuery.features.vectorSpace.get.queryKey(),
+    queryFn: () => consoleClient.features.vectorSpace.get(),
     enabled,
   })
 }
```
lyzno1: use queryOptions instead.

lyzno1 (changes requested):

lyzno1 (changes requested):
After the migration, the thin wrapper is no longer necessary. You should use consoleQuery directly at the point of call.

web/service/use-billing.spec.ts
EvanYao826: Removed the unnecessary hook-level test as requested.

web/service/use-billing.ts
EvanYao826: Updated all three call sites to use `consoleQuery.features.vectorSpace.get.queryOptions(...)` directly and removed the thin wrapper.

EvanYao826:
Thanks for the review. I removed the thin billing wrappers and now use `consoleQuery` with the generated vector-space query options directly at each call site. The updated CI is green. Could you please take another look?

lyzno1 (approved):

PR #39569, opened 2026-07-25
fix(dataset): stop the legacy permission gate from shadowing RBAC

### LINKED ISSUE #39379: RBAC: per-resource access control is not initialized on most App/Dataset creation paths — team members get 403 on resources they didn't create
### Self Checks

### Self Checks
  - [x] Searched existing issues; this concerns the RBAC access path.

  ## Summary

  When er-resource access is initialized on creation via
  `AppAccess/DatasetAccess.replace_whitelist(scope=ALL)` +
  `try_sync_creator_access_policy_member_bindings` + `initialize_created_app_rbac_access_task`.

  This initialization is only wired into the **primary console create controllers**. Every other creation entry point constructs the `App`/`Dataset` object directly and **skips it**. Resources created via those paths therefore have **no RBAC access record**, and any non-maintainer team member gets **403** when opening them.

  For datasets there is an additional layer: the legacy `check_dataset_permission` still runs unconditionally under RBAC and reads the legacy `dataset.permission` / `DatasetPermission` model, which the RBAC access-config UI does not update. A knowledge base created via the file-upload flow defaults to `permission=only_me` **and** has no RBAC whitelist, so it cannot be shared even after assigning "Partial team members" — the legacy gate rejects first with `You do not have permission to access this dataset.`

  The maintainer/creator keeps access via the maintainer short-circuit, so this is **silent for the creator** and only breaks collaboration. Because RBAC is default-off, it appears to be largely unexercised.

  ## Environment
  - `main` (verified on latest)

  ## Coverage matrix (verified on `main`)

  ### App creation
  | Entry point | RBAC init on create? | Ref |
  | --- | --- | --- |
  | Console create app | ✅ | `controllers/console/app/app.py:676` (+ `app_service.py:540`) |
  | DSL import (`import_app`) | ❌ | `services/app_dsl_service.py:442` |
  | Agent duplicate | ❌ | `services/agent/roster_service.py:456` |
  | Workflow convert | ❌ | `services/workflow/workflow_converter.py:94` |

  `import_app` is reused by console / `inner_api/app/dsl.py` / `openapi/app_dsl.py`; none
  initialize RBAC.

  ### Dataset creation
  | Entry point | RBAC init on create? | Ref |
  | --- | --- | --- |
  | Console empty dataset | ✅ | `controllers/console/datasets/datasets.py:603,611` |
  | Public API (`service_api`) | ✅ | `controllers/service_api/dataset/dataset.py:536,544` |
  | **File upload** (`save_document_without_dataset_id`) | ❌ (also defaults `permission=only_me`) | `services/dataset_service.py:3007` |
  | RAG pipeline empty dataset | ❌ | `services/dataset_service.py:532` |
  | External knowledge base | ❌ | `services/external_knowledge_service.py:295` |
  | RAG pipeline DSL import | ❌ | `services/rag_pipeline/rag_pipeline_dsl_service.py:281,414` |

  Legacy gate: `check_dataset_permission` has no `RBAC_ENABLED` guard and is invoked on read paths (8 call sites in `controllers/console/datasets/datasets_document.py`).

  ## Reproduction — dataset (file upload)
  1. User A creates a knowledge base by uploading a file.
  2. Set it to "Partial team members" and add user B.
  3. User B opens it → **403 `You do not have permission to access this dataset.`**
  4. User B opens it → **403 `You do not have permission to access this dataset.`**

  ## Reproduction — app (DSL import)
  1. User A imports an app from DSL.
  2. User B (a member who can normally open A's console-created apps) → **403** on the imported app.

  ## Root cause
  The RBAC-on-create logic is duplicated inline in two controllers rather than being a single shared step invoked by all creation paths, so each additional creation entry point silently omits it.

  ## Questions for maintainers (the reason for this issue)
  1. Is RBAC-on-create initialization intended for **all** creation paths, or only console create? If all, would you accept a shared `apply_<resource>_creation_rbac()` helper called by every path (App ×4, Dataset ×6)?
  2. For datasets, is `check_dataset_permission` meant to stay active under RBAC? If yes, should non-console create paths force `permission=ALL_TEAM` (as console empty-create does) to keep it a pass-through — or should the check be gated behind `if not RBAC_ENABLED`? The two differ in rollback semantics.
  3. Should the `inner_api` / `openapi` DSL-import endpoints also initialize RBAC?

  Happy to open a PR once the intended direction is confirmed.

_No response_

BenjaminX:
Fixes #39379

When RBAC_ENABLED is true, per-dataset access is owned by the RBAC whitelist and the @rbac_permission_required decorators. get_datasets already reflects that: its RBAC branch skips the legacy visibility filter and documents that "permission control is enforced via permission_keys on each item and @rbac_permission_required decorators".

check_dataset_permission and check_dataset_operator_permission never got the same treatment, so they keep applying the legacy only_me / partial_team rules on every read and write path. The RBAC access config UI only writes the RBAC whitelist and never touches dataset.permission, so the two models cannot agree: after granting a member access from the RBAC UI, the legacy gate still rejects them with "You do not have permission to access this dataset." before the whitelist is ever consulted. Only the creator gets through, via the maintainer short-circuit.

Skip the legacy rules when RBAC is enabled, keeping the tenant check in front of the guard so cross-tenant access is still rejected. Behaviour without RBAC is unchanged.

GetProcessRuleApi resolves its dataset from a query-string document id and therefore cannot use the rbac_permission_required decorator; it was relying on the legacy gate alone. Enforce RBAC explicitly there once the dataset is known, following the existing enforce_rbac_access pattern, so neutralising the legacy gate does not widen access on that route.

Audited every caller of both helpers: the console routes all carry rbac_permission_required, the service_api routes authenticate as the tenant owner (the legacy rules were already skipped for them), and the remaining call sites are helpers reached from decorated routes.


BenjaminX:
Ref issue link
https://github.com/langgenius/dify/issues/39379

crazywoola:
Hi @BenjaminX! Thanks so much for taking the time to contribute to Dify — we really appreciate it.

BenjaminX:
Thanks for taking a look, and sorry for the missing link — the issue did exist, I just referenced it in a comment instead of the description.

I've added `Fixes #39379` to the PR description now. For context on the guideline: #39379 was opened on 2026-07-21, four days before this PR (2026-07-25), and it is still open. So the "created and discussed beforehand" requirement was actually met — it simply wasn't linked in the right place.

I also re-verified against current `main` (`6edc8ba147`), and the problem still reproduces:

- `DatasetService.check_dataset_permission` (`api/services/dataset_service.py`) still applies the legacy `only_me` / `partial_team` rules with no `RBAC_ENABLED` guard.
- `DatasetService.check_dataset_operator_permission` — same.
- `GetProcessRuleApi` (`api/controllers/console/datasets/datasets_document.py`) still relies on the legacy gate alone.

`get_datasets` is still the only place carrying the RBAC branch, so the inconsistency this PR addresses is unchanged.

I'm not able to reopen the PR myself. Could you reopen it, or would you prefer a fresh PR against current `main` with the issue linked from the start? Happy to do either — just let me know which is easier for your review queue.

BenjaminX:
For context, here is how this PR relates to the other RBAC reports I filed — they are easy to conflate:

**#39379** (open) — the issue this PR is linked to. It describes two independent gaps:

1. RBAC access is not initialized on most App/Dataset creation paths.
2. The legacy dataset permission gate runs unconditionally under RBAC and shadows the whitelist.

This PR only addresses (2). Gap (1) is untouched and still open.

**#39736** (closed as completed) → **#39735** (merged 2026-08-12) — an adjacent problem: setting a dataset's access scope to "all" only replaced the whitelist scope without granting the per-member policies. That fix landed in `api/controllers/console/workspace/rbac.py`.

Worth flagging that the two interact. #39735 ensures members actually hold a policy, but `check_dataset_permission` still evaluates the legacy `only_me` / `partial_team` rules before the whitelist is ever consulted, and #39735 did not touch `dataset.permission`. So for a dataset created through a path that defaults to `permission=only_me` — the file-upload flow, for instance — a non-maintainer member is still rejected by the legacy gate even though the RBAC side is now correct. The merged fix does not fully take effect on those datasets until this gate is handled.

That is the complete picture from my side; I have no other open PRs in this area.

crazywoola:
@copilot resolve the merge conflicts in this pull request

PR #39579, opened 2026-07-25
fix(workflow): stop dropping unsaved edits when collaboration never c…

BenjaminX:
Draft autosave is debounced by 5s, so the save performed while leaving the editor is what persists everything typed in that window. Since #38997 that save is gated on collaborationManager.canFlushGraphOnPageClose(), which requires the session to be the elected draft leader.

Leader election only happens over the collaboration socket. When collaboration is enabled but the socket never connects, there is no leader, the guard can never pass, and both the unmount save and the keepalive page-close save return early. The edits made since the last autosave are discarded with no request, no error and no toast.

The guard replaced an isFollower check that was explicitly scoped to connected sessions, which is why this only regressed the disconnected case: the check went from "skip when we know we are a follower" to "skip unless we know we are the leader". Restore the original scope by consulting the collaborative rules only while actually connected. A connected follower is still blocked, so the stale-hidden-tab save that #38997 fixed stays fixed.


hjlarry:
Thanks for identifying this regression. I pushed a follow-up commit after tracing both exit paths.

The original patch still had two edge cases:

1. On SPA unmount, the outer guard allowed the save, but `handleSyncWorkflowDraft()` eventually reached `performLocalSync()`, where `canPersistLocalGraph()` rejected the disconnected collaboration session. As a result, no request was actually sent.
2. `isConnected() === false` covers both “never connected” and “connected, then disconnected.” Only the former is safe for a local fallback. The latter may contain a stale graph and must not bypass the protection introduced for #38997.

The follow-up now:

- tracks whether the current collaboration lifecycle has ever established a connection;
- allows the keepalive fallback only when collaboration never connected;
- keeps previously connected-but-currently-disconnected sessions behind the collaboration recovery guard;
- routes the unmount fallback through the existing keepalive save path.

hjlarry (dismissed):

hjlarry (approved):

PR #39589, opened 2026-07-26
feat(web): show default values in User Input field list

### LINKED ISSUE #39563: User Input node: show whether a field has a default value in the field list
### Self Checks

When building a workflow, I set default values on several User Input (Start node) fields so the flow can run without every field being filled in each time. But once the defaults are set, there's no way to tell from the field list which fields actually have a default and which don't — the list shows the field name, type, and whether it's required, but nothing about the default. To check a default I have to open each field's edit panel one at a time.

This gets frustrating on a Start node with several fields: I can see at a glance which fields are required, but not which ones will fall back to a default at run time, so I lose track of which inputs are safe to leave empty. The information already exists in the field config — it just isn't surfaced in the list.

To be clear, this is a display gap, not a functional bug: the defaults themselves apply correctly at run time. This is only about making an existing property visible in the field list.

Proposed change: in the User Input field list, show a field's default value beneath the field name when one is set (nothing shown when there's no default), with the value truncated if long and the full value available on hover. "required" stays where it is on the right.

I have a working implementation of this and will open a PR with it. I'm also happy to build whichever shape you'd prefer if you'd rather it look different. The attached screenshot shows what it looks like. The screenshot shows the default on its own line under each field name, a long value truncated with an ellipsis, and the full value revealed on hover.

Image

I considered adding the default values to the User Input nodes itself but didn't do it because it would make user nodes taller, which might interfere with existing users' flow diagrams.

IshankDev:
## Summary
- Show each User Input (Start node) field default beneath the field name in the panel field list when one is set, truncated with ellipsis and the full value available on hover.
- Keep `required` on the right; leave the canvas Start node unchanged so node height stays the same.
- Add `nodes.start.default` i18n keys across locales and cover the display behavior in `var-item` tests.

Fixes #39563

## Test plan
- [ ] Open a workflow with a User Input / Start node that has several fields
- [ ] Set defaults on some fields and leave others unset
- [ ] Confirm fields with defaults show `default: ...` under the name in the Input Field list
- [ ] Confirm long defaults truncate and reveal the full value on hover
- [ ] Confirm fields without defaults show no default line
- [ ] Confirm `required` still appears on the right
- [ ] Confirm the canvas Start node does not show defaults
- [ ] Run `pnpm test -- run app/components/workflow/nodes/start/components/__tests__/var-item.spec.tsx` from `web/`

### DIFF
--- a/web/app/components/workflow/nodes/start/components/__tests__/var-item.spec.tsx
+++ b/web/app/components/workflow/nodes/start/components/__tests__/var-item.spec.tsx
@@ -32,4 +32,36 @@ describe('StartVarItem', () => {
     fireEvent.click(screen.getByRole('button', { name: 'common.operation.remove' }))
     expect(handleRemove).toHaveBeenCalledTimes(1)
   })
+
+  it('shows the default value under the field name when one is set', () => {
+    render(
+      <VarItem
+        readonly
+        payload={createPayload({
+          default: 'langgenius/dify-official-plugins-and-extensions',
+          required: true,
+        })}
+      />,
+    )
+
+    const defaultLine = screen.getByTitle('langgenius/dify-official-plugins-and-extensions')
+    expect(defaultLine).toHaveTextContent(
+      'workflow.nodes.start.default: langgenius/dify-official-plugins-and-extensions',
+    )
+    expect(screen.getByText('workflow.nodes.start.required')).toBeInTheDocument()
+  })
+
+  it('hides the default line when no default value is set', () => {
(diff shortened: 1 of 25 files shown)


crazywoola:
Hi @IshankDev, thanks for your contribution.

### Why this is being closed
Issue #39563 already has PR #39564 taking priority under our author-first/FIFO rule because it was submitted by the issue author.

### Next steps
Please coordinate with the author of #39564 if you have additional findings or improvements that could help the active implementation.

PR #39591, opened 2026-07-26
fix(webapp-auth): drop 24x multiplier on web-app login token expiry

sergioperezcheco:
WebAppAuthService._get_account_jwt_token built the web-app login JWT expiry as ACCESS_TOKEN_EXPIRE_MINUTES * 24, so the default 60-minute setting issued a 24-hour token rather than the configured lifetime. The console token path (account_service.get_account_jwt_token) already uses the config value directly, so the multiplier was inconsistent with the documented behaviour ("Expiration time for access tokens in minutes") and the rest of the codebase.

This removes the multiplier so the web-app login token respects the operator-set expiry. I added a unit test that asserts the exp claim lands at roughly now + ACCESS_TOKEN_EXPIRE_MINUTES across a few values, and fails when the 24x multiplier is present — the existing testcontainers test only checked that an exp key existed, which is why this slipped through.

### DIFF
--- a/api/services/webapp_auth_service.py
+++ b/api/services/webapp_auth_service.py
@@ -116,7 +116,7 @@ class WebAppAuthService:
 
     @classmethod
     def _get_account_jwt_token(cls, account: Account) -> str:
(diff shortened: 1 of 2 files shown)


crazywoola:

PR #39614, opened 2026-07-27
fix(api): wait for workflow worker cleanup before task completion

linw1995:
> [!IMPORTANT]
>

## Summary

- wait for workflow worker threads to finish before Celery tasks complete
- defer worker thread joins until streaming responses are fully consumed
- clarify that workflow execution tasks consume streams instead of returning generators

From Codex

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

### DIFF
--- a/api/core/app/apps/advanced_chat/app_generator.py
+++ b/api/core/app/apps/advanced_chat/app_generator.py
@@ -636,13 +636,13 @@ class AdvancedChatAppGenerator(MessageBasedAppGenerator):
                     invoke_from=invoke_from,
                 )
             except BaseException:
-                worker_thread.join()
+                self._join_worker_thread(worker_thread)
                 raise
 
             if isinstance(converted_response, Generator):
                 return self._wrap_stream_with_worker_thread_join(converted_response, worker_thread)
 
-            worker_thread.join()
+            self._join_worker_thread(worker_thread)
             return converted_response
 
     def _generate_worker(
--- a/api/core/app/apps/advanced_chat/app_generator.py
+++ b/api/core/app/apps/advanced_chat/app_generator.py
@@ -616,23 +616,34 @@ class AdvancedChatAppGenerator(MessageBasedAppGenerator):
             message_snapshot = MessageSnapshot.from_message(message)
             session.close()
 
-            # return response or stream generator
-            response = self._handle_advanced_chat_response(
-                application_generate_entity=application_generate_entity,
-                workflow=workflow_snapshot,
-                queue_manager=queue_manager,
-                conversation=conversation_snapshot,
-                message=message_snapshot,
-                user=user,
-                stream=stream,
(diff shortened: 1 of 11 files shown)


api/core/app/apps/advanced_chat/app_generator.py
```diff
@@ -616,23 +616,34 @@ def _generate(
...
+                        account=user,
+                        tenant_id=application_generate_entity.app_config.tenant_id,
+                    ),
+                )
+                converted_response = AdvancedChatAppGenerateResponseConverter.convert(
+                    response=response,
+                    invoke_from=invoke_from,
+                )
+            except BaseException:
+                worker_thread.join()
```
QuantumGhost: This might block the celery task forever.

linw1995: If an error occurs and causes the task to block, removing this line may cause a thread leak. To prevent this, I replaced join with a timeout and added logic to skip the task and log a warning if it becomes blocked.

api/core/app/apps/base_app_generator.py
```diff
@@ -64,6 +65,17 @@ def save(self, process_data: Mapping[str, Any] | None, outputs: Mapping[str, Any
 class BaseAppGenerator:
     _file_access_controller: DatabaseFileAccessController = DatabaseFileAccessController()
 
+    @staticmethod
+    def _wrap_stream_with_worker_thread_join[ResponseT](
+        response_stream: Generator[ResponseT, None, None],
+        worker_thread: threading.Thread,
+    ) -> Generator[ResponseT, None, None]:
+        """Keep the producer owned by the response stream until both finish."""
+        try:
```
QuantumGhost, line 91: **P1: handle close before the first iteration**

This wrapper is itself a lazy generator. If it is closed before the first `next()` call, its body has not started, so this `finally` block is not entered and `worker_thread.join()` is never called. That can leave the producer thread running after its response owner has already been closed, which is the lifecycle mismatch this helper is intended to prevent.

A minimal reproduction is:

```python
joined = []


def managed_stream():
    try:
        yield 1
    finally:
        joined.append(True)


stream = managed_stream()
stream.close()
assert joined == []
```

The new close-path test calls `next(managed_stream)` before `close()`, so it does not cover this case. Please use an owner with an explicit `close()` implementation that closes the underlying response stream and joins the worker even before iteration starts, and add a close-before-first-`next()` regression test. Close propagation through the HTTP response wrappers should be covered as well.

Source: [Python generator iterator methods](https://docs.python.org/3/reference/expressions.html#generator-iterator-methods).

**AI disclosure**: This review comment was drafted with Codex. I have reviewed the analysis, and I am responsible for the content.

linw1995: It may happen, but it is not related to this PR. With or without these changes, the behavior remains the same.

QuantumGhost (approved):

Issue #39565, opened 2026-07-25
HTTP Request node: pressing Enter in a Params/Headers key or value field corrupts the row; no keyboard way to advance to the next param


dparkmit24:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.16.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

On an **HTTP Request** node in the workflow editor, in the **Params** key/value table.

**A. Enter in the KEY field of the trailing row splits the row**

1. Add a param row and type a key (e.g. `state`).
2. With the cursor still in the **Key** field, press **Enter**.

Result: a newline is inserted into the key, the row is split, and the value you had entered migrates onto a new empty-key row. The key/value pair you were editing is broken up.

https://github.com/user-attachments/assets/2ef31752-ce54-4ab4-b15e-2d55143fa4ee

**B. Enter in the VALUE field of the trailing row moves the value into a new row's key**

1. In the trailing row, type a key (e.g. `state`) and a value (e.g. `open`).
2. With the cursor in the **Value** field, press **Enter**.

Result: the value you just typed (`open`) becomes the **key of a new second row**, and the original row (`state`) is left with an empty value. What you typed as a value ends up in the wrong column entirely. Tab in the value field also fails to advance to the next row.

https://github.com/user-attachments/assets/a4c9ef08-ccdb-4ab8-932a-8b13203bceec

Because of this there is no keyboard way to commit a param and move to the next one — the natural expectation that **Enter or Tab advances to the next param** is unmet, and pressing Enter actively corrupts the row by relocating the value into a new key. Entering multiple params requires a mouse click to add each new row.

Both behaviours reproduce on a stock self-hosted 1.16.0 install. The **Headers** table on the same node uses the same component and behaves identically.

[...]

PR #39566, opened 2026-07-25
fix(http): stop Enter from corrupting Params/Headers rows; support Enter/Tab to advance

dparkmit24:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

Fixes #39565.

The HTTP Request node's Params/Headers key-value table had no keyboard handling, and pressing Enter corrupted the row being edited:

- **Enter in a KEY field** inserted a Lexical newline, so the key became `key\n`; because params/headers are stored as a newline-joined `key:value` string, `strToKeyValueList` then split that into a separate row.
- **Enter in a VALUE field** relocated the typed value into a new row's key, leaving the original row with an empty value.
- **Enter/Tab in a VALUE field** did nothing useful, so there was no keyboard way to commit a row and move to the next one — every additional param required a mouse click.

This intercepts keydown in the capture phase on each field's wrapper (HTTP-node-local; no shared editor component is touched):

- **KEY field:** Enter is suppressed, so no newline is inserted and the row is not split.
- **VALUE field:** Enter or plain Tab commits the current row and advances focus to the next row's key, creating a new trailing row when on the last row — mirroring the node's existing click-to-add-on-value-focus behavior.

Both handlers stand down while the variable-insert typeahead menu is open (detected via its stable `#typeahead-menu` portal), so Enter/Tab still select a variable there. Blank rows may briefly accumulate but are dropped by the existing serialize round-trip (`stringifyList` filters empty entries). Both the Params and Headers tables share this component, so the fix covers both.

A regression test (`key-value-keyboard.spec.tsx`, Vitest + happy-dom) covers: Enter in a key field does not insert a newline or split the row; Enter and plain Tab in a value field advance to a new row; and the typeahead stand-down leaves Enter/Tab free to select a variable.

`From Claude Code`

### Fix 1 — Enter in the KEY field no longer splits the row

**Before:** pressing Enter in the trailing row's key field inserts a newline and splits the row.

https://github.com/user-attachments/assets/9acec0cb-b72b-4b7e-bfb8-345032da5789

**After:** Enter in the key field is suppressed — no newline is inserted, the key stays intact, and the row is not split. (The fixed behavior is the absence of the split, so there is nothing to show in motion.)

[...]

PR #39644, opened 2026-07-27
fix(http): stop Enter from corrupting Params/Headers rows in HTTP Request node

sergioperezcheco:
On the HTTP Request node, the Params and Headers tables store their rows as a newline-joined "key:value" string, but each field is rendered with the multi-line PromptEditor. Pressing Enter while focused on a key or value inserted a literal newline, which the list<->string round-trip (strToKeyValueList splits on newline) then re-read as a row boundary. The result was that Enter split the row and relocated the value you had just typed into the next row's key, leaving the original row with an empty value. This is #39565 (and the same newline-joined-string root cause behind #39551 / #38860).

This adds a singleLine flag to the variable-aware input used by the key/value table. In single-line mode Enter is intercepted before it reaches the editor so no newline is inserted, unless the variable-insert menu is currently open, in which case Enter still selects the highlighted variable as before. On the value field, Enter also appends a new trailing row so params can be added from the keyboard. The menu-open state is detected via a data attribute on the typeahead portal.

Verified with tsc, the http and prompt-editor vitest suites (683 tests), and added a component test for the single-line commit behavior. I deliberately kept the scope to preventing the newline corruption and the value-field "add row" affordance; full Tab-to-next-row focus management felt out of scope and risky to bundle here.

### DIFF
--- a/web/app/components/base/prompt-editor/plugins/component-picker-block/index.tsx
+++ b/web/app/components/base/prompt-editor/plugins/component-picker-block/index.tsx
@@ -245,6 +245,7 @@ const ComponentPicker = ({
             <div className="size-0" data-prompt-editor-typeahead-menu>
               <div
                 className="w-65 rounded-lg border-[0.5px] border-components-panel-border bg-components-panel-bg-blur p-1 shadow-lg"
+                data-visible={isPositioned ? 'true' : 'false'}
                 style={{
                   ...floatingStyles,
                   visibility: isPositioned ? 'visible' : 'hidden',
--- a/web/app/components/base/prompt-editor/plugins/component-picker-block/index.tsx
+++ b/web/app/components/base/prompt-editor/plugins/component-picker-block/index.tsx
@@ -242,7 +242,7 @@ const ComponentPicker = ({
             // The `LexicalMenu` will try to calculate the position of the floating menu based on the first child.
             // Since we use floating ui, we need to wrap it with a div to prevent the position calculation being affected.
             // See https://github.com/facebook/lexical/blob/ac97dfa9e14a73ea2d6934ff566282d7f758e8bb/packages/lexical-react/src/shared/LexicalMenu.ts#L493
-            <div className="size-0">
+            <div className="size-0" data-prompt-editor-typeahead-menu>
               <div
                 className="w-65 rounded-lg border-[0.5px] border-components-panel-border bg-components-panel-bg-blur p-1 shadow-lg"
                 style={{
--- a/web/app/components/workflow/nodes/_base/components/__tests__/input-support-select-var.spec.tsx
+++ b/web/app/components/workflow/nodes/_base/components/__tests__/input-support-select-var.spec.tsx
@@ -1,0 +1,85 @@
+import type { ReactNode } from 'react'
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import Editor from '../input-support-select-var'
+
+vi.mock('@/app/components/base/prompt-editor', () => ({
+  __esModule: true,
+  default: ({ value }: { value: string }) => (
+    <div data-testid="prompt-editor">
+      <input data-testid="editor-input" defaultValue={value} />
+    </div>
+  ),
+}))
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: (cb: (k: { $: Record<string, string> }) => string) => cb({ $: {} }),
+  }),
+}))
+
+vi.mock('@langgenius/dify-ui/cn', () => ({
+  cn: (...args: unknown[]) => String(args.filter(Boolean).join(' ')),
+}))
+
+vi.mock('@langgenius/dify-ui/tooltip', () => ({
+  Tooltip: ({ children }: { children?: ReactNode }) => <div>{children}</div>,
+  TooltipContent: () => null,
+  TooltipTrigger: ({ children }: { children?: ReactNode }) => <div>{children}</div>,
+}))
+
+vi.mock('@/app/components/base/icons/src/vender/solid/development', () => ({
+  Variable02: () => null,
+}))
+
+vi.mock('@/app/components/workflow/store', () => ({
+  useStore: () => null,
+}))
+
+const wrap = (overrides: Record<string, unknown> = {}) =>
+  render((<Editor value="open" onChange={() => {}} {...overrides} />) as ReactNode)
+
+describe('input-support-select-var singleLine Enter handling', () => {
+  it('commits on Enter when singleLine and no typeahead menu is open', async () => {
+    const user = userEvent.setup()
+    const onCommit = vi.fn()
+
+    wrap({ singleLine: true, onCommit })
+
+    const input = screen.getByTestId('editor-input')
+    await user.click(input)
+    await user.keyboard('{Enter}')
+
+    expect(onCommit).toHaveBeenCalledTimes(1)
+  })
+
+  it('does not commit on Enter while the typeahead menu is visible', async () => {
+    const user = userEvent.setup()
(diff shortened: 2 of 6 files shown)


dparkmit24:
Disclosure before anything else: I filed #39565 and have an overlapping PR open at #39566, so treat everything below as coming from an interested party. I think your scoping decision is the better one — the `singleLine` flag on the shared editor is cleaner than the capture-phase wrapper handlers I used, and deliberately leaving focus management out is the right call for a first change. Three observations, one of which I think matters.

**1. The new test doesn't exercise `handleKeyDown`.**

`input-item.spec.tsx` mocks `input-support-select-var` and reimplements the behaviour inside the mock:

```js
onKeyDown={(e) => {
  if (singleLine && e.key === 'Enter') {
    e.preventDefault()
    onCommit?.()
  }
}}
```

So the assertions confirm that the mock calls `onCommit` — which it does by construction. The real `handleKeyDown`, including the menu-open guard, is never run; the test would still pass if that function were deleted. It's a good prop-plumbing test, and worth keeping as one, but the logic this PR adds is currently uncovered. A test that renders the real editor and asserts Enter is suppressed (and that it is *not* suppressed while the typeahead is open) would pin the part that can actually regress.

**2. The menu-open guard is a global lookup with an inline-style string match.**

```js
document.querySelector('[data-prompt-editor-typeahead-menu] > div:not([style*="visibility: hidden"])')
```

Two things I'd want to be sure of. It's document-scoped rather than scoped to this editor, so it matches any open typeahead on the page — probably fine in practice since only one can hold focus, but it's an assumption worth stating. More concerning, `[style*="visibility: hidden"]` depends on the exact serialisation of an inline style. If that ever becomes `display: none`, or the spacing changes, the selector silently stops matching and Enter gets swallowed while the menu is open — breaking variable insertion, with no test to catch it. A data attribute that reflects open state directly would be sturdier than inferring it from how the element is hidden.

**3. As I read it, the keyboard-advance half of #39565 isn't closed yet — and the issue has been.**

`onCommit={handleValueContainerClick}` resolves to:

```js
if (isLastItem && hasValuePayload) onAdd()
```

So Enter in a value field appends a trailing row only on the last row, and never moves focus. On any earlier row it does nothing. That matches your scope note exactly and I'm not asking you to widen it — but #39565 listed "no keyboard way to commit a param and advance" as half the defect, and that issue was closed as completed on 2026-08-03 while both behaviours still reproduce (I re-verified on 1.16.1 today). Might be worth not treating the issue as closed by this change, so the remainder doesn't get lost.

I have a local reproduction environment on 1.16.1 and am happy to test a revision against it if that's useful.

sergioperezcheco:
Thanks for the detailed pass — all three points are fair, and I've addressed 1 and 2 in 5e10cd2912.

**1. Real test for `handleKeyDown`** — agreed, the input-item spec only proved the mock's own behavior. Added `input-support-select-var.spec.tsx` that renders the real editor and drives Enter through the actual handler: commit when singleLine with no menu, suppressed while the menu is visible, and no commit when singleLine is off.

**2. Style-string guard** — you're right that `[style*="visibility: hidden"]` is brittle. The picker block now exposes `isPositioned` as `data-visible` on the menu, and the guard selects `[data-prompt-editor-typeahead-menu] > div[data-visible="true"]`. The test above pins the suppression path, so a styling change that breaks the selector fails CI instead of silently swallowing Enter.

**3. #39565 keyboard-advance half** — that's a fair catch on the issue state; this PR intentionally only covers the corruption half. I'll leave a note on the issue so the advance behavior isn't lost.

If you're still up for testing on 1.16.1, the new commit is on the branch now.

dparkmit24:
Tested the branch head (5e10cd2912) on 1.16.1. Setup: the three commits cherry-picked onto a fresh 1.16.1 worktree — one trivial conflict (main renamed the typeahead menu's width class on an adjacent line; I kept the tag's class, `data-visible` line added, nothing else) — web dev build against a stock 1.16.1 Docker API, driven in the browser on macOS.

**The result is browser-dependent:** Videos below demonstrate the issues

**Chrome** 
Enter "foo" in the key field. Tab to value field. Type "b". "foo" vanishes, and b was never visible.

https://github.com/user-attachments/assets/7aebfed6-35dc-4466-b038-8415f39c499d

Enter "foo" and press Enter in the key field. Key field becomes 2 lines. Tab to value field. Type "ba". New param with blank key and "ba" value is created.

https://github.com/user-attachments/assets/f7408c04-5de8-4658-9922-10b09eea6391

---

**Safari:** 
Enter "foo" in the key field. Tab to value field. Type "b". Both foo and b vanish.

https://github.com/user-attachments/assets/fdf73d75-0e87-469d-9d75-63c00ec965ce

Enter "foo" and press Enter in the key field. Key field remains just one line. Tab to value field. Type "b". "foo" vanishes, and "b" was never visible.

https://github.com/user-attachments/assets/0f851e60-d92f-472a-9e06-9b4da048f78d

**Important caveat on the videos:** the field-vanish behavior (typing in the value field destroying the trailing row, key included) also reproduces on **stock 1.16.1** with a virgin workflow, in both browsers — it's the pre-existing defect tracked in #39551 (root cause: each keystroke round-trips the newline-joined storage and remounts the row's editor mid-keystroke), with a fix open at #39552; it's not introduced by this PR. The PR-relevant delta in these videos is the **Enter behavior only**: in Chrome, Enter still inserts the newline (second video — the split and the ghost row); in Safari, Enter is correctly suppressed (fourth video — key field stays one line), which stock Safari does not do. So the fix works on WebKit and is inert on Chromium, per the mechanism below.

**Why (code-anchored):** the guard is a React synthetic `onKeyDown` on a wrapper div, but Lexical registers a *native* keydown listener directly on the contentEditable (`Lexical.dev.mjs:4119`, v0.47.0), which runs before any delegated React handler. On non-WebKit browsers, `LexicalRichText`'s Enter handler calls `preventDefault()` and inserts the paragraph synchronously inside that native listener (`LexicalRichText.dev.mjs:1306–1341`) — so by the time `handleKeyDown` runs, the newline already exists and its `preventDefault()` is a no-op. On Apple WebKit the same handler returns `false` without `preventDefault()` and lets the browser's default beforeinput path insert *after* propagation — which the guard can and does suppress. That matches the Chrome/Safari split exactly. Since `prompt-editor-content.tsx` is byte-identical between the tag and main and Lexical is 0.47.0 on both, I'd expect the same Chrome behavior on main (inferred from code identity — I only browser-tested 1.16.1).

**On the tests:** the new `input-support-select-var.spec.tsx` is a real improvement — it drives the actual `handleKeyDown`. But it mocks `@/app/components/base/prompt-editor` (line 6), so Lexical's own keydown handling never participates, and this ordering problem can't surface in any test that stubs the editor. Not a flaw in the spec so much as a limit of what's testable without the real editor mounted.

**A possible direction:** suppress Enter *inside* Lexical rather than outside it — e.g. register `KEY_ENTER_COMMAND` at high priority in the editor (or a small plugin) and return `true` when `singleLine`, so the suppression lives in the same pipeline that performs the insertion and is browser-independent. The typeahead case should compose naturally since the menu's selection runs through Lexical's command priorities too.

Happy to re-test any revision on the same rig, both browsers.

sergioperezcheco:
Excellent analysis — the native-listener ordering explains exactly why it worked on WebKit but not Chromium. Fixed by moving the interception to `onKeyDownCapture` with `stopPropagation()`: the capture phase fires on the wrapper before Lexical's native keydown on the contentEditable, so the event never reaches the paragraph-insert handler on Chromium either.

The variable-menu passthrough and the row-commit (`onCommit`) behavior are unchanged, and the single-line guard still only applies to the Params/Headers key-value fields.

Thanks again for the browser-by-browser repro videos — that saved a lot of blind debugging.

PR #39453, opened 2026-07-23
fix: fix mcp output_schema is optional

fatelei:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

fix #39418

Now the `structuredOutput` goes into json output if the schema is not defined.

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [ ] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [ ] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods

PR #39649, opened 2026-07-27
fix(tools): surface MCP structuredContent without a declared outputSchema

### LINKED ISSUE #39418: MCP Server Tool does not output structuredContent as json data when there is no outputSchema
### Self Checks

### Dify version

1.16.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Add an HTTP MCP Server as a tool according to https://docs.dify.ai/en/self-host/use-dify/workspace/tools#mcp

The MCP server should support results with `content` and `structuredContent`. I'm using https://github.com/ProfessionalWiki/mediawiki-mcp-server

Use the MCP server tool in a workflow.

Only the `content` is output as `text` from the workflow node.

### ✔️ Expected Behavior

Both the `content` and `structuredContent` are output from the node in `text` and json `data` as appropriate.

I know that `structuredContent` was only recently added for Dify Workflows exported as MCP servers, but from what I can tell from the code base, it should be supported for MCP Servers as tools.

Ah, I think I've traced it out it only gets output if there is an `output_schema`:

https://github.com/langgenius/dify/blob/4da4fa72cd9d011559fad3746547c935941724b2/api/core/tools/mcp_tool/tool.py#L106-L109

Does this need to be implemented by the MCP server? Is it _required_ by Dify, or could you still add `structuredContent` if it exists? `outputSchema` is not required by [the spec](https://modelcontextprotocol.io/specification/draft/server/tools#output-schema).

### ❌ Actual Behavior

Only the `content` is output as `text` from the workflow node.

Image

SpiliosDimakopoulos:
Fixes #39418

## Problem
MCPTool only yielded result.structuredContent when
self.entity.output_schema was also truthy:

    if self.entity.output_schema and result.structuredContent:

outputSchema is optional per the MCP spec
(https://modelcontextprotocol.io/specification/draft/server/tools#output-schema),
so a server can return structuredContent without declaring an
outputSchema for the tool. This silently dropped valid, already-present
data whenever the server didn't declare one.

## Why this guard wasn't needed
- create_variable_message(k, v) doesn't use output_schema for anything;
  it just wraps a key/value pair.
- No workflow node references entity.output_schema when handling tool
  outputs either.

So the condition served no functional purpose beyond gatekeeping.

## Fix
Removed the `self.entity.output_schema and` condition. structuredContent
is now surfaced whenever the MCP result includes it, regardless of
whether an outputSchema was declared.

## Testing
Added a regression test with output_schema=None. It fails against
unmodified dify (0 messages instead of 2) and passes after this fix
(verified against a separate, untouched checkout to be sure).

Ran the existing 40 MCPTool tests (all still pass) plus the full
tests/unit_tests/tools/ and tests/unit_tests/core/tools/ suites (378
tests, all pass). ruff check on the changed file: clean.


wylswz:
dup: #39453, will close this one.

PR #39652, opened 2026-07-27
fix: detect actual audio format for TTS Content-Type instead of assuming mpeg

SpiliosDimakopoulos:
invoke_tts() returns Iterable[bytes] with no format metadata, but AudioService.transcript_tts always hardcoded content_type="audio/mpeg" on the streamed response, in both places it returns a Generator. If a model/plugin provider streams a different container (WAV, OGG, FLAC), the browser receives a Content-Type that doesn't match the actual bytes, which can break playback.

Added a best-effort sniff of the first response chunk's magic bytes (RIFF/WAVE, OggS, fLaC, ID3 tag or raw MP3 frame sync) to set the correct Content-Type, falling back to the previous audio/mpeg default when the format isn't recognized. This is purely additive — streams that were already being labeled correctly see no behavior change.

Note: I wasn't able to confirm from this repo alone whether a currently-shipped provider plugin returns non-MP3 audio in practice (provider implementations live in the separate plugin repos) — happy to narrow scope if a maintainer says this isn't currently reachable.


crazywoola:
Hi @SpiliosDimakopoulos! Thanks so much for taking the time to contribute to Dify — we really appreciate it.

PR #39656, opened 2026-07-27
perf(web): skip unauthenticated SSR console prefetch

### LINKED ISSUE #39610: [Refactor/Chore] web: skip SSR console API prefetch when the request carries no auth cookies (avoids guaranteed 401 round-trips)
### Self Checks

### Description

`CommonLayoutHydrationBoundary` (`web/app/(commonLayout)/hydration-boundary.tsx`) unconditionally prefetches three console APIs during SSR for **every** console page render, forwarding the incoming `cookie` header as-is (`getServerConsoleClientContext` in `web/service/server.ts`):

- `GET /console/api/account/profile`
- `POST /console/api/workspaces/current`
- `GET /console/api/workspaces/current/rbac/my-permissions`

When the incoming request carries **no console auth cookies** (`access_token` / `refresh_token`, or their `__Host-` variants — see `_real_cookie_name` in `api/libs/token.py`), all three calls are guaranteed to return 401. The boundary then redirects to `/auth/refresh`, whose route handler (`web/app/auth/refresh/route.ts`) calls `POST /console/api/refresh-token` whenever the cookie header is merely **non-empty** — even if it only contains unrelated cookies such as `locale` — which also fails, before finally redirecting to `/signin`.

Net effect: every unauthenticated hit on a console URL (logged-out users, expired sessions in stale tabs, crawlers/scanners) costs **3–4 upstream API round-trips that cannot possibly succeed**. The Next.js server can see the cookies before issuing these fetches, so the outcome is decidable up front.

**Proposal (existence check only, no client-side validity check):**

1. In `CommonLayoutHydrationBoundary`, inspect cookies before prefetching:
   - neither `access_token` nor `refresh_token` present → `redirect('/signin?redirect_url=...')` directly, skip all three prefetches;
   - `refresh_token` only → redirect to `/auth/refresh` directly (skip the three guaranteed 401s, go straight to token refresh);
   - `access_token` present → keep current behavior (token validity is still decided solely by the API).
2. In the `/auth/refresh` route handler, only call `/refresh-token` when a `refresh_token` cookie is actually present, instead of "any non-empty cookie header".
3. Cookie-name resolution must mirror `api/libs/token.py` `_real_cookie_name` (plain vs `__Host-` prefix depending on HTTPS / `NEXT_PUBLIC_COOKIE_DOMAIN`). `CSRF_COOKIE_NAME()` in `web/config/index.ts` already implements this exact pattern, so an `ACCESS_TOKEN_COOKIE_NAME()` / `REFRESH_TOKEN_COOKIE_NAME()` helper can follow it.

Because this is an existence check only, users holding an expired-but-present token keep the exact current 401 → refresh flow; only requests that cannot possibly authenticate are short-circuited.

### Motivation

- Removes guaranteed-to-fail API traffic: 3–4 doomed upstream calls per anonymous console page view, on every deployment.
- Cleaner observability: these 401s inflate access-log and APM error-rate figures for the two console endpoints (4xx entry spans are recorded as errors by the tracing instrumentation), which makes real regressions harder to spot in any monitoring setup.
- Marginally faster redirect-to-signin for anonymous visitors, and less wasted SSR work in dify-web.

### Additional Context

Happy to submit a PR for this — the change is small (~20 lines plus a cookie-name helper and tests) and behavior-preserving for all authenticated flows.

qdivan:
## Summary

- skip the three SSR console API prefetches when the request has no access-token cookie
- route refresh-only requests directly through `/auth/refresh`, while requests without auth cookies go directly to sign-in
- avoid calling `/console/api/refresh-token` when the request only contains unrelated cookies
- keep the existing API-driven validation path whenever an access-token cookie is present

Closes #39610.

## Validation

- `NODE_ENV=test pnpm --dir web exec vp test run 'app/(commonLayout)/__tests__/hydration-boundary.spec.tsx' app/auth/refresh/__tests__/route.spec.ts` — 43 passed
- `NODE_ENV=test pnpm --dir web exec vp test run service/__tests__/server.spec.ts features/account-profile/__tests__/server.spec.ts config/__tests__/server.spec.ts` — 9 passed
- changed-file `vp check` — passed with no warnings, lint errors, or type errors
- `pnpm --dir web lint:tss` — 6000 passed
- `git diff --check` — passed
- independent candidate review — PASS
- independent patch review — PASS

The regression tests were also applied independently to the current upstream base without the production changes: 4 intended failures reproduced, then all 43 focused tests passed with this patch.

## Duplicate Check

Checked immediately before publication against upstream `main` at `f68cabe2267295fbfb10341ea53f31b65f0a97c5`.

- issue #39610 remains open, unassigned, and has no implementation claim or contribution-lock label
- no active PR was found by issue number, `CommonLayoutHydrationBoundary`, auth-cookie SSR, or refresh-token/hydration-boundary searches
- related merged PRs #39308, #38570, and #36818 address different hydration/base-path/auth-boot behavior and do not implement this cookie short-circuit

## Browser / Playwright

Not run. This change only selects server-side redirect/fetch branches and does not alter browser layout or interaction behavior; focused Vitest coverage deterministically verifies redirects and that the relevant fetch/prefetch calls do or do not occur.

### DIFF
--- a/web/app/(commonLayout)/__tests__/hydration-boundary.spec.tsx
+++ b/web/app/(commonLayout)/__tests__/hydration-boundary.spec.tsx
@@ -18,6 +18,8 @@ const mocks = vi.hoisted(() => ({
   headers: vi.fn(),
   resolveServerConsoleApiUrl: vi.fn(),
   basePath: '',
+  accessTokenCookieName: 'access_token',
+  refreshTokenCookieName: 'refresh_token',
 }))
 
 vi.mock('@/context/query-client-server', async (importOriginal) => {
@@ -41,6 +43,11 @@ vi.mock('@/utils/var', () => ({
   get basePath() {
     return mocks.basePath
   },
+}))
+
+vi.mock('@/config', () => ({
+  ACCESS_TOKEN_COOKIE_NAME: () => mocks.accessTokenCookieName,
+  REFRESH_TOKEN_COOKIE_NAME: () => mocks.refreshTokenCookieName,
 }))
 
 vi.mock('@/features/account-profile/server', () => ({
@@ -76,6 +83,8 @@ describe('CommonLayoutHydrationBoundary', () => {
   beforeEach(() => {
     vi.clearAllMocks()
     mocks.basePath = ''
+    mocks.accessTokenCookieName = 'access_token'
+    mocks.refreshTokenCookieName = 'refresh_token'
     mocks.rootQueryClient = new QueryClient({ defaultOptions: { queries: { retry: false } } })
     mocks.headers.mockResolvedValue(
       new Headers({
@@ -107,7 +116,7 @@ describe('CommonLayoutHydrationBoundary', () => {
       dataset: { default_permission_keys: [], overrides: [] },
     })
     mocks.getServerConsoleClientContext.mockResolvedValue({
-      cookie: 'session=abc',
+      cookie: 'access_token=abc; csrf_token=csrf-token',
       csrfToken: 'csrf-token',
     })
     mocks.workspaceQueryOptions.mockReturnValue({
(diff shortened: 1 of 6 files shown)


lyzno1:
Sorry, we plan to handle this issue internally.

PR #39659, opened 2026-07-28
chore: improve archived logs hint in logs list page

hjlarry:
> [!IMPORTANT]
>

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist


lyzno1 (commented):
Reviewed the retention behavior across accessibility, Dify UI primitives, state/query ownership, and the backend cleanup contract. I found two user-visible boundary issues.

web/app/components/app/log/cloud-sandbox-retention.ts
```diff
@@ -0,0 +1,21 @@
+'use client'
+
+import { useSuspenseQuery } from '@tanstack/react-query'
+import { Plan } from '@/app/components/billing/type'
+import { useProviderContext } from '@/context/provider-context'
+import { systemFeaturesQueryOptions } from '@/features/system-features/client'
+
+export const CLOUD_SANDBOX_TIME_PERIOD_KEYS = new Set(['1', '2', '3'])
```
lyzno1, line 8: [P1] Keep the full 30-day retention window reachable. The authoritative Cloud Sandbox setting retains both messages and workflow runs for 30 days (`SANDBOX_EXPIRED_RECORDS_RETENTION_DAYS`, consumed by `clean_messages` and `clean_workflow_runs_task`), but key `3` in both period mappings is only 28 days. Once this allowlist removes every longer option, logs from days 29–30 still exist in the backend but cannot be queried from either App Log or Workflow Log because neither page offers a custom range. Please align the allowed option with the 30-day retention contract (for example, add/use Last 30 days) and cover both list surfaces.

web/app/components/app/log/cloud-sandbox-retention.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function useIsCloudSandboxPlan() {
+  const { data: deploymentEdition } = useSuspenseQuery({
+    ...systemFeaturesQueryOptions(),
+    select: ({ deployment_edition }) => deployment_edition,
+  })
+  const { enableBilling, isFetchedPlan, plan } = useProviderContext()
+
+  return (
+    deploymentEdition === 'CLOUD' && enableBilling && isFetchedPlan && plan.type === Plan.sandbox
```
lyzno1: [P1] Do not collapse an unresolved plan into the unrestricted branch. `isFetchedPlan` starts false while `/features` is in flight, so both filters initially expose all nine periods. If a Cloud Sandbox user selects `4`–`9` during that window, or App Log restores one from `logsStateCache`, the settled Sandbox render only removes the item; `queryParams.period` stays unchanged. The real `Chip` then has no matching selected item (blank generic Select and no clear action), while the parent request still uses the hidden range; `9` even omits the time bound. Please preserve the pending/known state and let the `Logs` owner derive one valid effective period for both the Chip and the request, apply the same fix to Workflow Log, and add a pending-to-Sandbox rerender test with the real Chip.

lyzno1 (approved):

PR #39664, opened 2026-07-28
fix: prevent Safari from clipping the settings close button

iamjoel:
## Summary

- Move the account settings Close/ESC controls outside the scroll area so Safari no longer clips them when tab content becomes scrollable.
- Preserve the existing click and Escape close behavior, verified by the focused account settings test suite.

Fixes SNP-584

## Checklist

### DIFF
--- a/packages/dify-ui/src/avatar/__tests__/index.spec.tsx
+++ b/packages/dify-ui/src/avatar/__tests__/index.spec.tsx
@@ -7,17 +7,6 @@ function stubImageLoader() {
 
   function TestImage(_width?: number, _height?: number): HTMLImageElement {
     const image = document.createElement('img')
-    Object.defineProperties(image, {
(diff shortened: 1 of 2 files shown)


lyzno1 (dismissed):

linw1995 (dismissed):

packages/dify-ui/src/avatar/__tests__/index.spec.tsx
```diff
@@ -7,6 +7,17 @@ function stubImageLoader() {
...
+      complete: {
+        configurable: true,
+        value: false,
+      },
+      src: {
+        configurable: true,
+        value: '',
+        writable: true,
+      },
+    })
```
lyzno1:
```suggestion
```

lyzno1 (approved):

PR #39816, opened 2026-07-30
fix(account): stop sharing one additional_data dict across token calls

### LINKED ISSUE #39815: Verification code can leak between concurrent requests via mutable default `additional_data` in AccountService
### Self Checks

### Dify version

1.16.1 (source checkout at `3aa937de`)

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Three `AccountService` methods take the dict that carries the verification code as a mutable default argument, then write the code into it:

https://github.com/langgenius/dify/blob/3aa937de/api/services/account_service.py#L893-L906

```python
def generate_reset_password_token(cls, email, account=None, code=None, additional_data: dict[str, Any] = {}):
    if not code:
    additional_data["code"] = code          # writes into the one shared dict
    token = TokenManager.generate_token(..., additional_data=additional_data)
    return code, token
```

Same shape in `generate_email_register_token` (L909-L921) and `generate_owner_transfer_token` (L936-L951). The default dict is created once at import time, so every call that does not pass `additional_data` shares it. `SERVER_WORKER_CLASS` defaults to `gevent`, so a worker runs many greenlets over that one object, and `TokenManager.generate_token` reads `additional_data` after a Redis round-trip — a suspend point.

The caller returns `code` (its own local) to be emailed, but the token stored in Redis carries whatever is in the shared dict when `generate_token` reads it. Two overlapping requests therefore disagree.

Reproduced against the real methods with two threads, handing control from the first caller to the second at the point where `generate_token` would hit Redis. `stored` is what each token payload ended up with, `returned` is what each caller got back to email:

```
AssertionError: assert stored == returned
  Differing items:
  {'first@example.com': '906060'} != {'first@example.com': '517404'}
```

The first caller was emailed `517404` and its token stored the second caller's `906060`, so that user can never complete the reset.

There is also a deterministic effect with no concurrency needed — a caller that passes its own dict gets it mutated, and the default keeps the last code:

```python
>>> data = {"invite_id": "abc"}
>>> AccountService.generate_reset_password_token("u@x.com", additional_data=data)[0]
'164794'
>>> data
{'invite_id': 'abc', 'code': '164794'}
>>> AccountService.generate_reset_password_token("v@x.com")
>>> AccountService.generate_reset_password_token.__func__.__defaults__
(None, None, {'code': '798569'})
```

The call sites that rely on the default are the three that generate and email a fresh code: `send_reset_password_email` (L707), `send_email_register_email` (L740) and `send_owner_transfer_email` (L840). The controller call sites pass a fresh dict literal (`{"phase": "reset"}` / `{}`) and so escape the shared-default part, though their literal still gets mutated.

### ✔️ Expected Behavior

Each call keeps its own verification code: the code returned to the caller (and emailed) is the code stored in the token payload, and a caller-supplied dict is not modified.

### ❌ Actual Behavior

Under concurrent password-reset / email-register / owner-transfer requests, the token payload can hold another request's code, so the recipient's emailed code fails verification. The last code generated also stays in the module-level default dict for the process lifetime.

ErenAta16:
Fixes #39815

Three token generators in `api/services/account_service.py` share one mutable default and then write into it:

```python
def generate_reset_password_token(..., additional_data: dict[str, Any] = {}):
    additional_data["code"] = code                      # writes into the default

def generate_email_register_token(..., additional_data: dict[str, Any] = {}):
    additional_data["code"] = code

def generate_owner_transfer_token(..., additional_data: dict[str, Any] = {}):
    additional_data["code"] = code
```

This PR changes all three to `dict[str, Any] | None = None` and copies the caller's dict:

```python
additional_data = {**additional_data} if additional_data else {}
```

## What this fixes

**Reachable today: the caller's dict is mutated.** Every call writes `code` into whatever dict was passed in. All four production call sites pass a fresh literal that is discarded immediately, so nothing observable goes wrong right now, but the function silently modifies its argument and nothing says so.

**Latent: the shared default collapses concurrent calls.** When the default is used, two overlapping calls write into the same object, and the caller can be told one code while the token is minted with another:

```
default dict before any call: {}
after alice's call:           {'code': '909393'}
after bob's call:             {'code': '844774'}
-> same dict object reused:   True

codes returned to callers:       alice -> 251573,  bob -> 787541
code the token was minted with:  alice -> 787541
```

No production caller reaches that today. Verified on `main`, the callers are:

| generator | call site | argument |
|---|---|---|
| reset_password | `controllers/console/auth/forgot_password.py` | `{"phase": "reset"}` |
| reset_password | `controllers/web/forgot_password.py` | `{"phase": "reset"}` |
| email_register | `controllers/console/auth/email_register.py` | `{"phase": "register"}` |
| owner_transfer | `controllers/console/workspace/members.py` | `{}` |

so this is a latent trap rather than a live incident, and it should be merged on that basis rather than as a security fix.

## Why the latent argument is stronger than usual here

`api/.ruff.toml` selects all of `B` (line 14) but lists `B006`, `mutable-argument-default`, in `ignore` (line 72). So this pattern has no lint backstop in this repo, and it has already been copy-pasted into three functions. The next code-carrying generator added without an explicit `additional_data` argument inherits it, and the symptom would be a user occasionally receiving a code that does not work while someone else's token carries it. That is a bad thing to learn about from a support ticket.

## Tests

`api/tests/unit_tests/services/test_account_service.py`, parametrized over all three generators:

- the default is not a mutable dict (`inspect.signature(...).default is None`)
- a caller-supplied dict comes back without `code` written into it
- two interleaved calls each keep their own code

The concurrency test uses `threading` rather than gevent. The bug only needs the two calls to interleave between reading and writing the shared dict, which threads reproduce deterministically, so modelling the production worker type is not necessary here.

### DIFF
--- a/api/services/account_service.py
+++ b/api/services/account_service.py
@@ -895,10 +895,11 @@ class AccountService:
         email: str,
         account: Account | None = None,
         code: str | None = None,
-        additional_data: dict[str, Any] = {},
+        additional_data: dict[str, Any] | None = None,
     ):
         if not code:
             code = "".join([str(secrets.randbelow(exclusive_upper_bound=10)) for _ in range(6)])
+        additional_data = {**additional_data} if additional_data else {}
         additional_data["code"] = code
         token = TokenManager.generate_token(
             account=account, email=email, token_type="reset_password", additional_data=additional_data
@@ -910,10 +911,11 @@ class AccountService:
         cls,
         email: str,
         code: str | None = None,
-        additional_data: dict[str, Any] = {},
+        additional_data: dict[str, Any] | None = None,
     ):
         if not code:
             code = "".join([str(secrets.randbelow(exclusive_upper_bound=10)) for _ in range(6)])
+        additional_data = {**additional_data} if additional_data else {}
         additional_data["code"] = code
         token = TokenManager.generate_token(email=email, token_type="email_register", additional_data=additional_data)
         return code, token
@@ -938,10 +940,11 @@ class AccountService:
         email: str,
         account: Account | None = None,
         code: str | None = None,
-        additional_data: dict[str, Any] = {},
+        additional_data: dict[str, Any] | None = None,
     ):
         if not code:
             code = "".join([str(secrets.randbelow(exclusive_upper_bound=10)) for _ in range(6)])
+        additional_data = {**additional_data} if additional_data else {}
         additional_data["code"] = code
         token = TokenManager.generate_token(
             account=account, email=email, token_type="owner_transfer", additional_data=additional_data
--- a/api/tests/unit_tests/services/test_account_service.py
+++ b/api/tests/unit_tests/services/test_account_service.py
@@ -1,4 +1,6 @@
+import inspect
 import json
+import threading
 from collections.abc import Iterator, Sequence
 from datetime import datetime, timedelta
 from unittest.mock import MagicMock, patch
@@ -2876,3 +2878,74 @@ class TestIsEmailSendIpLimit:
             patch.object(dify_config, "EMAIL_SEND_IP_LIMIT_PER_MINUTE", 60),
         ):
             assert AccountService.is_email_send_ip_limit("1.2.3.4") is False
+
+
+class TestTokenGenerationAdditionalData:
+    """The code-carrying token generators must not share one dict across calls."""
+
+    _GENERATORS = (
+        "generate_reset_password_token",
+        "generate_email_register_token",
+        "generate_owner_transfer_token",
+    )
+
+    @pytest.mark.parametrize("method_name", _GENERATORS)
+    def test_additional_data_default_is_not_mutable(self, method_name: str) -> None:
+        """The default must not be a dict the function then writes the code into."""
+        parameter = inspect.signature(getattr(AccountService, method_name)).parameters["additional_data"]
+        assert parameter.default is None
+
+    @pytest.mark.parametrize("method_name", _GENERATORS)
+    def test_caller_dict_is_not_mutated(self, method_name: str) -> None:
+        """A dict passed in by the caller must come back without the code in it."""
+        generate = getattr(AccountService, method_name)
+        caller_data = {"invite_id": "abc"}
+
+        with patch("services.account_service.TokenManager.generate_token", return_value="token") as mock_generate:
+            code, _ = generate("user@example.com", additional_data=caller_data)
+
+        assert caller_data == {"invite_id": "abc"}
+        assert mock_generate.call_args.kwargs["additional_data"] == {"invite_id": "abc", "code": code}
+
+    @pytest.mark.parametrize("method_name", _GENERATORS)
+    def test_concurrent_calls_keep_their_own_code(self, method_name: str) -> None:
+        """Two overlapping calls must each store the code they returned to their caller.
+
+        The generators run under gevent workers, so a second greenlet can write to a
+        shared default dict between the first one's write and TokenManager reading it.
+        """
+        generate = getattr(AccountService, method_name)
+        first_entered = threading.Event()
+        second_wrote = threading.Event()
+        stored: dict[str, str] = {}
+
+        def fake_generate_token(
+            *,
+            email: str,
+            additional_data: dict[str, str] | None = None,
+            **_: object,
+        ) -> str:
+            if email == "first@example.com":
+                first_entered.set()
+                second_wrote.wait(timeout=5)
+            else:
+                second_wrote.set()
+            assert additional_data is not None
+            stored[email] = additional_data["code"]
+            return "token"
+
+        returned: dict[str, str] = {}
+
+        def call(email: str) -> None:
+            returned[email] = generate(email)[0]
+
+        with patch("services.account_service.TokenManager.generate_token", fake_generate_token):
+            first = threading.Thread(target=call, args=("first@example.com",))
+            second = threading.Thread(target=call, args=("second@example.com",))
+            first.start()
+            first_entered.wait(timeout=5)
+            second.start()
+            first.join(timeout=5)
+            second.join(timeout=5)
+
+        assert stored == returned


chuenchen309:
Checked this against `main` since it's been sitting a week. The mechanism is real and the fix is right, but the reachability is narrower than the PR reads — worth pinning down, because it changes which argument gets this merged.

**The mechanism, reproduced.** Copying `generate_reset_password_token` verbatim from `api/services/account_service.py` on `main` (only `TokenManager` stubbed, since it doesn't participate) and forcing the interleaving:

```
default dict before any call: {}
after alice's call:           {'code': '909393'}
after bob's call:             {'code': '844774'}
-> the same dict object is reused: True

codes returned to callers:       {'alice@example.com': '251573', 'bob@example.com': '787541'}
codes the token was minted with: alice@example.com -> {'code': '787541'}
                                 bob@example.com   -> {'code': '787541'}

MISMATCH: alice was told the code is 251573, but her token was minted with 787541
```

So when the default is used, two overlapping calls do collapse onto one code.

**But no caller uses the default.** All four production call sites pass their own dict literal:

| call site | argument |
|---|---|
| `controllers/console/auth/forgot_password.py:139` | `additional_data={"phase": "reset"}` |
| `controllers/web/forgot_password.py:123` | `additional_data={"phase": "reset"}` |
| `controllers/console/auth/email_register.py:144` | `additional_data={"phase": "register"}` |
| `controllers/console/workspace/members.py:492` | `additional_data={}` |

A fresh literal each call, so the shared default is never reached today, and the mutation of the caller's dict lands on an object that's discarded immediately after. `test_concurrent_calls_keep_their_own_code` calls `generate(email)` with no `additional_data`, which is the one shape production never produces — it's a valid test of the mechanism, but it isn't evidence of a live incident.

I'd say that makes this worth merging on the latent-bug argument rather than the security one, and the latent argument is stronger here than usual: `api/.ruff.toml` selects all of `B` but then lists `B006` (`mutable-argument-default`) in `ignore`. So this exact pattern has no lint backstop in this repo — the next call site added without an `additional_data` argument gets the shared dict, and the symptom would be a user occasionally receiving a code that doesn't work while someone else's token carries it. That's a bad thing to find out about from a bug report.

Two small notes on the change itself:

- `{**additional_data} if additional_data else {}` also fixes caller-dict mutation, which is a separate defect from the shared default and worth keeping in the description — it's the one that's reachable today, even if currently harmless.
- The docstring on the concurrency test says the generators run under gevent workers, and the test uses `threading`. That's fine for this bug, since it only needs the two calls to interleave, but the reasoning is worth stating so a reviewer doesn't have to work out whether the test models the real concurrency model.

Framing it as "remove a latent shared-state trap in a lint-blind spot, plus stop mutating the caller's dict" would let a maintainer merge it without first having to rule on whether it's a live security issue. As written it invites that question, and the answer — as far as I can verify — is not yet.

*AI-assisted review; the reproduction output, the four call sites and the ruff config above are from running and reading them on `main`, not inferred. I have no stake in this PR.*

ErenAta16:
You are right on the reachability, and the reframing was the right call. I have rewritten the description to lead with the latent-trap and caller-mutation arguments and to drop the security framing, since as you say it invites a question whose answer is "not yet".

Confirmed both of your load-bearing claims on `main` rather than taking them:

```
api/.ruff.toml:14   "B",       # flake8-bugbear rules      <- all of B selected
api/.ruff.toml:72   "B006",    # mutable-argument-default   <- then ignored
```

and the four call sites are as you list them, each passing its own literal.

**One thing to add, which I think strengthens your own argument.** You enumerated callers for `generate_reset_password_token`. The same mutable default and the same `additional_data["code"] = code` write appear in **three** functions in that file:

```
line 913   generate_reset_password_token
line 928   generate_email_register_token
line 956   generate_owner_transfer_token
```

I checked the callers of the other two as well, and they are clean in the same way (`email_register.py` passes `{"phase": "register"}`, `members.py` passes `{}`), so your conclusion holds for all three. But the pattern having been copy-pasted twice, in a file where the lint that would flag it is switched off, is a sharper version of the point you were making: this is not one function that happened to get it wrong, it is a shape that already propagated. The PR fixes all three and the tests are parametrized over all three.

On your two notes:

- The caller-dict mutation is now called out in its own section as the part that is reachable today. Agreed it is the stronger half of the "why merge this" case, and I had it buried.
- Fixed the concurrency-test docstring. It said gevent workers while the test uses `threading`; it now says the bug only needs the two calls to interleave between reading and writing the shared dict, which threads reproduce deterministically, so modelling the production worker type is not required.

Thanks for reproducing it rather than reasoning about it, and for going and reading the ruff config. The lint-blind-spot point is the argument I should have led with and did not have.

crazywoola:
Hi @ErenAta16! Thanks so much for taking the time to contribute to Dify — we really appreciate it.

ErenAta16:
Fair catch on the process, and thanks for the friendly close.

The issue does exist, I just failed to link it from the description: #39815, opened 30 July, the same finding this PR fixes. That was my mistake, not a missing discussion. I have added `Fixes #39815` to the top of the description now.

Could you reopen this one, or would you rather I open a fresh PR with the link in place from the start? Happy either way, whichever keeps your queue tidier.

One thing worth flagging for whoever picks it up, since it is easy to misread the severity in both directions. @chuenchen309 reviewed this in detail on 7 August and reproduced it against `main` rather than reasoning about it, and the outcome of that exchange was that the original security framing was too strong: all four production call sites pass their own dict literal, so the shared default is not reached today. What is reachable today is that every call mutates the caller's dict. The description has been rewritten around that, so it should be mergeable on the latent-trap argument without anyone having to rule on whether it is a live incident.

The part that made it worth doing at all is that `api/.ruff.toml` selects all of `B` and then lists `B006` (`mutable-argument-default`) in `ignore`, so this pattern has no lint backstop in the repo, and it has already been copy-pasted into three functions in that one file.

Thanks for taking the time to explain rather than just closing.

PR #39828, opened 2026-07-30
fix: refresh mcp server code by app id

zhangx1n:
## What changed

- Changed MCP server code refresh to `POST /apps/{app_id}/server/refresh`.
- Resolve the MCP server by tenant and app, and pass `app_id` from the Web client.
- Regenerated the Console API contract and preserved the existing `APP_VIEW_LAYOUT` permission.

## Why

The previous route exposed `server_id`, while app-scoped RBAC expects `app_id`, causing requests to fail before the controller ran. Regenerating the server code is also a state change, so the endpoint should use POST.


GareArc (approved):
LGTM

PR #39852, opened 2026-07-31
fix: don't use mutable default arguments (shared dict on the password-reset path)

### LINKED ISSUE #39851: Mutable default argument shared across requests in the token generators (password reset path)
### Self Checks

- [x] I have searched the existing issues and found none matching this.
- [x] I am using the latest `main`.

### Dify version

`main` (f9966c2)

### Cloud or Self Hosted

Self Hosted (Docker), also applies to Cloud

### Steps to reproduce

The three token generators in `api/services/account_service.py` take a **mutable default argument** and then write into it:

```python
@classmethod
def generate_reset_password_token(
    cls,
    email: str,
    account: Account | None = None,
    code: str | None = None,
    additional_data: dict[str, Any] = {},      # <- created once, at def time
):
    if not code:
    additional_data["code"] = code             # <- mutates that shared dict
    token = TokenManager.generate_token(
        account=account, email=email, token_type="reset_password", additional_data=additional_data
    )
    return code, token
```

Same shape in `generate_email_register_token` (L909) and `generate_owner_transfer_token` (L936).

The default `{}` is **one object shared by every call that omits the argument**, and all three in-tree callers omit it:

- `account_service.py:707` — `send_reset_password_email`
- `account_service.py:740` — `send_email_register_email`
- `account_service.py:840` — `send_owner_transfer_email`

### Why this is not just a style problem here

There is a yield point between the write and the read. `additional_data` is written in `account_service`, but only read later inside `TokenManager.generate_token` (`api/libs/helper.py:519`, `token_data.update(additional_data)`). In between, when an `account` is supplied — which `send_reset_password_email` does — `generate_token` runs:

```python
if account_id:
    old_token = cls._get_current_token_for_account(account_id, token_type)   # Redis round-trip
    if old_token:
        cls.revoke_token(old_token, token_type)                              # Redis round-trip
```

The default worker is gevent (`SERVER_WORKER_CLASS: geventwebsocket.gunicorn.workers.GeventWebSocketWorker` in `docker/docker-compose.yaml`, `--worker-class` in `api/docker/entrypoint.sh`), so those Redis calls yield to other greenlets **in the same process, sharing the same dict**. Two password-reset requests can interleave:

| | greenlet A | greenlet B | shared default dict |
|---|---|---|---|
| 1 | `additional_data["code"] = "111111"` | | `{"code": "111111"}` |
| 2 | Redis in `generate_token` → yields | | |
| 3 | | `additional_data["code"] = "222222"` | `{"code": "222222"}` |
| 4 | `token_data.update(additional_data)` | | |

A's token is now stored with B's verification code, while the email A receives contains `"111111"` (the local variable). A's reset then fails, and B's code is bound to A's token.

I want to be straight about severity: this is a race with a narrow window, not a guaranteed leak on every request. But it is on the password-reset path, it needs no attacker to trigger, and the fix is two lines per function.

### Second, unconditional effect

When a caller *does* pass a dict, `additional_data["code"] = code` writes into **the caller's** dict. `api/controllers/console/workspace/members.py:527` passes `additional_data={}` and gets a `"code"` key back that it never asked for.

### Expected behavior

Callers that omit `additional_data` should each get a fresh dict, and a caller's own dict should not be modified.

### Actual behavior

One dict is shared process-wide across all calls that omit the argument, and callers' dicts are mutated in place.

### Note

Three more mutable defaults exist elsewhere and are currently harmless (the functions only read them), but they are the same footgun one edit away from mattering:

- `api/configs/remote_settings_sources/apollo/python_3x.py:24` `http_request(headers={})`
- `api/core/rag/extractor/notion_extractor.py:92` `_get_notion_database_data(query_dict={})`
- `api/services/plugin/oauth_service.py:15` `create_proxy_context(extra_data={})`

I have a PR ready that fixes all six with regression tests.

noron12234:
Closes #39851

## Summary

Six functions take a mutable default argument (`= {}`). Three of them write into it, and one of those three is on the password-reset path.

A `dict` default is created **once, when the function is defined**, and shared by every call that omits the argument. `additional_data["code"] = code` therefore mutates process-wide shared state.

## The three that actually mutate

`api/services/account_service.py` — `generate_reset_password_token`, `generate_email_register_token`, `generate_owner_transfer_token`. All three in-tree callers (`send_reset_password_email` L707, `send_email_register_email` L740, `send_owner_transfer_email` L840) omit the argument, so they all share one dict.

The write and the read are separated by a yield point. `additional_data` is written in `account_service` but only read inside `TokenManager.generate_token` at `api/libs/helper.py:519` (`token_data.update(additional_data)`), and in between — when an `account` is supplied, which `send_reset_password_email` does — `generate_token` makes two Redis round-trips:

```python
if account_id:
    old_token = cls._get_current_token_for_account(account_id, token_type)
    if old_token:
        cls.revoke_token(old_token, token_type)
```

The default worker class is gevent (`geventwebsocket.gunicorn.workers.GeventWebSocketWorker`), so those calls yield to other greenlets in the same process, which share the same dict. Two concurrent password resets can interleave so that A's token is stored with B's code while A's email carries A's code.

Being straight about it: narrow window, not a guaranteed leak on every request. But it is on the password-reset path, needs no attacker, and the fix is two lines per function.

**Second, unconditional effect:** when a caller *does* pass a dict, the old code wrote `"code"` into **the caller's** dict. `api/controllers/console/workspace/members.py:527` passes `additional_data={}` and gets a key back it never asked for. The fix copies the input instead.

## The three that are currently harmless

Same footgun, but these only read their default, so nothing is wrong today — they are fixed so the pattern doesn't survive to bite later:

- `api/configs/remote_settings_sources/apollo/python_3x.py:24` — `http_request(headers={})`, does `dict(headers)`
- `api/core/rag/extractor/notion_extractor.py:92` — `_get_notion_database_data(query_dict={})`, does `query_dict.copy()`
- `api/services/plugin/oauth_service.py:15` — `create_proxy_context(extra_data={})`, spreads with `**extra_data`

## What changed

`= {}` → `| None = None` with an explicit fresh dict inside. For the three mutating ones the input is copied, so a caller's dict is never touched:

```python
additional_data: dict[str, Any] | None = None,
):
    additional_data = dict(additional_data) if additional_data else {}
```

No behaviour change for any existing caller beyond no longer being mutated.

## Tests

`api/tests/unit_tests/services/test_account_service.py` — `TestTokenGeneratorsDoNotShareState`, parametrised over all three generators:

1. the default argument is not mutated after a call
2. a caller-supplied dict is not mutated
3. the code still reaches `TokenManager.generate_token` (guards against fixing the bug by dropping the payload)

Tests 1 and 2 fail on `main` and pass with this change.

## Verification

- `pytest tests/unit_tests/services/test_account_service.py -k TokenGeneratorsDoNotShareState`
- Full file re-run to confirm the existing tests are unaffected.


asukaminato0721 (approved):

asukaminato0721 (changes requested):
no getattr

noron12234:
Fair — dropped in cfdfc76. The parametrize now carries the bound classmethods themselves:

```python
GENERATORS = [
    pytest.param(AccountService.generate_reset_password_token, id="generate_reset_password_token"),
    pytest.param(AccountService.generate_email_register_token, id="generate_email_register_token"),
    pytest.param(AccountService.generate_owner_transfer_token, id="generate_owner_transfer_token"),
]

@pytest.mark.parametrize("generator", GENERATORS)
def test_default_argument_is_not_mutated(self, generator) -> None:
    defaults = generator.__func__.__defaults__
    ...
```

No name lookups left in the class, and a type checker can follow it now. The `id=` keeps the test ids readable in the output.

Verified the regression value survived the rewrite — reverting the `account_service.py` fix still fails 6 of the 9:

```console
$ pytest tests/unit_tests/services/test_account_service.py -k TokenGenerators -q   # fix reverted
6 failed, 3 passed
$ pytest tests/unit_tests/services/test_account_service.py -q                      # fix in place
117 passed
```

`ruff check` and `ruff format --check` clean.

noron12234:
@asukaminato0721 The `no getattr` change request is addressed in cfdfc76 (current head of this PR) — there are now zero `getattr` occurrences in the file. The parametrize carries the bound classmethods directly:

```python
GENERATORS = [
    pytest.param(AccountService.generate_reset_password_token, id="generate_reset_password_token"),
    ...
]
```

Could you take another look when you get a chance? The review still shows as changes-requested, so I don't think it'll move without a re-review.

noron12234:
@asukaminato0721 I went back at the `no getattr` note properly.

My earlier fix only removed the literal `getattr` call — the test still read `generator.__func__.__defaults__`, which is the same introspection wearing a different hat. That's on me.

6362b32 drops introspection entirely. The test now calls each generator twice, scribbles on the dict the first call handed to `TokenManager.generate_token`, and asserts the second call isn't holding that same object. With the mutable default put back, it fails like this:

```
assert 'scribble' not in {'code': '222222', 'scribble': True}
```

which states the actual defect — two requests sharing one dict — rather than asserting on a function's internals. There are now zero `getattr`/`__defaults__`/`__func__` occurrences in the file.

Could you take another look when you have a moment? The `CHANGES_REQUESTED` is still recorded against 6491568, which is three commits behind.

noron12234:
Rebased onto current `main` and the conflict is resolved.

Upstream moved the email-registration flow into `account_email_registration_adapters.py`, where `additional_data` is built fresh inside `issue()`, so `generate_email_register_token` is gone and that instance of the bug with it. I took upstream's removal and dropped that generator from the parametrized test. The two remaining generators, `generate_reset_password_token` and `generate_owner_transfer_token`, are unchanged.

@asukaminato0721 the `getattr` you flagged was removed in cfdfc76 back on Aug 1 — the test now parametrizes over the function objects themselves. Sorry for not saying so at the time; the review is still marked as requesting changes, so this needs another look when you have a moment.

Verified locally: `pytest tests/unit_tests/services/test_account_service.py -k TokenGeneratorsDoNotShareState` is 6 passed; putting both signatures back to `= {}` fails 4 of the 6. `ruff format --diff` and `ruff check` are clean on both touched files.

ErenAta16 (approved):
Independent check of this one, since it is the candidate fix pointed at from #39851 and the shape of the bug makes it easy to underestimate.

**The shared default is reachable.** Both controller call sites pass a dict literal, `api/controllers/web/forgot_password.py:125` with `{"phase": "reset"}` and `api/controllers/console/workspace/members.py:497` with `{}`, so reading only those it is easy to conclude the mutable default is never actually used and to file this as hardening. It is used. Both generators are also called from inside `AccountService` with the argument omitted: `send_reset_password_email` at `api/services/account_service.py:604` and `send_owner_transfer_email` at `:703`. Those are the paths that send the verification email, so the shared dict sits on the main flow rather than on an edge case.

**Why the symptom is a wrong code rather than a crash.** In both generators the value that gets emailed is the local `code`, while the value that reaches the token comes out of `additional_data`. `send_reset_password_email` hands the local straight to `send_reset_password_mail_task.delay(..., code=code)`, and the token payload is read from the dict inside `TokenManager.generate_token`. With one dict shared across calls, a switch between `additional_data["code"] = code` and that read leaves the two out of step, so a user is mailed one code while their token carries another. That is the "the code in my email does not work" class of report, which is why this is worth more than a lint cleanup.

**The copy is safe.** `dict(additional_data) if additional_data else {}` does change behaviour for callers that pass a dict, since `["code"]` is no longer written back into the object they handed in. I enumerated every call site on `main@79effdd4` and none of them reads the dict afterwards: the two controllers pass an inline literal and discard it, the two internal callers pass nothing at all, and the tests patch the generator itself. So nothing depends on the old write-through, and dropping it is a straight improvement given these are request-scoped dictionaries.

**Coverage against current main.** `api/services/account_service.py` has exactly two mutable-default parameters left on main, at lines 761 and 794, and those are the two this PR fixes. `generate_change_email_token` already takes a typed payload and `generate_invite_token` builds its dict inside the function body, so account_service is complete once this lands.

**On the stale review state.** The changes-requested above was submitted against 6491568 and asked for the `getattr` to go. The next commit on the branch, cfdfc76e, "test: parametrize over the functions themselves instead of getattr", landed about eight hours later, and there is no `getattr` left at the current head eb5bdaa9. @asukaminato0721, a re-review would clear that state if you are happy with where it ended up.

To be clear about the basis for this: it is a read of the code at `main@79effdd4` and of the PR at `eb5bdaa9`, plus the call-site enumeration above. I did not run the api suite locally.

asukaminato0721:
ci

asukaminato0721 (changes requested):
^

PR #39933, opened 2026-08-03
feat: support azure keyvault

### LINKED ISSUE #32328: dify secret integrated with 3rd part provider
### Self Checks

i was using n8n before trying dify and i really miss a feature.
in n8n it was possible to integrate AWS Secrets Manager, Azure Key Vault, HashiCorp Vault (and other, now i only try azure and hashicorp to be honest) and fetch a set of keys from the vault.
keys are then considered as 'secret' and in workflow in n8n can be used like any other secret.

What's the differente between storing them into dify secret? Well management is much more simple as you have only one source of truth, if db password rotate or apikey for http call rotate, you don't have to do anything basically, as rotation happen on vault (so yes, vault should also refresh values every now and then or use the api that automatically get the secret when change, but to be honest i don't know how they work internally)

_No response_

wylswz:
> [!IMPORTANT]
>

## Summary

Partially fixes #32328 
resolves ENG-740

### Goals
- Provided an abstraction of KMS provider, including key provision, encryption and decryption.
- Direct access to `libs.rsa` from biz packages is prohibited.

### Non-Goals
- cypher text storage including `encrypted_xxx` fields of credentials.

## Screenshots

image

## Checklist


api/libs/key_providers/azure_keyvault_key_provider.py
```diff
@@ -0,0 +1,231 @@
...
+            # caller of decrypt_token_with_decoding (core/provider_manager.py,
+            # services/model_load_balancing_service.py) already only expects/suppresses
+            # ValueError for "this particular credential can't be decrypted right now", so Azure
+            # SDK errors must be translated here rather than left to escape as a different type
+            # and crash the whole call chain (e.g. building a tenant's full provider
+            # configuration just to create an unrelated new credential).
+            crypto_client, _ = self._get_crypto_client(tenant_id, version=metadata.get("key_version"))
+            aes_key = crypto_client.unwrap_key(wrap_alg, wrapped_key).key
+        except AzureError as exc:
+            raise ValueError(f"Failed to unwrap credential via Azure Key Vault: {exc}") from exc
```
CourTeous33, line 200: AzureError will be re-raised as a ValueError, which would be suppressed by here. https://github.com/langgenius/dify/blob/342776202c276dd2ba9390053d781f03f3ca606e/api/core/provider_manager.py#L1508-L1511
This means every error comes from Azure side will be ignored

But I do see this [PR](https://github.com/langgenius/dify/pull/38950/changes) that un-suppress valueerrors, so not sure what is the design purpose of this re-raise.

wylswz: When adding credentials to a model provider, the provider manager attempts to decrypt the existing credentials and falls back to an empty value if decryption raises a `ValueError`.

With the Azure Key Vault implementation, each ciphertext is associated with a specific key version. If the key is rotated and the previous version is disabled, decrypting credentials encrypted with that version raises an `AzureError`. Because this exception is not currently handled, the operation fails and prevents users from adding new credentials.

This change gracefully handles Azure Key Vault decryption failures and preserves the existing fallback behavior.

CourTeous33 (dismissed):

CourTeous33 (dismissed):

api/libs/key_providers/azure_keyvault_key_provider.py
```diff
@@ -0,0 +1,231 @@
...
+        vault_url = dify_config.AZURE_KEYVAULT_VAULT_URL
+        if not vault_url:
+            raise ValueError("AZURE_KEYVAULT_VAULT_URL must be configured when KEY_PROVIDER_TYPE=azure-keyvault")
+
+        self._credential = DefaultAzureCredential()
+        self._key_client = KeyClient(vault_url=vault_url, credential=self._credential)
+        # Cached per (tenant_id, *concrete* version). A version's key material is immutable
+        # once created, so entries never need to expire -- there's deliberately no cache entry
+        # keyed by "current" (version=None): see _get_crypto_client().
+        self._crypto_clients: dict[tuple[str, str], CryptographyClient] = {}
```
laipz8200: This cache is missing an eviction policy, which could lead to a memory leak.

laipz8200 (changes requested):

api/libs/key_providers/azure_keyvault_key_provider.py
wylswz: Switched to `LRUCache`.

api/libs/key_providers/azure_keyvault_key_provider.py
```diff
@@ -0,0 +1,233 @@
...
+    def __init__(self):
+        vault_url = dify_config.AZURE_KEYVAULT_VAULT_URL
+        if not vault_url:
+            raise ValueError("AZURE_KEYVAULT_VAULT_URL must be configured when KEY_PROVIDER_TYPE=azure-keyvault")
+
+        self._vault_url = vault_url
+        self._credential = DefaultAzureCredential()
+        self._key_client = KeyClient(vault_url=vault_url, credential=self._credential)
+        # Cached per (tenant_id, *concrete* version).
+        self._crypto_clients: LRUCache[tuple[str, str], CryptographyClient] = LRUCache(
```
laipz8200: When an LRU cache evicts items, it doesn't actively call the client's `close()` method. This can cause sockets to stay open and prevent them from being reclaimed in a timely manner.

While this might not be a serious issue on its own, I'm looking forward to a better solution.

laipz8200 (dismissed):
Almost LGTM

api/libs/key_providers/azure_keyvault_key_provider.py
wylswz: We could subclass the `LRUCache` to add custom eviction behavior.

api/libs/key_providers/azure_keyvault_key_provider.py
```diff
@@ -0,0 +1,239 @@
...
+# self-describing, even though the two providers never decode each other's payloads.
+_PREFIX = b"HYBRID:"
+
+# Bump this if the *binary layout* below ever changes. Adding new metadata fields does NOT
+# require a bump (metadata is a JSON object -- old readers just ignore unknown keys via .get()).
+_ENVELOPE_VERSION = 1
+
+_DEFAULT_WRAP_ALGORITHM = KeyWrapAlgorithm.rsa_oaep_256
+
+_CRYPTO_CLIENT_CACHE_MAXSIZE = 1
```
laipz8200: The KeyProvider is a singleton. If the cache max size is set to 1, the cache loses its effect.

api/libs/key_providers/azure_keyvault_key_provider.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+class _CryptoClientCache(LRUCache[tuple[str, str], CryptographyClient]):
+    @override
+    def popitem(self) -> tuple[tuple[str, str], CryptographyClient]:
```
laipz8200: This operation introduces a race condition: if a client is popped while being held by a method, it can be closed before unwrap, causing unwrap to fail.

laipz8200 (changes requested):

api/libs/key_providers/azure_keyvault_key_provider.py
wylswz: That was for debug purpose, will rollback.

api/libs/key_providers/azure_keyvault_key_provider.py
wylswz: Will use `KeyClient.get_cryptography_client` instead, so the underlying client instance could be shared.

wylswz: Removed caching. 

The version lookup only happens in encryption where version parameter is `None`. Given that encryption is less frequent compared to decryption, so the cost is acceptable.

laipz8200 (dismissed):

laipz8200 (approved):

PR #39953, opened 2026-08-03
fix(vdb): bound the TiDB Cloud API timeouts

he-yufeng:
## What

The TiDB-on-Qdrant provider's two TiDB Cloud calls (`create_tidb_serverless_cluster` and `change_tidb_serverless_root_password`) run `httpx.post`/`put` without a timeout, so a slow or hanging TiDB Cloud API blocks the provisioning or rotation thread forever. Both now use a bounded timeout with the same shape as the recent extractor timeout fixes (30s read, 5s connect).

## Tests

New cases assert both calls pass a bounded `timeout` kwarg through. I verified the kwargs end to end through the real factory with `httpx.post`/`put` mocked (`Timeout(connect=5.0, read=30.0, write=30.0, pool=30.0)` on both). The full provider test file could not run locally here: the installed graphon predates `TTSModelVoice` and `graphon.variables.template_resolution`, and the package index is unreachable from this machine, so the heavy parent modules were stubbed for the local verification. CI covers the real suite.

From Kimi Code


laipz8200 (approved):

PR #39995, opened 2026-08-04
fix(api): reserve batch document retries atomically

### LINKED ISSUE #39975: Batch document retry can leave documents waiting without an indexing task
### Self Checks

### Dify version

main (545bdecff2)

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

1. Prepare two documents in the same dataset with a retryable indexing status.
2. Start a retry for the second document so its Redis retry flag already exists.
3. Send a batch retry request containing both document IDs, with the first document listed before the document that is already being retried.
4. Observe the document states after the request fails.

The same check also uses a separate Redis GET followed by SETEX, so two concurrent retry requests for the same document can both pass the initial check.

### ✔️ Expected Behavior

The complete batch should be reserved before any document state is changed. If any selected document is already being retried, the request should reject the batch without updating any document or dispatching a retry task.

### ❌ Actual Behavior

The service commits each document as waiting before checking the next document. If a later document is already being retried, the method raises before dispatching the task, leaving earlier documents in waiting without an indexing task. The non-atomic GET/SETEX sequence can also admit concurrent retries.

he-yufeng:
Fixes #39975.

`retry_document` committed each document as `waiting` one at a time and only then checked the next document's retry flag. A batch where a later document was already being retried raised mid-loop and left the earlier documents in `waiting` with no indexing task ever dispatched, since the task goes out only after the loop. The flag check itself was a separate GET followed by SETEX, so two concurrent retry requests for the same document could both pass.

The method now claims every flag atomically with SET NX up front, commits all statuses in one transaction, and releases the claimed flags plus rolls back on any failure, so a partially-retryable batch changes nothing at all.

## Verification

Three unit tests in `test_dataset_service_document.py`: a lost atomic claim rejects without any commit, a mid-batch claim failure releases the earlier claim and leaves the first document untouched, and a free batch commits all documents once and dispatches the task. Full file: 80 passed with the repo conftest.

From Kimi Code

### DIFF
--- a/api/services/dataset_service.py
+++ b/api/services/dataset_service.py
@@ -2099,18 +2099,27 @@ class DocumentService:
 
     @staticmethod
     def retry_document(dataset_id: str, documents: list[Document], session: Session):
-        for document in documents:
-            # add retry flag
-            retry_indexing_cache_key = f"document_{document.id}_is_retried"
-            cache_result = redis_client.get(retry_indexing_cache_key)
-            if cache_result is not None:
-                raise ValueError("Document is being retried, please try again later")
-            # retry document indexing
-            document.indexing_status = IndexingStatus.WAITING
-            session.add(document)
-            session.commit()
+        # Reserve every document before changing any state. Committing one
+        # document as waiting and then finding the next one already retried
+        # would leave the first in waiting with no task ever dispatched.
+        claimed_keys: list[str] = []
+        try:
+            for document in documents:
+                retry_indexing_cache_key = f"document_{document.id}_is_retried"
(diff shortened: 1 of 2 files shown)


crazywoola:
Hi @he-yufeng, thank you very much for taking the time to contribute this fix.

We handle overlapping pull requests on a first-in, first-out basis. Since #39978 was opened earlier for #39975 and has now been merged, this later pull request has been closed as a duplicate.

We really appreciate your contribution and hope to see more contributions from you in the future!

Issue #32450, opened 2026-02-21
bug: FC agent runner silently loses data when LLM calls the same tool multiple times


trongtrandp:
## Self Checks

- [x] This is only a bug report, not a feature request or other type of issue.
- [x] I have searched for existing issues to avoid creating duplicates.

## Dify version

1.13.0

## Cloud or Self Hosted

Self Hosted (Docker)

## Steps to reproduce

1. Create an Agent app with Function Calling mode
2. Add a tool (e.g., `search`) that the LLM might call multiple times
3. Send a prompt that causes the LLM to call the same tool 2+ times in a single thought (e.g., "Search for 'Python tutorial' and also search for 'JavaScript tutorial'")
4. Check the agent thought logs

## ✅ Expected Behavior

All tool calls should be preserved with their individual inputs and outputs. For example:
- `search("Python tutorial")` → result A
- `search("JavaScript tutorial")` → result B

## ❌ Actual Behavior

Only the **last** tool call's data survives. Previous calls are silently overwritten.

In the example above, only `search("JavaScript tutorial")` → result B is stored. The first search call is completely lost.

## Root Cause

The FC agent runner uses **dict comprehensions keyed by tool name** to store tool data. When the same tool is called multiple times, later entries overwrite earlier ones:

```python
# api/core/agent/fc_agent_runner.py

# Line ~128 (streaming path) and ~155 (blocking path):
tool_call_inputs = json.dumps(
    {tool_call[1]: tool_call[2] for tool_call in tool_calls},  # ← dict key collision!
    ensure_ascii=False,
)

# Line ~287-292 (observation and meta):
tool_invoke_meta={
    tool_response["tool_call_name"]: tool_response["tool_invoke_meta"]
    for tool_response in tool_responses  # ← same collision
},
observation={
    tool_response["tool_call_name"]: tool_response["tool_response"]
    for tool_response in tool_responses  # ← same collision
},
```

**4 collision sites** in total — affecting `tool_input`, `observation`, and `tool_invoke_meta` storage.

## Impact

- **Data loss**: Tool call inputs, outputs, and metadata are silently discarded
- **Misleading logs**: Agent thought logs show only one tool call when multiple were actually executed
- **Affects all FC agent apps** where the LLM decides to call the same tool more than once in a single thought

### Degraded agent reasoning in multi-turn conversations

This bug doesn't just affect log display — it **directly degrades agent quality** in subsequent turns.

[...]

Issue #16220, opened 2025-03-19
Agent API Streaming - if a tool gets called twice `tool_input` and `observation` are returned for just one call.


tognee:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.0.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Create an Agent with ddg image search.
Ask the Agent to see a picture of a dog and a picture of a cat.
See the result.

### ✔️ Expected Behavior

ddgo_img is called two times, see the query "dog" and the query "cat", see each call output.

### ❌ Actual Behavior

ddgo_img is called two times, you can only see the "cat" query and output.

tognee:
The issue is that `observation` and `tool_input` uses the tool name as the key, it should use the tool_call_id instead.
I think the agent_thinking event should be rethinked.

crazywoola:
What is your model provider and it's version?

tognee:
> What is your model provider and it's version?

tried with openai, azure openai and vertexai

but I think this is not linked with a specific provider

tognee:
The issue is still present because the code hasn't been changed.

The problem is in the file [fc_agent_runner.py](https://github.com/langgenius/dify/blob/main/api/core/agent/fc_agent_runner.py):
- On [line 129](https://github.com/langgenius/dify/blob/main/api/core/agent/fc_agent_runner.py#L129) and [line 156](https://github.com/langgenius/dify/blob/main/api/core/agent/fc_agent_runner.py#L156) each tool call is appended in an array with indexes `0` for `tool_call_id`, `1` for `tool_name` and `2` for `tool_arguments`
[...]

PR #40000, opened 2026-08-04
fix(agent): retain every tool call's input and result when a tool repeats in a turn

dparkmit24:
> [!IMPORTANT]
>

## Summary

When a model calls the same tool twice in one turn — two searches, two file reads — one of the results is discarded, the agent log shows the surviving one twice, and on the next turn the model is shown two identical calls it never made.

**Reproduction** (from #16220): give an Agent app in function-calling mode an image-search tool and ask it for a picture of a dog and a picture of a cat. The tool is invoked twice, but the agent log shows only the "cat" query and only the "cat" result.

**Cause.** `fc_agent_runner` stores a turn's tool data in dicts keyed by tool name:

```python
tool_call_inputs = json.dumps({tool_call[1]: tool_call[2] for tool_call in tool_calls}, ...)
observation={tool_response["tool_call_name"]: tool_response["tool_response"] for ...}
```

Two calls to `search`, one surviving key. Four sites: the streaming and blocking input paths and the observation and meta builds.

Replay compounds it. `base_agent_runner.organize_agent_history` splits the `";"`-joined names back apart and looks each one up **by name**, so both calls come back carrying the surviving call's arguments and the surviving call's result. The model made two different calls and got two different answers; the next turn shows it two identical ones.

The display has the same shape of bug one layer out. `AgentService.get_agent_logs` builds one log entry per name in the `";"`-joined column and then looks each name up in a name-keyed dict, so both entries read the same key.

**Fix, in two parts.**

1. **Storage and replay.** Keep one value per call, in call order, when a name repeats in a turn. A tool called once keeps its bare value. `tool_input`, `observation` **and `tool_invoke_meta`** are grouped the same way.
2. **Display.** `MessageAgentThought` gains `tool_inputs_per_call`, `tool_outputs_per_call` and `tool_metas_per_call` — one entry per entry in `tools`, in call order — and `get_agent_logs` walks them alongside the names, so a repeated tool renders one entry per call with that call's own input, output and provider fields.

**`tool_invoke_meta` is included, and an earlier revision of this PR was wrong to exclude it.** It is not merely diagnostic: `get_agent_logs` reads `tool_config` → the provider icon, `time_cost` → the duration, and `error` → the status icon and error box **for each displayed call**. The reason it was left out was that widening it alone breaks `get_agent_logs`, which calls `.get()` on each tool's meta entry — and that reason dissolves once the display path reads per call, which this PR now does. Its other reader, `ops_trace_manager.tool_trace`, covers the tool rather than one call of it and has no call to attach to, so it takes the last call's meta — the only one it had before.

**Correction to an earlier revision of this description.** It said that for a repeated tool "the panel now shows both calls' data instead of one call's data twice". **That was true only in the narrowest sense and it misled.** Grouping the payloads without changing the display path meant both rendered entries received the whole per-call list, so the panel went from showing one call's data twice to showing **both calls' data twice** — a different wrong answer, not a right one. The panel now shows **one entry per call, each with its own data**, which is what that sentence should have claimed. Thanks to @chuenchen309 for catching it; he traced it on `main` without building anything and he was right.

**Compatibility — the part worth reviewing.** These are existing `LongText` columns with existing rows, so the fix must not change how any of them read. No schema change and no migration.

- A tool called once in a turn serialises exactly as before. Single-call turns and turns using several *different* tools are unchanged, byte for byte.
- Rows written by earlier versions hold one value under a repeated name. The reader only consumes positionally when the name repeats **and** the stored value is a list of matching length, so those rows replay — and render — exactly as they do today.
- The only rows whose shape changes are the ones that are currently losing data, where no faithful record exists to stay compatible with.
- `tool_inputs_dict` / `tool_outputs_dict` / `tool_meta` are left exactly as they are; the per-call readers are new properties layered on top.
- The Agent v2 runner writes one row per tool call, so no name ever repeats in a row and nothing about it changes.

Compatibility is covered by tests rather than by argument: the legacy records are constructed as literals — never by calling the new writer — at both layers, `test_legacy_record_replays_unchanged` for the model-facing replay and `test_legacy_repeated_tool_renders_exactly_as_it_did` for the rendered agent log.

**Why not key by `tool_call_id`?** Both prior reports suggested it, and it is the obvious answer, but `MessageAgentThought.tool_inputs_dict`, `.tool_outputs_dict` and `AgentService.get_agent_logs` all look these payloads up by tool name. With id keys those lookups miss and the agent-log panel goes blank for exactly the turns this fixes. Grouping under the existing name key keeps every reader working.

**Not covered.**

- `cot_agent_runner` does not have this defect. Its scratchpad holds a single `Action` per iteration, so a turn can only contain one tool call and there is nothing to collide.
- The Agent v2 runner is unaffected — it already binds results by `tool_call_id` (#38180, #38556).
- **Open question, raised by @chuenchen309:** the per-call selector discriminates positionally — `occurrences > 1 and isinstance(value, list) and len(value) == occurrences` — which cannot distinguish a per-call list from a legacy single-call value that happens to be a list of the same length. He suggested a sentinel-tagged shape (`{"__calls__": [...]}`). It is deliberately **not** changed in this PR: a sentinel would be a third on-disk shape alongside bare-value and bare-list that every reader would need to know about, and the trade-off deserves its own decision. Happy to take it either way.
- The live chat thought panel is unchanged. It renders the `";"`-joined name as a single entry with the raw payload string and does not split per call; making it do so means changing the format of the `tool` column itself, which is a much wider change than this one. No frontend file is touched.

**Prior reports of this defect:** #16220 (2025-03, with a reproduction and a correct root-cause diagnosis) and #32450 (2026-02, which enumerated all four collision sites). Both reports were accurate and both still describe `main`. Credit to @tognee and @trongtrandp — the diagnosis in this PR is theirs.

**Testing.** The full `api/tests/unit_tests` tree: **16804 passed, 2 failed, 6 skipped** (8m11s). Both failures are pre-existing and unrelated — `test_hitl_service_api.py`'s two `expiration_time` assertions are off by exactly 28800s on a non-UTC machine, and they fail identically on this branch's base commit. `ruff check`, `ruff format --check` and `mypy` are clean on every changed file. **No frontend check was run because no frontend file is changed.**

## Screenshots

| Before | After |
|--------|-------|
| agent log: two `search` entries, both showing the same input and output | agent log: two `search` entries, one per call, each with its own input, output and provider |
| replay shows two identical `search` calls with the same result | replay shows two distinct calls with their own results |

## Checklist

- [ ] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods — backend only; see **Testing** above

From Claude Code

### DIFF
--- a/api/core/agent/base_agent_runner.py
+++ b/api/core/agent/base_agent_runner.py
@@ -1,8 +1,9 @@
 import json
 import logging
 import uuid
+from collections import Counter
 from decimal import Decimal
-from typing import Union, cast
+from typing import Any, Union, cast
 
 from sqlalchemy import func, select
 from sqlalchemy.orm import Session
@@ -46,6 +47,21 @@ from models.model import Conversation, Message, MessageAgentThought, MessageFile
 
 logger = logging.getLogger(__name__)
 _file_access_controller = DatabaseFileAccessController()
+
+
+def _select_tool_occurrence(value: Any, occurrence: int, occurrences: int) -> Any:
+    """
+    Pick one call's value out of a persisted agent thought payload.
+
+    A tool called several times in one turn stores one value per call, in call
+    order. Records written before those calls were kept apart store a single
+    value for the tool name, and every occurrence replays it — the behaviour
+    those records were written with.
+    """
+    if occurrences > 1 and isinstance(value, list) and len(value) == occurrences:
+        return value[occurrence]
+
+    return value
 
 
 class BaseAgentRunner(AppRunner):
@@ -408,7 +424,11 @@ class BaseAgentRunner(AppRunner):
                         else:
                             tool_responses = dict.fromkeys(tool_names, observation_payload)
 
+                        tool_occurrences = Counter(tool_names)
+                        seen_tools: Counter[str] = Counter()
                         for tool in tool_names:
+                            occurrence = seen_tools[tool]
+                            seen_tools[tool] += 1
                             # generate a uuid for tool call
                             tool_call_id = str(uuid.uuid4())
                             tool_calls.append(
@@ -417,13 +437,21 @@ class BaseAgentRunner(AppRunner):
                                     type="function",
                                     function=AssistantPromptMessage.ToolCall.ToolCallFunction(
                                         name=tool,
-                                        arguments=json.dumps(tool_inputs.get(tool, {})),
+                                        arguments=json.dumps(
+                                            _select_tool_occurrence(
+                                                tool_inputs.get(tool, {}), occurrence, tool_occurrences[tool]
+                                            )
+                                        ),
                                     ),
                                 )
                             )
                             tool_call_response.append(
                                 ToolPromptMessage(
-                                    content=tool_responses.get(tool, agent_thought.observation),
+                                    content=_select_tool_occurrence(
+                                        tool_responses.get(tool, agent_thought.observation),
+                                        occurrence,
+                                        tool_occurrences[tool],
+                                    ),
                                     name=tool,
                                     tool_call_id=tool_call_id,
                                 )
--- a/api/core/agent/base_agent_runner.py
+++ b/api/core/agent/base_agent_runner.py
@@ -57,6 +57,19 @@ def _select_tool_occurrence(value: Any, occurrence: int, occurrences: int) -> An
     order. Records written before those calls were kept apart store a single
     value for the tool name, and every occurrence replays it — the behaviour
     those records were written with.
+
+    `observation` values are always `str`: `ToolEngine.agent_invoke` is typed
+    `-> tuple[str, list[str], ToolInvokeMeta]` and the runners store element 0,
+    so a list there is not a shape any writer produces and the length check is
+    defensive. `tool_input` values are `json.loads` of the model's `arguments`
+    with no shape check, so a legacy list-valued input is possible in principle;
+    on a length collision the rule reads per call, not whole —
+    `test_a_legacy_list_of_matching_length_is_read_per_call_not_whole` pins it.
+
+    The same function is defined, identically, in `models/model.py`.
+    The duplication is deliberate: `models/` importing from `core/agent/` is the
+    worse layering trade and the reverse is odd, so neither copy is in the wrong
+    place and neither should move. A change to this rule must be applied in both.
     """
     if occurrences > 1 and isinstance(value, list) and len(value) == occurrences:
         return value[occurrence]
--- a/api/core/agent/base_agent_runner.py
+++ b/api/core/agent/base_agent_runner.py
@@ -3,7 +3,7 @@ import logging
 import uuid
 from collections import Counter
 from decimal import Decimal
-from typing import Any, Union, cast
+from typing import Union, cast
 
 from sqlalchemy import func, select
 from sqlalchemy.orm import Session
@@ -43,25 +43,17 @@ from graphon.model_runtime.entities.message_entities import ImagePromptMessageCo
 from graphon.model_runtime.entities.model_entities import ModelFeature
 from graphon.model_runtime.model_providers.base.large_language_model import LargeLanguageModel
 from models.enums import CreatorUserRole
-from models.model import Conversation, Message, MessageAgentThought, MessageFile, load_annotation_reply_config
+from models.model import (
+    Conversation,
+    Message,
+    MessageAgentThought,
+    MessageFile,
+    load_annotation_reply_config,
+    select_tool_occurrence,
+)
 
 logger = logging.getLogger(__name__)
 _file_access_controller = DatabaseFileAccessController()
-
-
-def _select_tool_occurrence(value: Any, occurrence: int, occurrences: int) -> Any:
-    """
-    Pick one call's value out of a persisted agent thought payload.
-
-    A tool called several times in one turn stores one value per call, in call
-    order. Records written before those calls were kept apart store a single
-    value for the tool name, and every occurrence replays it — the behaviour
-    those records were written with.
-    """
-    if occurrences > 1 and isinstance(value, list) and len(value) == occurrences:
-        return value[occurrence]
-
-    return value
 
 
 class BaseAgentRunner(AppRunner):
@@ -438,7 +430,7 @@ class BaseAgentRunner(AppRunner):
                                     function=AssistantPromptMessage.ToolCall.ToolCallFunction(
                                         name=tool,
                                         arguments=json.dumps(
-                                            _select_tool_occurrence(
+                                            select_tool_occurrence(
                                                 tool_inputs.get(tool, {}), occurrence, tool_occurrences[tool]
                                             )
                                         ),
@@ -447,7 +439,7 @@ class BaseAgentRunner(AppRunner):
                             )
                             tool_call_response.append(
                                 ToolPromptMessage(
-                                    content=_select_tool_occurrence(
+                                    content=select_tool_occurrence(
                                         tool_responses.get(tool, agent_thought.observation),
                                         occurrence,
                                         tool_occurrences[tool],
--- a/api/core/agent/base_agent_runner.py
+++ b/api/core/agent/base_agent_runner.py
@@ -3,7 +3,7 @@ import logging
 import uuid
 from collections import Counter
 from decimal import Decimal
-from typing import Union, cast
+from typing import Any, Union, cast
 
 from sqlalchemy import func, select
 from sqlalchemy.orm import Session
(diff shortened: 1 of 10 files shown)


chuenchen309:
I read through this against `main`. The diagnosis is right and the persistence design is careful — keeping a bare value for the single-call case so existing rows keep replaying exactly as before, rather than migrating them, is the part I'd have gotten wrong.

One thing I think matters before this lands: **the path your reproduction describes is not the path this PR fixes.**

The repro is "the agent log shows only the `cat` query and only the `cat` result". That log is built in `agent_service.py`, not in `organize_agent_history`:

```python
for agent_thought in agent_thoughts:
    tools = agent_thought.tools                      # ["search", "search"]
    tool_inputs = agent_thought.tool_inputs_dict
    tool_outputs = agent_thought.tool_outputs_dict or {}
    for tool in tools:                               # iterates twice, same name
        tool_input = tool_inputs.get(tool_name, {})  # same value both times
        tool_output = tool_outputs.get(tool_name, {})
        tool_meta_data = tool_meta.get(tool_name, {})
```

`tool_inputs_dict` and `tool_outputs_dict` (`models/model.py`) still read the old shape — `for tool in tools: result[tool] = data[tool]` — so with the new list-valued payload both iterations get the same list. After this PR the log stops showing only "cat", but it shows `[dog, cat]` twice rather than one call each. The model-facing history is fixed; the user-facing log changes shape without becoming correct.

Two smaller things in the same area:

**`tool_invoke_meta` is the third per-call field and wasn't grouped.** It sits two lines above the `observation` you did change, with the identical comprehension:

```python
tool_invoke_meta={
    tool_response["tool_call_name"]: tool_response["meta"] for tool_response in tool_responses
},
observation=_group_by_tool_name(...)   # changed
```

So `tool_meta` in the loop above has the same last-one-wins collision, and that's what feeds `tool_provider_type` / `tool_provider_id` per displayed call.

**`_select_tool_occurrence`'s discriminator is positional.** `occurrences > 1 and isinstance(value, list) and len(value) == occurrences` can't distinguish a per-call list from an old single-call value that happens to be a list of the same length. `extract_tool_calls` annotates the args as `dict[str, Any]` but actually assigns `json.loads(...)` unchecked, so a model emitting a JSON array as `arguments` would land there. I couldn't construct a realistic case — it needs a non-object `arguments` *and* a length collision — so I'd treat it as a note rather than a blocker, but a sentinel-tagged shape (e.g. `{"__calls__": [...]}`) would remove the ambiguity entirely if you'd rather not rely on that.

None of this argues against the change — the storage format has to be fixed first either way, and doing it back-compatibly is the right call. It does suggest `agent_service.py` and the two `models/model.py` properties belong in the same PR, otherwise #16220 stays open from the reporter's point of view.

*This is a read-only review: I traced the paths above on `main` and did not build Dify or run the test suite, so I haven't independently confirmed the pass/fail claims in your description. AI-assisted with agentic coding tools; all conclusions above were reviewed by me.*

dparkmit24:
@chuenchen309 You were right on all three, and the display half is now in — `c881abff` plus `8b91f5bc`.

`get_agent_logs` now walks three new per-call readers on `MessageAgentThought` alongside the names, so a repeated tool renders one entry per call with its own input, output and provider fields. `tool_invoke_meta` is grouped with the other two: your point that it drives the icon, duration and status per displayed call is what changed my mind — I'd excluded it because widening it alone breaks `get_agent_logs`, and that reason dissolves once the display path reads per call.

One more reader turned up while doing it: `ops_trace_manager.tool_trace` also calls `.get()` on the meta and would have thrown once it became a list. It covers the tool rather than one call of it and has no call index to attach to, so it takes the last call's meta — which is the only one it had before this change anyway.

I've corrected the description in place rather than quietly editing the sentence out, since you'd already read it.

On the discriminator — I've left it alone for now, and here's my hesitation rather than a decision. A sentinel would be a third on-disk shape alongside bare-value and bare-list, and after this change there are four readers that would each need to recognise it: `organize_agent_history`, `get_agent_logs`, `tool_trace`, and anything parsing the raw strings off the two API response models, which I can't see the far side of. `test_a_stored_list_that_is_not_one_value_per_call_is_replayed_whole` pins the current behaviour by length so that whichever way this goes, it shows up as a test change rather than silently. If you think the ambiguity is worth the third shape I'll do it — you've been right about the parts I got wrong so far.

chuenchen309:
On the discriminator: **I'd leave it out too**, and I don't think that's a close call. The call count is already recorded in `tools`, so a sentinel would be a second encoding of information the row already carries — you'd be paying four readers plus an unbounded set of raw-string consumers for redundancy, not for new signal. `_select_tool_occurrence`'s `occurrences > 1 and isinstance(value, list) and len(value) == occurrences` is the right shape for that.

What I'd do instead is cheaper and gets you most of what the sentinel was for: **pin the colliding case as well as the non-colliding one.**

`test_a_stored_list_that_is_not_one_value_per_call_is_replayed_whole` uses a 3-element list against 2 calls, so it pins the side where the length rule is unambiguous. The side that decides whether the rule is safe is the other one — a tool called *n* times whose single stored value happens to be a list of length *n*:

```python
tool="search;search",
observation=json.dumps({"search": ["first result", "second result"]}),
```

Read as one value, both calls should replay `["first result", "second result"]`. Read by length, call 1 gets `"first result"` and call 2 gets `"second result"` — plausible-looking history that never happened. Right now that case is decided by the rule but not described by any test, so it reads as an oversight rather than a decision.

A test asserting the length-based reading, named for what it concedes, does the job a sentinel would have done: it makes the ambiguity explicit and turns any future change to it into a visible test change. That's the same argument you already made for the existing test, applied to the half that actually carries the risk.

**One thing I could not settle from the diff**, and it's what decides how much residual risk is left: whether a stored `tool_response` is ever list-valued in rows written before this change. If tool responses were always strings there, the collision is unreachable for outputs and this is purely defensive; if some tools stored lists, then old multi-call rows for those tools are exactly where it bites. You or a maintainer can answer that much faster than I can from outside — and if the answer is "always strings", I'd say say so in the docstring, because that is the fact the length rule is leaning on.

Nice catch on `ops_trace_manager.tool_trace`, and taking last-call meta there is right — it matches what the field already meant.

*AI-assisted review; the reader logic, the writer shapes and the new tests above were read on this branch rather than inferred.*

dparkmit24:
@chuenchen309 : On the discriminator — agreed, and your reason is better than mine. `tools` already carries the call count, so a sentinel would be a second encoding of something the row states; I was weighing it as new signal and it isn't.

On your open question, I can settle it: `tool_response` is never list-valued. `ToolEngine.agent_invoke` is typed `-> tuple[str, list[str], ToolInvokeMeta]`, and `fc_agent_runner` unpacks the first element straight into the observation dict. So for `observation` the collision is unreachable and the length rule is purely defensive.

`tool_input` is the half that carries the risk, and it's the one the length rule is actually leaning on. `extract_tool_calls` assigns `json.loads(function.arguments)` into a field annotated `dict[str, Any]` without checking, so a model emitting a top-level JSON array does land there. It's rare and it needs a length collision on top, but it isn't structurally impossible the way the output side is.

I'll put both facts in the docstring rather than leaving them implicit, and add the test you describe — a tool called n times whose single stored value is a list of length n, asserting the length-based reading, named for what it concedes. You're right that the existing test pins the unambiguous side and leaves the risky one decided-but-undescribed, and that's the same argument I made for the legacy tests. Will push shortly.

chuenchen309:
*(Edited: I posted this a minute after your reply without having read it, and most of what was here — the `ToolEngine.agent_invoke` / `tool_response_to_str` trace showing observations are never list-valued — you had already worked out yourself. Sorry for the duplicate. Trimmed to the one thing it added.)*

While you're in the docstring anyway: **`_select_tool_occurrence` is now defined twice**, identically including its docstring — `core/agent/base_agent_runner.py` and `models/model.py`.

I don't think either copy is in the wrong place. `models/` importing from `core/agent/` would be a worse layering trade, and the reverse is odd. But it does mean the discriminator we just agreed to keep now lives in two places, so a future change to it is a four-reader problem *and* a two-definition problem — and the two facts you're about to write down are exactly the kind of thing that ends up in one copy and not the other. A line on each pointing at the other would be enough for a future editor to find both.

dparkmit24:
@chuenchen309 : No need to apologise — you were replying to what was on the page.

Agreed on the cross-reference, and the context makes it sharper than it looks: the duplication is deliberate but nothing in the code says so. I'd originally moved the function into `models/` so both readers shared one definition, then reverted it precisely because your review thread anchored to it in `base_agent_runner` and moving a symbol out from under an active review costs more than two lines of duplication. So both copies exist on purpose — and you're right that a reader has no way to know that, which is exactly how one copy ends up drifting.

I'll put the same docstring in both and have each name the other, with a line stating the duplication is intentional and why. That way a future editor finds both, and anyone tempted to unify them has to think about the layering first rather than discovering it from a cycle.

dparkmit24:
@chuenchen309 : Confirmed and pushed — `2c162109`.

Your meta point is the one thing that landed after the push. Good to have, and it makes the picture cleaner: two of the three per-call payloads are structurally unreachable and only `tool_input` can get there at all. The docstring makes that argument from the `observation` side and names the test, so I've left it rather than lengthening it for a third case that reaches the same conclusion.

One thing that turned up while checking, which you wouldn't have seen from the `ToolEngine` chain: there's a second writer into the observation column outside `core/agent/`. `agent_app/app_runner.py` `json.dumps` arbitrary plugin content into it. It doesn't falsify "always strings" for the selector's purposes — only `fc_agent_runner` joins names with `";"`, so `occurrences == 1` on those rows and the length rule is unreachable regardless of shape — but it's why the docstring is scoped to what the selector reads rather than to the column itself. Worth knowing if anyone later reasons about that column from the writer side.

Both docstrings are byte-identical apart from the pointer line naming the other module, and each says the duplication is deliberate and why. The test is at both readers — `test_base_agent_runner.py` and `test_model.py` — a two-call record whose single stored value is a two-element list, asserting call 1 reads element 0 and call 2 reads element 1, named for what it concedes rather than what it asserts.

No behaviour change: the condition, the ordering and the fallback are untouched. +64 lines, 0 deletions.

PR #40016, opened 2026-08-05
chore: mock to real class

asukaminato0721:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

#35872

## Screenshots

| Before | After |
|--------|-------|
| ... | ... |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [ ] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [ ] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods

Issue #35872, opened 2026-05-07
too much `Mock(spec=`


asukaminato0721:
Many of them is outdated.

<img width="577" height="363" alt="Image" src="https://github.com/user-attachments/assets/f23ec90b-4338-4e32-9584-16d87a72353f" />

need to clean them.

@dosu give me a task list to solve it

asukaminato0721:
@dosu based on current unit test changes, how to fix them.

asukaminato0721:
@dosu , no, in unit test, now we have sqlite session, we can use db

asukaminato0721:
@dosu what's current progress.

asukaminato0721:
@dosu count `MagicMock(` `Mock(` and `SimpleNamespace(`

PR #40018, opened 2026-08-05
test: use real ORM models in dataset service tests

asukaminato0721:
## Summary

- Replace ORM `Mock(spec=...)` doubles in dataset services, console dataset controllers, hit testing, and datasource tests with real SQLAlchemy models.
- Keep shared dataset helpers and their consumers together so this PR is independently reviewable.
- Populate mapped fields through constructors without generic `setattr` model factories.

Part 2 of the split of #40016. Related to #35872.

## Validation

- `make lint`
- `./dev/pyrefly-check-local`
- Dataset service tests passed before the split (`200 passed` for the main dataset-service group).

From Codex


41tair (approved):
LGTM

PR #40077, opened 2026-08-06
fix: report unavailable agent tools in publish and workflow checks

iamjoel:
## Summary

- Block Agent v2 publishing when configured tools are uninstalled or unauthorized without calling the publish API
- Use the Tools UI display name in unavailable-tool errors, including marketplace plugin names
- Surface unavailable inline agent tools in the workflow checklist and open their configuration panel from the fix action

Fixes DIFY-2862

## Checklist


lyzno1 (approved):

liejiayong:
In which version update will this bug be fixed?

Issue #39993, opened 2026-08-04
[Refactor/Chore] API Machinery: Migrate API endpoints to layered application services


41tair:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

Migrate Dify API endpoints toward layered application services, using [#39822](https://github.com/langgenius/dify/pull/39822) as the initial reference implementation.
The goal is to make API dependencies, business boundaries, database access, and side effects explicit while preserving existing behavior unless an API contract change is intentionally approved.
This is an umbrella issue for the complete API Machinery migration. Individual controller/resource families can be claimed and implemented through separate PRs.

escape0707:
@41tair  I've revised #39974 and marked it as Ready for Review. Although the CI is failing due to unrelated toolchain downloading process.
Could you please give review and advice?

fszcd:
I'd like to take the Console tag resource family (`controllers/console/tag/tags.py`) as a separate layer of this migration. I could not find an open PR covering this boundary.

Planned scope:
- move tag list/create/update/delete and binding orchestration behind a framework-neutral application service;
- give persistence its own short-lived SQLAlchemy session/transaction boundary instead of passing `db.session` from controllers;
- keep tenant scoping, role checks, snippet RBAC behavior, routes, and response contracts unchanged;
- add focused service/repository/controller regression tests and architecture checks.

I'll keep this as one atomic PR and mark it as part of #39993.

PR #40094, opened 2026-08-06
refactor(api): extract recommended app query service

hjlarry:
> [!IMPORTANT]
>

## Summary

Part of https://github.com/langgenius/dify/issues/39993

## Intentional migration boundaries

Several parts of this implementation are transitional rather than the intended final layering.

| Transitional state | Why it remains in this PR | Follow-up condition |
| --- | --- | --- |
| `LegacyRecommendedAppCatalogGateway` adapts dictionaries and `App` ORM objects | The builtin, remote, and database retrieval implementations currently return different shapes. The remote Learn Dify path can also fall back to the database and return ORM-backed data. Rewriting every provider would significantly expand this controller-focused PR. | Once every provider implements the typed catalog port and returns persistence-neutral records, the compatibility gateway and its mapping code can be deleted. |
| Legacy retrieval methods still accept a SQLAlchemy `Session` | The existing shared protocol requires a session even for builtin and remote implementations that do not use it. It also contains existing cross-provider fallback behavior. | Once providers are migrated independently, only the database adapter should own a session; builtin and remote adapters should no longer depend on SQLAlchemy. |
| Remote retrieval still reads Flask request/config state directly | Origin forwarding and remote fallback behavior currently live inside the legacy retrieval implementation. Moving them here would mix an HTTP-client redesign into this PR. | Once the remote provider receives explicit configuration, request metadata, and an HTTP client, it can implement the catalog port directly. |
| Database retrieval still combines querying, ORM models, response shaping, and DSL export | These responsibilities already exist together and changing them would affect more than the recommended-app query controller contract. This PR only makes that dependency explicit and keeps it outside the application service. | A later database-provider migration can introduce typed read models, keep SQLAlchemy inside a repository, and expose DSL export through an explicit port. |
| Controllers still use `with_current_user` to read `interface_language` | The current `RequestContext` does not carry locale information. Changing admission/context construction would affect many unrelated controllers. The application service therefore receives only the scalar language value and does not receive the ORM `Account`. | Once locale becomes part of the stable request context, or account preferences are exposed through a query port, these controllers can stop depending on `current_user`. |
| `RecommendedAppService` still exists | Its remaining methods belong to different flows: runtime admission (`get_app`), trial usage writes (`add_trial_app_record`), and the deployment-level trial flag. They are not part of the read-path migration in this PR. | Trial writes should first move to a transactional command/repository, followed by a separate runtime-admission refactor. The legacy service can be removed after both paths are migrated. |

These compromises are intentionally contained behind the compatibility gateway and composition root. New query behavior can now be implemented in the application service without adding further dependencies to the legacy stack.

## Follow-up work

1. Move trial usage writes out of `RecommendedAppService`, with repository-owned transactions and an atomic usage increment.
2. Migrate runtime trial admission and `get_app_model_with_trial`.
3. Replace the legacy builtin, remote, and database retrieval implementations with typed catalog adapters.
4. Remove `LegacyRecommendedAppCatalogGateway`, the old retrieval factory, and `RecommendedAppService`.
5. Consolidate the recommended-app service and repository modules once the final feature boundary is established.

## Screenshots

| Before | After |
| ------ | ----- |
| ...    | ...   |

## Checklist

### DIFF
--- a/api/.importlinter
+++ b/api/.importlinter
@@ -125,3 +125,21 @@ forbidden_modules =
     repositories
     sqlalchemy
     werkzeug
+
+[importlinter:contract:recommended-app-query-service-boundary]
+name = Recommended app query application service is framework and persistence neutral
+type = forbidden
+source_modules =
+    services.recommended_app_query_service
+forbidden_modules =
+    configs
+    controllers
+    extensions
+    flask
+    models
+    repositories
+    services.recommend_app
+    services.recommended_app_query_compat
+    services.recommended_app_service
+    sqlalchemy
+    werkzeug
--- a/api/.importlinter
+++ b/api/.importlinter
@@ -58,3 +58,21 @@ forbidden_modules =
     services.workspace_member_role_resolver
     sqlalchemy
     werkzeug
+
+[importlinter:contract:recommended-app-query-service-boundary]
+name = Recommended app query application service is framework and persistence neutral
+type = forbidden
+source_modules =
+    services.recommended_app_query_service
+forbidden_modules =
+    configs
+    controllers
+    extensions
+    flask
+    models
+    repositories
+    services.recommend_app
+    services.recommended_app_query_compat
+    services.recommended_app_service
+    sqlalchemy
+    werkzeug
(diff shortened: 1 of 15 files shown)


hjlarry:
Superseded by #40207 so this refactor can serve as the bottom layer of native GitHub stack #40211. The replacement points to the same commit (`7c61bbed55f36ad36dc70395cafb841faea7fdd1`), and the next refactor layer is #40208. Review and development continue on the new stack.

PR #40137, opened 2026-08-07
feat: support accessibility checks in CI

iamjoel:
## Summary

- Add page-scoped WCAG A and AA Axe scans for the main console pages, with readable reports for detected violations.
- Add a manually triggered GitHub Actions workflow that can select a page and WCAG level without running accessibility scans on every push.
- Add a targeted `lint:a11y` command that reuses the existing `jsx-a11y` rules for selected frontend files or directories.

Implements SNP-632

## Checklist


e2e/features/step-definitions/common/navigation.steps.ts
```diff
@@ -4,6 +4,30 @@ import { expect } from '@playwright/test'
...
 
+type MainPageConfig = {
+  heading?: string
+  path: string
+  readyText?: string
+}
+
+const mainPages: Record<string, MainPageConfig> = {
+  Agents: { path: '/agents', heading: 'Agents' },
+  Home: { path: '/' },
```
lyzno1: Axe analyzes the DOM state that exists at the instant analyze() runs. For Home, this config only waits for the URL and active navigation link. The Home route can still be rendering its Suspense skeleton while HomeContent waits for the prefetched queries, so this scan can pass without inspecting the actual home content. Please make each page configuration wait for a feature-owned content-ready locator before running Axe; Home in particular needs a stable element that only appears after HomeContent resolves.

.github/workflows/accessibility-e2e.yml
```diff
@@ -0,0 +1,105 @@
...
+          - all
+      level:
+        description: WCAG level to scan
+        required: true
+        default: aa
+        type: choice
+        options:
+          - aa
+          - a
+          - both
```
lyzno1: The AA scan already includes every tag used by the A scan. With axe-core 4.12.1, the current A set selects 63 rules and the AA set selects 70 rules, with zero A rules missing from AA. Therefore both starts two full reset/build/runtime jobs and rescans the same A rules without adding coverage. Please remove this option, or produce separate A/AA reporting from one AA run if separate reporting is a product requirement.

e2e/package.json
```diff
@@ -4,6 +4,9 @@
   "type": "module",
   "scripts": {
     "e2e": "tsx ./scripts/run-cucumber.ts",
+    "e2e:accessibility": "pnpm run e2e:accessibility:a",
```
lyzno1: This generic command unexpectedly runs only Level A, while the workflow defaults to AA and the AA configuration is the stronger superset. A developer invoking e2e:accessibility will miss the seven AA-only rules. Please either remove the ambiguous alias or make it run the intended default, preferably AA.

web/scripts/a11y/oxlint.config.ts
```diff
@@ -0,0 +1,23 @@
+import type { OxlintConfig } from 'vite-plus/lint'
+import { lintConfig } from '../../../lint.config.ts'
+
+const webTsxOverride = lintConfig.overrides?.find(
+  (override) =>
```
lyzno1: This does not add an accessibility check that the repository lacks. The root lint configuration already enables the same 32 jsx-a11y rules for web/**/*.tsx, vp check already loads that configuration, and vp check accepts scoped paths. This creates a second raw-Oxlint execution path that introspects the exact shape of the root override, while the new CI workflow does not call it. Unless an a11y-only local diagnostic has a documented consumer that cannot use scoped vp check, please remove lint:a11y and keep the root Vite+ configuration as the single owner.

e2e/features/accessibility/agents/a.feature
```diff
@@ -0,0 +1,9 @@
+@accessibility @axe @wcag-a @wcag-page-agents @authenticated
```
lyzno1: The audit matrix is currently represented by twelve nearly identical feature files: every page and level duplicates the same setup, navigation, assertion, and disclaimer. This is test-runner configuration rather than twelve distinct behavior specifications, and every new page must now update workflow choices, navigation config, tags, folders, and two files. Please consolidate the matrix with a Scenario Outline and tagged Examples, or keep one page scenario at the AA level if Level A does not need a separate report. Cucumber supports tags on Examples blocks, so page and level selection do not require copied feature files.

lyzno1 (commented):
I reviewed this as a Cucumber/Playwright E2E strategy, not only as an Axe API integration. Keeping Axe in a dedicated @axe lane and reusing the repository-owned runner lifecycle, authentication, diagnostics, and artifacts is the right boundary. The default functional suite should not pay for a full-page audit on every scenario.

The current workflow is still an opt-in audit rather than a CI regression gate because it only uses workflow_dispatch, and this head has no WCAG workflow run yet. That can be intentional, but the PR should validate at least the AA/all path once before merge. The selection commands should also include @axe explicitly so that @wcag-* and @wcag-page-* remain qualifiers inside the accessibility suite rather than becoming independent global tags.

The inline comments cover the concrete issues I recommend addressing before relying on this signal.

e2e/features/step-definitions/common/navigation.steps.ts
```diff
@@ -4,6 +4,30 @@ import { expect } from '@playwright/test'
 import { waitForAppsConsole } from '../../../support/apps'
 import { waitForConsoleHome } from '../../../support/home'
 
+type MainPageConfig = {
+  heading?: string
+  path: string
+  readyText?: string
+}
+
+const mainPages: Record<string, MainPageConfig> = {
```
lyzno1: This shared registry is premature in common/. The two generic steps and this page map are consumed only by the new accessibility features, while common/ is reserved for genuinely cross-capability behavior. It also makes this file a second owner of route paths and readiness contracts next to feature helpers such as waitForAppsConsole and waitForConsoleHome. Keep the page matrix under the accessibility capability until another feature genuinely reuses the same behavior, and delegate readiness to feature-owned helpers instead of centralizing feature details here.

lyzno1 (commented):
Additional strict pass against the repository E2E contract and the Cucumber/Playwright authoring guidance:

- The tag taxonomy itself can be sound: @accessibility is the existing capability group, @axe is the opt-in expensive audit lane, and @wcag-* / @wcag-page-* are selectors. The commands need to preserve that hierarchy explicitly.
- A dedicated Axe lane is preferable to injecting a global scan into every functional scenario. Global after-each scans would slow unrelated journeys, scan arbitrary transient states, and blur behavior failures with audit failures.
- Reusing the Cucumber runner for lifecycle, authentication, isolation, diagnostics, and reports is reasonable. That does not require representing the page/level matrix as many copied behavior specifications.
- workflow_dispatch alone creates a manual audit, not a regression gate. If that is the intended product, name and document it as such. If CI enforcement is intended, give it an automatic scheduled or post-merge consumer rather than implying that this PR adds a continuously enforced check.

The additional inline comments focus on ownership, locator/assertion quality, thin step glue, audit completeness, and tag selection semantics.

e2e/features/step-definitions/common/navigation.steps.ts
```diff
@@ -20,6 +48,21 @@ Then('I should stay on the apps console', async function (this: DifyWorld) {
...
+
+  await expect.poll(() => new URL(page.url()).pathname).toBe(config.path)
+  await expect(page.getByRole('link', { name: pageName, exact: true })).toHaveAttribute(
+    'aria-current',
+    'page',
+  )
+  if (config.heading)
+    await expect(page.getByRole('heading', { name: config.heading, exact: true })).toBeVisible()
+  if (config.readyText)
+    await expect(page.getByText(config.readyText, { exact: true }).first()).toBeVisible()
```
lyzno1: Using getByText(...).first() turns an ambiguous text match into an arbitrary readiness signal. That is exactly the locator pattern Playwright recommends avoiding. The Integrations UI already exposes a named region for Model Provider through IntegrationSectionLayout, so this can use getByRole("region", { name: "Model Provider" }) and wait for the actual content owner. The URL assertion above should likewise use expect(page).toHaveURL(...) instead of a custom expect.poll around page.url(), which gives better Playwright diagnostics and follows the web-first assertion model.

e2e/features/step-definitions/accessibility/axe.steps.ts
```diff
@@ -0,0 +1,62 @@
...
+    )
+    .join('\n\n')
+
+const checkCurrentPage = async (world: DifyWorld, level: WcagLevel) => {
+  const results = await new AxeBuilder({ page: world.getPage() })
+    .withTags(wcagTagsByLevel[level])
+    .analyze()
+  const formattedViolations = formatViolations(results.violations)
+
+  if (results.violations.length > 0) {
```
lyzno1, line 29: For a manually consumed accessibility audit, dropping results.incomplete loses the items that Axe could not decide automatically and that require manual review. They should not fail this automated gate, but they should be attached to the Cucumber report with the scanned URL and level. Otherwise a green artifact presents only violations and hides the manual-review queue that Axe intentionally returns.

e2e/features/step-definitions/accessibility/axe.steps.ts
```diff
@@ -0,0 +1,62 @@
...
+        2,
+      ),
+      'application/json',
+    )
+  }
+
+  expect(results.violations, formattedViolations).toEqual([])
+}
+
+Then(
```
lyzno1, line 61: These two step definitions differ only by the level literal, so the glue is encoding a tiny second language instead of using the Cucumber parameter model. Use one Cucumber Expression such as Level {word}, narrow the value to A | AA at the boundary, and call checkCurrentPage once. This keeps the step definition thin and removes another paired edit whenever the wording changes.

e2e/package.json
```diff
@@ -4,6 +4,9 @@
...
+    "e2e:accessibility:a": "tsx ./scripts/run-cucumber.ts --full -- --tags @wcag-a",
```
lyzno1: Selection should encode the suite boundary, for example @axe and @wcag-a (and the same for AA). @wcag-a is a qualifier of the Axe audit, not an independent global runtime tag. The one-page workflow expression and the documented command should also include @axe. This prevents future scenarios that happen to use a WCAG classification tag from being pulled into a full reset accessibility run.

.github/workflows/accessibility-e2e.yml
```diff
@@ -0,0 +1,105 @@
+name: WCAG Accessibility E2E
+run-name: WCAG ${{ inputs.level }} · ${{ inputs.page }}
+
+on:
+  workflow_dispatch:
```
lyzno1, line 5: This trigger makes the job a manual audit only; it will not detect regressions unless somebody remembers to run it, and the PR head currently has no WCAG run. If the requirement is a CI quality signal, add an automatic scheduled or post-merge invocation. If manual execution is intentional, rename the workflow and PR wording to avoid presenting it as an enforced CI check, and document who runs it and when.

web/scripts/a11y/oxlint.config.ts
iamjoel: Thanks — agreed that this does not add missing repository-wide accessibility coverage, and CI is not its intended consumer.

The consumer is a local page-scoped diagnostic:

`pnpm lint:a11y --deps <entry.tsx>`

Unlike `vp check <path>`, this mode resolves the entry file's transitive local imports, including path aliases, re-exports, and dynamic imports, and then runs only the jsx-a11y rules against the resulting JSX/TSX files. A scoped `vp check` checks only the supplied path or directory and also reports unrelated lint, formatting, and type diagnostics.

The rule definitions still come from the root lint configuration, so this does not maintain a second copy of the 32 rules. I agree that the local-only use case needs to be documented. I can add it to `docs/lint.md` and refactor the root config to export the Web jsx-a11y rules explicitly, so this script does not depend on finding a particular override shape.

web/scripts/lint-a11y.mjs
```diff
@@ -0,0 +1,202 @@
...
+    for (const importedFile of importedFiles) {
+      const resolvedModule = ts.resolveModuleName(
+        importedFile.fileName,
+        currentFile,
+        compilerOptions,
+        ts.sys,
+        resolutionCache,
+      ).resolvedModule
+
+      if (!resolvedModule || resolvedModule.isExternalLibraryImport) continue
```
lyzno1, line 157: [P1] This graph is not actually page-scoped. TypeScript marks workspace imports such as `@langgenius/dify-ui/button` as `isExternalLibraryImport`, so this branch discards local source outside `web/`; the `isWithinWebRoot` check would discard it as well. Running the same algorithm from the Home page entry also omits both `web/app/layout.tsx` and `web/app/(commonLayout)/layout.tsx`, because Next App Router composes those files without importing them from the page. In a representative resolution it skipped 54 workspace source files, including Dify UI controls. The result therefore cannot describe the JSX rendered by a page, despite the documented claim, and the ordinary `vp check` already owns these exact `jsx-a11y` rules. Please remove this parallel 202-line resolver/config subsystem, or define a narrower truthful contract and cover its Next route/workspace resolution behavior with tests and an actual automated consumer.

web/app/components/apps/app-list-catalog.tsx
```diff
@@ -68,6 +68,7 @@ function CatalogSkeleton() {
       className={`relative grow content-start ${APP_LIST_GRID_CLASS_NAME}`}
       role="status"
       aria-label={t(($) => $.loading, { ns: 'common' })}
+      data-testid="app-list-loading"
```
lyzno1: [P2] This production-only test hook is redundant: the same element already has `role="status"` and an accessible loading name, which is precisely the Playwright locator contract we should prefer. The same pattern is added to both Knowledge skeletons. Please keep the test coupled to user-observable semantics (or expose `aria-busy` on the owning region if that is the real readiness contract) and remove all three `data-testid` additions. A hidden test-id only says that a marker is absent; it does not itself prove which resolved content/error/empty branch is ready for the Axe snapshot.

.github/workflows/accessibility-e2e.yml
```diff
@@ -0,0 +1,101 @@
+name: Manual WCAG Accessibility Audit
+run-name: WCAG ${{ inputs.level }} · ${{ inputs.page }}
+
+on:
+  workflow_dispatch:
```
lyzno1, line 5: [P1] This is the only workflow that executes the new `@axe` scenarios, but there is no run for the current implementation. All three WCAG runs visible on this branch failed, and the latest one failed the actual Axe assertion; meanwhile the green normal Full-Stack E2E check excludes `@axe` in both Cucumber defaults. Please run the final AA/all command against this head and attach the result before merging. Otherwise the PR can be green while its entire advertised audit path remains unverified.

lyzno1 (commented):
I am not approving this head yet.

The overall Cucumber/Playwright boundary is now mostly sound: keeping Axe in an explicit `@axe` lane, excluding it from the default functional suite, and treating `@wcag-*` / `@wcag-page-*` as qualifiers is preferable to running a global Axe scan after every scenario. The workflow is also now accurately named as a manual audit rather than a regression gate.

There are still three blocking review points:

1. The new dependency-aware static lint command does not model a rendered Next.js page, while duplicating rules already enforced by `vp check`.
2. The readiness changes add test-only production attributes even though the loading elements already expose accessible status semantics.
3. The final Axe path has not been validated on this head. Every WCAG workflow run visible on this branch failed; the latest run is https://github.com/langgenius/dify/actions/runs/31141507287. The current green Full-Stack E2E job is not evidence for this change because both default Cucumber paths explicitly exclude `@axe`. This PR's own E2E contract says the AA/all path should be run when changing the workflow, matrix, or readiness contracts. Please provide a current AA/all execution result before merge.

The inline comments contain the concrete code changes I recommend.

web/app/components/apps/app-list-catalog.tsx
iamjoel: Removed.

lyzno1 (approved):

PR #40145, opened 2026-08-07
fix(web): prevent formatTime from overflowing units array past 60h

zl86790:
## Summary

Found this while looking into how the dataset document metadata panel renders indexing time. `formatTime()` in `web/utils/format.ts` walks a `['sec', 'min', 'h']` array, dividing by 60 and bumping an index each time, until the value drops below 60. Problem is the loop condition (`index < units.length`) lets `index` advance one step past the last valid slot. Once a duration hits 60 hours (216000 seconds) or more, it ends up reading `units[3]`, which is `undefined`, so instead of `"60.00 h"` you get `"1.00 undefined"` printed straight into the UI.

It's not just a theoretical edge case either — `use-metadata.ts` calls `formatTime()` directly to render `indexing_latency` on the document metadata panel, so any dataset document whose indexing takes 60+ hours will show this broken string to users.

Fix is a one-liner: cap the loop at `index < units.length - 1` instead of `index < units.length`, so once we're on the last unit (`h`) we stop advancing the index and just keep dividing the value down. Everything under 60h formats exactly the same as before — `30 -> "30.00 sec"`, `60 -> "1.00 min"`, `3600 -> "1.00 h"`.

Added a test case for the 60h boundary (`216000 -> "60.00 h"`) alongside the existing `formatTime` cases in `format.spec.ts`.

## Screenshots

N/A - logic-only fix, nothing changes visually for the normal (< 60h) path.

## Checklist


anujbolewar (commented):
Correct — the guard now stops the loop before it overflows past the hours unit, so 216000 seconds correctly renders as 60.00 h instead of wrapping to 1.0 h. The added boundary test is good. Worth also covering the value just below the next threshold (for example 215999 becomes 59.99 h) to guard the rounding at the transition, since that is the exact spot floating-point drift tends to show up.

crazywoola:
Hi @zl86790! Thanks so much for taking the time to contribute to Dify — we really appreciate it.

PR #40152, opened 2026-08-07
fix(web): match IPv6 loopback hostname with brackets in isPrivateOrLo…

zl86790:
## Summary

`isPrivateOrLocalAddress()` in `web/utils/urlValidation.ts` is supposed to flag localhost, private IP ranges, and `.local` addresses so the UI can warn users. The IPv6 loopback check never actually works though:

```ts
if (hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '::1') return true
```

`URL.hostname` serializes IPv6 addresses with brackets, so for `http://[::1]/` the hostname is the string `"[::1]"`, not `"::1"`. The comparison never matches, so `isPrivateOrLocalAddress('http://[::1]/')` silently returns `false` — it falls through every other check and ends up treated as a public address.

```js
new URL('http://[::1]:8080/').hostname // "[::1]"
```

Ran it before/after the fix:

```
http://[::1]/       before: false   after: true
http://[::1]:8080/  before: false   after: true
```

Two places use this to warn users about private/local endpoints, and both silently skip the warning for IPv6 loopback:
- `web/app/components/workflow/nodes/trigger-webhook/panel.tsx:199` (webhook debug URL)
- `web/app/components/plugins/plugin-detail-panel/subscription-list/create/hooks/use-common-modal-state.helpers.ts:171` (plugin subscription callback URL)

Fix is just matching the bracketed form instead:

```ts
if (hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '[::1]') return true
```

Kept this scoped to just the loopback comparison. `isPrivateOrLocalAddress` also doesn't cover `fe80::/10` (link-local) or `fd00::/8` (ULA) IPv6 ranges, but that's a separate gap, not something this bug touches — didn't want to bundle it in here.

`isPrivateOrLocalAddress` had zero test coverage before this, so added a small one covering the loopback case this PR fixes.

## Screenshots

N/A — logic fix, no UI change.

## Checklist


anujbolewar (commented):
The IPv6 loopback recognition is right, and the test for the bracket form is welcome given the parser here returns the hostname without braces for ::1. A cleaner approach is to strip and normalize brackets once at the top of this function, then compare against the bare values, which avoids two subtly different comparisons. Also note this only covers IPv6 loopback, not other private v6 ranges such as fe80; worth deciding whether that is in scope or a follow-up.

crazywoola:
Hi @zl86790! Thanks so much for taking the time to contribute to Dify — we really appreciate it.

PR #40172, opened 2026-08-08
fix(api): avoid shared verification-code token payload dict

### LINKED ISSUE #39815: Verification code can leak between concurrent requests via mutable default `additional_data` in AccountService

Souravrajvi0:
## Summary

Fixes #39815.

`AccountService.generate_reset_password_token`, `generate_email_register_token`, and `generate_owner_transfer_token` used a mutable default argument (`additional_data: dict[str, Any] = {}`) and then mutated it in place with `additional_data["code"] = code`.

Under concurrent requests, that shared dict could leak one request's verification code into another request's token payload.

This change:
- replaces the mutable default with `additional_data: dict[str, Any] | None = None`
- builds a fresh payload per call via `token_data = {**(additional_data or {}), "code": code}`
- avoids mutating caller-supplied dicts

## Test plan

- [x] Added unit tests in `api/tests/unit_tests/services/test_account_service.py::TestVerificationTokenGeneration`
- [x] Verified consecutive calls do not share default `additional_data`
- [x] Verified caller-provided `additional_data` is not mutated
- [x] `uv run pytest tests/unit_tests/services/test_account_service.py::TestVerificationTokenGeneration -q`


Souravrajvi0:
Closing — no maintainer engagement after several weeks. May reopen if still relevant.

PR #40216, opened 2026-08-09
refactor(web): simplify the dataset menu item

lyzno1:
## Summary

- replace the React namespace import with a named memo import in DatasetInfo MenuItem
- keep the rendered DOM, click handling, styles, tests, and lint suppressions unchanged

## Contract

This is an import-only cleanup. MenuItem rendering and behavior must remain identical.

## Scope

- Dataset Sidebar MenuItem only
- two equivalent import/export lines

## Non-goals

- no accessibility claim in this layer
- no DOM, event, or visual change
- no suppression cleanup
- no adjacent DatasetInfo components

## Verification

- pnpm exec vp test run app/components/app-sidebar/dataset-info/__tests__/index.spec.tsx (20/20)
- pnpm exec vp fmt --write app/components/app-sidebar/dataset-info/menu-item.tsx
- git diff --check

## Visual regression review

The JSX and class list are unchanged, so no visual change is intended.

## Rollback

Revert this PR independently to restore the React namespace import.

## Stack

This is the dependency-free cleanup layer. The following layer will make MenuItem a native button and remove its two accessibility suppressions.

### DIFF
--- a/web/app/components/app-sidebar/dataset-info/menu-item.tsx
+++ b/web/app/components/app-sidebar/dataset-info/menu-item.tsx
@@ -1,5 +1,5 @@
 import type { RemixiconComponentType } from '@remixicon/react'
-import * as React from 'react'
+import { memo } from 'react'
 
 type MenuItemProps = {
   name: string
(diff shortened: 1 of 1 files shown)


lyzno1:
Closing because this React.memo import rewrite is not a real prerequisite for the dataset menu accessibility repair. The retained fix is #40248 directly on main.

PR #40220, opened 2026-08-09
test(web): rely on the plugin loading status

lyzno1:
## Summary

- remove per-card data-testid attributes from the plugin list skeleton
- stop asserting the internal six-card placeholder count
- query the existing loading status through its common.loading accessible name
- keep skeleton rendering and spacing unchanged

## Contract

While installed plugins are loading, PluginsPanel exposes one named status. The focused tests continue to protect default and compact spacing without depending on the number or markup of internal visual placeholders.

## Scope

- PluginListSkeleton child test hooks
- two loading-spacing assertions in PluginsPanel tests

## Non-goals

- no skeleton count, layout, animation, or visual change
- no PluginsPanel loading policy change
- no suppression change

## Verification

- pnpm exec vp test run app/components/plugins/plugin-page/__tests__/plugins-panel.spec.tsx (34/34)
- pnpm exec vp check app/components/plugins/plugin-page/plugin-list-skeleton.tsx app/components/plugins/plugin-page/__tests__/plugins-panel.spec.tsx
- git diff --check

## Visual regression review

The six-card Array rendering and every class are unchanged. Only non-visual data-testid attributes and implementation-detail assertions are removed.

## Rollback

Revert this PR independently to restore the child test hooks and exact-count assertions.

## Stack

### DIFF
--- a/web/app/components/plugins/plugin-page/__tests__/plugins-panel.spec.tsx
+++ b/web/app/components/plugins/plugin-page/__tests__/plugins-panel.spec.tsx
@@ -312,10 +312,9 @@ describe('PluginsPanel', () => {
 
     render(<PluginsPanel />)
 
-    const loadingState = screen.getByRole('status')
+    const loadingState = screen.getByRole('status', { name: 'common.loading' })
 
     expect(loadingState).toHaveClass('px-12')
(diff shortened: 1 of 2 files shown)


lyzno1:

PR #40242, opened 2026-08-09
refactor(web): simplify the tool category filter component

lyzno1:
## Summary

- replace the legacy `React.FC` declaration with a named function
- import `ReactNode` directly for the option icon type
- keep the tool category filter's public behavior and rendered DOM unchanged

## Scope

Only `web/app/components/base/tab-slider-new/index.tsx`.

This is a mechanical preparation layer. It does not change props, classes, handlers, test ids, or output.

## Verification

- `pnpm exec vp test run app/components/base/tab-slider-new/__tests__/index.spec.tsx` — 1/1
- `pnpm check` — 0 errors, existing warnings only
- `git diff --check`

## Visual regression

None: the generated JSX and class names are unchanged.

## Stack

Base: `main`

Child: #40243 adds the semantic filter behavior. Review and revert this PR independently; if reverting the complete stack, start from the top child.

### DIFF
--- a/web/app/components/base/tab-slider-new/index.tsx
+++ b/web/app/components/base/tab-slider-new/index.tsx
@@ -1,10 +1,10 @@
-import type { FC } from 'react'
+import type { ReactNode } from 'react'
 import { cn } from '@langgenius/dify-ui/cn'
 
 type Option = {
   value: string
   text: string
-  icon?: React.ReactNode
+  icon?: ReactNode
 }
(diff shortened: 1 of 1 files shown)


lyzno1:
Closing because this FC/ReactNode declaration cleanup is not a real prerequisite. #40243 has been rebased onto current main; its two normal Storybook states are pixel-identical to main, with only the semantic SegmentedControl/keyboard contract retained.

PR #40249, opened 2026-08-09
fix(web): make tool provider cards buttons

lyzno1:
## Summary

- Render each tool provider card as a native `button type="button"`.
- Map the existing disabled presentation to the native `disabled` state.
- Preserve the block-card geometry with explicit `w-full text-start`.
- Cover public button semantics, keyboard activation, and disabled behavior in the owner test.
- Prune the exact two obsolete `jsx-a11y` suppressions.

## Behavior contract

- A tool card is reachable by Tab and opens its read-only details with pointer or Enter activation.
- A disabled card is exposed as disabled, is skipped by browser activation, and does not open details.
- Closing the detail view keeps the existing behavior.

## Visual regression review

A temporary Storybook comparison rendered the previous div and the new button under the same production CSS. It was removed before commit.

- Card geometry: 320 x 63 px for both
- Padding: 12 x 16 px for both
- Border: 0.5 px with 12 px radius for both
- Title and description geometry: identical relative coordinates and dimensions
- Text alignment: `start` for both
- Hover background: `rgb(249, 250, 251)` for both
- Disabled state: opacity 0.5 and `cursor: not-allowed` for both

The only intended new visual state is the browser focus indicator during keyboard navigation.

## Validation

- `pnpm exec vp check app/components/tools/provider/tool-item.tsx app/components/tools/provider/__tests__/tool-item.spec.tsx`
- `node scripts/lint-a11y.mjs app/components/tools/provider/tool-item.tsx`
- `pnpm exec vp test run app/components/tools/provider/__tests__/tool-item.spec.tsx` (3/3)
- `pnpm check` (0 errors; 2059 existing warnings)

From Codex

### DIFF
--- a/oxlint-suppressions.json
+++ b/oxlint-suppressions.json
@@ -3616,14 +3616,6 @@
       "count": 1
     }
   },
-  "web/app/components/tools/provider/tool-item.tsx": {
-    "jsx_a11y/click-events-have-key-events": {
-      "count": 1
-    },
-    "jsx_a11y/no-static-element-interactions": {
-      "count": 1
-    }
-  },
   "web/app/components/tools/setting/build-in/config-credentials.tsx": {
     "typescript/no-explicit-any": {
       "count": 3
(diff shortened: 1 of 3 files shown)


lyzno1:
Closing as superseded by #40362. The retained implementation has native disabled behavior, valid button content, focus-visible styling, semantic tests, and preserves RTL text-start.

PR #40262, opened 2026-08-09
test(web): query plugin item actions by name

lyzno1:
## Summary

- Replace every remaining single-action positional locator in the plugin item action tests with a helper that queries the intended action by role and accessible name.
- Remove the generic getActionButtons helper so future tests cannot silently bind behavior to toolbar order.
- Keep absence checks on queryAllByRole because those assertions verify action visibility rather than one named action.

## Behavior contract

- Delete scenarios explicitly select Delete.
- Plugin information scenarios explicitly select Plugin info.
- Update scenarios explicitly select Check for updates.
- This PR changes no production behavior and depends on #40261, which introduces the accessible names used by these locators.

## Visual regression

This is a test-only PR. It has no production bundle, DOM, CSS, layout, interaction, or runtime visual effect.

## Validation

- pnpm exec vp check app/components/plugins/plugin-item/__tests__/action.spec.tsx (0 errors, 0 warnings)
- pnpm exec vp test run app/components/plugins/plugin-item/__tests__/action.spec.tsx (32/32)
- git diff codex/a11y-plugin-item-actions --stat (1 test file; 25 insertions, 27 deletions)

From Codex

### DIFF
--- a/web/app/components/plugins/plugin-item/__tests__/action.spec.tsx
+++ b/web/app/components/plugins/plugin-item/__tests__/action.spec.tsx
@@ -140,8 +140,6 @@ const getDeleteButton = () => screen.getByRole('button', { name: 'plugin.action.
 
 // ==================== Tests ====================
 
-// Helper to find action buttons (real ActionButton component uses type="button")
-const getActionButtons = () => screen.getAllByRole('button')
 const queryActionButtons = () => screen.queryAllByRole('button')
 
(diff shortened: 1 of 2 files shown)


lyzno1:
Closing after folding the semantic role/name interaction assertions into #40261. Production behavior and its public-interface test now live in the same reviewable layer.

PR #40265, opened 2026-08-09
fix(web): name plugin credential actions

lyzno1:
## Summary

- Give the plugin credential Rename, Edit, and Delete icon buttons the accessible names already shown in their tooltips.
- Hide each CSS icon from the accessibility tree because its button now owns the action semantics.
- Make the owner tests activate those actions by role and name instead of button order or icon class.

## Behavior contract

- Rename enters the existing inline rename flow.
- Edit passes the same credential values to the existing edit callback.
- Delete passes the same credential id to the existing delete callback and still stops row-click propagation.
- Permission, enterprise, borrowed-credential, disabled, and OAuth visibility rules remain unchanged.

## Scope boundary

This PR intentionally does not change the clickable credential row or the inline rename input. The row keyboard model and the Dify UI field/form migration are separate higher-risk owner changes.

## Visual regression

No visual change is expected. This PR does not change elements, classes, styles, layout, dimensions, hover visibility, node order, event handlers, or focus styling; it only adds ARIA attributes.

## Validation

- pnpm exec vp check app/components/plugins/plugin-auth/authorized/__tests__/item.spec.tsx (0 errors, 0 warnings)
- pnpm exec vp test run app/components/plugins/plugin-auth/authorized/__tests__/item.spec.tsx (37/37)
- Focused production lint exposes 3 pre-existing owner errors: deprecated Input import plus the clickable credential row missing role and keyboard handling.
- node scripts/lint-a11y.mjs confirms only the same 2 pre-existing clickable-row accessibility errors.

From Codex

### DIFF
--- a/web/app/components/plugins/plugin-auth/authorized/__tests__/item.spec.tsx
+++ b/web/app/components/plugins/plugin-auth/authorized/__tests__/item.spec.tsx
@@ -191,8 +191,7 @@ describe('Item Component', () => {
       )
 
       const enterRenameMode = () => {
-        const firstButton = result.container.querySelectorAll('button')[0] as HTMLElement
-        fireEvent.click(firstButton)
+        fireEvent.click(screen.getByRole('button', { name: 'common.operation.rename' }))
       }
 
       return { ...result, onRename, enterRenameMode }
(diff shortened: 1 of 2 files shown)


lyzno1:
Closing to avoid a partial accessibility fix. The named actions remain hidden with display:none until pointer hover, while the clickable row and inline rename flow still lack a complete keyboard/Form model. This needs an owner-level interaction redesign, not labels alone.

PR #40266, opened 2026-08-09
test(web): query core credential actions by name

lyzno1:
## Summary

- Query the core plugin credential Delete, Rename, and Edit flows by role and accessible name instead of SVG class, button order, or generic icon traversal.
- Remove conditional branches that allowed those tests to pass without finding or activating the intended action.
- Keep this PR test-only and scoped to the primary confirmation, rename, edit modal, close, and remove flows.

## Behavior contract

- Delete tests must activate Delete and observe the confirmation, cancellation, mutation, notification, and update callback paths.
- Rename tests must activate Rename and observe the textbox, save mutation, notification, and update callback paths.
- Edit tests must activate Edit and observe the modal open, close/reset, and remove-to-confirmation paths.
- This PR depends on #40265, which supplies the accessible names used by these locators.

## Visual regression

This is a test-only PR. It has no production bundle, DOM, CSS, layout, interaction, or runtime visual effect.

## Validation

- pnpm exec vp check app/components/plugins/plugin-auth/authorized/__tests__/index.spec.tsx (0 errors, 0 warnings)
- pnpm exec vp test run app/components/plugins/plugin-auth/authorized/__tests__/index.spec.tsx (62/62)
- pnpm check (0 errors, 2059 existing warnings)
- git diff codex/a11y-plugin-credential-actions --stat (1 test file; 118 insertions, 338 deletions)

From Codex

PR #40268, opened 2026-08-09
test(web): query credential coverage actions by name

lyzno1:
## Summary

- Replace the remaining concurrent and additional-coverage credential action entry locators with role-and-name queries.
- Remove hidden-button traversal, SVG inspection, and conditional fallbacks that allowed coverage tests to skip the intended action.
- Delete one outside-click test that was proven to be a no-op because its overlay selector never matched the rendered dialog.

## Behavior contract

- Delete coverage explicitly selects Delete for concurrency, confirmation reset, Escape close, and credential-id flows.
- Rename concurrency coverage explicitly selects Rename before driving the existing textbox and Save behavior.
- Edit/remove and modal-reset coverage explicitly selects Edit before using the existing textual modal actions.
- This PR depends on #40266 because both layers intentionally simplify the same large owner spec in reviewable stages; #40266 in turn depends on the production names from #40265.

## Visual regression

This is a test-only PR. It has no production bundle, DOM, CSS, layout, interaction, or runtime visual effect.

## Validation

- pnpm exec vp check app/components/plugins/plugin-auth/authorized/__tests__/index.spec.tsx (0 errors, 0 warnings)
- pnpm exec vp test run app/components/plugins/plugin-auth/authorized/__tests__/index.spec.tsx (61/61)
- No remaining Rename, Edit, or Delete SVG-class locators or hidden-action-button traversal in this spec.
- git diff codex/a11y-plugin-credential-action-tests --stat (1 test file; 222 insertions, 685 deletions)

From Codex

### DIFF
--- a/web/app/components/plugins/plugin-auth/authorized/__tests__/index.spec.tsx
+++ b/web/app/components/plugins/plugin-auth/authorized/__tests__/index.spec.tsx
@@ -897,38 +897,13 @@ describe('Authorized Component', () => {
       // Verify API Keys section is shown
       expect(screen.getByText('API Keys'))!.toBeInTheDocument()
 
-      // Find edit button - look for buttons in the action area
-      const actionAreaButtons = Array.from(
-        document.querySelectorAll('.group-hover\\:flex button, .hidden button'),
-      )
-
-      for (const btn of actionAreaButtons) {
-        const svg = btn.querySelector('svg')
-        if (
-          svg &&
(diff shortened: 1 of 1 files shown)


lyzno1:
Closing during the test audit. This continues a split cleanup of the same oversized spec without adding a distinct behavior contract, and still leaves substantial duplicated/private-implementation coverage.

PR #40271, opened 2026-08-09
refactor(web): remove annotation control React.FC

lyzno1:
## Summary

- Replace React.FC with an explicit Props annotation on the annotation reply control component.
- Replace the React namespace import with a named memo import.
- Preserve the existing memoized default export and all render behavior.

## Behavior contract

- Prop requirements remain exactly the same.
- Memoization remains exactly the same.
- This PR contains no accessibility, service, state, or test behavior change and depends on #40270 only because both PRs intentionally modify the same component in review order.

## Visual regression

No visual change is possible from this refactor. The JSX, DOM, classes, styles, layout, icons, ARIA attributes, and event handlers are unchanged relative to #40270.

## Validation

- pnpm exec vp check app/components/base/features/new-feature-panel/annotation-reply/annotation-ctrl-button.tsx (0 errors, 2 existing icon warnings)
- pnpm exec vp test run app/components/base/features/new-feature-panel/annotation-reply/__tests__/annotation-ctrl-button.spec.tsx (7/7)
- git diff codex/a11y-annotation-actions --stat (1 production file; 4 insertions, 5 deletions)

From Codex

### DIFF
--- a/web/app/components/base/features/new-feature-panel/annotation-reply/annotation-ctrl-button.tsx
+++ b/web/app/components/base/features/new-feature-panel/annotation-reply/annotation-ctrl-button.tsx
@@ -1,9 +1,8 @@
 'use client'
-import type { FC } from 'react'
 import { toast } from '@langgenius/dify-ui/toast'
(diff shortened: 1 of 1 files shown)


lyzno1:

PR #40277, opened 2026-08-09
refactor(workflow): adapt to Graphon engine architecture

laipz8200:
## Summary

- Upgrade the API integration to the refactored Graphon engine architecture.
- Migrate engine, runtime state, events, layers, commands, and container ownership handling to the new APIs.
- Preserve Dify workflow, Chatflow, pipeline, pause/resume, and nested container behavior across the migration.
- Preserve Workflow Tool tracing across worker handoffs with worker-scoped context activation and explicit container span parentage.
- Restore node ordering and expandable execution details when console runs resume after Human Input.
- Add internal Workflow Tool tracing to Test Run, run history, and Logs. Correlate child nodes with their exact parent execution so repeated and nested tool calls remain separate, and refresh the details across pause/resume.

## Dependency

The API pins the official [Graphon 0.8.0 release](https://github.com/langgenius/graphon/releases/tag/v0.8.0) from PyPI, which includes the engine and runtime changes required by this migration. The dify-agent directory remains unchanged from main and keeps Graphon 0.5.2.

## Session

[Implementation session](https://chatgpt.com/s/cx_6a9e97687d848191aacd04f580c3eb07)


api/controllers/console/app/workflow_run.py
```diff
@@ -311,6 +313,35 @@ def get(self, request_context: RequestContext, app_model: App, run_id: UUID):
...
+        "Internal node executions retrieved successfully",
+        console_ns.models[WorkflowToolNodeExecutionListResponse.__name__],
+    )
+    @console_ns.response(404, "Workflow run not found")
+    @console_account_admission(
+        rbac_checks=[RBACCheck(RBACPermission.APP_CREATE_AND_MANAGEMENT, PlainApp())],
+    )
+    @get_app_model(mode=[AppMode.ADVANCED_CHAT, AppMode.WORKFLOW])
+    def get(self, request_context: RequestContext, app_model: App, run_id: UUID, node_execution_id: str):
+        executions = application_services().workflow_runs.get_workflow_tool_node_executions(
```
41tair, line 334: We only checks permissions for the root application A, but this endpoint also returns internal node traces from the invoked Workflow Tool’s source application B. A user who can manage A but lacks access to B can therefore invoke B as a shared workflow tool and read its internal inputs, process_data, and outputs through this endpoint, bypassing the permission check on B’s own execution-details endpoint.

api/core/workflow/workflow_tool_node.py
```diff
@@ -0,0 +1,156 @@
...
+        self,
+        *,
+        result: ContainerRunResult,
+    ) -> Generator[NodeEventPayload | ContainerAwaitRequest, None, None]:
+        if not isinstance(result, ContainerExecutionResult):
+            raise TypeError(f"Unsupported Workflow Tool container result {type(result).__name__}")
+
+        for _ in range(result.steps):
+            self.runtime_state.increment_node_run_steps()
+        node_run_result = result.node_run_result.to_node_run_result()
```
41tair, line 113: to_node_run_result() converts output segments to plain Python values, so an empty files output becomes [] and enters the variable pool as ArrayAnyVariable.

api/core/tools/workflow_as_tool/repository.py
```diff
@@ -0,0 +1,28 @@
...
+@dataclass(frozen=True, slots=True)
+class WorkflowToolSource:
+    """Database-independent definition of a pinned Workflow Tool source."""
+
+    app_id: str
+    workflow_id: str
+    graph_config: Mapping[str, Any]
+    features_dict: Mapping[str, Any]
+    environment_variables: Sequence[VariableBase]
+    workflow_kind: str
```
QuantumGhost, line 17: Maybe we should use the `WorkflowKind` enum here?

api/core/app/apps/common/workflow_response_converter.py
```diff
@@ -143,6 +144,7 @@ def __init__(
             self._truncator = VariableTruncator.default()
 
         self._node_snapshots: dict[NodeExecutionId, _NodeSnapshot] = {}
+        self._node_sequence = 0
```
QuantumGhost, line 147: What's the difference between `event.node_run_index` and `_node_sequence`?

api/core/app/apps/workflow_app_runner.py
```diff
@@ -288,20 +295,25 @@ def _get_graph_and_variable_pool_for_single_node_run(
...
 
         # filter nodes only in the specified node type (iteration or loop)
-        main_node_config = next((n for n in graph_config.get("nodes", []) if n.get("id") == node_id), None)
-        start_node_id = main_node_config.get("data", {}).get("start_node_id") if main_node_config else None
+        all_node_configs = graph_config.get("nodes", [])
+        nodes_by_id = {node["id"]: node for node in all_node_configs}
+        container_ids = {node["id"]: resolve_container_id(node, nodes_by_id=nodes_by_id) for node in all_node_configs}
+        node_ids = {node_id}
+        while (
+            child_node_ids := {node.get("id") for node in all_node_configs if container_ids[node["id"]] in node_ids}
```
QuantumGhost, line 303: This code is too smart and obscure. Maybe we should extract it into a dedicated function?

or refactor it to something like this:

```
# Include all nested nodes needed to execute the selected container.
node_ids = {node_id}

while True:
    # Find new children recursively.
    new_node_ids = {
        child_id
        for child_id, container_id in container_ids.items()
        if container_id in node_ids
    } - node_ids

    if not new_node_ids:
        break

    node_ids.update(new_node_ids)
```

api/core/app/apps/workflow_app_runner.py
```diff
@@ -288,20 +295,25 @@ def _get_graph_and_variable_pool_for_single_node_run(
...
+            node_ids.update(child_node_ids)
+        # Preserve ownership resolved with the complete ancestry before removing it
+        # from the debug subtree. Nested legacy IDs can otherwise become ambiguous.
         node_configs = [
-            node
-            for node in graph_config.get("nodes", [])
-            if node.get("id") == node_id
-            or node.get("data", {}).get(node_type_filter_key, "") == node_id
-            or (start_node_id and node.get("id") == start_node_id)
+            {**node, "data": {**node.get("data", {}), "container_id": container_ids[node["id"]]}}
```
QuantumGhost, line 310: This part is obscure too.

Please expand and simplify it, or extract it to a dedicated function.

api/core/workflow/workflow_entry.py
```diff
@@ -146,51 +164,71 @@ def __init__(
...
 
-    def run(self) -> Generator[GraphEngineEvent, None, None]:
+    def run(self, *, pause_state_layer: PauseStatePersistenceLayer | None = None) -> Generator[EngineEvent, None, None]:
         graph_engine = self.graph_engine
 
         try:
             # Preserve Dify's response-stream semantics on top of Graphon 0.5.0.
             generator = iter_dify_graph_engine_events(graph_engine, self._response_stream_filter)
-            yield from generator
+            paused_event = None
```
QuantumGhost, line 218: The `paused_event` is both captured by the `pause_state_layer` AND the `run` function. It seems that there is no clear responsibility for pause event capturing.

I recommend only capturing `GraphRunPausedEvent` inside `pause_state_layer`, and invoking the `persist_pending_pause` method here.

The `pause_state_layer` also seems to break the layer abstraction. However, I would suggest keeping the current implementation as-is.

api/core/app/apps/workflow_app_runner.py
```diff
@@ -415,13 +431,149 @@ def _build_agent_strategy_info(event: NodeRunStartedEvent) -> AgentStrategyInfo
...
+
+        engine = workflow_entry.graph_engine
+        variable_pool = engine.runtime_state.variable_pool
+        run_context = resolve_dify_run_context(engine.graph.root_node.run_context)
+        run_id = get_system_text(variable_pool, SystemVariableKey.WORKFLOW_EXECUTION_ID)
+        repository = HumanInputFormSubmissionRepository()
+        for form_id, reason in pending_forms.items():
+            if form_id in published_form_ids:
+                continue
+            form = repository.get_by_form_id(form_id)
```
QuantumGhost: Maybe we should introduce a `get_by_form_ids` method to batch the form retrieval process?

api/core/app/apps/workflow_app_runner.py
```diff
@@ -656,9 +788,9 @@ def _handle_event(self, workflow_entry: WorkflowEntry, event: GraphEngineEvent):
                         node_type=event.node_type,
                         node_title=event.node_title,
                         start_at=event.start_at,
-                        node_run_index=workflow_entry.graph_engine.graph_runtime_state.node_run_steps,
+                        node_run_index=workflow_entry.graph_engine.runtime_state.node_run_steps,
                         inputs=event.inputs,
-                        metadata=event.metadata,
+                        metadata={**event.metadata, **node_metadata},
```
QuantumGhost: What's the purpose of overwriting event.metadata with node_metadata here?

api/core/app/apps/workflow_app_runner.py
```diff
@@ -681,10 +813,10 @@ def _handle_event(self, workflow_entry: WorkflowEntry, event: GraphEngineEvent):
                         node_type=event.node_type,
                         node_title=event.node_title,
                         start_at=event.start_at,
-                        node_run_index=workflow_entry.graph_engine.graph_runtime_state.node_run_steps,
+                        node_run_index=workflow_entry.graph_engine.runtime_state.node_run_steps,
                         inputs=event.inputs,
                         outputs=event.outputs,
-                        metadata=event.metadata,
+                        metadata={**event.metadata, **node_metadata},
```
QuantumGhost: ditto

api/core/app/apps/workflow_app_runner.py
```diff
@@ -697,9 +829,9 @@ def _handle_event(self, workflow_entry: WorkflowEntry, event: GraphEngineEvent):
                         node_type=event.node_type,
                         node_title=event.node_title,
                         start_at=event.start_at,
-                        node_run_index=workflow_entry.graph_engine.graph_runtime_state.node_run_steps,
+                        node_run_index=workflow_entry.graph_engine.runtime_state.node_run_steps,
                         inputs=event.inputs,
-                        metadata=event.metadata,
+                        metadata={**event.metadata, **node_metadata},
```
QuantumGhost: ditto

api/core/app/apps/workflow_app_runner.py
```diff
@@ -722,10 +854,10 @@ def _handle_event(self, workflow_entry: WorkflowEntry, event: GraphEngineEvent):
                         node_type=event.node_type,
                         node_title=event.node_title,
                         start_at=event.start_at,
-                        node_run_index=workflow_entry.graph_engine.graph_runtime_state.node_run_steps,
+                        node_run_index=workflow_entry.graph_engine.runtime_state.node_run_steps,
                         inputs=event.inputs,
                         outputs=event.outputs,
-                        metadata=event.metadata,
+                        metadata={**event.metadata, **node_metadata},
```
QuantumGhost: ditto

api/core/app/layers/execution_context_layer.py
```diff
@@ -0,0 +1,30 @@
...
+        self._context = capture_current_context()
+
+    @contextmanager
+    def enter_context(self) -> Generator[None, None, None]:
+        """Create a fresh manager for a node worker or child-frame construction."""
+        file_runtime = peek_workflow_file_runtime()
+        # Host context restoration can include an older file binding. Keep the
+        # engine's adapter during the body and restore it after host teardown.
+        with use_workflow_file_runtime(file_runtime), self._context:
+            with use_workflow_file_runtime(file_runtime):
```
QuantumGhost: Why is the `with use_workflow_file_runtime(file_runtime):` statement repeated?

api/core/app/layers/pause_state_persist_layer.py
```diff
@@ -99,35 +99,27 @@ def __init__(
...
     def on_graph_start(self) -> None:
-        """
-        Called when graph execution starts.
-
-        This is called after the engine has been initialized but before any nodes
-        are executed. Layers can use this to set up resources or log start information.
-        """
-        pass
+        """Clear any pause captured by a prior run."""
+        self._paused_event = None
```
QuantumGhost: This is not necessary since we do not reuse PauseStatePersistenceLayer.

api/core/app/workflow/layers/observability.py
```diff
@@ -86,45 +81,39 @@ def _get_parser(self, node: Node) -> NodeOTelParser:
     @override
     def on_graph_start(self) -> None:
         """Called when graph execution starts."""
-        self._node_contexts.clear()
+        self.on_graph_end(None)
```
QuantumGhost: This is obscure. Why do we invoke  `self.on_graph_end` inside the `on_graph_start` handler?

If the purpose is to end existing spans and clear the `_node_span` mapping, why not extract it into a dedicated function?

api/core/app/workflow/layers/observability.py
```diff
@@ -52,7 +47,7 @@ class ObservabilityLayer(GraphEngineLayer):
 
     def __init__(self) -> None:
         super().__init__()
-        self._node_contexts: dict[str, _NodeSpanContext] = {}
```
QuantumGhost, line 55: If we only keep spans inside the `ObservabilityLayer`, How should we ensure those span survive the workflow suspension / resumption? AFAIK, states in Layers are not persisted as part of the `Engine` state.

If supporting tracing association after resumption is not currently supported, I would suggest add a short docstring of the current limitation here.

api/core/app/workflow/layers/observability.py
```diff
@@ -86,45 +81,39 @@ def _get_parser(self, node: Node) -> NodeOTelParser:
...
-            )
-
-            new_context = set_span_in_context(span)
-            token = context_api.attach(new_context)
-
-            self._node_contexts[execution_id] = _NodeSpanContext(span=span, token=token)
-
+            if not self._is_disabled and self._tracer and (execution_id := node.execution_id):
+                span = self._node_spans.get(execution_id)
+                if span is None:
```
QuantumGhost, line 101: This code could be simplified with early-returning.

api/core/app/workflow/layers/persistence.py
```diff
@@ -68,20 +70,8 @@ class PersistenceWorkflowInfo:
...
-    title: str
-    predecessor_node_id: str | None
-    iteration_id: str | None
-    loop_id: str | None
-    created_at: datetime
-
-
-class WorkflowPersistenceLayer(GraphEngineLayer):
-    """GraphEngine layer that persists workflow and node execution state."""
+class WorkflowPersistenceLayer(Layer):
```
QuantumGhost: This part is obscure. I don't understand why we need nested `WorkflowPersistenceLayer`. Could you elaborate this?

hepeng92:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

QuantumGhost:
A question to the data modeling: AFAIK, the node execution associate with their parent WorkflowTool nodes' execution via the `process_data["workflow_tool_parent_execution_id"]` field stored in a Text Field.

I see the difficulty to add a column to the `workflow_node_executions` table in large deployment. However, the [current query](https://github.com/langgenius/dify/pull/40277/changes#diff-68acef937c30cb3bb74b313727470f3246fbdc1387f36b1525b6b28869df973cR168) seems far from ideal and requires a cast to retrieve the corresponding parent id.

Maybe we should introduce another table to associate node executions in subgraphs to their parents? This approach avoids the complexity of migrating `workflow_node_executions` table and 
provide a way to fast retrieve node executions generated by a WorkflowTool execution.

It seems that some knowledge of how those special fields are stored has already leaded into other parts of the codebase. For example, the truncation needs to aware of those special fields and preserve them while truncating.

api/core/app/workflow/layers/persistence.py
```diff
@@ -464,6 +487,9 @@ def _update_node_execution(
         self._workflow_node_execution_repository.save_execution_data(domain_execution)
 
     def _fail_running_node_executions(self, *, error_message: str) -> None:
+        for tool_layer in self._workflow_tool_layers.values():
```
QuantumGhost: This would issue many SQL update queries for deeply nested workflow invocations.

api/core/ops/unified_trace/hierarchy.py
```diff
@@ -63,20 +72,55 @@ def execution_metadata(execution: object) -> Mapping[str, Any]:
...
+        parent_id = process_data.get(WORKFLOW_TOOL_PARENT_EXECUTION_ID_KEY)
+        if isinstance(parent_id, str) and parent_id in execution_ids:
+            parents[execution_id(item)] = execution_ids[parent_id]
+    _remove_cycles(parents)
+    return parents
+
+
+def _unique_execution_by_node_id(executions: Sequence[WorkflowExecutionLike]) -> dict[tuple[str, str, str], str]:
+    result: dict[tuple[str, str, str], str] = {}
+    repeated_node_keys: set[tuple[str, str, str]] = set()
```
QuantumGhost: A tuple of three or more elements hurts readability.

Consider use `namedtuple` or frozen `dataclass`

api/core/repositories/sqlalchemy_workflow_node_execution_repository.py
```diff
@@ -506,18 +522,29 @@ def get_db_models_by_workflow_run(
...
+        if include_workflow_tools and not self._app_id:
+            raise ValueError("app_id is required to include Workflow Tool executions")
         with self._session_factory() as session:
+            trigger = triggered_from or self._triggered_from or WorkflowNodeExecutionTriggeredFrom.WORKFLOW_RUN
+            owner_filter: ColumnElement[bool] = WorkflowNodeExecutionModel.triggered_from == trigger
+            if self._app_id:
+                owner_filter &= WorkflowNodeExecutionModel.app_id == self._app_id
+            if include_workflow_tools:
+                assert self._app_id is not None
+                owner_filter = or_(
```
QuantumGhost: This part of the query is rather complex and may not fully utilize the indexes. Given the mass volume of 
`WorkflowNodeExecutionModel`, maybe we should find a better way to query `workflow_tools` executions?

However, I haven't done any benchmark yet. So my judgement does not have data support.

api/controllers/console/app/workflow_run.py
laipz8200: It looks like we need to make a design decision: should we allow users to view the execution details of called workflow tools for convenience, or restrict access to ensure more robust permission management?

laipz8200: Perhaps we can keep the ability to view workflow tool tracing, but check permissions again when drilling down into the tracing details.

api/core/app/apps/workflow_app_runner.py
laipz8200: This isn't overwriting; it's a merge. Event metadata only contains information about that specific node running, while node metadata contains things only the engine knows about, such as information related to the outer container.

laipz8200:
> I see the difficulty to add a column to the workflow_node_executions table in large deployment. However, the [current query](https://github.com/langgenius/dify/pull/40277/changes#diff-68acef937c30cb3bb74b313727470f3246fbdc1387f36b1525b6b28869df973cR168) seems far from ideal and requires a cast to retrieve the corresponding parent id.
> 
> Maybe we should introduce another table to associate node executions in subgraphs to their parents? This approach avoids the complexity of migrating workflow_node_executions table and
provide a way to fast retrieve node executions generated by a WorkflowTool execution.

@QuantumGhost We've already put some work into reducing the row count of the log table, so adding new columns is worth considering now.

Besides, the way I see it, running WorkflowTool is part of the workflow execution, and I don't want to add a separate table just for that.

PR #40279, opened 2026-08-09
refactor(web): use css icon for generated result copy

lyzno1:
## Summary

- Replace the generated-result copy action Remix React SVG with the lint-recommended `i-ri-clipboard-line` CSS icon.
- Remove the now-unused `@remixicon/react` import.
- Preserve the accessible name and decorative icon contract introduced by #40259.

## Dependency

This PR is stacked on #40259 because both layers intentionally modify the same copy action. The parent owns accessibility semantics; this child owns only icon implementation.

## Behavior contract

- Copy still invokes the same clipboard and success-toast behavior.
- The button accessible name remains Copy.
- The icon remains decorative and exposes no separate accessibility node.

## Visual regression

This PR intentionally changes only the icon rendering implementation. The button element, padding, dimensions, color class, placement, handler, and surrounding layout are unchanged. The CSS icon uses the exact `i-ri-clipboard-line` mapping required by the repository lint rule and is already used by other 16px copy actions in the codebase.

Reviewer focus: compare the 16x16 clipboard silhouette, optical centering, color, button padding, and alignment with Apply in light and dark themes. No layout shift or ordinary-state difference is acceptable.

## Validation

- `pnpm exec vp check app/components/app/configuration/config/automatic/result.tsx app/components/app/configuration/config/automatic/__tests__/result.spec.tsx` (0 warnings, 0 errors; parent had 1 icon warning)
- `pnpm lint:a11y app/components/app/configuration/config/automatic/result.tsx` (0 diagnostics)
- `pnpm exec vp test run app/components/app/configuration/config/automatic/__tests__/result.spec.tsx` (3/3)
- `pnpm check` (0 errors, 2058 warnings; exactly one fewer than #40259)
- `git diff codex/a11y-generated-result-copy --stat` (1 production file; 1 insertion, 2 deletions)

From Codex

PR #40281, opened 2026-08-09
refactor(web): remove generated result React.FC

lyzno1:
## Summary

- Replace `React.FC` with an explicit `Props` annotation on the generated result component.
- Replace the React namespace import with the named `memo` import.
- Preserve the memoized export and every render path.

## Dependency

This PR is stacked on #40279 only because all three PRs intentionally touch the same component in review order. It has no accessibility or icon behavior dependency.

## Behavior contract

The component accepts the same props and remains memoized. There are no state, service, event-handler, or test changes.

## Visual regression

No visual change is expected: JSX, DOM, classes, styles, layout, icons, ARIA attributes, and handlers are unchanged relative to #40279.

Reviewer focus: confirm the diff is limited to the React import, explicit props annotation, and memo export.

## Validation

- Focused `vp check`: 0 warnings, 0 errors
- Focused Vitest: 3/3 tests passed
- Full `pnpm check`: 0 errors, 2058 existing warnings
- Diff relative to #40279: 1 production file, 4 insertions, 5 deletions

From Codex

### DIFF
--- a/web/app/components/app/configuration/config/automatic/result.tsx
+++ b/web/app/components/app/configuration/config/automatic/result.tsx
@@ -1,9 +1,8 @@
 'use client'
-import type { FC } from 'react'
 import type { GenRes } from '@/service/debug'
 import { Button } from '@langgenius/dify-ui/button'
 import copy from 'copy-to-clipboard'
-import * as React from 'react'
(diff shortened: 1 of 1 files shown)


lyzno1:

PR #40305, opened 2026-08-09
test(web): query external API card actions by name

lyzno1:
## Summary

- Replace every remaining External Knowledge API card `button[0]` and `button[1]` lookup with Edit/Delete role-and-name queries.
- Keep the existing edit, delete, confirmation, query invalidation, error, and permission assertions unchanged.
- Remove DOM-order coupling from the behavior suite.

## Dependency

Depends only on #40304, which introduces the accessible names consumed by these tests.

## Test contract

This test-only layer makes the public semantic contract the locator for 11 interaction paths. It does not change production code, add test ids, or alter service mocks and expected payloads.

## Visual regression

Not applicable: this layer changes one test file only. Production JSX, styles, assets, and runtime behavior are identical to #40304.

## Validation

- Focused Vitest: 17/17 tests passed
- Focused code check: 0 errors and 3 existing production icon warnings
- Full `pnpm check`: 0 errors and 2059 existing warnings
- `git diff --check`: passed
- Scope against parent: 1 test file, 22 insertions and 25 deletions

## Rollback

Revert `1f6f8d1978`; #40304's production semantics remain intact.

From Codex

### DIFF
--- a/web/app/components/datasets/external-api/external-knowledge-api-card/__tests__/index.spec.tsx
+++ b/web/app/components/datasets/external-api/external-knowledge-api-card/__tests__/index.spec.tsx
@@ -128,9 +128,8 @@ describe('ExternalKnowledgeAPICard', () => {
       }
       vi.mocked(fetchExternalAPI).mockResolvedValue(mockResponse)
 
-      const { container } = render(<ExternalKnowledgeAPICard {...defaultProps} />)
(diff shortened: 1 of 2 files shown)


lyzno1:
Closing after folding the semantic edit/delete interaction assertions into #40304. Production behavior and its public-interface test now live in the same reviewable layer.

PR #40306, opened 2026-08-09
refactor(web): use css icons for external API card

lyzno1:
## Summary

- Replace the API connection, Edit, and Delete SVG components with the equivalent project CSS icons.
- Preserve decorative semantics for all three glyphs.
- Delete the test that asserted only that some SVG existed.

## Dependency

Depends only on #40305. This is an optional visual-implementation cleanup above the completed semantic and test contracts.

## Test boundary

The removed test inspected an implementation detail and could not verify icon identity, alignment, color, or behavior. The remaining 16 tests cover the card's public rendering, permissions, edit flow, delete flow, confirmations, invalidation, and errors; visual equivalence is reviewed explicitly below.

## Visual regression

This layer is expected to be pixel-equivalent: the three CSS classes map to the same project icon names and retain the same `size-4` and text-color classes.

Reviewer checklist:

- Compare manageable and read-only cards in light and dark themes.
- Check API connection, Edit, and Delete glyph shape, 16px size, alignment, color, and surrounding spacing.
- Check Edit/Delete default, hover, focus-visible, and destructive hover states.
- Confirm card height, border, background, name, endpoint, and confirmation flow are unchanged.

## Validation

- Focused Vitest: 16/16 behavior tests passed
- Focused accessibility lint: 0 diagnostics
- Focused code check: 0 errors and 0 warnings
- Full `pnpm check`: 0 errors and 2056 warnings, exactly 3 fewer warnings than the parent
- `git diff --check`: passed
- Scope against parent: 1 production file and 1 test file, 8 insertions and 13 deletions

## Rollback

Revert `3092b775b8` to restore only the icon implementation and SVG-trivia test.

From Codex

PR #40307, opened 2026-08-09
refactor(web): remove external API card React.FC

lyzno1:
## Summary

- Replace the External Knowledge API card's `React.FC` annotation with an explicit props parameter type.
- Remove the now-unused React namespace import.

## Dependency

Depends only on #40306 because it is a separate optional cleanup on the same component. It does not change the semantic or icon contracts.

## Behavior contract

Runtime output, props, callbacks, state, data fetching, mutations, dialogs, JSX, classes, icons, and conditional rendering are identical. Only TypeScript component typing changes.

## Visual regression

No JSX, classes, styles, assets, or runtime branches change. A visual difference is not possible from this type-only refactor; the card should render exactly as #40306.

## Validation

- Focused Vitest: 16/16 behavior tests passed
- Focused code check: 0 errors and 0 warnings
- Full `pnpm check`: 0 errors and 2056 warnings, unchanged from the parent
- `git diff --check`: passed
- Scope against parent: 1 production file, 2 insertions and 3 deletions

## Rollback

Revert `4e4b65efe1` to restore only the component typing style.

From Codex

### DIFF
--- a/web/app/components/datasets/external-api/external-knowledge-api-card/index.tsx
+++ b/web/app/components/datasets/external-api/external-knowledge-api-card/index.tsx
@@ -10,7 +10,6 @@ import {
   AlertDialogTitle,
 } from '@langgenius/dify-ui/alert-dialog'
 import { useQueryClient } from '@tanstack/react-query'
-import * as React from 'react'
 import { useState } from 'react'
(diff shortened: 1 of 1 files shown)


lyzno1:

PR #40310, opened 2026-08-09
refactor(web): use css icons for external API panel

lyzno1:
## Summary

- replace the External API panel documentation, close, and add RemixIcon components with equivalent CSS icon spans
- keep all glyphs decorative while preserving existing size and color classes
- remove exactly three prefer-tailwind-icons warnings

## Dependency

This PR is stacked on #40309 because it preserves the accessible close-action contract introduced there. It contains only icon implementation changes.

## Visual regression review

Please compare before and after in light and dark themes:
- documentation, close, and add glyph shapes
- 12px documentation icon and 16px close/add icon dimensions
- accent, tertiary, and primary-button icon colors
- inline alignment, gaps, and button box geometry
- close action hover, active, and focus-visible states
- panel header and create-button layout

No text, event, accessibility name, or component geometry is intentionally changed.

## Validation

- focused Vitest: 14/14 passed
- full pnpm check: 0 errors, 2056 existing warnings
- warning delta from parent: exactly -3
- git diff --check: passed

## Rollback

Revert this PR alone to restore the component icons. The semantic fix in #40309 remains intact.

### DIFF
--- a/web/app/components/datasets/external-api/external-api-panel/index.tsx
+++ b/web/app/components/datasets/external-api/external-api-panel/index.tsx
@@ -1,6 +1,5 @@
 import { Button } from '@langgenius/dify-ui/button'
 import { cn } from '@langgenius/dify-ui/cn'
-import { RiAddLine, RiBookOpenLine, RiCloseLine } from '@remixicon/react'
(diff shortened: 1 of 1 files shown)


lyzno1:

PR #40311, opened 2026-08-09
refactor(web): remove external API panel React.FC

lyzno1:
## Summary

- type ExternalAPIPanel props directly on the function parameter
- remove the unused React namespace import and React.FC annotation
- preserve the component API, JSX, and runtime behavior

## Dependency

This cleanup is stacked on #40310 only because it edits the same component after the icon conversion. It has no accessibility or runtime dependency and remains a separate review layer.

## Visual regression review

This is a type-only refactor. The emitted JSX, classes, visible content, dimensions, interaction states, and accessibility tree are unchanged. A visual difference would be a regression.

## Validation

- focused Vitest: 14/14 passed
- full pnpm check: 0 errors, 2056 existing warnings
- git diff --check: passed

## Rollback

Revert this PR alone to restore React.FC without affecting either lower stack layer.


lyzno1:

PR #40318, opened 2026-08-09
refactor(web): use css icon for external retrieval close

lyzno1:
## Summary

- replace the external retrieval close RemixIcon component with the equivalent CSS icon span
- keep the glyph decorative and preserve its 16px size and shrink behavior
- remove exactly one prefer-tailwind-icons warning

## Dependency

This PR is stacked on #40317 because it preserves the accessible close-action contract introduced there. It contains only the icon implementation change.

## Visual regression review

Please compare before and after in light and dark themes:
- close glyph shape and 16px dimensions
- header alignment and spacing
- ActionButton box geometry
- hover, active, and focus-visible states
- retrieval settings and footer layout

No text, event, accessible name, retrieval value, or component geometry is intentionally changed.

## Validation

- focused Vitest: 10/10 passed
- warning delta from parent: exactly -1
- git diff --check: passed

## Rollback

Revert this PR alone to restore the component icon. The semantic fix in #40317 remains intact.

### DIFF
--- a/web/app/components/datasets/hit-testing/modify-external-retrieval-modal.tsx
+++ b/web/app/components/datasets/hit-testing/modify-external-retrieval-modal.tsx
@@ -1,5 +1,4 @@
 import { Button } from '@langgenius/dify-ui/button'
-import { RiCloseLine } from '@remixicon/react'
 import { useState } from 'react'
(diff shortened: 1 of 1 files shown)


lyzno1:

PR #40322, opened 2026-08-09
refactor(web): remove human input preview React.FC

lyzno1:
## Summary

- type FormContentPreview props directly on the function parameter
- remove the FC type import and annotation
- preserve the component API, JSX, preview behavior, and runtime output

## Dependency

This cleanup is stacked on #40321 only because it edits the same component after the semantic fix. It has no accessibility or runtime dependency and remains a separate review layer.

## Visual regression review

This is a type-only refactor. The emitted JSX, classes, visible content, panel geometry, interaction states, preview behavior, and accessibility tree are unchanged. A visual difference would be a regression.

## Validation

- focused Vitest: 3/3 passed
- full pnpm check: 0 errors, 2059 existing warnings
- git diff --check: passed

## Rollback

Revert this PR alone to restore React.FC without affecting the semantic fix in #40321.

### DIFF
--- a/web/app/components/workflow/nodes/human-input/components/form-content-preview.tsx
+++ b/web/app/components/workflow/nodes/human-input/components/form-content-preview.tsx
@@ -1,5 +1,4 @@
 'use client'
-import type { FC } from 'react'
 import type { FormInputItem, UserAction } from '../types'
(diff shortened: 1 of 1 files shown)


lyzno1:

PR #40325, opened 2026-08-09
refactor(web): use css icon for code generator

lyzno1:
## Summary

- replace the code generator SVG component with the equivalent CSS icon span
- keep the glyph decorative and preserve its 16px size and primary color
- remove exactly one prefer-tailwind-icons warning

## Dependency

This PR is stacked on #40324 because it preserves the named generator action and its behavior test. It contains only the icon implementation change.

## Visual regression review

Please compare before and after:
- generator glyph shape and 16px dimensions
- primary color in light and dark themes
- inline alignment and ActionButton box geometry
- hover, active, and focus-visible states
- surrounding code editor layout

No text, event, accessible name, modal state, or component geometry is intentionally changed.

## Validation

- focused Vitest: 1/1 passed
- warning delta from parent: exactly -1
- git diff --check: passed

## Rollback

Revert this PR alone to restore the SVG component. The semantic fix and behavior test in #40324 remain intact.


lyzno1:

PR #40330, opened 2026-08-09
refactor(web): remove prompt generator React.FC

lyzno1:
## Summary

- type PromptGeneratorBtn props directly on the function parameter
- remove the FC type import and annotation
- preserve the component API, JSX, state transition, and runtime behavior

## Dependency

This cleanup is stacked on #40329 only because it edits the same component after the icon conversion. It has no accessibility or runtime dependency and remains a separate review layer.

## Visual regression review

This is a type-only refactor. The emitted JSX, classes, visible content, interaction states, modal behavior, and accessibility tree are unchanged. A visual difference would be a regression.

## Validation

- focused Vitest: 1/1 passed
- git diff --check: passed

## Rollback

Revert this PR alone to restore React.FC without affecting either lower stack layer.


lyzno1:

PR #40338, opened 2026-08-09
refactor(web): use css icon for assigner add

lyzno1:
## Summary

- replace the single decorative Remix add glyph with its matching CSS icon
- preserve the exact 16px size, shrink behavior, and tertiary text color
- remove the now-unused icon import

## Boundary

This is an owner-local visual-resource cleanup only. The accessible name remains owned by #40337, and the React component typing cleanup remains in the next PR.

## Validation

- focused lint — the one icon-component warning is removed
- `pnpm --dir web exec vp test run app/components/workflow/nodes/assigner/__tests__/panel.spec.tsx` — 1 passed
- `pnpm --dir web lint:a11y app/components/workflow/nodes/assigner/panel.tsx` — passed
- `pnpm check` — 0 errors, 2058 warnings (exactly one fewer)
- `git diff --check` — passed

## Visual regression review

Compare the add glyph shape, 16px bounds, tertiary color, centering, shrink behavior, button box, and hover/active/light/dark states. No visual difference is intended; any mismatch is a regression.

## Stack

Depends on #40337. Second PR in stack #40340.

### DIFF
--- a/web/app/components/workflow/nodes/assigner/panel.tsx
+++ b/web/app/components/workflow/nodes/assigner/panel.tsx
@@ -1,7 +1,6 @@
 import type { FC } from 'react'
 import type { AssignerNodeType } from './types'
 import type { NodePanelProps } from '@/app/components/workflow/types'
-import { RiAddLine } from '@remixicon/react'
 import * as React from 'react'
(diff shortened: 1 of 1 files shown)


lyzno1:

PR #40342, opened 2026-08-09
refactor(web): use css icons for chat settings

lyzno1:
## Summary

- replace the two decorative SVG components in the chat-with-history settings popover with matching CSS icons
- preserve the trigger glyph at 18px and the panel glyph at 24px
- remove the now-unused icon imports

## Boundary

This PR contains only visual-resource substitutions. The accessible name and state contract remain owned by #40341; the embedded chatbot already uses the matching CSS trigger icon.

## Validation

- focused lint — both icon-component warnings removed
- the two focused popover suites — 4 passed
- standalone a11y lint on both production components — passed
- `pnpm check` — 0 errors, 2057 warnings (exactly two fewer)
- `git diff --check` — passed

## Visual regression review

Compare the chat-settings glyph and filled message glyph shape, 18px/24px bounds, current-color rendering, alignment, shrink behavior, button box, panel spacing, hover/open states, and light/dark themes. No visual difference is intended; any mismatch is a regression.

## Stack

Depends on #40341. Final PR in stack #40343.


lyzno1:

PR #40345, opened 2026-08-09
refactor(web): use css icons for chat history header

lyzno1:
## Summary

- replace the header's expand-sidebar, new-chat, and reset-chat SVG components with matching CSS icons
- preserve all three glyphs at 18px
- remove the now-unused Remix icon import

## Boundary

This PR changes only the header's decorative icon resources. The five accessible names and behavior-focused test selectors remain owned by #40344; sidebar icons were already CSS icons.

## Validation

- focused lint — all three header icon warnings removed
- header and sidebar focused suites — 78 passed
- standalone a11y lint on both production components — passed
- `pnpm check` — 0 errors, 2056 warnings (exactly three fewer)
- `git diff --check` — passed

## Visual regression review

Compare all three glyph shapes and directions, 18px bounds, current-color rendering, centering, disabled appearance, tooltips, sidebar transition, hover/active states, and light/dark themes. No visual difference is intended; any mismatch is a regression.

## Stack

Depends on #40344. Final PR in stack #40346.


lyzno1:

PR #40347, opened 2026-08-09
fix(web): use button for human input content toggle

lyzno1:
## Summary

- replace the mouse-only Human Input content toggle `div` with a native button
- expose the localized expand/collapse name and `aria-expanded` state
- support pointer, Enter, and Space through native button behavior
- migrate the owner test away from `data-testid` and icon-class selectors
- prune the two matching jsx-a11y suppressions

## Boundary

This PR owns only the collapsible wrapper control. It does not change Human Input fields, validation, submission, child content, initial-state ownership, or form layout. The wrapper-level test id used by other owner tests is retained; only the obsolete toggle test id is removed.

## Test audit

The existing tests checked icon classes and pointer clicks. They now assert role, localized name, expanded state, content visibility, and both keyboard activation keys. JSX children are rendered as actual children, removing three pre-existing test warnings while preventing the new keyboard case from adding another.

## Validation

- semantic tests changed first: 3 of 4 failed against the mouse-only baseline
- focused owner suite — 4 passed
- standalone a11y lint — 0 diagnostics
- focused lint — 0 warnings
- `pnpm check` — 0 errors, 2056 warnings (three fewer than main)
- `git diff --check` — passed
- suppression delta — exactly one `click-events-have-key-events` and one `no-static-element-interactions` entry removed

## Visual regression review

The former wrapper and new button both occupy a 16×16 box. The button explicitly uses `appearance-none`, zero padding, centered flex alignment, and the same 16px CSS arrow glyphs. Verify collapsed/expanded arrows, header gap and vertical alignment, panel geometry, light/dark colors, and pointer behavior. No normal-state visual difference is intended; the keyboard-only focus ring is the sole expected delta.

## Rollback

Revert `90bb3962fc`; native semantics, semantic tests, warning cleanup, and suppressions roll back together.

From Codex


lyzno1:
Closing as superseded by #40202. The retained implementation gives each same-screen toggle a unique accessible name including nodeTitle and validates the consumer contract, while removing the same suppressions.

PR #40350, opened 2026-08-09
refactor(web): use css icons for tool detail

lyzno1:
## Summary

- replace the Tool Detail hammer, loader, collapsed arrow, and expanded arrow SVG components with matching CSS icons
- preserve the 14px status glyphs, 16px arrows, margins, auto alignment, and loader spin
- remove the now-unused Remix icon import

## Boundary

This PR changes only decorative icon resources. Native disclosure semantics, valid button content, state assertions, and suppression pruning remain owned by #40349.

## Validation

- focused lint — all four icon warnings removed
- focused owner suite — 4 passed
- standalone a11y lint — 0 diagnostics
- `pnpm check` — 0 errors, 2055 warnings (exactly four fewer)
- `git diff --check` — passed

## Visual regression review

Compare hammer and loader shapes, 14px bounds, loader rotation, 16px arrow shapes/directions, margins, right alignment when expanded, current-color rendering, header geometry, and light/dark states. No visual difference is intended; any mismatch is a regression.

## Stack

Depends on #40349. Final PR in stack #40351.

### DIFF
--- a/web/app/components/base/chat/chat/answer/tool-detail.tsx
+++ b/web/app/components/base/chat/chat/answer/tool-detail.tsx
@@ -1,6 +1,5 @@
 import type { ToolInfoInThought } from '../type'
 import { cn } from '@langgenius/dify-ui/cn'
-import { RiArrowDownSLine, RiArrowRightSLine, RiHammerFill, RiLoader2Line } from '@remixicon/react'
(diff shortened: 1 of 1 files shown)


lyzno1:

PR #40366, opened 2026-08-09
fix(web): hide citation measurement text

lyzno1:
## Summary

- hide opacity-zero citation measurement clones from the accessibility tree
- keep their text and refs intact for width measurement
- retain `data-testid="citation-measurement-item"` only as the geometry seam used by the layout test

## Boundary

- citation measurement clones only
- no rendered citation, overflow behavior, CSS class, size, position, or measurement algorithm changed
- no suppression entry changed in this layer

## Validation

- `pnpm --dir web exec vp test run app/components/base/chat/chat/citation/__tests__/index.spec.tsx` — 21/21 passed
- `./node_modules/.bin/vp lint web/app/components/base/chat/chat/citation/index.tsx web/app/components/base/chat/chat/citation/__tests__/index.spec.tsx`
- `pnpm check` — 0 errors, 2059 warnings
- `git diff --check`

## Visual regression review

- verify citation chips and overflow count render exactly as before
- measurement clones retain the same text, ref, dimensions, opacity, and absolute positioning
- `aria-hidden` has no CSS or layout effect

## Stack

- root PR
- #40367 adds the citation overflow button semantics in a separate interaction layer
- this layer can be reviewed and reverted independently

## Rollback

Revert commit `8216b1deb2d9791f58451143e2478b99e8d894e5`.

From Codex

### DIFF
--- a/web/app/components/base/chat/chat/citation/__tests__/index.spec.tsx
+++ b/web/app/components/base/chat/chat/citation/__tests__/index.spec.tsx
@@ -99,11 +99,13 @@ describe('Citation', () => {
       expect(screen.getAllByTestId('citation-measurement-item')).toHaveLength(2)
     })
 
-    it('should display the document name inside each measurement item', () => {
+    it('should hide measurement text from the accessibility tree', () => {
       mockClientWidths({ container: 500, item: 50 })
       setupContainer()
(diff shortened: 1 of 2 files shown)


lyzno1:
Closing as a duplicate of #40203, which already owns the same citation measurement accessibility-tree contract with the focused regression coverage.

Issue #40372, opened 2026-08-09
rewrite property to accept session: Session as parameter


asukaminato0721:
example: #40370 

@dosu show me the file list to clean and what to do.

a-yeyang:
I'd like to work on this as a follow-up to #40370. Opening a PR that converts the remaining `@property` accessors on `models/web.py:SavedMessage` and `models/trigger.py:WorkflowTriggerLog` that internally call `db.session...` into plain methods taking `session: Session` explicitly, matching the pattern from #40370:

- `SavedMessage.message`
- `WorkflowTriggerLog.created_by_account`
- `WorkflowTriggerLog.created_by_end_user`

Confirmed (via repo-wide grep) there are no other call sites of these three accessors outside their own model files and tests, so this is a self-contained, non-breaking change. Will link the PR here once it's up.

Note: there are other `@property`-with-`db.session` accessors elsewhere in the codebase (e.g. `models/comment.py`, `models/dataset.py`, `models/workflow.py` — some already partially migrated by #40370/#40379/#40453) that are out of scope for this PR; happy to pick up further batches in follow-ups if useful.

ddddajie:
Hi, I'd like to work on the api/models/account.py part of this issue.

I'll refactor Account.get_by_openid to accept a caller-provided Session instead of accessing db.session internally, and update all related call sites and tests.

I'll keep the scope focused on account.py to avoid overlapping with the ongoing web.py / trigger.py work.

ddddajie:
Update: I found that #40453 already covers the api/models/account.py scope, so I won't duplicate that work.

I'd like to take the api/models/comment.py part instead, refactoring the remaining account-related properties that access db.session internally to accept an explicit Session, and updating the related call sites and tests.

I'll keep the PR scoped to comment.py and its required callers.

morning-verlu:
I'd like to work on the `api/models/tools.py` part of this issue. I'll refactor the 6 `@property` accessors that internally call `db.session` (ApiToolProvider.user, ApiToolProvider.tenant, WorkflowToolProvider.user, WorkflowToolProvider.tenant, WorkflowToolProvider.app, MCPToolProvider.load_user) into plain methods taking `session: Session` explicitly, matching the pattern from #40370. Will link the PR here once it's up.

dvd233:
Hi, I'd like to take a narrow, non-overlapping slice in `api/models/dataset.py`: `DocumentSegment.previous_segment` and `DocumentSegment.next_segment`.

[...]

PR #40379, opened 2026-08-09
refactor(models): pass session into CustomizedSnippet accessors

anneheartrecord:
## Summary

Part of #40372 — this one takes `api/models/snippet.py`. It does not touch `api/models/model.py`, so it doesn't overlap #40370.

`CustomizedSnippet` reached for the Flask-global `db.session` inside five properties:
`graph_dict`, `tags`, `created_by_account`, `author_name`, `updated_by_account`. Those are now
`get_*(*, session: Session)` methods, matching the accessors already extracted in `models/dataset.py`
and `models/workflow.py`.

Since the response models validate by attribute, the session is bound at the request boundary with a
`SnippetResponseSource` adapter in `fields/snippet_fields.py` — same shape as `_WorkflowResponseSource`
in `controllers/console/app/workflow.py` and `_SessionResponseSource` in `fields/conversation_fields.py`.
All four snippet response sites in `controllers/console/workspace/snippets.py` go through it.

`input_fields_list` and `version_str` are pure and stay properties.

Two things worth calling out for review:

- The PATCH handler now serializes **inside** the session block, and its `except ValueError` is narrowed
  to the update itself. `pydantic.ValidationError` subclasses `ValueError`, so leaving `dump_response`
  inside that `except` would report an already-committed write back to the client as a 400. There's a
  regression test for this.
- `tests/unit_tests/models/test_snippet.py` no longer needs to monkeypatch `db.session` to inject the
  SQLite session — that hack was the symptom this refactor removes. I also dropped the
  `@pytest.mark.parametrize("sqlite_session", [...], indirect=True)` decorators there, which the
  `sqlite_session` fixture docstring asks to remove on review ("Legacy indirect model parameters remain
  accepted by pytest but are ignored").

No user-visible behavior change: response payload shapes and values are identical.

## Screenshots

Not applicable — backend refactor with no API surface change.

## Checklist

- [x] This change requires a documentation update, included: no docs change needed (no API surface change)

`ruff format`/`ruff check`, `lint-imports`, `lint_response_contracts.py` and `pyrefly` are clean;
829 tests pass across `tests/unit_tests/{models,fields,controllers/console/workspace,controllers/console/snippets}`
and `test_snippet_service.py`. Frontend untouched.

From Claude Code

### DIFF
--- a/api/controllers/console/workspace/snippets.py
+++ b/api/controllers/console/workspace/snippets.py
@@ -36,7 +36,12 @@ from controllers.console.wraps import (
 from core.plugin.entities.plugin import PluginDependency
 from extensions.ext_database import db
 from fields.base import ResponseModel
-from fields.snippet_fields import SnippetListItemResponse, SnippetPaginationResponse, SnippetResponse
+from fields.snippet_fields import (
+    SnippetListItemResponse,
+    SnippetPaginationResponse,
+    SnippetResponse,
+    SnippetResponseSource,
+)
 from libs.helper import dump_response
 from libs.login import login_required
 from models import Account
@@ -116,10 +121,11 @@ class CustomizedSnippetsApi(Resource):
         """List customized snippets with pagination and search."""
         query = _snippet_list_query_from_request()
 
+        session = db.session()
         snippet_service = _snippet_service()
         snippets, total, has_more = snippet_service.get_snippets(
             tenant_id=current_tenant_id,
-            session=db.session(),
+            session=session,
             page=query.page,
             limit=query.limit,
             keyword=query.keyword,
@@ -131,7 +137,7 @@ class CustomizedSnippetsApi(Resource):
         return dump_response(
             SnippetPaginationResponse,
             {
-                "data": snippets,
+                "data": [SnippetResponseSource(snippet, session=session) for snippet in snippets],
                 "page": query.page,
                 "limit": query.limit,
                 "total": total,
@@ -177,7 +183,7 @@ class CustomizedSnippetsApi(Resource):
         except ValueError as e:
             return {"message": str(e)}, 400
 
-        return dump_response(SnippetResponse, snippet), 201
+        return dump_response(SnippetResponse, SnippetResponseSource(snippet, session=db.session())), 201
 
 
 @console_ns.route("/workspaces/current/customized-snippets/<uuid:snippet_id>")
@@ -200,7 +206,7 @@ class CustomizedSnippetDetailApi(Resource):
         if not snippet:
             raise NotFound("Snippet not found")
 
-        return dump_response(SnippetResponse, snippet), 200
+        return dump_response(SnippetResponse, SnippetResponseSource(snippet, session=db.session())), 200
 
     @console_ns.doc("update_customized_snippet")
     @console_ns.expect(console_ns.models.get(UpdateSnippetPayload.__name__))
@@ -236,9 +242,9 @@ class CustomizedSnippetDetailApi(Resource):
         if not update_data:
             return {"message": "No valid fields to update"}, 400
 
-        try:
-            with Session(db.engine, expire_on_commit=False) as session:
-                snippet = session.merge(snippet)
+        with Session(db.engine, expire_on_commit=False) as session:
+            snippet = session.merge(snippet)
+            try:
                 snippet = SnippetService.update_snippet(
                     session=session,
                     snippet=snippet,
@@ -246,10 +252,10 @@ class CustomizedSnippetDetailApi(Resource):
                     data=update_data,
                 )
                 session.commit()
-        except ValueError as e:
-            return {"message": str(e)}, 400
+            except ValueError as e:
+                return {"message": str(e)}, 400
 
-        return dump_response(SnippetResponse, snippet), 200
+            return dump_response(SnippetResponse, SnippetResponseSource(snippet, session=session)), 200
 
     @console_ns.doc("delete_customized_snippet")
     @console_ns.response(204, "Snippet deleted successfully")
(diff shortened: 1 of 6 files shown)


api/controllers/console/workspace/snippets.py
```diff
@@ -200,7 +206,7 @@ def get(self, current_tenant_id: str, snippet_id: UUID):
         if not snippet:
             raise NotFound("Snippet not found")
 
-        return dump_response(SnippetResponse, snippet), 200
+        return dump_response(SnippetResponse, SnippetResponseSource(snippet, session=db.session())), 200
```
asukaminato0721: you may try with_session to inject

api/fields/snippet_fields.py
```diff
@@ -72,3 +76,39 @@ class SnippetPaginationResponse(ResponseModel):
...
+    methods, so the session is bound here at the request boundary instead of inside
+    the model.
+    """
+
+    def __init__(self, snippet: CustomizedSnippet, *, session: Session) -> None:
+        self._snippet = snippet
+        self._session = session
+
+    def __getattr__(self, name: str) -> object:
+        return getattr(self._snippet, name)  # guard-ignore: no-new-getattr -- delegates model fields
```
asukaminato0721: no reflection plz

api/tests/unit_tests/controllers/console/workspace/test_snippets.py
```diff
@@ -306,6 +326,42 @@ def __init__(self, engine, *args, **kwargs):
...
+def test_patch_snippet_does_not_report_a_committed_write_as_a_bad_request(app: Flask, monkeypatch: pytest.MonkeyPatch):
+    """Only the update itself maps ValueError to 400.
+
+    ``ValidationError`` subclasses ``ValueError``, so serializing the committed row must not be
+    covered by that ``except`` — otherwise a persisted update is reported back as a 400.
+    """
+    user = _account("account-1")
+    snippet = _snippet()
+    updated_snippet = _snippet(get_graph_dict="not-a-dict")
+    session = SimpleNamespace(merge=Mock(return_value=snippet), commit=Mock())
```
asukaminato0721: read conftest

asukaminato0721 (changes requested):

anneheartrecord:
All three done in 26f4e4ba.

**`with_session`** — list/create/detail/update now take the injected session; no more `db.session()` in this file. Update also dropped its own `Session(db.engine, expire_on_commit=False)`; `session_factory` is already `expire_on_commit=False`, so the semantics match and it picks up the decorator's rollback.

One behavior change worth flagging: update's `except ValueError` now raises `BadRequest` instead of returning `{"message": ...}, 400`. `with_session` commits on a normal return, so returning there would persist whatever `update_snippet` wrote before it rejected the payload. Status and `message` are unchanged, the body gains the usual `code`/`status` envelope. There's a regression test for it.

**No reflection** — `SnippetResponseSource` is gone, replaced by explicit `snippet_response()` / `snippet_list_item_response()` builders, same `x_response(obj, *, session)` shape as `document_fields.py` but without the `__getattr__`. I checked the output is byte-identical to the old adapter for both response models, `validation_alias`es included.

**conftest** — those four handlers' tests now use `sqlite_session` and real `CustomizedSnippet` rows, so the `get_*(session=...)` accessors run real queries instead of stubs. The rejected-update test runs through the decorator rather than `unwrap`, since the commit being tested lives in the decorator.

Left `delete` / `export` / `imports` / `check-dependencies` / `use-count` on their own `Session(db.engine)` with their existing fakes — they don't touch the accessors this PR changes, so it looked like a separate PR. Happy to fold them in here if you'd rather.

anneheartrecord:
Rebased onto current main — the branch had drifted into a conflict.

The only real collision was with #41012, which rewrote `test_snippet_list_fields_include_author_name` to monkeypatch `snippet_module.db.session`. That symbol is exactly what this PR removes from `models/snippet.py`, so I pointed the test at `snippet_list_item_response(snippet, session=sqlite_session)` and left the assertion alone. The export-history changes from that commit are untouched.

Verified on the new base: 35 passed in the three snippet suites, 146 passed across the adjacent snippet/tag/DSL suites, ruff check + format clean, and `check_no_new_controller_sqlalchemy.py --base-rev main` back to exit 0 (moving `session.merge` out of the `try` block had made it a net-new hunk).

No behaviour changes since the last push — your three asks are still addressed the same way. Happy to take another pass whenever you have time.

anneheartrecord:
@asukaminato0721 one thing worth a quick decision before this sits any longer: #41531 (opened Aug 31) converts the same five `CustomizedSnippet` accessors in `snippet.py`. Its CI went red because `snippet_fields.py` reads those accessors through `from_attributes`, so once they take a `session` the responses have to be built explicitly. The plan its author sketched there on Sep 2 — session-aware `get_*` methods, a `SnippetResponseSource`-style wrapper in `snippet_fields.py`, wrapping the `dump_response` call sites in the controller, then dropping the properties — is what this PR has contained since Aug 9. The one difference: after your "no reflection" note the wrapper became explicit `snippet_response()` / `snippet_list_item_response()` builders instead of a `__getattr__` source.

So the two PRs can't stack, and I'd rather neither of us keeps waiting on the other. Two options:

- Re-review this one. It's still mergeable against today's main (rebased Aug 25), your three items from the first review are addressed in 26f4e4ba, and locally the three snippet suites plus `check_no_new_getattr`, `check_no_new_controller_sqlalchemy` and `lint_response_contracts` are green. Fork CI on the current head is waiting behind the approval gate, so a click there would also give you the real signal.
- Or, if you'd rather take the ResponseSource shape from #41531 after all, say so and I'll close this one instead of having two PRs fight over the file.

Mainly I don't want the other author to spend time re-deriving a diff that already exists.

asukaminato0721 (approved):

PR #40395, opened 2026-08-10
refactor(api): replace legacy recommended app retrieval stack

hjlarry:
## Summary

Part of #39993.

This is PR 4 of the recommended-app controller stack. Its Git base is #40257.

- Replace the legacy `services/recommend_app` factory and retrieval hierarchy plus `recommended_app_query_compat` with a typed catalog query port, built-in and remote gateways, a source router, and a database repository wired at the application-services composition root.
- Make source behavior explicit: regular remote list failures fall back to bundled data, empty recommendation pages fall back to bundled en-US, and Learn Dify remote failures fall back to bundled Learn Dify data. Built-in mode no longer reads Learn Dify entries from the database.
- Preserve the legacy remote-detail contract: every HTTP non-200 response is an authoritative catalog miss, while transport and JSON failures fall back to bundled data and decoded schema errors remain visible.
- Preserve the remote template fetch cache introduced by #40176: it remains process-local, uses the configured TTL and a URL-plus-Origin key, caches only successfully decoded 200 responses, and can be disabled with a non-positive TTL.
- Preserve database category ordering and its existing Redis key, but close the database Session before reading Redis. Learn Dify does not read category ordering because its response omits page categories.
- Make database catalog membership a lightweight existence query so preview admission does not export a DSL before loading the admitted app.
- Rename the read-only admission decorator to `get_previewable_app_model`. Preview accepts either TrialApp registration or catalog membership, while execution remains protected by the existing feature, registration, and quota checks.
- Delete the replaced compatibility, factory, retrieval, and legacy test modules, and migrate their behavioral coverage to the new adapters and repository.

## Validation

- 173 focused controller, composition, repository, gateway, and query-service tests passed.
- Ruff and Pyrefly passed for the changed implementation and tests.


api/services/recommended_app_catalog_gateway.py
```diff
@@ -0,0 +1,373 @@
...
+import json
+import logging
+import threading
+from collections.abc import Callable, Mapping, Sequence
+from pathlib import Path
+from typing import cast, override
+
+import httpx
+from cachetools import TTLCache
+from flask import current_app, has_request_context, request
```
41tair: If could we shouldn't import anything from flask in service layer.

api/services/recommended_app_catalog_gateway.py
```diff
@@ -0,0 +1,373 @@
...
+    @override
+    def list_learn_dify(self, language: str) -> RecommendedAppCatalogPage:
+        source = self._source()
+        if source is not self._remote:
+            return source.list_learn_dify(language)
+        try:
+            return self._remote.list_learn_dify(language)
+        except _RecommendedAppSourceUnavailableError as error:
+            logger.warning("fetch learn dify apps from dify official failed: %s, switch to built-in.", error)
+            return self._builtin.list_learn_dify(language)
```
41tair, line 237: If failures will return the content in recommended_apps.json, but there no learn_dify_apps  in there.

api/repositories/recommended_app_catalog_repository.py
```diff
@@ -0,0 +1,191 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Keep the legacy "explore" Redis key: Explore was the former UI name for this recommended-app surface.
+_CATEGORY_ORDER_KEY_PREFIX = "explore:apps:category_order"
+
+
+def _order_categories(categories: set[str], language: str) -> list[str]:
+    try:
+        raw_categories = redis_client.get(f"{_CATEGORY_ORDER_KEY_PREFIX}:{language}")
```
41tair: Maybe we should use redis client from injection.

api/services/recommended_app_catalog_gateway.py
hjlarry: I added `is_learn_dify` to three existing bundled templates and now derive the built-in Learn Dify list from the same recommended_apps data

api/services/recommended_app_catalog_gateway.py
hjlarry: the previous implementation tried to read Origin from request.headers, but this gateway is not guaranteed to run within a Flask request context, so the header was not reliably available in this layer. It now uses CONSOLE_WEB_URL directly as the outbound Origin

41tair (dismissed):

41tair (approved):

PR #40396, opened 2026-08-10
fix: add semantic document titles across pages

iamjoel:
## Summary

- Add localized semantic document titles across authentication, dataset creation, OAuth, device authorization, education, and human-input flows.
- Provide branded root fallbacks and dynamic titles based on application and page state.
- Remove the accidental integrations component route and add focused title coverage.

Fixes SNP-626

## Checklist


lyzno1 (commented):
The direction is good, but title ownership should be split between server metadata for route-level defaults and client updates only for genuinely state-driven views. A few state and ownership issues remain below.

web/app/layout.tsx
```diff
@@ -40,10 +40,15 @@ export default async function RootLayout({ children }: { children: React.ReactNo
...
-
+  const systemFeatures = queryClient.getQueryData(systemFeaturesQuery.queryKey)
+  const documentTitle =
+    systemFeatures?.branding.enabled && systemFeatures.branding.application_title
+      ? systemFeatures.branding.application_title
+      : 'Dify'
   return (
     <html lang={locale ?? 'en'} className="h-full" suppressHydrationWarning>
       <head>
+        <title>{documentTitle}</title>
```
lyzno1: Could this be the single server-side metadata owner instead? A root `generateMetadata` can provide the branded default and `title.template`, while fixed routes provide page metadata. The current setup sends only the brand in the initial HTML, then relies on `useTitle` after hydration and duplicates the brand fallback logic in `useDocumentTitle`. See: https://nextjs.org/docs/app/api-reference/functions/generate-metadata

web/app/reset-password/set-password/page.tsx
```diff
@@ -8,11 +8,13 @@ import { useCallback, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import Input from '@/app/components/base/input'
 import { validPassword } from '@/config'
+import useDocumentTitle from '@/hooks/use-document-title'
 import { useRouter, useSearchParams } from '@/next/navigation'
 import { changePasswordWithToken } from '@/service/common'
 
 const ChangePasswordForm = () => {
   const { t } = useTranslation()
+  useDocumentTitle(t(($) => $.changePassword, { ns: 'login' }))
```
lyzno1: This stays “Set a password” after `showSuccess` replaces the page with the password-changed view for ~5 seconds. Please derive the title from `showSuccess` (the webapp reset flow has the same issue) and cover the success state. WCAG 2.4.2 guidance: https://www.w3.org/WAI/WCAG22/Understanding/page-titled.html

web/app/forgot-password/page.tsx
```diff
@@ -10,9 +11,14 @@ import Header from '../signin/_header'
 import ForgotPasswordForm from './ForgotPasswordForm'
 
 const ForgotPassword = () => {
-  useDocumentTitle('')
+  const { t } = useTranslation()
   const searchParams = useSearchParams()
   const token = searchParams.get('token')
+  useDocumentTitle(
```
lyzno1: Token presence is not enough to describe the current view: an invalid token and a successful reset still keep the “Set a password” title. Please let the client state owner update the title for loading/invalid/success states.

web/app/components/explore/try-app/app/index.tsx
```diff
@@ -13,11 +14,12 @@ type Props = Readonly<{
 }>
 
 function TryApp({ appId, appDetail }: Props) {
+  const { t } = useTranslation()
   const mode = appDetail?.mode
   const isChat = ['chat', 'advanced-chat', 'agent-chat'].includes(mode!)
   const isCompletion = !isChat
 
-  useDocumentTitle(appDetail?.site?.title || '')
+  useDocumentTitle(appDetail?.site?.title || t(($) => $.explore, { ns: 'login' }))
```
lyzno1: This dialog competes with the underlying Home/Apps route title owner, and `useTitle` does not restore the previous title when the dialog unmounts. The new fallback can also produce `Explore Dify - Dify` (or leak `Dify` in a branded deployment). Prefer keeping document-title ownership at the route level.

web/app/signin/invite-settings/page.tsx
```diff
@@ -115,6 +116,15 @@ export default function InviteSettingsPage() {
...
   const requiresAccountSetup =
     checkRes?.data?.requires_setup ?? checkRes?.data?.account_status === 'pending'
+  const setupAccountTitle = t(($) => $.setYourAccount, { ns: 'login' })
+  const documentTitle = !checkRes
+    ? setupAccountTitle
+    : !checkRes.is_valid
+      ? t(($) => $.invalid, { ns: 'login' })
+      : requiresAccountSetup || !checkRes.data?.workspace_name
+        ? setupAccountTitle
+        : `${t(($) => $.join, { ns: 'login' })}${checkRes.data.workspace_name}`
```
lyzno1: Please use a translated interpolation here instead of concatenation. Several locales do not include a trailing space or require a different word order, so this can produce titles such as `RejoindreAcme`; the current key-based test does not catch that.

web/app/components/datasets/create/index.tsx
```diff
@@ -55,6 +56,9 @@ const DEFAULT_CRAWL_OPTIONS: CrawlOptions = {
 
 const DatasetUpdateForm = ({ datasetId }: DatasetUpdateFormProps) => {
   const { t } = useTranslation()
+  useDocumentTitle(
```
lyzno1: This component is also rendered by /datasets/[datasetId]/documents/create, where the client dataset-detail layout already owns Add Documents · {dataset}. This later-mounted effect overwrites that title with the top-level “Create a ready-to-use knowledge base” copy and drops the dataset context. Please keep a single client owner here—for example, do not claim the title when datasetId is present—and cover the composed route rather than testing both owners in isolation.

web/app/(commonLayout)/app/(appDetailLayout)/[appId]/layout-main.tsx
```diff
@@ -59,8 +76,10 @@ const AppDetailLayout: FC<IAppDetailLayoutProps> = (props) => {
   const [isLoadingAppDetail, setIsLoadingAppDetail] = useState(false)
   const [appDetailRes, setAppDetailRes] = useState<App | null>(null)
   const routeAppDetail = appDetailRes ?? (appDetail?.id === appId ? appDetail : null)
+  const pageTitle = appDetailPageTitle(pathname, t)
+  const appName = routeAppDetail?.id === appId ? routeAppDetail.name : undefined
```
Comment (lyzno1) [inline web/app/(commonLayout)/app/(appDetailLayout)/[appId]/layout-main.tsx:81]: On a direct load, appDetailRes remains non-null after bootstrapping the store, so it permanently wins the current fallback order. Renaming the app—or receiving a collaboration metadata update—updates appDetail, but appName still comes from the initial response and the document title keeps the old name. Please keep this client-owned title reactive by preferring the same-ID store value and using appDetailRes only as the bootstrap fallback, with a direct-load-then-rename regression test.

web/app/signin/page.tsx
```diff
@@ -1,11 +1,19 @@
...
 
 const SignIn = () => {
+  const { t } = useTranslation()
   const searchParams = useSearchParams()
   const step = searchParams.get('step')
+  const documentTitle =
+    step === 'next'
+      ? t(($) => $.oneMoreStep, { ns: 'login' })
+      : t(($) => $.signBtn, { ns: 'login' })
+  useDocumentTitle(documentTitle)
```
lyzno1: This semantic title is fully determined by the request-visible step search parameter, but it is currently applied only after client hydration. Please do not move SignInLayout or its provider/query ownership to the server. A narrow option is to keep this interactive implementation unchanged in a Client Component and let only a thin page route shell export generateMetadata({ searchParams }), as recommended by the Next.js Metadata API: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#why-generatemetadata-is-server-component-only

lyzno1 (approved):

web/app/signin/sign-in-page.tsx
```diff
@@ -0,0 +1,13 @@
+'use client'
+import { useSearchParams } from '@/next/navigation'
+import NormalForm from './normal-form'
+import OneMoreStep from './one-more-step'
+
+const SignInPage = () => {
```
lyzno1, line 6: The sign-in header calls `setLocaleOnClient(value, false)`, which changes react-i18next without a navigation or reload. Before this split, `useDocumentTitle` reran from the translated title; now `generateMetadata` only reflects the server request, so selecting another language translates the form but leaves the tab title in the previous language until the next navigation. Please keep a small client title reconciliation here while retaining server metadata (or explicitly refresh metadata on locale change), and cover the no-reload locale switch.

web/app/(commonLayout)/app/(appDetailLayout)/[appId]/layout-main.tsx
Comment (lyzno1) [reply web/app/(commonLayout)/app/(appDetailLayout)/[appId]/layout-main.tsx:81]: The title now reacts, but this preference also changes the data source used by the ACL/redirect effect below. The app-update endpoint returns AppDetailWithSite without the GET endpoint's permission_keys enrichment, so onEdit stores a same-ID app with permission_keys: []. A per-app RBAC editor who renames an app will then be treated as unauthorized and redirected away from workflow/logs/overview. Please prefer the live store only for the title, or merge the returned metadata into the existing fully authorized detail instead of replacing the ACL source. The regression test should model the actual update response with empty permission keys.

web/app/signin/sign-in-page.tsx
iamjoel: Not important...

PR #40424, opened 2026-08-10
fix(vdb): narrow bare except in text_exists() to NotFoundError

### LINKED ISSUE #40423: bare except in VDB text_exists() swallows KeyboardInterrupt/SystemExit
### Self Checks

## 1. Environment

- Dify version: main (f5e1f1590f)

## 2. Describe the problem

The `text_exists()` methods in the OpenSearch and Lindorm vector backends use a bare `except:` that catches every exception, including `KeyboardInterrupt` and `SystemExit`:

- `api/providers/vdb/vdb-opensearch/src/dify_vdb_opensearch/opensearch_vector.py:194`
- `api/providers/vdb/vdb-lindorm/src/dify_vdb_lindorm/lindorm_vector.py:251`

```python
def text_exists(self, id: str) -> bool:
    try:
        self._client.get(index=self._collection_name.lower(), id=id)
        return True
    except:          # catches KeyboardInterrupt / SystemExit too
        return False
```

A user pressing Ctrl+C (or a worker shutting down) mid-call gets the interrupt silently swallowed and the document reported as "does not exist".

## 3. Steps to reproduce

1. Configure the vector store as OpenSearch or Lindorm.
2. Call `text_exists()` and interrupt the process during the `get` call.

## 4. ✔️ Expected Behavior

Only "document not found" (HTTP 404) errors are caught and reported as `False`; `KeyboardInterrupt`/`SystemExit` and other transport errors propagate normally.

## 5. ❌ Actual Behavior

The bare `except:` swallows `KeyboardInterrupt`/`SystemExit`, hiding real exits.

## 6. Fix

`opensearchpy` raises `NotFoundError` for HTTP 404 responses (missing documents), so narrow the handler to `except NotFoundError:`. Keep the existing `BulkIndexError` import (still used by `delete_by_ids()`) and add the `NotFoundError` import. Unit tests are updated accordingly to raise `NotFoundError` instead of a generic `RuntimeError`.

zrh805:
> [!IMPORTANT]
>

## Summary

Replace the bare `except:` in `OpenSearchVector.text_exists()` and `LindormVectorStore.text_exists()` with `except NotFoundError:`. `opensearchpy` maps `NotFoundException`/`index_not_found_exception` (both HTTP 404) to `NotFoundError`, so only missing-document / missing-index lookups are reported as `False`.

The wider consequence: previously a bare `except:` also swallowed transport errors. The production caller `_filter_duplicate_texts` (`api/core/rag/datasource/vdb/vector_base.py:65`, `vector_factory.py:303`) loops over every document without a `try`, so an OpenSearch connection hiccup used to be misread as "document absent" and the document silently re-indexed — a fast path to duplicates you cannot explain later. With this change the transport error propagates and aborts the run instead. Tests now pin this behaviour on both sides: `NotFoundError` returns `False`, while a `ConnectionError` propagates.

Per review feedback (thanks @chuenchen309): `NotFoundError` is imported from the `opensearchpy` package root to match the neighbouring imports; the real `opensearch-py` hierarchy is confirmed (`404 -> NotFoundError`, `ConnectionError -> ConnectionTimeout -> TransportError`, etc.).

```diff
 def text_exists(self, id: str) -> bool:
     try:
         self._client.get(index=self._collection_name.lower(), id=id)
         return True
-    except:
+    except NotFoundError:
         return False
```

Fixes #40423

## Screenshots

| Before | After |
| ------ | ----- |
| Missing doc and transport errors both return False (silent re-index on hiccup) | Only NotFoundError (404) returns False; connection errors propagate and abort indexing |

## Checklist


chuenchen309:
The narrowing is right, and the case that would have made it wrong holds up. Checking the client rather
than assuming it (`opensearch-py` 3.2.0):

```
HTTP_EXCEPTIONS = {400: RequestError, 401: AuthenticationException, 403: AuthorizationException,
                   404: NotFoundError, 409: ConflictError}
```

404 is mapped by status alone, so `NotFoundError` covers both a missing document **and** a missing index.
That second one is the case worth naming, because it is what a first indexing run into a not-yet-created
collection hits — `index_not_found_exception` also comes back as 404, so `text_exists()` still returns
`False` there instead of raising.

**One consequence that is wider than the description.** The body argues from `KeyboardInterrupt` /
`SystemExit`, but those were never plausible here. What the bare `except:` was really swallowing is the
rest of the transport hierarchy — and none of it is a `NotFoundError` subclass:

```
ConnectionError       -> TransportError -> OpenSearchException
ConnectionTimeout     -> ConnectionError -> TransportError
AuthenticationException -> TransportError -> OpenSearchException
NotFoundError         -> TransportError -> OpenSearchException
```

The production caller is `_filter_duplicate_texts`, at `api/core/rag/datasource/vdb/vector_base.py:65` and
again at `vector_factory.py:303`. It loops over every document in the batch and calls `text_exists(doc_id)`
with no `try` around it. So the change is: an OpenSearch hiccup mid-batch used to be read as "document
absent" and the document was re-indexed; now it aborts the whole indexing run.

I think that is the better behaviour — silently re-indexing on a connection error is how you get duplicates
you cannot explain later — but it is the part a reviewer should be weighing, and the description does not
mention it. Worth saying so explicitly, and a test asserting that `ConnectionError` propagates would pin it
next to the two that assert `NotFoundError` does not.

**Completeness.** These two are the only truly bare `except:` in any vdb `text_exists`. Scanning the first
12 lines of every `def text_exists` on `main` under `api/providers/vdb/`: lindorm and opensearch had
`except:`, oceanbase (`oceanbase_vector.py`) and analyticdb (`analyticdb_vector_sql.py`) use
`except Exception as e`, which is a different and milder question. So this closes the family rather than
sampling it.

Small nit: the two files import their other opensearch names from the package root
(`from opensearchpy import OpenSearch, helpers`), and `NotFoundError` is re-exported there too, so
`from opensearchpy import NotFoundError` would match the neighbours. Purely cosmetic.

*AI-assisted; the exception table and the caller locations are output from running and grepping, not
recalled.*

zrh805:
Thanks for the detailed review @chuenchen309 — all three points are addressed in `a288ec2c`:

1. **Wider consequence now documented + pinned by tests.** The PR description now states that transport errors (`ConnectionError`/`ConnectionTimeout`/…) propagate out of `text_exists()` and abort an indexing run in `_filter_duplicate_texts`, instead of silently re-indexing duplicates. I agree with you that propagating is the better behaviour; both unit tests now assert both sides — `NotFoundError` returns `False`, and a `ConnectionError` propagates.
2. **Completeness** — confirmed lindorm/opensearch were the only truly bare `except:` in the VDB `text_exists()` family (oceanbase/analyticdb already use `except Exception as e`).
3. **Import nit** — `NotFoundError` is now imported from the `opensearchpy` package root (it is re-exported there), matching the neighbouring imports.

chuenchen309:
Checked `a288ec2c` and `26974384`. The import alignment and the propagation tests are in, and the
`26974384` follow-up is the right catch — moving the import to the package root does break any test
that fabricates its own `opensearchpy`, and all three that exist on `main` are now patched
(`test_opensearch.py`, `test_opensearch_vector.py`, `test_lindorm_vector.py`; `git grep -l
'types.ModuleType("opensearchpy")'` finds no others).

One problem with the two new assertions, though.

**They raise the builtin `ConnectionError`, not the client's.** Neither fake module defines a
`ConnectionError`, and neither test file imports one, so the bare name in

```python
vector._client.get.side_effect = ConnectionError("connection failed")
with pytest.raises(ConnectionError, match="connection failed"):
```

resolves to `builtins.ConnectionError`, an `OSError`. The exception the real client raises is
`opensearchpy.exceptions.ConnectionError`, and the two are unrelated (`opensearch-py` 3.1.0, the
pinned version):

```
builtins.ConnectionError        -> OSError -> Exception
opensearchpy ConnectionError    -> TransportError -> OpenSearchException -> Exception
issubclass(opensearchpy.exceptions.ConnectionError, builtins.ConnectionError) is False
```

Both propagate today, so the tests are green and the shipped behaviour is right. What's lost is the
ability to fail. The regression these tests exist to catch is someone later deciding a transport blip
shouldn't abort an indexing run and widening the handler — and that change is invisible to them:

| raised by the client | `except NotFoundError` (this PR) | `except (NotFoundError, TransportError)` |
|---|---|---|
| `builtins.ConnectionError` — what the tests raise | propagates | propagates |
| `opensearchpy` `ConnectionError` — what the client raises | propagates | **swallowed** |
| `opensearchpy` `ConnectionTimeout` | propagates | **swallowed** |

Row 1 is the test, rows 2 and 3 are production. The fix is small: give each fake module a
`ConnectionError` deriving from its `TransportError` (or just from `Exception` if you'd rather not
model the hierarchy), expose it at the package root next to `NotFoundError`, and have the tests raise
`module.ConnectionError`. Then widening the handler turns the tests red, which is the point of them.

Also a correction to my own earlier comment: I labelled the exception table `opensearch-py` 3.2.0,
but both `pyproject.toml`s pin `3.1.0`. I re-ran it on 3.1.0 — `HTTP_EXCEPTIONS` is identical, so
nothing in that comment changes except the version string.

*AI-assisted with agentic coding tools; the class relationships and the table are output from running
this against `opensearch-py` 3.1.0, not recalled.*

PR #40436, opened 2026-08-10
refactor(api): add account profile application service

41tair:
## Summary

Establish the Account application boundary and the repository foundation for thinning `controllers/console/workspace/account.py`.

part of #39993.

- add framework-neutral Account contracts, ports, and a stable request context
- introduce `AccountRepository` and an Account unit of work backed by short-lived SQLAlchemy sessions
- move profile mutation orchestration into `AccountProfileService`
- add `PATCH /console/api/account/profile` as the consolidated profile mutation endpoint
- keep existing profile routes and mark them as deprecated in code for a later removal cycle
- register the service and repository implementations in the application-service composition root
- enforce the framework and persistence boundary with import-linter

## Architecture

The application service declares the ports it needs and does not import Flask, HTTP types, ORM models, SQLAlchemy, or repositories. The repository and unit-of-work implementations remain outside the application layer. No code falls back to `db.session`.

This is the foundation for a future IAM service containing multiple repositories, including Account and RBAC repositories.

## Stack

1. #40436 — Account repository, UoW, and profile service
2. #40437 — Account access, password, avatar, and integrations
3. #40438 — Account lifecycle, invitations, and deletion
4. #40439 — Change-email state machine
5. #40440 — Education, deletion feedback, and Admission completion


api/controllers/console/workspace/account.py
```diff
@@ -123,6 +129,42 @@ def validate_timezone(cls, value: str) -> str:
...
+
+    name: Annotated[str, Field(min_length=3, max_length=30)] | SkipJsonSchema[None] = None
+    avatar: str | SkipJsonSchema[None] = None
+    interface_language: str | SkipJsonSchema[None] = None
+    interface_theme: Literal["light", "dark"] | SkipJsonSchema[None] = None
+    timezone: str | SkipJsonSchema[None] = None
+
+    @field_validator("*", mode="before")
+    @classmethod
+    def reject_null(cls, value: object) -> object:
```
asukaminato0721, line 138: then why the field has | None

api/controllers/console/workspace/account.py
```diff
@@ -123,6 +129,42 @@ def validate_timezone(cls, value: str) -> str:
...
+        if value is None:
+            raise ValueError("Account profile fields cannot be null")
+        return value
+
+    @field_validator("interface_language")
+    @classmethod
+    def validate_language(cls, value: str | None) -> str | None:
+        return supported_language(value) if value is not None else None
```
asukaminato0721: so... can this field be None ?

api/services/entities/account_entities.py
```diff
@@ -0,0 +1,42 @@
...
+@dataclass(frozen=True, slots=True)
+class AccountProfileChanges:
+    name: str | None = None
+    avatar: str | None = None
+    interface_language: str | None = None
+    interface_theme: str | None = None
+    timezone: str | None = None
+
+    def has_changes(self) -> bool:
+        return any(
```
asukaminato0721, line 33:
```suggestion
        return any(v is not None for v in asdict(self).values())
```

maybe ?

api/controllers/console/workspace/account.py
41tair: None is used internally to represent an omitted PATCH field. Explicit JSON null is hidden from the schema and rejected by reject_null.

api/services/entities/account_entities.py
41tair: I’d prefer to keep the fields explicit here. asdict() recursively copies the dataclass and would make future fields silently affect this predicate.

api/controllers/console/workspace/account.py
41tair: fixed.

api/controllers/console/workspace/account.py
```diff
@@ -123,6 +129,42 @@ def validate_timezone(cls, value: str) -> str:
...
+        if value is None:
+            raise ValueError("Account profile fields cannot be null")
+        return value
+
+    @field_validator("interface_language")
+    @classmethod
+    def validate_language(cls, value: str) -> str:
+        return supported_language(value)
+
+    @field_validator("timezone")
```
asukaminato0721, line 148: https://pydantic.dev/docs/validation/latest/concepts/validators/#field-before-validator ?

41tair: reject_null is already a field before validator via @field_validator("*", mode="before"). So i think the timezone we get is a validated string.

asukaminato0721: I mean use Annotation style type.

api/controllers/console/workspace/account.py
```diff
@@ -310,21 +361,29 @@ class AccountProfileApi(Resource):
     def get(self, current_user: Account):
         return dump_response(AccountResponse, current_user)
 
+    @console_ns.expect(console_ns.models[AccountProfilePatchPayload.__name__])
+    @console_ns.response(HTTPStatus.OK, "Success", console_ns.models[AccountResponse.__name__])
+    @console_account_admission()
+    def patch(self, request_context: RequestContext):
+        payload = console_ns.payload or {}
+        args = AccountProfilePatchPayload.model_validate(payload)
```
asukaminato0721: maybe model_validate decorator?

asukaminato0721 (dismissed):

api/controllers/console/workspace/account.py
41tair: got

hjlarry (approved):

PR #40449, opened 2026-08-10
fix(web): render agent image icons from icon_url

### LINKED ISSUE #40442: Agent image icon fails to load (404): AppIcon receives the raw file id instead of icon_url
### Self Checks

### Dify version

1.16.1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Enable Agent v2 (`ENABLE_AGENT_V2=true`) and create an Agent, e.g. `Test-agent`.
2. Set the Agent icon to an uploaded **image** (not an emoji). The upload itself succeeds: `POST /console/api/files/upload` returns 200 and a row is created in `upload_files`.
3. Reload the page and look at the Agent icon on the Agents list, or on the Agent configure page.

Requests observed in the browser devtools and in the nginx access log:

```
GET /29bdb007-4d8c-4888-83a2-7587abcafb26                                       404
GET /agents/019fd69e-7b20-79ee-8e15-b289e27928bc/29bdb007-4d8c-4888-83a2-7587abcafb26   404
```

The requested path is the **bare `upload_files.id`**. Because it is not an absolute path, the browser resolves it relative to the current page, which is why the same id produces a different (and always wrong) URL on `/agents` versus `/agents/<agent_id>/configure`.

Agents whose icon is an **emoji** are not affected. Apps, datasets and the webapp site icon are not affected either — only Agent icons of type `image`.

### ✔️ Expected Behavior

The Agent icon renders using the signed preview URL that the backend already computes and returns as `icon_url`, e.g.

```
/files/29bdb007-4d8c-4888-83a2-7587abcafb26/file-preview?timestamp=...&nonce=...&sign=...
```

### ❌ Actual Behavior

A 404 and a broken image placeholder, everywhere an Agent icon of type `image` is rendered.

### Root cause

The backend is fine. `AgentAppDetailWithSite` and `AgentAppPartial` both inherit the `icon_url` computed field from `AppDetailWithSite`, and it resolves correctly. Verified in a 1.16.1 container:

```python
>>> AgentAppDetailWithSite.model_validate(app_model, from_attributes=True, ...).model_dump(mode="json")["icon_url"]
'/files/29bdb007-4d8c-4888-83a2-7587abcafb26/file-preview?timestamp=1786355545&nonce=...&sign=...'
```

The problem is on the web side. `AppIcon` uses the `imageUrl` prop directly as the `` source:

```tsx
// web/app/components/base/app-icon/index.tsx
112: const isValidImageIcon = iconType === 'image' && imageUrl
138: <img src={imageUrl} className="size-full" alt="app icon" />
```

But the Agent components pass `agent.icon` — the raw `upload_files` id — into `imageUrl`, where every other caller passes `icon_url`:

```tsx
// web/app/components/workflow/block-selector/agent-selector.tsx:230
imageUrl={agent.icon ?? undefined}

// web/app/components/workflow/nodes/agent-v2/components/agent-roster-field.tsx:73
imageUrl={agent.icon ?? undefined}
```

Compare with the correct pattern used elsewhere, e.g. app / dataset / site icons:

```tsx
imageUrl={app.icon_url}
```

The same `imageUrl: X.icon` pattern also appears in the shipped `dify-web:1.16.1` bundle in the Agent configure page header component (the chunk containing `i-custom-vender-agent-v2-configure`), so there is likely a third call site besides the two above. In the built output the bad pattern occurs only in Agent-related chunks:

```
$ grep -oE "imageUrl:[A-Za-z_$]+\.icon(\?\?|,|\}|\))" <chunk>.js
# matches only in the agent header and block-selector chunks;
# all app/dataset/site chunks use imageUrl:X.icon_url
```

### Suggested fix

Pass `icon_url` instead of `icon` in the Agent call sites, falling back to `undefined`:

```diff
- imageUrl={agent.icon ?? undefined}
+ imageUrl={agent.icon_url ?? undefined}
```

`icon_url` is already present on both the Agent detail and the Agent list (roster) responses, so no backend change is needed.

Possibly related: #39447 (closed, reported against 1.16.0).

Souravrajvi0:
## Summary

Fixes #40442

Agent image icons were rendered with the raw `upload_files.id` as the ``, causing 404s on Agents roster, configure sidebar, and workflow agent surfaces. This change uses the signed `icon_url` for image icons (and keeps external link icons unchanged).

## Changes

- Add `getAgentAppIconImageUrl()` helper for consistent icon URL resolution
- Update Agent UI surfaces to pass `icon_url` into `AppIcon.imageUrl`:
  - Agent roster list
  - Agent detail navigation
  - Workflow agent selector, node, and roster field
- Include `icon_url` in roster/invite-option API payloads via `AgentRosterService.serialize_agent`
- Add frontend and backend unit tests

## Test plan

- [x] `pnpm exec vitest run` on affected frontend test files (37 tests passed)
- [x] `pytest tests/unit_tests/services/agent/test_agent_services.py::test_serialize_agent_includes_icon_url_for_image_icons`
- [ ] Manual: create Agent with uploaded image icon, verify icon renders on roster + configure pages after reload

### DIFF
--- a/api/fields/agent_fields.py
+++ b/api/fields/agent_fields.py
@@ -81,6 +81,7 @@ class AgentRosterResponse(ResponseModel):
     icon_type: AgentIconType | None = None
     icon: str | None = None
     icon_background: str | None = None
+    icon_url: str | None = None
     agent_kind: AgentKind
     scope: AgentScope
     source: AgentSource
--- a/api/openapi/markdown/console-openapi.md
+++ b/api/openapi/markdown/console-openapi.md
@@ -14239,6 +14239,7 @@ Supported icon storage formats for Agent roster entries.
 | icon | string |  | No |
 | icon_background | string |  | No |
 | icon_type | [AgentIconType](#agenticontype) |  | No |
+| icon_url | string |  | No |
 | id | string |  | Yes |
 | in_current_workflow_count | integer |  | No |
 | is_in_current_workflow | boolean |  | No |
@@ -14703,6 +14704,7 @@ section may be empty, which is how callers express "no knowledge layer".
 | icon | string |  | No |
 | icon_background | string |  | No |
 | icon_type | [AgentIconType](#agenticontype) |  | No |
+| icon_url | string |  | No |
 | id | string |  | Yes |
 | name | string |  | Yes |
 | published_node_reference_count | integer |  | No |
--- a/api/services/agent/roster_service.py
+++ b/api/services/agent/roster_service.py
@@ -103,6 +103,8 @@ class AgentRosterService:
         reference_count: int | None = None,
     ) -> dict[str, Any]:
         published_references = published_references or []
+        from libs.helper import build_icon_url
+
         return {
             "id": agent.id,
             "name": agent.name,
@@ -111,6 +113,7 @@ class AgentRosterService:
             "icon_type": agent.icon_type.value if agent.icon_type else None,
             "icon": agent.icon,
             "icon_background": agent.icon_background,
+            "icon_url": build_icon_url(agent.icon_type, agent.icon),
             "agent_kind": agent.agent_kind.value,
             "scope": agent.scope.value,
             "source": agent.source.value,
--- a/api/tests/unit_tests/services/agent/test_agent_services.py
+++ b/api/tests/unit_tests/services/agent/test_agent_services.py
@@ -10,6 +10,7 @@ from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session
 
 from core.workflow.nodes.agent_v2.validators import WorkflowAgentNodeValidationError
+from libs.helper import build_icon_url
 from models.account import Account
 from models.agent import (
     Agent,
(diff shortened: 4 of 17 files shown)


dparkmit24:
Disclosure first: I had independently built a fix for #40442 (never opened) before finding this PR, so I arrived at this review with the same code paths fresh in mind.

**Runtime check.** I cherry-picked this branch's two commits onto a 1.16.1 stack (clean on the web side) and verified in the browser: an image-icon Agent that renders as a broken image on stock 1.16.1 renders the uploaded image on `/agents` with this change — `getAgentAppIconImageUrl` produced the signed `/files/...` URL. The block-selector wasn't runtime-exercisable on my rig (its API was stock), so I traced that path on `main` instead: `GET /agent/invite-options` builds its payload entirely from `serialize_agent()`'s dict (controller → `dump_response` → `model_validate(<dict>)`), so the new `icon_url` field is populated on that endpoint too. Traced, not executed — but the trace is unambiguous.

**One suggestion: an unmocked assertion in the backend test.** `test_serialize_agent_includes_icon_url_for_image_icons` monkeypatches `build_icon_url`, which leaves one real dependency untested: `agent.icon_type` here is `AgentIconType`, a different enum from `models.model.IconType`, so `build_icon_url` takes its `str(icon_type)` fallback — and that yields `'image'` only because `AgentIconType` is a `StrEnum`. If either enum's base class ever changed to a plain `(str, Enum)`, every icon would silently serialize `icon_url=None` while the mocked test stayed green. A single assertion calling the real `build_icon_url` with `AgentIconType.IMAGE` would pin it. (For transparency: the parallel fix I'd drafted had the identical dependency — this is a property of the shared helper, not of this patch.)

**A strength worth naming, so a future refactor doesn't "simplify" it away:** keeping `icon` as the raw file id and adding `icon_url` additively is load-bearing, not just tidy. The Agent edit dialog reads `icon` back (`createAgentIconSelection` → `fileId`) and writes it to `PUT /agent/{agent_id}` (`applyIconPayload`) — if `icon` ever carried the signed URL, opening Edit and pressing Save would persist an expiring signed URL into `apps.icon`. This PR gets that right, and it also covers the `/agents/<id>` detail sidebar (`navigation.tsx`), a surface my own draft had missed.

Souravrajvi0:
Rebased onto latest `main` — ready for review.

PR #40455, opened 2026-08-10
fix(agent-app): deliver tool-generated images to the chat instead of dropping them

### LINKED ISSUE #40425: Agent App: Images generated by third-party tools are not displayed in chat — image URLs discarded in _convert_tool_response_to_text
## What happened

When an Agent App (the new Agent backend path, not the legacy Agent Chat) calls a third-party tool that returns an image (e.g. an image generation tool that returns `MessageType.IMAGE_LINK` or `MessageType.IMAGE`), the image is never displayed to the user in the chat interface. The user only sees a text response like "image has been created and sent to user already, you do not need to create it, just tell the user to check it now."

## Root Cause

There are two layers where image data is lost:

### 1. dify-agent SDK side — `_convert_tool_response_to_text` discards image URLs

`dify-agent/src/dify_agent/layers/dify_plugin/tools_layer.py`, lines 636-681

```python
def _convert_tool_response_to_text(tool_response: Sequence[DifyPluginToolInvokeMessage]) -> str:
    ...
    elif response.type in {
        DifyPluginToolInvokeMessage.MessageType.IMAGE_LINK,
        DifyPluginToolInvokeMessage.MessageType.IMAGE,
    }:
        parts.append(
            "image has been created and sent to user already, "
            "you do not need to create it, just tell the user to check it now."
        )
        # ❌ Image URL in message.text is discarded
```

The plugin daemon returns `DifyPluginToolInvokeMessage` with `message.text` containing the image URL, but `_convert_tool_response_to_text` replaces it with a fixed text string. The image URL is lost at this point.

### 2. Dify API side — no compensation mechanism to push images to frontend

The legacy Agent Chat path (`FunctionCallAgentRunner` / `CotAgentRunner`) has a compensation chain that runs **before** converting tool output to text:
- `ToolFileMessageTransformer.transform_tool_invoke_messages()` downloads and stores images as `ToolFile`
- `ToolEngine._create_message_files()` creates `MessageFile` DB records
- The runner publishes `QueueMessageFileEvent` to the SSE stream
- The frontend receives `message_file` SSE events and renders the image

The new Agent App path (`AgentAppRunner` in `api/core/app/apps/agent_app/app_runner.py`) has **no equivalent mechanism**:
- `_AgentProcessRecorder._handle_tool_result_event()` stores tool results as observation text only
- `MessageAgentThought.message_files` is always `""` (empty string)
- `QueueMessageFileEvent` is never imported or published
- `_message_end_to_stream_response()` queries `MessageFile` table → 0 records → `files=[]`
- Frontend `onFile` callback never fires

## Reproduction

1. Create an Agent App (not Agent Chat / not Workflow)
2. Configure a third-party image generation tool (e.g. DALL-E, Stable Diffusion, or any tool that returns `MessageType.IMAGE` / `MessageType.IMAGE_LINK`)
3. Send a message requesting image generation
4. Agent calls the tool, tool successfully generates an image
5. Chat only shows text "image has been created and sent to user already..." — no image is displayed

## Expected Behavior

The image generated by the tool should be displayed in the chat interface, similar to how the legacy Agent Chat path handles tool-generated images.

## Flow Comparison

```
Legacy Agent Chat (images display correctly ✅):
  ToolEngine.agent_invoke()
    → ToolFileMessageTransformer: download image, store as ToolFile
    → _create_message_files(): create MessageFile record (type=IMAGE, belongs_to=ASSISTANT)
    → publish QueueMessageFileEvent → SSE: message_file event → frontend renders ✅
    → tool_response_to_str(): "image has been created..." (text for LLM only)
    → QueueMessageEndEvent → message_end.files = [image list] → frontend renders ✅

New Agent App (images lost ❌):
  dify-agent SDK: invoke_tool closure
    → _convert_tool_response_to_text(): "image has been created..." ← ❌ URL discarded
    → pydantic-ai ToolReturnPart.content = fixed text (no image)
    → emit_run_succeeded(output="text response")
  Dify API: AgentAppRunner._consume_stream()
    → _AgentProcessRecorder: stores observation text only, message_files=""
    → No MessageFile created, No QueueMessageFileEvent published
    → SSE: agent_message (text only), message_end (files=[])
    → Frontend: no image ❌
```

## Suggested Fix

Two approaches:

**Option A: Fix in dify-agent SDK** — Return image information from `_convert_tool_response_to_text` alongside the text, so it can flow through pydantic-ai events to the Dify API side, which then creates `MessageFile` records and publishes `QueueMessageFileEvent`.

**Option B: Add compensation in Dify API** — In `AgentAppRunner._consume_stream()` / `_AgentProcessRecorder._handle_tool_result_event()`, extract file references from tool results, create `MessageFile` records (similar to legacy `ToolEngine._create_message_files`), and publish `QueueMessageFileEvent` to the SSE stream.

## Environment

- Dify version: latest `main` branch (cloned for source analysis)
- Path affected: Agent App (new Agent backend path)
- Key files:
  - `dify-agent/src/dify_agent/layers/dify_plugin/tools_layer.py` (lines 636-681)
  - `api/core/app/apps/agent_app/app_runner.py` (lines 250-570)
  - `api/core/workflow/nodes/agent_v2/agent_node.py` (lines 502-590)
  - `api/core/tools/tool_engine.py` (legacy reference, lines 49-157)

dparkmit24:
Fixes #40425

## Summary

Credit where it is due: @iMLe0n's source analysis in #40425 identified both loss layers precisely — `_convert_tool_response_to_text` discarding the URL carried in `message.text`, and `AgentAppRunner`/`_AgentProcessRecorder` having no equivalent of the legacy `ToolFileMessageTransformer → ToolEngine._create_message_files → QueueMessageFileEvent` chain. This PR implements their Option A, with one refinement: rather than changing the observation text, the SDK reports URL-bearing `IMAGE`/`IMAGE_LINK` messages on pydantic-ai's application-only `ToolReturn.metadata` channel (`dify_tool_files` key, new client-safe constant in `dify_agent.layers.dify_plugin.configs`). The LLM-facing observation is byte-identical to before — covered by a test — so model behaviour cannot change, and `ToolReturnPart.metadata` already exists in the wire schema, so the protocol gains no new field.

On the API side, `_AgentProcessRecorder` now extracts those references from tool-result events and replays the existing legacy chain: `ToolFileMessageTransformer.transform_tool_invoke_messages` downloads the image into a `ToolFile` (same SSRF-protected path as legacy Agent Chat), `ToolEngine._create_message_files` creates the `MessageFile` rows, the thought row gets `message_files`, and `QueueMessageFileEvent` is published — from which the existing EasyUI pipeline already produces `message_file` SSE events and a populated `message_end.files`. No frontend changes needed. Any failure in the file chain degrades to the text-only observation instead of failing the run.

Two findings from implementation, both about the code rather than the report:

- An API-only fix (issue's Option B) is not possible: the tool closure returns only the converted text, so `ToolReturnPart.content` — the only thing the API receives — carries no file information. The SDK half of this change is what makes the API half possible.
- `agent_v2/agent_node.py` (workflow Agent node) loses the images too, but has no Message/MessageFile surface — exposing tool files there means file-typed node outputs, which is a separate feature. This PR puts the references on the wire for that path as well, so a follow-up can build on it without another SDK change.

Version-skew safe in both directions: an older API ignores the metadata key; an older agent backend sends no metadata and the API behaves exactly as today.

Known limitation: blob-type tool messages (raw bytes on the daemon stream) are not collected — only URL-bearing image messages. In `dify-official-plugins`, 28 tools emit `IMAGE` via `create_image_message` and one (`ddgo_img`) emits `IMAGE_LINK`, so the URL-bearing case covers the tool ecosystem's common path; blob-emitting tools (e.g. the QR-code generator) are a follow-up.

## Verification

**Runtime, end to end (self-hosted Docker, 1.16.1 images):** with the fal `FLUX.1 [dev]` tool configured on an Agent App and the prompt "Use the FLUX.1 [dev] tool to generate an image of a golden retriever":

- **Without this change** (1.16.1): the tool returned a real image URL; the chat showed only "image has been created and sent to user already…" and no image — reproducing the reported behaviour at runtime, since the issue's original analysis was source-only.
- **With this change** (same images, the branch's four source files applied): the image renders in the chat. The exercised path is `IMAGE`-with-URL over a live agent backend; `IMAGE_LINK` follows the identical branch but was not runtime-exercised (the only credential-free `IMAGE_LINK` tool, DuckDuckGo image search, rate-limited during testing).
- The two runs were on two stacks built from the same 1.16.1 images (the unfixed one additionally had `INTERNAL_FILES_URL` set explicitly, unrelated to this path), so this is a before/after demonstration rather than a single-variable toggle on one deployment.

**Unit tests, both failing before and passing after:** SDK-side (daemon `image_link` → `ToolReturn` with unchanged observation text and the file reference in metadata) and API-side (tool-result event with metadata → `MessageFile` ids on the thought, `QueueMessageFileEvent` published; transformer failure → clean degradation to the text-only observation). Full `dify-agent` local suite and the `agent_app` / `clients/agent_backend` / `core/tools` API suites pass.

## Screenshots

**Before**
40425-before

**After**
40425-after

## Checklist

- [x] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods — ruff clean on all changed files; the api-wide `make type-check` reports 194 pre-existing errors on any file including untouched ones, so I have not claimed that gate passes

From Claude Code

### DIFF
--- a/api/core/app/apps/agent_app/app_runner.py
+++ b/api/core/app/apps/agent_app/app_runner.py
@@ -16,6 +16,7 @@ from decimal import Decimal
 from typing import Any, Literal
 
 from dify_agent.layers.ask_human import AskHumanToolArgs
+from dify_agent.layers.dify_plugin import DIFY_PLUGIN_TOOL_FILES_METADATA_KEY
 from dify_agent.protocol import DeferredToolResultsPayload
 from pydantic import JsonValue
 
@@ -50,8 +51,12 @@ from core.app.entities.queue_entities import (
     QueueAgentThoughtEvent,
     QueueLLMChunkEvent,
     QueueMessageEndEvent,
+    QueueMessageFileEvent,
 )
 from core.repositories.human_input_repository import HumanInputFormRepository, HumanInputFormRepositoryImpl
+from core.tools.entities.tool_entities import ToolInvokeMessage
+from core.tools.tool_engine import ToolEngine
+from core.tools.utils.message_transformer import ToolFileMessageTransformer
 from core.workflow.nodes.agent_v2.ask_human_hitl import AskHumanFormBuildError, create_ask_human_form
 from core.workflow.nodes.agent_v2.ask_human_resume import build_deferred_tool_results, resolve_ask_human_form
 from extensions.ext_database import db
@@ -435,9 +440,21 @@ class _AgentProcessRecorder:
         content = part.get("content")
         if content is None:
             content = part
-        self._record_tool_observation(tool_call_id=tool_call_id, tool_name=tool_name, observation=content)
+        self._record_tool_observation(
+            tool_call_id=tool_call_id,
+            tool_name=tool_name,
+            observation=content,
+            file_references=_tool_file_references(part.get("metadata")),
+        )
 
-    def _record_tool_observation(self, *, tool_call_id: str | None, tool_name: str | None, observation: Any) -> None:
+    def _record_tool_observation(
+        self,
+        *,
+        tool_call_id: str | None,
+        tool_name: str | None,
+        observation: Any,
+        file_references: list[dict[str, Any]] | None = None,
+    ) -> None:
         self._close_thinking_segments()
         thought_id = self._lookup_observation_thought(tool_call_id=tool_call_id, tool_name=tool_name)
         if thought_id is None:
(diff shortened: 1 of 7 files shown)


crazywoola:
Please resolve the conflicts

crazywoola (approved):

PR #40462, opened 2026-08-11
refactor(api): extract installed app definition queries

hjlarry:
## Summary

Part of #39993.

- define one App Definition query boundary for installed-app parameters and tool-icon metadata
- keep SQLAlchemy session ownership and ORM projection inside a single repository
- expose one composition capability instead of endpoint-shaped parameter and meta services
- preserve the existing response and error contracts while leaving shared admission wrappers unchanged

### DIFF
--- a/api/.importlinter
+++ b/api/.importlinter
@@ -93,6 +93,20 @@ forbidden_modules =
     sqlalchemy
     werkzeug
 
+[importlinter:contract:app-definition-query-service-boundary]
+name = App definition query application service is framework and persistence neutral
+type = forbidden
+source_modules =
+    services.app_definition_query_service
+forbidden_modules =
+    controllers
+    extensions
+    flask
+    models
+    repositories
+    sqlalchemy
+    werkzeug
+
 [importlinter:contract:feature-query-service-boundary]
 name = Feature query application service is framework and persistence neutral
 type = forbidden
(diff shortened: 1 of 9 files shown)


41tair:
My main concern is that the App-related boundaries may be becoming a little too endpoint-oriented. Each endpoint is locally well layered, but across the stack we are gradually creating one service, port, repository, DTO, and error model per API response. It may be worth defining a few stable App subdomains first—such as App Definition, App Runtime, and Web App Access—and organizing the new abstractions around those capabilities.

hjlarry:
> My main concern is that the App-related boundaries may be becoming a little too endpoint-oriented. Each endpoint is locally well layered, but across the stack we are gradually creating one service, port, repository, DTO, and error model per API response. It may be worth defining a few stable App subdomains first—such as App Definition, App Runtime, and Web App Access—and organizing the new abstractions around those capabilities.

Thanks. I’ve reworked the stack around stable capabilities rather than API endpoints:
- AppDefinitionQueryService/Repository now owns app parameters, tool metadata, and basic app information. The later PRs extend and reuse this boundary instead of adding separate services, repositories, DTOs, and errors for each endpoint.
- WebAppAccessQueryService/Repository now owns both access-mode and permission decisions. I removed the separate permission-oriented service.
please review again.

41tair (approved):

PR #40482, opened 2026-08-11
refactor(api): extract web app access queries

hjlarry:
## Summary

- define a stable Web App Access query capability for resolving configured access modes
- move app-code lookup into a repository-owned read session
- expose `WebAppAccessMode` from the shared `enums` module and adapt Enterprise settings at the composition root
- snapshot the deployment-wide Web App Auth feature gate when application services are initialized
- keep the controller focused on request validation, error translation, and response serialization
- stabilize access errors as `invalid_param`, `app_not_found`, or `web_app_access_unavailable` based on the failing boundary

This PR is independent of the App Definition stack. It only migrates `/webapp/access-mode`; `/webapp/permission` extends the same capability in a stacked follow-up.

## Compatibility

- Web App Auth disabled still returns `public` before resolving an app
- `appCode` still takes precedence over `appId`
- an unknown `appCode` does not fall back to `appId` and now returns `app_not_found` (404) without echoing the code
- a missing app reference returns `invalid_param` (400)
- known database and Enterprise availability or response failures return `web_app_access_unavailable` (503)
- unknown programming errors remain internal errors instead of being hidden as availability failures
- the successful `accessMode` response is unchanged

## Validation

- 120 focused controller, service, repository, Enterprise, and composition tests
- Ruff check and format check
- Pyrefly
- import-linter: 9 contracts kept
- response-contract linter: 478 valid, 0 mismatch
- no-new-getattr and no-new-controller-SQLAlchemy guards
- `git diff --check`

### DIFF
--- a/api/controllers/web/app.py
+++ b/api/controllers/web/app.py
@@ -8,6 +8,7 @@ from werkzeug.exceptions import Unauthorized
 
 from constants import HEADER_NAME_APP_CODE
 from controllers.common import fields
+from controllers.common.errors import InvalidArgumentError
 from controllers.common.fields import AccessModeResponse, Parameters
 from controllers.common.schema import query_params_from_model, register_response_schema_models, register_schema_models
 from extensions.ext_application_services import application_services
@@ -19,10 +20,15 @@ from models.model import App, EndUser
 from services.app_definition_query_service import AppDefinitionNotPublishedError, AppDefinitionUnavailableError
 from services.enterprise.enterprise_service import EnterpriseService
 from services.feature_service import FeatureService
+from services.webapp_access_query_service import (
+    WebAppAccessAppNotFoundError,
+    WebAppAccessReferenceRequiredError,
+    WebAppAccessUnavailableError,
+)
 from services.webapp_auth_service import WebAppAuthService
 
 from . import web_ns
-from .error import AgentNotPublishedError, AppUnavailableError
+from .error import AgentNotPublishedError, AppUnavailableError, WebAppAccessServiceUnavailableError, WebAppNotFoundError
 from .wraps import WebApiResource
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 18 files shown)


GareArc (dismissed):

api/enums/webapp_access_mode.py
```diff
@@ -0,0 +1,8 @@
+from enum import StrEnum
+
+
+class WebAppAccessMode(StrEnum):
```
41tair: Just put this into `__init__.py`

api/controllers/web/app.py
```diff
@@ -122,21 +121,11 @@ class AppAccessMode(Resource):
...
-            raise ValueError("appId or appCode must be provided")
-
-        res = EnterpriseService.WebAppAuth.get_app_access_mode_by_id(app_id)
-
-        return {"accessMode": res.access_mode}
+        access_mode = application_services().webapp_access_mode_queries.get_access_mode(
+            app_id=args.app_id,
+            app_code=args.app_code,
+        )
+        return {"accessMode": access_mode.value}
```
41tair: Use dumpResponse and SimpleResultResponse instead of dict.

41tair (changes requested):

api/controllers/web/app.py
```diff
@@ -17,13 +18,17 @@
...
+from services.webapp_access_query_service import (
+    WebAppAccessAppNotFoundError,
+    WebAppAccessReferenceRequiredError,
+    WebAppAccessUnavailableError,
+)
 from services.webapp_auth_service import WebAppAuthService
 
 from . import web_ns
-from .error import AgentNotPublishedError, AppUnavailableError
+from .error import AgentNotPublishedError, AppUnavailableError, WebAppAccessServiceUnavailableError, WebAppNotFoundError
```
41tair: i prefer use the absolute reference.

41tair (approved):

PR #40149, opened 2026-08-07
fix(api): bind dataset operations to owners

WH-2099:
## Summary

- Bind dataset API enablement, processing, execution-log, and operational reads to tenant-scoped dataset and document owner chains.
- Resolve every document and metadata ID before metadata writes, and use canonical metadata names from the database.
- Carry `tenant_id`, `dataset_id`, and `document_id` through website sync tasks and scope segment cleanup to that owner chain.
- Scope document metadata serialization through tenant, dataset, and document binding predicates.

Refs #37985

## Screenshots

Not applicable.

## Test plan

- `258 passed` across the affected controller, service, metadata, and task unit-test files.
- `3 passed` for the website sync service boundary after the final owner-mismatch regression was added.
- `71 tests collected` for the affected container-integration files.
- Ruff check and format check passed for all changed Python files.
- Pyrefly and Mypy passed for all changed production files.
- `make api-contract-lint` passed with zero response mismatches.

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues.
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] No documentation update is required for this internal ownership-binding change.
- [ ] I ran the complete `make lint && make type-check` backend suite.

PR #40503, opened 2026-08-11
fix(api): bind remaining RAG resources to owners

WH-2099:
## Summary

- Bind customized RAG template detail, raw export, and publish flows to tenant-owned resources.
- Resolve and authorize transform datasets before plugin or database side effects.
- Restrict single-dataset router output to the validated allowlist while retaining a tenant-scoped fresh read.
- Add a short resource-boundary change guide based on the accepted review patterns from #40149.

## Compatibility

- Preserve existing routes, request payloads, success responses, and valid empty icon metadata.
- Add no database migration, Redis format, Celery task, or other cross-process contract change.
- Keep RBAC and legacy permission paths separate for mixed deployment configurations.

## Test plan

- 239 targeted RAG, schema, and contract tests passed.
- 19 affected regression tests were rerun after the final review fixes.
- `make type-check-core`
- `make api-contract-lint`
- Ruff check and format check
- `git diff --check`

Related to #37985.
Together with #40149, this completes the three remaining RAG checklist items.


api/controllers/console/datasets/rag_pipeline/rag_pipeline_workflow.py
```diff
@@ -1016,19 +1019,31 @@ def get(self, pipeline: Pipeline, node_id: str):
 @console_ns.route("/rag/pipelines/transform/datasets/<uuid:dataset_id>")
 class RagPipelineTransformApi(Resource):
     @console_ns.response(200, "Success", console_ns.models[RagPipelineOpaqueResponse.__name__])
+    @console_ns.response(404, "Dataset or pipeline not found")
     @setup_required
     @login_required
     @account_initialization_required
     @with_current_user
+    @with_current_tenant_id
+    @rbac_permission_required(RBACResourceScope.DATASET, RBACPermission.DATASET_EDIT)
```
41tair, line 1028: why we don't use the decorator get_rag_pipeline like what we did in https://github.com/langgenius/dify/pull/40503/changes#diff-450361505f89bc3082706dd3e4b0c89e7fe6fe9809a9ef968c737280e3708e5eR225

WH-2099: This route starts from a legacy Dataset that has not been transformed yet, so it only has `dataset_id` and normally has no `pipeline_id` for `get_rag_pipeline` to resolve. The publish route is different because its path already identifies an existing pipeline.

Here we apply the same trusted-root pattern at the correct boundary: resolve the Dataset with `dataset_id + current_tenant_id`, perform RBAC/legacy ACL checks, and pass the validated Dataset object to the service. Reusing `get_rag_pipeline` would require changing the HTTP contract or introducing a misleading fallback for datasets without a pipeline.

41tair (approved):

PR #40535, opened 2026-08-11
test: migrate segment service sessions and ORM models to SQLite

asukaminato0721:
## Summary
- replace segment and child-chunk SQLAlchemy session doubles with shared SQLite sessions
- use real Dataset, Document, DocumentSegment, ChildChunk, Account, and Tenant models
- retain mocks only for vector indexing, Redis coordination, and asynchronous cleanup tasks
- consolidate SQL-shape assertions into persisted ownership, filtering, ordering, rollback, and mutation behavior

## Real persistence coverage
The suite verifies full tenant/dataset/document/segment ownership chains, keyword and status filtering, pagination, child-chunk position assignment and CRUD, vector-failure rollback, segment creation and error persistence, embedding-token updates, document word-count adjustments, scoped bulk deletion, delete-in-progress guards, and enable/disable transitions with decoy rows.

## Production changes
None.

## Size
- 537 additions, 1,016 deletions
- 1,553 changed lines

The module is slightly above the normal upper bound because its query, child-chunk, and segment mutation clusters share one interleaved file and cannot be split without overlapping independent PRs.

## Validation
- PASS: make test TARGET_TESTS=./api/tests/unit_tests/services/test_dataset_service_segment.py (36 passed)
- PASS: uv run --project api pytest -q api/tests/unit_tests/services/test_dataset_service_segment.py --no-cov (36 passed)
- PASS: make lint
- BLOCKED: make type-check — mypy 1.20.2 exits with an internal error in its bundled typeshed stdlib/zipimport.pyi:17; no project type diagnostic is emitted
- PASS: structural audit found no mocked SQLAlchemy sessions/factories, db.session chains, or mapped-model stand-ins in the migrated module

The full test suite was not run, per request.


41tair (changes requested):
Maybe we should check the boundary, seems a lots of case just deleted.

41tair (approved):

PR #40565, opened 2026-08-12
test: migrate console snippets sessions and ORM models to SQLite

asukaminato0721:
## Summary

- migrate console snippet workflow and draft-variable controller tests to SQLite-backed SQLAlchemy sessions
- replace mapped-model stand-ins with real Account, CustomizedSnippet, Workflow, and WorkflowDraftVariable instances
- verify persisted updates, deletes, tenant/user/node filtering, and rollback behavior with the real ORM
- keep workspace snippet endpoint tests in their dedicated controller PR to avoid overlapping files

## Production corrections

- inject request-owned sessions into snippet controllers instead of constructing sessions from `db.engine` or using `db.session`
- serialize snippet and workflow ORM properties through the request session
- keep translated validation failures transaction-safe and fix unreachable file-variable mapping code exposed during migration

## Size

382 additions, 322 deletions (704 changed lines).

## Validation

- targeted run covering both current test modules (executed before the overlapping workspace module was removed) — 52 passed
- `make lint` — passed
- `make type-check` — pyrefly passed; mypy 1.20.2 hit its existing internal error at `typeshed/stdlib/zipimport.pyi:17`
- structural audit — no mocked SQLAlchemy sessions/factories or mapped ORM stand-ins remain in the migrated tests

The full test suite was not run, per request.

### DIFF
--- a/api/controllers/console/snippets/snippet_workflow.py
+++ b/api/controllers/console/snippets/snippet_workflow.py
@@ -12,6 +12,7 @@ from controllers.common.controller_schemas import WorkflowUpdatePayload
 from controllers.common.fields import GeneratedAppResponse, SimpleResultResponse
 from controllers.common.rbac import RBACCheck, Workspace
 from controllers.common.schema import query_params_from_model, register_response_schema_models, register_schema_models
+from controllers.common.session import with_session
 from controllers.console import console_ns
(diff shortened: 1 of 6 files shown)


41tair (changes requested):
CI Failed.

41tair (approved):

PR #40584, opened 2026-08-12
fix(ops): keep the path of self-hosted LangSmith and Weave endpoints

GareArc:
## Summary

A self-hosted LangSmith endpoint containing a path — e.g. `https://langsmith.internal/api` — could not be configured. Saving appeared to succeed, but the endpoint was silently stored as `https://langsmith.internal`, and since the LangSmith SDK builds request URLs as `f"{api_url}/sessions"`, the connection check hit `https://langsmith.internal/sessions` and the reverse proxy answered `405`. Reported by a self-hosted enterprise user; no public issue was filed.

The cause is `core/ops/utils.validate_url()`, which rebuilds the URL from scheme + netloc only:

```python
normalized_url = f"{parsed.scheme}://{parsed.netloc}"
```

The path-preserving alternative `validate_url_with_path()` already existed — its own docstring has marked `validate_url()` deprecated for a while — and most trace providers were migrated to it. LangSmith and Weave were not, because they are HTTPS-only and the path-preserving helper had no way to express a scheme restriction.

Worth noting for anyone debugging a report of this shape: the validator runs on **decrypt as well as encrypt** (`ops_trace_manager.py:384` and `:428`, since `endpoint` is in `other_keys`), so the path was stripped on write *and* re-stripped on every read. Patching the stored row directly would not have worked around it.

### Changes

- `validate_url_with_path()` gains an `allowed_schemes` parameter, which is what LangSmith and Weave were missing. It is keyword-only: `str.endswith` accepts tuples, so a positional tuple would otherwise bind silently to `required_suffix` and fail later with a confusing `URL should end with ('https',)`.
- `LangSmithConfig.endpoint`, `WeaveConfig.endpoint` and `WeaveConfig.host` now use it. Weave's two fields keep their differing scheme policies (`endpoint` HTTPS-only, `host` either), passed explicitly rather than inherited from the default, since that asymmetry is deliberate provider policy.
- `validate_url_with_path()` now strips **surrounding whitespace** before parsing. This affects the five providers already using the helper and is a fix in its own right: `urlparse("  https://host/x  ")` reports a valid scheme, so a pasted trailing space was being persisted verbatim, and the frontend does not trim the field. It also means `trace-opik`'s `required_suffix` check now runs on the cleaned value, so `"https://host/api/ "` no longer fails validation.
  No **structural** normalization was added — trailing slashes, query strings and fragments are still preserved byte-for-byte, which `trace-opik`'s `required_suffix="/api/"` depends on and which the LangSmith SDK handles itself via `rstrip("/")`.

### Behavior changes

- Self-hosted LangSmith and Weave endpoints with a path now work. Previously unconfigurable unless served at the domain root.
- Scheme-validation error text changed for these two providers: `URL scheme must be one of: https` → `URL must start with https://`, and for Weave's `host`, `URL scheme must be one of: https, http` → `URL must start with https:// or http://`.

## Checklist


api/providers/trace/trace-langsmith/src/dify_trace_langsmith/config.py
```diff
@@ -16,5 +16,5 @@ class LangSmithConfig(BaseTracingConfig):
     @field_validator("endpoint")
     @classmethod
     def endpoint_validator(cls, v, info: ValidationInfo):
-        # LangSmith only allows HTTPS
-        return validate_url(v, "https://api.smith.langchain.com", allowed_schemes=("https",))
+        # LangSmith only allows HTTPS; self-hosted deployments may sit under a path prefix
+        return validate_url_with_path(v, "https://api.smith.langchain.com", allowed_schemes=("https",))
```
wylswz, line 20: Is `validate_url_with_path` without `required_suffix` equivalent to `validate_url`?

GareArc: `validate_url` rebuilds the url from two parsed components and ignore everything after the host, which is why this fix lands. For  example `https://langsmith.internal/api/v1` gets converted to `https://langsmith.internal`.

wylswz (approved):

PR #40676, opened 2026-08-13
feat(web): redesign the Marketplace home experience

CodingOnStar:
## Summary

Redesigns the Marketplace home experience and makes the discovery flow reusable across standalone and embedded contexts.

- Adds a branded Marketplace home with live search, autocomplete, trending content, sticky navigation, catalog tabs, and category filters.
- Adds template discovery alongside plugins, including template routes, collections, cards, search contracts, and detail dialogs.
- Unifies standalone and embedded Marketplace behavior for browsing, detail dialogs, plugin installation, headers, and the Creator Center entry.
- Integrates Marketplace Cloud authentication, silent OAuth authorization, and safe post-login redirects.
- Preserves server rendering and query hydration, improves cache behavior, defers offscreen content, and expands unit and E2E coverage.

## Checklist


web/app/(commonLayout)/marketplace/page.tsx
```diff
@@ -1,19 +1,30 @@
...
-const MarketplacePage = ({ searchParams }: MarketplacePageProps) => {
   return (
     <div
       id="marketplace-container"
-      className="flex h-full min-h-0 flex-col overflow-y-auto bg-background-default-subtle pr-1"
+      className="flex h-full min-h-0 flex-col overflow-y-auto bg-background-default"
     >
       <MarketplaceInstallPermissionProvider>
-        <Marketplace searchParams={searchParams} isMarketplacePlatform showInstallButton />
+        <EmbeddedMarketplace
```
lyzno1: [P1] Preserve Marketplace list hydration

This route now prefetches only banners and mounts `EmbeddedMarketplace` directly, so the collection/search query consumed by `ListWrapper` has no dehydrated data. On a direct request to `/marketplace` (including `?q=` or `?category=`), `useQuery` does not execute during server rendering and the first response contains only the loading spinner; the previous route passed `searchParams` through `Marketplace -> HydrateQueryClient`. Please keep the new home shell, but restore a server prefetch/HydrationBoundary for the list query using the request search params.

web/app/components/plugins/marketplace/detail-dialog/frame.tsx
```diff
@@ -0,0 +1,92 @@
...
+              <div className="h-4 w-5/6 animate-pulse rounded-md bg-state-base-hover motion-reduce:animate-none" />
+              <div className="mt-8 h-72 w-full animate-pulse rounded-xl bg-state-base-hover motion-reduce:animate-none" />
+            </div>
+            <div className="hidden w-60 flex-col gap-4 lg:flex">
+              <div className="h-24 animate-pulse rounded-xl bg-state-base-hover motion-reduce:animate-none" />
+              <div className="h-52 animate-pulse rounded-xl bg-state-base-hover motion-reduce:animate-none" />
+            </div>
+          </div>
+        </div>
+        <iframe
```
lyzno1: [P1] Keep the loading iframe out of the interaction tree

While `isLoading` is true, this iframe is only hidden with `opacity-0`. It remains focusable and pointer-active, and it precedes the close button, so opening the dialog on a slow connection can move focus into an invisible cross-origin frame and clicks on the skeleton can reach it. Please make the iframe inert and pointer-disabled until `onLoad`, and set an explicit visible initial-focus target such as the close control or dialog surface.

web/app/components/plugins/marketplace/home/marketplace-search-autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+          query: debouncedSearch,
+          sort_by: 'install_count',
+          sort_order: 'DESC',
+          category: category !== 'all' && !isBundleSearch ? category : '',
+        },
+      },
+      retry: false,
+    }),
+    enabled: hasQuery && searchesPlugins,
+    placeholderData: keepPreviousData,
```
lyzno1: [P2] Do not expose stale suggestions as selectable results

`keepPreviousData` intentionally returns the previous query's data while the new key is pending. Because this Autocomplete also uses `filter={null}`, typing a second term after the first response leaves the old term's items visible and keyboard-selectable until the new request finishes. Please remove `keepPreviousData` for suggestion queries, or pass no items while `isPlaceholderData` and show only the searching status.

lyzno1 (commented):
I found two blocking regressions and several reproducible interaction/test-contract issues in the current draft. The inline comments below are ordered by severity; please address the P1/P2 items before marking this ready for review.

web/app/components/plugins/marketplace/list/carousel.tsx
```diff
@@ -151,23 +211,111 @@ const Carousel = ({
...
         />
       )}
       <div ref={carouselRef} className="overflow-hidden rounded-[inherit]">
         <div className="flex" style={{ columnGap: '12px' }}>
-          {children}
+          {pages.map((page) => {
+            const isMounted = !deferMountPages || mountedPageIds.has(page.id)
+
+            return (
+              <div
```
lyzno1, line 367: [P2] Remove off-screen pages from keyboard and AT navigation

Every mounted page remains in the accessibility tree and tab order even when Embla has translated it out of view. In a multi-page template/plugin collection, Tab therefore moves into invisible card buttons/links and screen readers expose all pages as current content. Track `selectedIndex` here and mark non-current pages `inert`/`aria-hidden`; also give pages the slide role and accessible name.

web/app/components/plugins/marketplace/list/carousel.tsx
```diff
@@ -103,44 +110,97 @@ const CarouselControls = ({
...
-        stopOnInteraction: false,
-        stopOnMouseEnter: true,
-      }),
-    ]
-  }, [autoPlay, autoPlayInterval])
+  const carouselRootRef = useRef<HTMLDivElement>(null)
+  const autoplay = useMemo(() => {
+    if (!autoPlay) return undefined
+
+    return Autoplay({
```
lyzno1, line 148: [P2] Provide a persistent stop for collection rotation

This carousel rotates indefinitely every five seconds, but exposes only previous/next/page controls. Hover and reduced-motion handling do not give keyboard, touch, or assistive-technology users a persistent way to stop the auto-updating content. Please remove collection autoplay, or add a Pause/Play control and keep rotation stopped after focus enters until the user explicitly resumes it.

web/app/components/plugins/marketplace/home/home-trending.tsx
```diff
@@ -0,0 +1,649 @@
...
+
+    return () => {
+      progressAnimation.onfinish = null
+      progressAnimation.cancel()
+      if (progressAnimationRef.current === progressAnimation) progressAnimationRef.current = null
+    }
+  }, [onNext, selectedIndex])
+
+  useEffect(() => {
+    const carouselRoot = carouselRootRef.current
```
lyzno1: [P2] Pause when focus enters the whole carousel

The focus/hover listeners are attached to the content viewport, but `TrendingNavigation` is its sibling. Keeping focus on a pagination or Pause/Play button for five seconds therefore does not add the `focus` pause reason and the slide still changes during control interaction. Attach this ref/listener boundary to the enclosing carousel region so every control and slide participates.

web/app/components/plugins/marketplace/home/home-trending.tsx
```diff
@@ -0,0 +1,649 @@
...
+            onSelect={selectSlide}
+            onNext={selectNextSlide}
+          />
+          <div
+            ref={carouselRootRef}
+            className="h-full overflow-hidden rounded-2xl"
+            data-home-trending-carousel-root
+          >
+            <div
+              aria-live="polite"
```
lyzno1: [P2] Disable live announcements during autoplay

This live region stays `polite` while the carousel automatically changes `aria-hidden`/`inert` every five seconds, so screen-reader users can be repeatedly interrupted by unsolicited slide announcements. Use `aria-live="off"` while rotation is running and switch to `polite` only after it is paused or manually controlled; `web/features/home/banner/banner.tsx` already follows that contract.

web/app/components/plugins/marketplace/embedded.tsx
```diff
@@ -0,0 +1,40 @@
...
+      page: 'plugins' as const,
+      language: locale,
+    },
+  }
+  const { data: banners = [] } = useQuery(
+    queryOptions({
+      queryKey: [...marketplaceQuery.banners.list.queryKey({ input }), locale],
+      queryFn: () => fetchPluginBanners(locale),
+      enabled: variant === 'home',
+      initialData: initialBanners,
```
lyzno1: [P2] Scope initial banner data to its locale

`initialBanners` came from the locale used by the server render, but it is supplied to whichever `useLocale()` key is current. During the existing client-side language change, the component can observe the new locale before `router.refresh()`, seed that new key with old-language banners, and mark them fresh for five minutes, so no new-locale request runs. Pass the initial locale and only provide `initialData` when it matches, or hydrate a locale-keyed query.

web/app/components/plugins/marketplace/home/home-header.tsx
```diff
@@ -0,0 +1,120 @@
...
+  } catch {
+    return PUBLIC_CREATOR_CENTER_URL
+  }
+}
+
+const CreatorCenter = () => {
+  const creatorCenterUrl = getCreatorCenterUrl(MARKETPLACE_URL_PREFIX)
+
+  return (
+    <Link href={creatorCenterUrl} target="_blank" rel="noopener noreferrer">
```
lyzno1: [P2] Keep this navigation as a single anchor

`Link` wrapping Dify UI `Button` renders nested interactive elements (`<a><button>`), producing conflicting semantics and focus behavior. The Guide link uses the same pattern. Please render one `Link`/`<a>` and, only if it needs button visuals, apply `buttonVariants(...)` to that anchor rather than composing a Button element.

e2e/features/step-definitions/marketplace-performance.steps.ts
```diff
@@ -0,0 +1,97 @@
...
+      await cdpSession.send('Network.emulateNetworkConditions', {
+        connectionType: 'cellular4g',
+        downloadThroughput: FAST_4G_DOWNLOAD_BYTES_PER_SECOND,
+        latency: 60,
+        offline: false,
+        uploadThroughput: FAST_4G_UPLOAD_BYTES_PER_SECOND,
+      })
+      await cdpSession.send('Emulation.setCPUThrottlingRate', { rate: 4 })
+
+      await page.goto('/marketplace', { waitUntil: 'domcontentloaded' })
```
lyzno1, line 59: [P2] Give the benchmark a deterministic Marketplace data owner

This step waits for a Marketplace card, but the E2E runner does not set the frontend's `NEXT_PUBLIC_MARKETPLACE_API_PREFIX`. A clean checkout therefore falls back to `http://localhost:5002/api`, produces an empty list, and times out; a developer `.env.local` instead makes the metric depend on live Marketplace content and service latency. Please inject a frozen fixture endpoint through the runner and keep it in the same lifecycle as this scenario.

e2e/cucumber.config.ts
```diff
@@ -3,7 +3,7 @@ import './scripts/env-register'
 
 const hasCliTags = process.argv.some((arg) => arg === '--tags' || arg.startsWith('--tags='))
 const defaultNonExternalTags =
-  'not @axe and not @prepared and not @external-model and not @external-tool'
+  'not @axe and not @prepared and not @external-model and not @external-tool and not @marketplace-performance'
```
lyzno1, line 6: [P2] Connect this budget to an executable lane

The tag is excluded from both the default and full suites, and no workflow invokes the new `e2e:marketplace-performance` script. As written, a regression in all three budgets still leaves every required/default check green. Once the data source is deterministic, add a dedicated Chromium lane; otherwise describe this as an opt-in diagnostic rather than a protected budget.

web/app/styles/globals.css
```diff
@@ -3,6 +3,7 @@
 @source '../../app';
 @source '../../context';
 @source '../../hooks';
+@source '../../../../app';
```
lyzno1: [P3] Remove the ineffective source registration

`@source` paths are relative to this stylesheet, so this resolves four levels up to an `app` directory outside the repository; it does not include the new `web/app` code, which is already covered by `@source '../../app'`. The two inline icon safelists are also redundant because both complete class names occur statically in `home-trending.tsx`. Compiling the PR head with and without these three lines produced byte-for-byte identical CSS. Please remove them rather than leave a misleading scan boundary.

.github/workflows/main-ci.yml
```diff
@@ -393,6 +394,66 @@ jobs:
           echo "Web full-stack E2E was not required, but the skip job finished with result: $SKIP_RESULT" >&2
           exit 1
 
+  marketplace-performance-run:
+    name: Run Marketplace Performance E2E
+    needs:
+      - pre_job
+      - check-changes
+    if: needs.pre_job.outputs.should_skip != 'true' && needs.check-changes.outputs.e2e-changed == 'true'
```
lyzno1: [P1] Remove the Marketplace performance lane from main CI

This condition reuses the repository-wide `e2e-changed` signal, whose filter includes every `api/**`, `web/**`, `packages/**`, and `e2e/**` change. Consequently almost every product PR now starts a second 60-minute full-stack lifecycle on a dedicated runner, including middleware/API/Celery and a forced production Web rebuild, and unrelated PRs can be blocked by a single-sample Marketplace timing budget. Please remove this benchmark and its supporting E2E/CI plumbing from this feature PR. If the experiment is retained later, keep it as an opt-in diagnostic rather than another required main-CI gate.

web/service/marketplace-template-discovery.ts
```diff
@@ -0,0 +1,85 @@
...
+  category,
+  query,
+  sortBy = 'usage_count',
+  sortOrder = 'DESC',
+}: SearchMarketplaceTemplatesOptions) {
+  try {
+    const response = await marketplaceClient.templateSearch({
+      body: {
+        page: 1,
+        page_size: 40,
```
lyzno1: [P1] Make template search results pageable

This fetch is permanently pinned to the first 40 records, while the page displays `searchResult.total` and exposes neither pagination nor infinite loading. Whenever `total > 40` (the current Marketplace response reports 286), every later template is unreachable, and filtering only this first page by locale can hide matching templates that exist later. Make the page or cursor part of route/query state and provide a way to fetch subsequent results.

web/app/components/plugins/marketplace/templates/template-language.ts
```diff
@@ -0,0 +1,40 @@
...
+  T extends Pick<MarketplaceTemplate, 'preferred_languages'>,
+>(templates: T[], locale: string) {
+  const languageFamily = getTemplateLanguageFamily(locale)
+
+  return templates.filter((template) => {
+    const preferredLanguages = (template.preferred_languages ?? []).map((language) =>
+      language.toLowerCase(),
+    )
+
+    if (languageFamily === 'other') {
```
lyzno1: [P1] Do not collapse unrelated locales into `other`

For every supported locale outside English, Chinese, and Japanese, this branch rejects all `en`/`zh`/`ja` templates and accepts every remaining language indiscriminately. For example, the current marketing response has 57 results but its first 40 are all English, Japanese, or Chinese, so a German `/templates/marketing` renders an empty state while still reporting 57 results. Match the normalized requested locale and define an explicit fallback, or move language filtering into the Marketplace query.

web/service/marketplace-template-discovery.ts
```diff
@@ -0,0 +1,85 @@
...
+export async function getMarketplaceTemplateCollectionsAndTemplates(): Promise<MarketplaceTemplateCollectionsResult> {
+  try {
+    const response = await marketplaceClient.templateCollections({
+      query: {
+        page: 1,
+        page_size: 100,
+      },
+    })
+    const collections = response.data?.collections ?? []
+    const entries = await Promise.all(
```
lyzno1: [P2] Avoid an uncached request per template collection

The default `/templates` render first loads up to 100 collections and then issues one POST per collection, each requesting up to 100 templates. `marketplaceClient` uses `cache: 'no-store'`, so every server render/navigation can fan out to as many as 101 external requests before the page is available. Please provide a single Marketplace response for collection previews, or introduce a bounded cache-owned fetch instead of performing this N+1 fan-out in the Server Component path.

lyzno1 (commented):
Reviewed the complete branch diff (`e3b3165e` → `f033532d`), not only the latest follow-up commit. The latest revision fixes several earlier issues, including server hydration, iframe focus isolation, single-anchor button styling, and the ineffective Tailwind sources. The branch still has blocking correctness and accessibility regressions in the new Templates/Marketplace flows.

The Marketplace performance benchmark should also be removed from this feature PR, including its E2E feature/steps, fixture stub, runner plumbing, reusable workflow, and main-CI run/skip/finalizer jobs. A single page-specific timing sample is not an appropriate repository-wide full-stack gate, especially when it consumes a second dedicated E2E runner for unrelated product changes.

web/app/(commonLayout)/templates/[[...category]]/page.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <div
+      id="marketplace-container"
+      className="flex h-full min-h-0 flex-col overflow-y-auto bg-background-default"
+    >
+      <EmbeddedTemplatesMarketplace
+        category={category}
+        locale={locale}
+        query={resolvedSearchParams.q ?? ''}
+        sortBy={resolvedSearchParams.sort_by}
```
Comment (lyzno1) [inline web/app/(commonLayout)/templates/[[...category]]/page.tsx]: [P2] Validate template view and sort parameters at the route boundary

These values come directly from a public URL and are forwarded unchanged. An unknown `view` already switches the page away from collections, and `?view=search&sort_order=garbage` makes the current Marketplace endpoint fail; the service then converts that failure into an empty result, so the UI falsely says no templates exist. Parse `view`, `sort_by`, and `sort_order` as supported enums with defaults before calling the service.

web/app/components/plugins/marketplace/home/home-catalog-tabs.tsx
```diff
@@ -0,0 +1,77 @@
...
+  language,
+}: HomeCatalogTabsProps) => {
+  const { t } = useTranslation()
+  const catalogParams = language ? { language } : undefined
+  const getRelativeCatalogHref = (path: string) => {
+    const searchParams = new URLSearchParams(catalogParams)
+    const queryString = searchParams.toString()
+    return queryString ? `${path}?${queryString}` : path
+  }
+  const pluginsHref = isMarketplacePlatform
```
lyzno1, line 30: [P2] Reset or serialize plugin search mode across catalog routes

The Plugins link returns to `/marketplace`, but `searchModeAtom` lives in the app-wide Jotai store and is neither reset nor represented in this URL. Repro: click View More on the current Featured or Latest collection (both have an empty query), switch to Templates, then return to Plugins; the server prefetches collections for `/marketplace`, but the client still renders empty-query search results until a full refresh. Make view/sort route-owned, or reset the plugin-only atom at the route boundary.

web/app/components/plugins/marketplace/home/marketplace-search-autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+  })
+  const pluginSuggestions = searchesPlugins
+    ? (pluginQuery.data?.data.bundles ?? pluginQuery.data?.data.plugins ?? []).map((plugin) =>
+        toPluginSuggestion(plugin, locale),
+      )
+    : []
+  const templateSuggestions = searchesTemplates
+    ? (templateQuery.data?.data?.templates ?? []).map(toTemplateSuggestion)
+    : []
+  const suggestions = [...templateSuggestions, ...pluginSuggestions]
```
lyzno1, line 142: [P2] Clear suggestions while the new value is still debouncing

Removing `placeholderData` only clears the previous query after `debouncedSearch` changes. For the first 300 ms after editing a populated search, `debouncedSearch` still points at the old term, so these arrays still expose the old query data; with `filter={null}`, those stale options remain visible and keyboard-selectable. Gate suggestions on `value.trim() === debouncedSearch` (and show the searching state while they differ), or clear the items immediately on input change. The new test mocks `useDebounce` to an identity function, so it cannot catch this window.

web/app/components/plugins/marketplace/home/home-trending.tsx
```diff
@@ -0,0 +1,664 @@
...
+      setPauseReason('reduced-motion', reducedMotionQuery.matches)
+    }
+
+    syncReducedMotion()
+    reducedMotionQuery.addEventListener('change', syncReducedMotion)
+
+    return () => reducedMotionQuery.removeEventListener('change', syncReducedMotion)
+  }, [setPauseReason])
+
+  const toggleAutoplay = () => {
```
lyzno1: [P2] Keep the Play control in sync with the actual pause reasons

When focus is inside the carousel, the `focus` pause reason remains set. Activating Play clears only `user` and `reduced-motion`, so the animation stays paused while `isExplicitlyPaused` becomes false and the control changes back to a Pause label. Keyboard repro: focus the control, activate Pause, then activate Play without moving focus; no slide advances until focus leaves even though the control reports the playing state. Explicit Play must override the focus pause, or the control state must be derived from all active pause reasons.

web/app/components/plugins/marketplace/list/carousel.tsx
```diff
@@ -98,49 +115,128 @@ const CarouselControls = ({
...
-        stopOnMouseEnter: true,
-      }),
-    ]
-  }, [autoPlay, autoPlayInterval])
+  const carouselRootRef = useRef<HTMLDivElement>(null)
+  const [isUserPaused, setIsUserPaused] = useState(false)
+  const autoplay = useMemo(() => {
+    if (!autoPlay) return undefined
+
+    return Autoplay({
```
lyzno1, line 148: [P2] Honor reduced motion for the first collection

The `prefers-reduced-motion` guard is installed only in the `pauseWhenOffscreen` effect. `ListWithCollection` passes `pauseWhenOffscreen={false}` for its first server-rendered collection, so that instance reaches `playOnInit: true` and rotates even when the OS requests reduced motion. Apply the media-query guard independently of the offscreen optimization and cover the production first-collection configuration.

web/app/components/plugins/marketplace/home/home-header.tsx
```diff
@@ -0,0 +1,123 @@
...
+  } catch {
+    return PUBLIC_CREATOR_CENTER_URL
+  }
+}
+
+const CreatorCenter = () => {
+  const creatorCenterUrl = getCreatorCenterUrl(MARKETPLACE_URL_PREFIX)
+
+  return (
+    <Link
```
lyzno1: [P2] Give the responsive Creator Center link an accessible, localized name

Below the `lg` breakpoint, the text is removed with `display: none` and the remaining icon is `aria-hidden`, leaving this icon-only link unnamed. At larger breakpoints the same label is hard-coded English. Add a localized `aria-label` to the Link and use that translation for the visible span as well. The CSS-less unit test currently derives a name from the `hidden` span, so it does not reproduce the real breakpoint.

web/app/components/plugins/marketplace/templates/index.tsx
```diff
@@ -0,0 +1,253 @@
...
+            catalogCategories={
+              <TemplateCategoryNavigation
+                activeCategory={category}
+                ariaLabel={tPlugin('allCategories' as never)}
+                labels={categoryLabels}
+                query={query}
+              />
+            }
+          />
+          <main
```
lyzno1: [P2] Do not nest another main landmark here

This route is already rendered inside `MainNavLayout`'s `<main id={MAIN_CONTENT_ID}>`, so this creates a second `main` nested inside the page's existing main landmark. That produces invalid landmark structure and a duplicate main region for assistive technology. Use a `div` or named `section`; the common layout owns the page-level main and skip-link target.

web/app/components/plugins/marketplace/home/home-trending.tsx
```diff
@@ -0,0 +1,664 @@
...
+
+function BlogBannerSlide({ banner }: { banner: BannerBlog }) {
+  const opensInNewTab = /^https?:\/\//.test(banner.content.link)
+
+  return (
+    <Link
+      href={banner.content.link}
+      target={opensInNewTab ? '_blank' : undefined}
+      rel={opensInNewTab ? 'noopener noreferrer' : undefined}
+      aria-label={`Read more about ${banner.content.blog_title}`}
```
lyzno1: [P2] Complete localization for the new banner actions

Both `Read more about …` and the visible `Read more` label are hard-coded English. In addition, `marketplace.home.trendingView`, `trendingPause`, and `trendingPlay` exist in only 3 of the 24 supported `plugin.json` locales. Please add feature-local keys for the blog action and provide the new action/control keys in every supported locale so non-English Marketplace pages and accessible names do not mix languages.

.github/workflows/main-ci.yml
CodingOnStar: Fixed in e7ca2735bf. Removed the three `marketplace-performance-*` jobs from `main-ci.yml` (and the workflow file from the e2e path filter), and switched `marketplace-performance-e2e.yml` to `workflow_dispatch` so the single-sample benchmark is an opt-in diagnostic lane instead of a required PR gate.

web/service/marketplace-template-discovery.ts
CodingOnStar: Fixed in e7ca2735bf. `searchMarketplaceTemplates` now accepts a `page` parameter (page size stays 40 via `TEMPLATE_SEARCH_PAGE_SIZE`) and returns `page` alongside `total`. The route parses `?page=` (invalid values fall back to 1) and the search view renders SSR-friendly `<Link>`-based previous/next pagination computed from `total`, so results beyond the first 40 are reachable.

web/app/components/plugins/marketplace/templates/template-language.ts
CodingOnStar: Fixed in e7ca2735bf. `filterTemplatesForLocale` now matches `preferred_languages` by the normalized language prefix of the requested locale (any locale, not just en/zh/ja), keeps templates without language metadata as language-agnostic, and explicitly falls back to English templates (and finally to the unfiltered list) when nothing matches — so e.g. `de-DE` no longer renders an empty state. Added unit tests covering the matching and both fallbacks.

web/service/marketplace-template-discovery.ts
CodingOnStar: Fixed in e7ca2735bf. `getMarketplaceTemplateCollectionsAndTemplates` is now wrapped in a module-level cache with a 5-minute TTL plus in-flight dedupe (concurrent renders share one fetch; failures are not cached), the per-collection `limit` dropped from 100 to 24 (the carousel preview needs far less than 100), and the fan-out runs in batches of 5 instead of firing one request per collection at once. Added tests for the caching/dedupe behavior.

web/app/(commonLayout)/templates/[[...category]]/page.tsx
Comment (CodingOnStar) [reply web/app/(commonLayout)/templates/[[...category]]/page.tsx]: Fixed in e7ca2735bf. The route now white-lists these params before passing them down: `view` (only `search`), `sort_by` (`usage_count`/`created_at`), `sort_order` (`ASC`/`DESC`) and `page` (positive integer); anything else falls back to the defaults so unsupported values never reach the Marketplace API. Added route tests for both the valid and fallback paths.

web/app/components/plugins/marketplace/home/home-catalog-tabs.tsx
CodingOnStar: Fixed in e7ca2735bf. `EmbeddedMarketplace` now resets `searchModeAtom` to `null` on mount (`useResetMarketplaceSearchModeOnMount`), so navigating View More → Templates → back to `/marketplace` shows the collections again instead of an empty-query search view. URL-owned state (q/tags/category) is unaffected, and same-route View More clicks still work since the component does not remount.

web/app/components/plugins/marketplace/home/marketplace-search-autocomplete.tsx
CodingOnStar: Fixed in e7ca2735bf. Added a `value.trim() === debouncedSearch` gate: while the edited value is still inside the debounce window the suggestion lists are emptied and `isSearching` is forced on, so stale options are never visible or keyboard-selectable during the 300ms lag. Added a test that exercises the real debounce (no identity mock) to cover this window.

web/app/components/plugins/marketplace/home/home-trending.tsx
CodingOnStar: Fixed in e7ca2735bf. The Play branch of `toggleAutoplay` now also clears the implicit `focus` and `hover` pause reasons, so an explicit Play resumes rotation even though the toggle itself is still focused/hovered; those reasons re-engage on the next `focusin`/`mouseenter`. Added a keyboard Pause → Play test asserting the rotation resumes without moving focus.

web/app/components/plugins/marketplace/list/carousel.tsx
CodingOnStar: Fixed in e7ca2735bf. Reduced motion is now tracked by a dedicated effect for the eager autoplay path (`pauseWhenOffscreen=false`, i.e. the production first collection), seeded from `matchMedia` on mount, and the eager play/stop effect honors it; the viewport-managed path keeps its existing handling. Added a test that the first-collection configuration stays stopped under `prefers-reduced-motion` and resumes when the preference is lifted.

web/app/components/plugins/marketplace/home/home-header.tsx
CodingOnStar: Fixed in e7ca2735bf. The Creator Center link now gets an explicit `aria-label` and its visible text from the new `marketplace.home.creatorCenter` key (translated in all 24 locales), so it keeps an accessible, localized name below the `lg` breakpoint where the text is hidden.

web/app/components/plugins/marketplace/templates/index.tsx
CodingOnStar: Fixed in e7ca2735bf. Replaced this `<main>` with a `<div>` (the app shell already provides the main landmark). Also audited the plugin home tree — no other nested `main` landmarks in the marketplace components.

web/app/components/plugins/marketplace/home/home-trending.tsx
CodingOnStar: Fixed in e7ca2735bf. The blog banner now uses `marketplace.home.trendingReadMoreAbout` (with `{{title}}` interpolation) for the accessible name and `marketplace.home.trendingReadMore` for the visible action, and I backfilled `trendingView`/`trendingPause`/`trendingPlay`/`creatorCenter` plus the two new keys across all 24 locales (previously only en-US/zh-Hans/zh-Hant had the trending keys).

web/app/components/plugins/marketplace/home/marketplace-search-autocomplete.tsx
```diff
@@ -0,0 +1,286 @@
...
+        )
+      : []
+  const templateSuggestions =
+    !isDebouncing && searchesTemplates
+      ? (templateQuery.data?.data?.templates ?? []).map(toTemplateSuggestion)
+      : []
+  const suggestions = [...templateSuggestions, ...pluginSuggestions]
+  const isSearching = isDebouncing || pluginQuery.isFetching || templateQuery.isFetching
+  // Open only when there is something to show; do not flash a searching status.
+  const showDropdown = isOpen && hasQuery && (!isSearching || suggestions.length > 0)
```
lyzno1, line 145: [P2] Keep async result visibility out of the controlled open state

`isOpen` records Base UI's requested interaction state, but `showDropdown` forces the controlled `open` prop back to `false` throughout the debounce/request window. Repro: type a query, click outside before the request resolves, then wait for the response. Because the popup was controlled closed, Base UI had no mounted focus manager/outside-press dismissal to emit `onOpenChange(false)`, so `isOpen` remains `true` and the popup can reopen after the user has left the field. Keep the Root open state tied to the interaction session (and the non-empty input), and represent the pending state inside the popup. The existing empty `suggestions` gate can still prevent stale options from being exposed while a request is pending.

web/app/components/plugins/marketplace/home/marketplace-search-autocomplete.tsx
```diff
@@ -0,0 +1,286 @@
...
+                  <span className="truncate system-xs-regular text-text-quaternary">
+                    {item.meta}
+                  </span>
+                )}
+              </span>
+              <AutocompleteItemIndicator />
+            </AutocompleteItem>
+          )}
+        </AutocompleteList>
+        {!isSearching && <AutocompleteEmpty>{emptyText}</AutocompleteEmpty>}
```
lyzno1, line 243: [P2] Keep the async live region mounted across state changes

Base UI explicitly requires the `Empty`/`Status` root to remain mounted for reliable screen-reader announcements; only its children should change. Here `AutocompleteEmpty` is conditionally mounted after searching, while the popup (and therefore the only `aria-busy` node) is absent for the entire pending state. This also means the tests currently codify that no status exists while results are loading. Please follow the Base UI async pattern: keep an `AutocompleteStatus` (or the `Empty` root where appropriate) mounted, update its children for searching/error/empty/result states, and keep the popup open during the active search so `aria-busy` is actually exposed.

lyzno1 (commented):
Reviewed the new Marketplace Autocomplete against Base UI's controlled-open contract and async Autocomplete guidance. I found two async interaction/accessibility issues below.

PR #40679, opened 2026-08-13
fix: hide knowledge retrieval from Agent V2 configuration

iamjoel:
## Summary

- Hide unsupported knowledge retrieval controls from roster and inline Agent V2 configuration panels behind a temporary feature flag.
- Remove the Knowledge category from the prompt editor slash menu and add regression coverage for the hidden entry.

Fixes DIFY-2884

## Checklist


lyzno1 (approved):

yasu-oh:
@iamjoel 
Hi, thanks for the work on Agent V2!

I noticed that Knowledge Retrieval is still disabled behind `ENABLE_AGENT_KNOWLEDGE_RETRIEVAL`, even though much of the runtime support appears to already be implemented.

Is there currently a plan to re-enable Knowledge Retrieval for Agent V2?

If so, is there any rough roadmap or are there specific blockers that still need to be resolved before this feature can be enabled again?

Knowledge Base integration is an important use case for us, so any insight into the current status would be greatly appreciated. Thanks!

Issue #37827, opened 2026-06-23
too many patch in test


asukaminato0721:
there are sooo many patch in the unit test / integration test, I think it's not good.

@dosu any idea?

asukaminato0721:
in test, build "real" fake outer api to avoid mock and patch

@dosu is this reasonable?

and I am really consider using sqlite for some query, since only part of the api need pg extension.

PR #40689, opened 2026-08-13
test(vdb): use real sessions and SDK clients

asukaminato0721:
## Summary

- replace the shared VDB integration Redis mock with a real Redis client, isolated by a unique per-test key prefix and cleaned up without `flushdb`
- exercise real SDK clients, domain models, response parsing, and query construction for Baidu, Tencent, Huawei Cloud, Upstash, Hologres, and VikingDB while keeping deterministic substitutes only at unavailable network/SQL transport boundaries
- keep `MOCK_SWITCH` as the credential-free cloud test mode and preserve Docker-free VDB unit tests
- serialize the Baidu primary-key field as a string so the real Mochow SDK can encode the query payload

This PR is intentionally independent of #40687.

Related to #37827.

## Verification

- `make test TARGET_TESTS='./api/providers/vdb/*/tests/unit_tests'` — 598 passed
- `make lint` — passed
- Upstash integration test in offline SDK mode — passed
- real-SDK offline harnesses for Baidu, Tencent, Huawei Cloud, Hologres, and VikingDB — passed
- Redis fixture lifecycle harness — passed (connection ping, prefixed key access, scoped cleanup, close, and prior-client restoration)

The Docker-backed cloud and Chroma/PGVector/Qdrant/Weaviate integration commands were attempted locally, but pytest could not start the services because this environment has no `docker` executable. They remain for CI validation. Type checking was intentionally not run for this change.


41tair (changes requested):
We should expose redis port in docker/docker-compose.pytest.ports.yaml.

PR #40792, opened 2026-08-14
fix: reclaim conversation resources on deletion

zyssyz123:
## Summary\n\n- soft-delete conversations before asynchronous cleanup and add a periodic sweeper for missed task dispatches\n- remove conversation-owned object storage and related rows while preserving ToolFiles referenced by Agent Drive\n- serialize cleanup with Drive commits and transfer lifecycle ownership to surviving Drive references\n- add cleanup configuration and a partial index for deleted-conversation scans\n\nApp deletion is intentionally out of scope.\n\n## Verification\n\n- 72 related unit tests passed\n- 31 PostgreSQL container integration tests passed, including a real ToolFile lock race\n- Ruff, Mypy, and Pyrefly passed for the changed backend paths\n- Alembic has a single head on current main\n- dev Service API E2E passed: list/delete (204), immediate hide (404 on repeat), async physical cleanup\n- dev Drive preservation, shared-to-owned transfer, storage cleanup, and sweeper fallback paths passed\n\nLinear: DIFY-2882

### DIFF
--- a/api/.env.example
+++ b/api/.env.example
@@ -829,6 +829,12 @@ ENABLE_HUMAN_INPUT_TIMEOUT_TASK=true
 # Human input timeout check interval in minutes
 HUMAN_INPUT_TIMEOUT_TASK_INTERVAL=1
 
+# Whether to recover soft-deleted conversation cleanup jobs periodically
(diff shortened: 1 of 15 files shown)


MRZHUH (approved):
LGTM

PR #40803, opened 2026-08-14
fix(api): bound connect phase to 3.0s in auth credential validation clients

### LINKED ISSUE #40802: fix(api): add explicit connect timeout to auth credential validation clients
## Problem

The three outbound credential-validation clients (`api/services/auth/{watercrawl,firecrawl,jina/jina}.py`) all use a plain `httpx.Timeout(10.0)` for their validation requests. That sets the overall request timeout to 10s but leaves the **connect** phase using httpx's default 5s. Because the connect phase and the read phase share the same 10s budget, a slow or unreachable remote endpoint can keep the worker waiting up to 10s per credential check before failing.

The same `api/services/` directory already has a tighter pattern in `operation_service.py`:

```python
OPERATION_REQUEST_TIMEOUT = httpx.Timeout(10.0, connect=3.0)
```

…where the connect phase is bounded at 3s and the read phase is 10s. The auth modules should follow the same shape so an unreachable endpoint fails fast and the worker is freed sooner.

## Motivation

`#37637` (cycle-6 sibling) added explicit read timeouts to the auth validation requests; the connect phase was left as httpx's default 5s, and the read timeout was set to a single 10s value with no `connect=` argument. Several other outbound modules in the same directory (`services/operation_service.py`, `services/rag_pipeline/...`, `services/recommend_app/...`) already use the `(read, connect)` split. Aligning the auth validation clients with the same shape is a small, low-risk consistency fix.

Faster failure also reduces the window for an attacker to keep a credential-validation request in flight, though the primary motivation is consistency and faster failure under network degradation.

## Current behavior

`httpx.Timeout(10.0)`:
- `connect` → httpx default (5s)
- `read` → 10s
- `write` → 10s
- `pool` → 10s

When the remote endpoint is unreachable (TCP RST or silent drop), the request blocks for up to 10s before the overall timeout fires. If the connect phase consumes most of the budget, the read never gets a chance to time out cleanly.

## Expected behavior

`httpx.Timeout(10.0, connect=3.0)`:
- `connect` → 3s (bounded)
- `read` → 10s
- `write` → 10s
- `pool` → 10s

Unreachable endpoints fail in ≤3s instead of up to 10s; the overall timeout still bounds the total request to 10s.

## Affected files

| File | Current | Target |
|------|---------|--------|
| `api/services/auth/watercrawl/watercrawl.py:11` | `httpx.Timeout(10.0)` | `httpx.Timeout(10.0, connect=3.0)` |
| `api/services/auth/firecrawl/firecrawl.py:10` | `httpx.Timeout(10.0)` | `httpx.Timeout(10.0, connect=3.0)` |
| `api/services/auth/jina/jina.py:12` | `httpx.Timeout(10.0)` | `httpx.Timeout(10.0, connect=3.0)` |

The pooled httpx clients in `jina.py` and `jina/jina.py` are constructed with `httpx.Client(timeout=...)` at import time, so the connect timeout is set once on the client. The `_CREDENTIAL_TIMEOUT` constants in watercrawl and firecrawl are passed directly to `httpx.get`/`httpx.post` calls, so the same value covers both code paths.

## Out of scope (separate concerns)

- `api/services/auth/jina.py` (the orphan file, not the package). It's been kept in sync with `jina/jina.py` for several releases but is shadowed by the package at the same name. The timeout fix in this issue should not touch it; deleting the orphan is a separate cleanup tracked under a follow-up.
- The wider `connect=` gap across `api/core/` modules (e.g. `api/core/rag/extractor/watercrawl/client.py:235` uses a bare `timeout=30`; `api/core/helper/creators.py:23` same). Those are in different code areas and are better handled as their own focused fix.

## Proposed solution

Single-commit, three-file change:

```python
# api/services/auth/watercrawl/watercrawl.py
- _CREDENTIAL_TIMEOUT = httpx.Timeout(10.0)
+ _CREDENTIAL_TIMEOUT = httpx.Timeout(10.0, connect=3.0)
```

```python
# api/services/auth/firecrawl/firecrawl.py
- _CREDENTIAL_TIMEOUT = httpx.Timeout(10.0)
+ _CREDENTIAL_TIMEOUT = httpx.Timeout(10.0, connect=3.0)
```

```python
# api/services/auth/jina/jina.py
- timeout=httpx.Timeout(10.0),
+ timeout=httpx.Timeout(10.0, connect=3.0),
```

Plus tests:
- `test_watercrawl_auth.py` — assert `_CREDENTIAL_TIMEOUT.connect == 3.0` and that `_get_request` is called with the new timeout.
- `test_firecrawl_auth.py` — same shape.
- `test_jina_auth.py` — assert the pooled `_http_client.timeout.connect == 3.0`.

The existing `test_jina_auth_standalone_module.py` is the orphan's test and is unaffected.

## Alternatives considered

- **Set both read and connect to 5s** (`httpx.Timeout(5.0, connect=3.0)`). Rejected: 5s read is a tighter bound than the 10s the existing fix at #37637 set, which would be a behavior change for callers that observe success latency.
- **Add a global default** via `app_factory.py` or similar. Rejected: the project already prefers per-module constants (`OPERATION_REQUEST_TIMEOUT`, `_CREDENTIAL_TIMEOUT`); adding a global would be a bigger refactor and out of scope for this fix.
- **Use `httpx.Timeout(connect=3.0)` only** (drop the read timeout). Rejected: leaves the request unbounded on read, which the existing fix at #37637 was specifically trying to bound.

## Acceptance criteria

- [ ] All three files use `httpx.Timeout(10.0, connect=3.0)`.
- [ ] Existing tests in `test_watercrawl_auth.py`, `test_firecrawl_auth.py`, `test_jina_auth.py` continue to pass.
- [ ] New tests assert the connect timeout is 3.0 in each affected client.
- [ ] No other source files change.
- [ ] Backward compatible: any existing caller passing a custom `timeout` to the auth methods is unaffected (none currently do).

## Backward compatibility

Public API of the `JinaAuth`, `FirecrawlAuth`, `WatercrawlAuth` classes is unchanged. The internal timeout constant changes from 10.0s to (10.0s read, 3.0s connect), so callers that observe failure latency on unreachable endpoints will see it drop from up to 10s to ≤3s. Callers that observe success latency are unaffected.

## Risks

- The 3s connect timeout is shorter than httpx's 5s default. If any of the three remote providers has a real connect latency > 3s under normal operation, the validation will start failing. This is unlikely for these well-known APIs (Jina, Firecrawl, WaterCrawl) but is the only behavior change.

Harsh23Kashyap:
## Problem

The three outbound credential-validation clients in `api/services/auth/` (`watercrawl`, `firecrawl`, `jina/jina`) all use a plain `httpx.Timeout(10.0)` for their validation requests. That leaves the **connect** phase on httpx's 5s default and the read/write/pool phases at 10s. A slow or unreachable remote endpoint can keep the worker waiting up to 10s per credential check before failing.

The same `api/services/` directory already has the tighter pattern:

```python
# api/services/operation_service.py:6
OPERATION_REQUEST_TIMEOUT = httpx.Timeout(10.0, connect=3.0)
```

## Fix

Add an explicit `connect=3.0` to all three auth client timeouts so unreachable endpoints fail in ≤3s instead of up to 10s, matching the shape already used elsewhere in `services/`.

```python
# before
_CREDENTIAL_TIMEOUT = httpx.Timeout(10.0)
# after
_CREDENTIAL_TIMEOUT = httpx.Timeout(10.0, connect=3.0)
```

```python
# before
timeout=httpx.Timeout(10.0),
# after
timeout=httpx.Timeout(10.0, connect=3.0),
```

## Scope

3 source files (one-line each), 3 test files (new assertions + update two existing test calls that hardcoded the old timeout):

| File | Change |
|------|--------|
| `api/services/auth/watercrawl/watercrawl.py` | `httpx.Timeout(10.0)` → `httpx.Timeout(10.0, connect=3.0)` |
| `api/services/auth/firecrawl/firecrawl.py` | same |
| `api/services/auth/jina/jina.py` | same, on pooled client constructor |
| `api/tests/unit_tests/services/auth/test_watercrawl_auth.py` | update existing assertion + add 2 new tests |
| `api/tests/unit_tests/services/auth/test_firecrawl_auth.py` | update existing assertion + add 2 new tests |
| `api/tests/unit_tests/services/auth/test_jina_auth.py` | add 1 new test |

Fixes #40802

## Out of scope (tracked separately)

- `api/services/auth/jina.py` is the orphan file shadowed by the `jina/` package. It still uses the old `httpx.Timeout(10.0)`. Deleting the orphan (or unifying the two) is a separate cleanup that touches test fixture and the duplicate-file situation.
- The wider `connect=` gap across `api/core/` (e.g. `api/core/rag/extractor/watercrawl/client.py:235` and `api/core/helper/creators.py:23` use a bare integer `timeout=30` without a connect timeout) is a different code area and a separate fix.

## Testing

- 69 tests in `test_watercrawl_auth.py`, `test_firecrawl_auth.py`, `test_jina_auth.py` pass locally (`--noconftest`).
- 4 new tests added (2 for watercrawl, 2 for firecrawl, 1 for jina; 1 is parametrized equivalent in firecrawl). One existing test in watercrawl and one in firecrawl that hardcoded the old timeout are updated to assert the new one.
- `ruff check` and `ruff format --check` clean on `services/auth/` and the corresponding test directory.
- `pyrefly check services/auth/` reports 0 diagnostics.

## Risks

The connect timeout is 3.0s versus httpx's 5.0s default. If any of the three remote providers has a real connect latency > 3s under normal operation, the validation will start failing. This is unlikely for the well-known Jina, Firecrawl, and WaterCrawl APIs, and the existing read=10s budget still bounds the total request.

## Backward compatibility

Public API of `JinaAuth`, `FirecrawlAuth`, `WatercrawlAuth` is unchanged. Only the internal timeout shape changes.

### DIFF
--- a/api/services/auth/firecrawl/firecrawl.py
+++ b/api/services/auth/firecrawl/firecrawl.py
@@ -7,7 +7,7 @@ from services.auth.api_key_auth_base import ApiKeyAuthBase, AuthCredentials
 
 # Explicit bounded timeout for credential-validation requests so a slow or
 # hanging Firecrawl endpoint cannot block the worker indefinitely.
-_CREDENTIAL_TIMEOUT = httpx.Timeout(10.0)
+_CREDENTIAL_TIMEOUT = httpx.Timeout(10.0, connect=3.0)
 
 
 class FirecrawlAuth(ApiKeyAuthBase):
--- a/api/services/auth/jina/jina.py
+++ b/api/services/auth/jina/jina.py
@@ -9,7 +9,7 @@ from services.auth.api_key_auth_base import ApiKeyAuthBase, AuthCredentials
 _http_client: httpx.Client = get_pooled_http_client(
     "auth:jina",
     lambda: httpx.Client(
-        timeout=httpx.Timeout(10.0),
+        timeout=httpx.Timeout(10.0, connect=3.0),
         limits=httpx.Limits(max_keepalive_connections=50, max_connections=100),
     ),
 )
--- a/api/services/auth/watercrawl/watercrawl.py
+++ b/api/services/auth/watercrawl/watercrawl.py
@@ -8,7 +8,7 @@ from services.auth.api_key_auth_base import ApiKeyAuthBase, AuthCredentials
 
 # Explicit bounded timeout for credential-validation requests so a slow or
 # hanging WaterCrawl endpoint cannot block the worker indefinitely.
-_CREDENTIAL_TIMEOUT = httpx.Timeout(10.0)
+_CREDENTIAL_TIMEOUT = httpx.Timeout(10.0, connect=3.0)
 
 
 class WatercrawlAuth(ApiKeyAuthBase):
--- a/api/tests/unit_tests/services/auth/test_firecrawl_auth.py
+++ b/api/tests/unit_tests/services/auth/test_firecrawl_auth.py
@@ -86,7 +86,7 @@ class TestFirecrawlAuth:
             "https://api.firecrawl.dev/v1/crawl",
             headers={"Content-Type": "application/json", "Authorization": "Bearer test_api_key_123"},
             json=expected_data,
-            timeout=httpx.Timeout(10.0),
+            timeout=httpx.Timeout(10.0, connect=3.0),
         )
 
     @pytest.mark.parametrize(
@@ -206,3 +206,24 @@ class TestFirecrawlAuth:
 
         # Verify the timeout exception is raised with original message
         assert "timed out" in str(exc_info.value)
+
+    def test_credential_timeout_bounds_connect_phase(self):
+        """Credential-validation timeout should bound the connect phase to 3.0s."""
+        from services.auth.firecrawl.firecrawl import _CREDENTIAL_TIMEOUT
+
+        assert _CREDENTIAL_TIMEOUT.connect == 3.0
+        assert _CREDENTIAL_TIMEOUT.read == 10.0
+
+    @patch("services.auth.firecrawl.firecrawl.httpx.post", autospec=True)
+    def test_post_request_passes_credential_timeout_with_connect(self, mock_post: MagicMock):
+        """`_post_request` should forward the connect-bounded timeout to httpx.post."""
+        mock_response = MagicMock()
+        mock_response.status_code = 200
+        mock_post.return_value = mock_response
+
+        credentials = {"auth_type": "bearer", "config": {"api_key": "test_api_key_123"}}
+        FirecrawlAuth(credentials).validate_credentials()
+
+        passed_timeout = mock_post.call_args.kwargs["timeout"]
+        assert passed_timeout.connect == 3.0
+        assert passed_timeout.read == 10.0
--- a/api/tests/unit_tests/services/auth/test_firecrawl_auth.py
+++ b/api/tests/unit_tests/services/auth/test_firecrawl_auth.py
@@ -208,7 +208,7 @@ class TestFirecrawlAuth:
         assert "timed out" in str(exc_info.value)
 
     def test_credential_timeout_bounds_connect_phase(self):
-        """Credential-validation timeout should bound the connect phase to 3.0s."""
+        """Test that the credential timeout bounds the connect phase to 3.0s"""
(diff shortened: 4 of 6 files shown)


crazywoola:

### Why this is being closed

### Next steps

Issue #39947, opened 2026-08-03
clean away unnecessary-type-conversion


asukaminato0721:
now in CI we ignore these errors.

@dosu how to fix them.

kobewl:
Hi! I'd like to work on this one. 👋

yrx1:
Heads-up: the actual remaining violations are larger than the ~11 originally reported (Dosu's list is stale). Running `pyrefly check` with the rule at `error` currently surfaces **404** occurrences across the codebase.

I've cleaned all 404 across three PRs, with `unnecessary-type-conversion` promoted `info -> error` in the final one:
- #41081 controllers/services (48 files)
- #41082 core/tasks/libs (63 files)
- #41083 providers/enterprise + rule promotion (17 files)

Full `./dev/pyrefly-check-local` passes with zero violations at the new severity.

Gambit-Checkmate:
Hi, I’d like to clean up the unnecessary type conversion mentioned here. I can start by locating the affected path, add or update a focused test, and keep the change limited to behavior-preserving cleanup. Is this still unassigned and suitable for a community PR?

LeonxLJX:
Hi! I'd like to work on this issue. Please let me know if I can be assigned. Thanks!

LeonxLJX:
Hi! I'd like to clean away the unnecessary type conversions. Thanks!

LeonxLJX:
Hi! I'd like to work on this. I'll analyze the relevant dify code path, reproduce, implement a minimal fix and add a regression test. Please let me know if I can proceed.

lianbie:
I'd like to work on this. Could you assign it to me?

Java123456com:
I'd like to take the remaining slice not covered by the open PRs (#39950, #41081, #41082, #41083, #41722). Cross-checking the current `pyrefly check --min-severity info` output against the files touched by those PRs leaves 15 sites in 10 files:

- `api/services/account_adapters.py` (4)
- `api/services/account_login_adapters.py` (1)
- `api/services/skill_management_service.py` (1)
- `api/repositories/oauth_access_token_repository.py` (2)
- `api/repositories/webapp_access_query_repository.py` (1)
- `api/events/event_handlers/update_provider_when_message_created.py` (1)
- `api/extensions/ext_application_services.py` (1)
- `api/tests/test_containers_integration_tests/controllers/console/test_apikey.py` (2)
- `api/tests/unit_tests/controllers/web/test_pydantic_models.py` (1)
[...]

PR #40822, opened 2026-08-15
refactor: clean away unnecessary-type-conversion (bool/str no-op wrappers)

akamabu:
Addresses #39947 — removes redundant `bool()`/`str()` wrappers (no-op type conversions) flagged by the `unnecessary-type-conversion` rule.

Changes:
- `workflow_collaboration_repository.py`: drop `bool()` on `redis.hexists`/`redis.exists` for session methods (lines 142, 145)
- `account.py`: drop `str()` on Click string options (lines 23, 65)
- `runtime_feature_manifest.py`: drop `bool(value)` and `bool(item)` inside `any()` (lines 39, 41)

Kept `bool()` in `server_heartbeat_exists` (line 240): the unit test asserts `is True` (strict identity), and redis returns int `1`, so the conversion is required there.

`api/pyproject.toml` left at `"info"` intentionally — promoting to `"error"` should wait until all instances repo-wide are cleared.

### DIFF
--- a/api/commands/account.py
+++ b/api/commands/account.py
@@ -20,7 +20,7 @@ def reset_password(email, new_password, password_confirm):
     Reset password of owner account
     Only available in SELF_HOSTED mode
     """
-    if str(new_password).strip() != str(password_confirm).strip():
+    if new_password.strip() != password_confirm.strip():
         click.echo(click.style("Passwords do not match.", fg="red"))
         return
     normalized_email = email.strip().lower()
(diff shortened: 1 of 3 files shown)


akamabu:
The CI runs on the latest commit are showing `action_required` — could a maintainer approve the workflow run so the checks can complete? The change is a small cleanup (no-op type-conversion wrappers) with all logic preserved.

asukaminato0721 (approved):

PR #39950, opened 2026-08-03
chore: remove unnecessary type conversions

tusharmotwani1718:
## Summary

This PR removes redundant type conversions reported by Pyrefly's `unnecessary-type-conversion` rule and promotes the rule from `info` to `error` so future occurrences are enforced in CI.

Fixes #39947

### Changes
- Removed redundant `bool()` wrappers around Redis method results in `workflow_collaboration_repository.py`.
- Removed redundant `str()` wrappers around `"\n".join(...)` in `dataset_retriever_tool.py`.
- Removed redundant `str()` conversions on Click string arguments in `account.py`.
- Removed unnecessary `bool()` calls used inside boolean contexts in `runtime_feature_manifest.py`.
- Updated `api/pyproject.toml` to set `unnecessary-type-conversion = "error"`.

## Screenshots

N/A (static analysis cleanup)

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [X] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [X] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods

PR #40832, opened 2026-08-16
chore(api): remove unnecessary type conversions and enforce the rule

### LINKED ISSUE #39947: clean away unnecessary-type-conversion

icn5381:
Closes #39947.

## What changed

Pyrefly's `unnecessary-type-conversion` rule was configured as `info` because the codebase carried instances of it. This PR removes **all 168 instances across 76 files** and promotes the rule to `error` in `api/pyproject.toml` so CI enforces it going forward — the complete sweep, not just the ~11 instances listed in the issue discussion:

- redundant `str()` around already-`str` expressions (tenant/document/upload-file/job ids, `"...".join(...)` results, config values, Click options)
- redundant `int()` around already-`int` expressions (token counts, thresholds, timestamps)
- redundant `bool()` around comparisons and truthy expressions, including two multi-line wrappers that keep their parenthesized multi-line form
- redundant `float()`/`bytes()` wrappers

Each removal is behavior-preserving per pyrefly's own analysis (the argument is already of the target type).

Secondary cleanups the sweep exposed, each following ruff's guidance: six ternaries that became bare-name `x if x else None` collapse to `x or None`, and eight comprehensions that became identity mappings (`list(x for x in y)`) become `list(y)` / `set(y)` / `dict(y)`.

## How I verified

- `pyrefly check --min-severity info` before: 168 `unnecessary-type-conversion` diagnostics; after: **0**. Total diagnostics dropped 10,331 → 10,163 (exactly the fixed count; no new diagnostics). The remaining error-severity diagnostics are unchanged pre-existing ones — this machine has no deps installed, so import-resolution noise dominates; CI runs with the full environment.
- With the rule at `error`, the pyrefly run stays clean for this rule.
- `python -m py_compile` passes on all 77 changed files.
- `ruff check` passes on every changed file.


crazywoola:
Closing as a duplicate of #39950, which is being retained for #39947 under the issue-author-first, then FIFO triage rule. Thanks for your contribution.

PR #40833, opened 2026-08-16
refactor(web): derive prompt-res editor key during render

Bruce-Yii:
Refs #25194

## Summary

`PromptRes` called a `useState` setter directly inside a `useEffect` to remount the prompt editor with a fresh key whenever the `value` prop changed. This triggers `eslint-react/hooks-extra-no-direct-set-state-in-use-effect` and costs an extra render pass.

Replaced it with React's recommended "adjusting state while rendering" pattern: track the previous `value` in state and bump the editor key synchronously during render when the value differs. Behavior is unchanged: the editor still remounts exactly when the prompt value changes.

Also removed this file's `eslint-react/set-state-in-effect` suppression from `oxlint-suppressions.json`, so the rule is enforced for the file again.

## Test plan

- `vp check web/app/components/app/configuration/config/automatic/prompt-res.tsx` — pass (formatting, lint, and type errors: none)
- `vp test run app/components/app/configuration/config/automatic` from `web/` — 18 tests passed (7 files), including `result.spec.tsx`

From ChatGPT


Bruce-Yii:
Hi maintainers — this PR addresses one remaining `setState`-in-`useEffect` occurrence tracked in #25194. GitHub currently shows the PR as mergeable, while the Actions runs are in `action_required`. If those runs need maintainer approval for a first-time external contributor, could you please approve them when convenient? I’m happy to adjust anything in the implementation if needed.

PR #40879, opened 2026-08-17
fix: compare file signatures in constant time and tolerate non-ASCII input

r0h1tb:
## Summary

File-signature verification compared the client-supplied `sign` with `!=` in four places.
`remote_fetcher` already used `hmac.compare_digest` for the same kind of check:

| site | before |
|---|---|
| `core/tools/signature.py:86` (`verify_tool_file_signature`) | `!=` |
| `core/tools/signature.py:159` (`verify_plugin_file_signature`) | `!=` |
| `core/app/workflow/file_runtime.py:153` (`verify_preview_signature`) | `!=` |
| `services/agent_drive_service.py:1216` (archive member) | `!=` |
| `core/file/remote_fetcher.py:232` | `compare_digest`, raises on non-ASCII |

Switching the four to `hmac.compare_digest` alone would have introduced a crash, which is
why the fix compares bytes. `compare_digest` rejects `str` carrying non-ASCII characters,
and `sign` comes straight off the query string (`controllers/files/tool_files.py:53` passes
`request.args`). `remote_fetcher` already has that bug today:

```
_verify_signed_file_url(..., sign="é")
-> TypeError: comparing strings with non-ASCII characters is not supported
   core/file/remote_fetcher.py:232
```

A request with `?sign=é` returns 500 instead of 403.

## Changes

All five sites compare the encoded bytes — constant time, and no assumption about what the
caller sent.

## Tests

- `tests/unit_tests/core/file/test_remote_fetcher.py` — 3 cases that **fail on main** with
  the `TypeError` above.
- `tests/unit_tests/core/tools/test_signature.py` — 6 cases that pass on main and guard
  against reintroducing the crash via a naive `compare_digest(str, str)`.

`tests/unit_tests/core` + `tests/unit_tests/services`: 10894 → 10903 passed, 0 failed
(+9 = the new cases). `ruff check` and `ruff format --check` clean.

Two modules fail to collect in a batch run both before and after this change
(`core/moderation/api/test_api.py`, `core/external_data_tool/test_base.py`,
`ImportError: attempted relative import`); both are excluded from the counts above and pass
when run individually.

`libs/token.py:98` compares a bearer token with `compare_digest` on `str` and has the same
non-ASCII exposure, but I could not confirm a reachable non-ASCII path there, so it is left
alone.

### DIFF
--- a/api/core/app/workflow/file_runtime.py
+++ b/api/core/app/workflow/file_runtime.py
@@ -150,7 +150,7 @@ class DifyWorkflowFileRuntime(WorkflowFileRuntimeProtocol):
     ) -> bool:
         payload = f"{preview_kind}-preview|{file_id}|{timestamp}|{nonce}"
         recalculated = hmac.new(self._secret_key(), payload.encode(), hashlib.sha256).digest()
-        if sign != base64.urlsafe_b64encode(recalculated).decode():
+        if not hmac.compare_digest(sign.encode(), base64.urlsafe_b64encode(recalculated)):
             return False
         return int(time.time()) - int(timestamp) <= dify_config.FILES_ACCESS_TIMEOUT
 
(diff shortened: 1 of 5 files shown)


crazywoola:
Hi @r0h1tb, thank you for your contribution to Dify!

We are closing this PR for now because it does not link a tracking issue in its description. Our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md#pull-request-process) ask contributors to discuss changes in an issue and link it in the PR description.

Please open an issue describing the problem and proposed change, or link an existing relevant issue, then add `Fixes #<issue_number>` to the PR description. Once updated, please request that this PR be reopened so review can continue.

Thank you for your work and understanding!

PR #40919, opened 2026-08-18
refactor(web): remove unnecessary setState in useEffect

gegemeimingzi:
## Summary

Part of #25194

Three cases where `useState` + `useEffect` can be replaced with simpler
derived values during render, per https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

## Changes

- **`verify-email.tsx`** — `shouldButtonDisabled` is purely derived from
  `verificationCode` and `isDeleting`. Replaced with a `const`.
- **`prompt-res.tsx`** — The `editorKey` state + effect existed to force
  remount `PromptEditor` on value change. Passing `value` as the React `key`
  achieves the same thing without extra state.
- **`settings-modal/index.tsx`** — `memberList` was synced from React Query
  data via useEffect but never independently mutated. Derived inline.

## Type of Change

- [x] Refactoring (no behavior change)

## Testing

- [x] `tsc --noEmit` passes
- [x] oxlint passes (no new diagnostics)
- [x] lint-staged pre-commit hook passes
- [ ] No full UI verification (requires backend services)

### DIFF
--- a/oxlint-suppressions.json
+++ b/oxlint-suppressions.json
@@ -138,9 +138,6 @@
     }
   },
   "web/app/account/(commonLayout)/delete-account/components/verify-email.tsx": {
-    "eslint-react/set-state-in-effect": {
-      "count": 1
-    },
     "no-restricted-imports": {
       "count": 1
     }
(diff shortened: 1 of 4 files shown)


gegemeimingzi:
Closing this PR - not pursuing this change further. Thanks for the automated review feedback.

PR #38653, opened 2026-07-10
refactor(chat): modernize speech-to-text pipeline

lyzno1:
## Summary

- replace the legacy `js-audio-recorder` and `lamejs` CJS pipeline with AudioWorklet capture and lazy-loaded Mediabunny MP3 encoding
- route speech-to-text through explicit App or Agent targets instead of inferring identity from the URL
- add `POST /console/api/agent/{agent_id}/audio-to-text` with draft and debug-build Agent Soul resolution
- preserve the existing MP3 upload contract required by the backend and plugin demo
- prevent empty `/console/api/` requests and Agent IDs being sent to `/apps/{app_id}/audio-to-text`
- preserve legacy Agent App speech-to-text by falling back to AppModelConfig when no Roster Agent backs the App
- load the encoder before requesting microphone access and abort pending recorder setup when the voice surface closes
- release MediaStream tracks and AudioContext resources before MP3 finalization, and make cancellation settle pending Worklet, write-queue, and finalize waits
- keep runtime cleanup independent so a failed disconnect, context close, or encoder cancellation cannot strand the remaining resources
- expose a localized, cancellable starting state and distinguish microphone permission denial from other recorder startup failures
- abort in-flight transcription and suppress upload when the user cancels during MP3 finalization, draft saving, or the request itself
- keep save-before-transcribe failures owned by the Agent draft flow instead of showing a duplicate generic toast
- discard stale recorder setup during React effect replay and keep the active recorder stable across callback and chat-input rerenders
- restore the Advanced Chat workflow debug voice entry with an explicit console App target
- restore focus to the chat textarea with the caret at the end after successful transcription without stealing focus when the user has moved elsewhere
- expose a specific localized `Stop recording` accessible name and announce asynchronous voice-input states
- raise the declared iOS support boundary to 16.4 instead of shipping a second non-SIMD encoder fallback
- add a real Cucumber + Playwright speech-to-text scenario using a deterministic Chromium fake microphone, production frontend build, active default STT model, real provider request, input assertion, and focus assertion
- wire the STT model selector through the E2E env schema, local example, and external-runtime CI job
- trigger the post-merge external suite when Agent STT steps, microphone fixtures, shared voice/chat code, generated Agent contracts, console transport, routes, or encoder dependency declarations change

## Ownership

- frontend owns microphone capture, MP3 encoding, local recording state, explicit request target selection, save-before-transcribe orchestration, cancellation, and post-transcription focus restoration
[...]

Issue #40927, opened 2026-08-18
[Bug] browserslist floor too narrow (last 1 Chrome / iOS >=16.4) makes Lightning CSS strip -webkit-mask prefixes, breaking masked icons on Chrome 112 and iOS 15


yiclong:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.16.1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

In Dify 1.16.x, web/package.json -> browserslist was tightened to last 1 Chrome version and iOS >=16.4 (was iOS >=15 in 1.14.2). Lightning CSS then treats the -webkit- prefixed forms of mask / mask-size / mask-image as unnecessary and strips them from the built CSS. Slightly older browsers that still need the prefixed properties (e.g. desktop Chrome 112, iOS 15 Safari) lose masked icons.
1、Checkout main (or any 1.16.x release) and build the web frontend.
2、Open the app in Chrome 112 (desktop).
3、Navigate to any page that renders icons via CSS mask / mask-image / mask-size.
4、Observe the icons.
5、(Cross-check) Open the same page in the latest Chrome, or in a 1.14.2 build on Chrome 112 - icons render correctly.

### ✔️ Expected Behavior

Masked icons render correctly on any browser a realistic user might still run, including desktop Chrome 112 and iOS 15 Safari. Lightning CSS should either keep the -webkit-mask-* prefixes for those targets, or the browserslist floor should be widened so the prefix is re-emitted.

### ❌ Actual Behavior

With the current browserslist, Lightning CSS drops -webkit-mask-size / -webkit-mask / -webkit-mask-image from the generated CSS. On Chrome 112 (and iOS 15.x), the unprefixed mask / mask-size / mask-image are not fully recognized, so masked icons render incorrectly - missing, oversized, or invisible.
[...]

PR #40931, opened 2026-08-18
fix(web): keep -webkit-mask-* prefixes in Lightning CSS output

yiclong:
## Problem
The browserslist floor was raised to `iOS >=16.4` in PR #38653 (for the WASM SIMD MP3 encoder). With that floor, Lightning CSS treats `-webkit-mask-*` vendor prefixes as unnecessary and strips them, breaking CSS mask icons on Chrome 112 and iOS 15.

## Fix
Force the `vendor-prefixes` feature so prefixed output is always emitted regardless of browserslist targets:
- `web/next.config.ts`: `experimental.lightningCssFeatures.include: ['vendor-prefixes']` (Turbopack path)
- `web/vite.config.ts`: `css.lightningcss.include: 262144` (= `Features.VendorPrefixes`, vinext path)

## Verification
- Tested: `-webkit-mask-size` now present in built CSS; icons render correctly on older browsers
- MP3 encoder floor (iOS 16.4) unchanged

### DIFF
--- a/web/next.config.ts
+++ b/web/next.config.ts
@@ -12,6 +12,18 @@ const allowedDevOrigins = process.env.NEXT_ALLOWED_DEV_ORIGINS?.split(',')
 const nextConfig: NextConfig = {
   basePath: env.NEXT_PUBLIC_BASE_PATH,
   ...(allowedDevOrigins?.length ? { allowedDevOrigins } : {}),
+  experimental: {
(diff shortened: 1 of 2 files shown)


lyzno1:
Refs https://github.com/langgenius/dify/issues/40927#issuecomment-5326554407

PR #40947, opened 2026-08-18
fix(logstore): normalize workflow timestamps to naive UTC

### LINKED ISSUE #40943: LogStore repositories convert workflow timestamps to local time, producing shifted timestamps, negative elapsed_time, and a TypeError
### Self Checks

### Dify version

1.16.1

### Cloud or Self Hosted

Self Hosted (Docker), Self Hosted (Source)

### Steps to reproduce

The LogStore repositories convert timestamps using timezone-naive local-time calls, while the rest of the codebase stores naive **UTC**:

`api/extensions/logstore/repositories/logstore_api_workflow_run_repository.py`

```python
started_at = data.get("started_at") or data.get("created_at")
if started_at:
    match started_at:
        case str():
            model.created_at = datetime.fromisoformat(started_at)   # L111
        case int() | float():
            model.created_at = datetime.fromtimestamp(started_at)   # L113  <- local time
        case _:
            model.created_at = started_at
else:
    model.created_at = datetime.now()                               # L118  <- local time

finished_at = data.get("finished_at")
if finished_at:
    match finished_at:
        case str():
            model.finished_at = datetime.fromisoformat(finished_at) # L124
        case int() | float():
            model.finished_at = datetime.fromtimestamp(finished_at) # L126  <- local time
...
if model.finished_at and model.created_at:
    model.elapsed_time = (model.finished_at - model.created_at).total_seconds()
```

The same pattern is in `logstore_api_workflow_node_execution_repository.py` at lines 93, 97 and 108.

`datetime.fromtimestamp(ts)` without `tz` returns the host's **local** time, and `datetime.now()` likewise. Every other epoch conversion in the codebase passes `tz=UTC` explicitly — `api/controllers/web/wraps.py:148`, `api/libs/device_flow_security.py:126`, `api/services/retention/workflow_run/clear_free_plan_expired_workflow_run_logs.py:520`, `api/extensions/ext_logging.py:86` — and `libs/datetime_utils.naive_utc_now()` is used in 277 places as the project-wide convention.

Because `str` and `int | float` are handled independently per field, the two timestamps on one record can be produced by different branches, which is where it stops being a simple offset shift.

Reproduction, driving the real `_dict_to_workflow_run` on a host in `Asia/Kolkata` (UTC+05:30):

```python
import datetime as dt
from extensions.logstore.repositories.logstore_api_workflow_run_repository import _dict_to_workflow_run

START = dt.datetime(2026, 8, 18, 2, 0, 0, tzinfo=dt.UTC)     # ground truth
FINISH = START + dt.timedelta(seconds=30)                     # a 30-second run
BASE = {"id": "run-1", "tenant_id": "t", "app_id": "a", "workflow_id": "w"}

for label, payload in [
    ("A both epoch",         {"started_at": START.timestamp(),                  "finished_at": FINISH.timestamp()}),
    ("B iso-offset + epoch", {"started_at": START.isoformat(),                  "finished_at": FINISH.timestamp()}),
    ("C naive-iso + epoch",  {"started_at": START.replace(tzinfo=None).isoformat(), "finished_at": FINISH.timestamp()}),
    ("D missing started_at", {                                                   "finished_at": FINISH.timestamp()}),
]:
    try:
        m = _dict_to_workflow_run({**BASE, **payload})
        print(label, "|", m.created_at, "|", m.finished_at, "|", m.elapsed_time)
    except Exception as e:
        print(label, "| RAISED", type(e).__name__, e)
```

Output:

```
A both epoch         | 2026-08-18 07:30:00 | 2026-08-18 07:30:30 | 30.0
B iso-offset + epoch | RAISED TypeError can't subtract offset-naive and offset-aware datetimes
C naive-iso + epoch  | 2026-08-18 02:00:00 | 2026-08-18 07:30:30 | 19830.0
D missing started_at | 2026-08-18 21:21:45 | 2026-08-18 07:30:30 | -49875.038848
```

### ✔️ Expected Behavior

All four cases should yield `created_at = 2026-08-18 02:00:00` (naive UTC, matching what the rest of the codebase writes) and `elapsed_time = 30.0`, independent of the host timezone.

### ❌ Actual Behavior

On a UTC+05:30 host:

- **A** — timestamps are shifted 5h30m into the future. `elapsed_time` is correct only because both values are shifted equally. Workflow run logs display the wrong time and sort inconsistently against records written through the Postgres path, which use `naive_utc_now()`.
- **B** — `TypeError: can't subtract offset-naive and offset-aware datetimes`. `fromisoformat` on a string carrying an offset returns an **aware** datetime, while `fromtimestamp` returns a **naive** one, so the `elapsed_time` subtraction raises.
- **C** — a 30-second workflow run is reported as **19830 seconds** (5h30m30s), because only one of the two values is shifted.
- **D** — `elapsed_time` is **negative** (`-49875.0`), since `created_at` defaults to local `datetime.now()` while `finished_at` comes from the epoch branch.

On a UTC host all four happen to agree, which is why this is easy to miss.

### Additional context

Scope: this code path is only reachable when the LogStore storage backend is enabled, so deployments on the default Postgres path are unaffected.

Suggested fix, consistent with existing helpers in `libs/datetime_utils`:

- `datetime.now()` → `naive_utc_now()`
- `datetime.fromtimestamp(ts)` → `datetime.fromtimestamp(ts, tz=UTC).replace(tzinfo=None)`
- `datetime.fromisoformat(s)` → wrap in `ensure_naive_utc(...)`, which already normalises aware values to naive UTC and leaves naive ones untouched

That makes every branch of both `match` statements produce naive UTC, which fixes the offset shift, the `TypeError` and both bad `elapsed_time` results together.

I'd be happy to open a PR.

amogh-nagri-11:
## Summary

Fixes #40943.

The two LogStore repositories deserialize workflow timestamps with calls that resolve against the **host's local clock**, while the rest of the codebase stores naive **UTC**:

- `datetime.fromtimestamp(ts)` without `tz` → naive local time
- `datetime.now()` → naive local time
- `datetime.fromisoformat(s)` → **aware** when the string carries an offset, naive otherwise

On its own that would just be a uniform offset. The real problem is that `created_at` and `finished_at` are matched **independently**, so one record's two timestamps can be produced by different branches and land in different frames — and `_dict_to_workflow_run` then subtracts them for `elapsed_time`.

Observed on a UTC+05:30 host (reproduction in the issue drives the real functions):

| Payload | Before | After |
| --- | --- | --- |
| both epoch | timestamps shifted +5:30 | correct naive UTC |
| aware ISO + epoch | `TypeError: can't subtract offset-naive and offset-aware datetimes` | `elapsed_time = 30.0` |
| naive ISO + epoch | 30-second run reported as **19830 s** | `elapsed_time = 30.0` |
| `started_at` missing | `elapsed_time` **negative** (`-49875.0`) | defaults to `naive_utc_now()` |

On a UTC host all four coincidentally agree, which is why this is easy to miss.

## Changes

Every branch of both `match` statements now terminates in naive UTC, using the helpers that already exist in `libs/datetime_utils`:

- `datetime.now()` → `naive_utc_now()` (the same call the Postgres write path uses)
- `datetime.fromtimestamp(ts)` → `datetime.fromtimestamp(ts, tz=UTC).replace(tzinfo=None)`
- `datetime.fromisoformat(s)` → wrapped in `ensure_naive_utc(...)`, which normalizes aware values and is a no-op for naive ones
- the already-a-`datetime` passthrough branch is wrapped in `ensure_naive_utc` too, so an aware value from upstream can't reintroduce the mixed-frame subtraction

Applied to all six call sites across `logstore_api_workflow_run_repository.py` and `logstore_api_workflow_node_execution_repository.py`. One change fixes the offset shift, the `TypeError`, and both bad `elapsed_time` results together.

Scope note: this path is only reachable when the Aliyun SLS LogStore backend is enabled, so deployments on the default Postgres path are unaffected. No behavior change on UTC hosts.

## Tests

Added regression coverage that pins `TZ=Asia/Kolkata` via `time.tzset()`, so the assertions are meaningful on the UTC CI host rather than passing by coincidence. It covers all four payload shapes plus the missing-timestamp default, for both repositories.

All 10 new tests fail on `main` and pass with this change:

```
uv run --project api --dev pytest api/tests/unit_tests/extensions/logstore/repositories/
13 passed
```

`ruff check`, `ruff format --check`, `mypy` and `pyrefly` are clean on the changed files (the two pre-existing `bad-instantiation` pyrefly diagnostics in the node-execution test file are on untouched lines).

## Screenshots

Not applicable — backend-only timezone correctness fix.

## Checklist

### DIFF
--- a/api/extensions/logstore/repositories/logstore_api_workflow_node_execution_repository.py
+++ b/api/extensions/logstore/repositories/logstore_api_workflow_node_execution_repository.py
@@ -8,7 +8,7 @@ WorkflowNodeExecutionModel operations using Aliyun SLS LogStore.
 import logging
 import time
 from collections.abc import Mapping, Sequence
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import Any, override
 
 from sqlalchemy.orm import sessionmaker
@@ -17,6 +17,7 @@ from extensions.logstore.aliyun_logstore import AliyunLogStore
 from extensions.logstore.repositories import safe_float, safe_int
 from extensions.logstore.sql_escape import escape_identifier, escape_logstore_query_value
 from graphon.enums import WorkflowNodeExecutionStatus
+from libs.datetime_utils import ensure_naive_utc, naive_utc_now
 from models.enums import CreatorUserRole
 from models.workflow import WorkflowNodeExecutionModel, WorkflowNodeExecutionTriggeredFrom
 from repositories.api_workflow_node_execution_repository import DifyAPIWorkflowNodeExecutionRepository
@@ -86,28 +87,29 @@ def _dict_to_workflow_node_execution_model(data: dict[str, Any]) -> WorkflowNode
     model.execution_metadata = data.get("execution_metadata")
 
     # Handle datetime fields
+    # Every branch must yield naive UTC, matching what the database path stores.
     created_at = data.get("created_at")
     match created_at:
         case None:
             # Provide default created_at if missing
-            model.created_at = datetime.now()
+            model.created_at = naive_utc_now()
         case str():
-            model.created_at = datetime.fromisoformat(created_at)
+            model.created_at = ensure_naive_utc(datetime.fromisoformat(created_at))
         case int() | float():
-            model.created_at = datetime.fromtimestamp(created_at)
+            model.created_at = datetime.fromtimestamp(created_at, tz=UTC).replace(tzinfo=None)
         case _:
-            model.created_at = created_at
+            model.created_at = ensure_naive_utc(created_at)
 
     finished_at = data.get("finished_at")
     match finished_at:
         case None:
             ...
         case str():
-            model.finished_at = datetime.fromisoformat(finished_at)
+            model.finished_at = ensure_naive_utc(datetime.fromisoformat(finished_at))
         case int() | float():
-            model.finished_at = datetime.fromtimestamp(finished_at)
+            model.finished_at = datetime.fromtimestamp(finished_at, tz=UTC).replace(tzinfo=None)
         case _:
-            model.finished_at = finished_at
+            model.finished_at = ensure_naive_utc(finished_at)
 
     return model
 
--- a/api/extensions/logstore/repositories/logstore_api_workflow_run_repository.py
+++ b/api/extensions/logstore/repositories/logstore_api_workflow_run_repository.py
@@ -17,7 +17,7 @@ import logging
 import os
 import time
 from collections.abc import Sequence
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import Any, cast, override
 
 from sqlalchemy.orm import sessionmaker
@@ -26,6 +26,7 @@ from extensions.logstore.aliyun_logstore import AliyunLogStore
 from extensions.logstore.repositories import safe_float, safe_int
 from extensions.logstore.sql_escape import escape_identifier, escape_logstore_query_value, escape_sql_string
 from graphon.enums import WorkflowExecutionStatus
+from libs.datetime_utils import ensure_naive_utc, naive_utc_now
 from libs.infinite_scroll_pagination import InfiniteScrollPagination
 from models.enums import CreatorUserRole, WorkflowRunTriggeredFrom
 from models.workflow import WorkflowRun, WorkflowType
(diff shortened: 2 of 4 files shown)


asukaminato0721 (approved):

amogh-nagri-11:
Thank you @asukaminato0721 for the quick review and merge!

PR #40949, opened 2026-08-19
refactor(api): extract web app runtime bootstrap

hjlarry:
## Summary

- add a stable Web App Runtime query capability for the public Web app bootstrap
- move App, Tenant, and Site reads into a repository-owned session and complete scalar mapping before external feature or storage work
- centralize Web app icon URL selection in FileService and keep the controller limited to admission, error mapping, and dump_response serialization

## Why

The Web /site controller currently performs ORM reads and then calls feature and file infrastructure from the request path. This change gives those responsibilities explicit owners while reusing the existing App Definition database adapter instead of adding an endpoint-specific repository.

## Compatibility

- preserves the existing WebAppSiteResponse schema and generic 403 behavior for unavailable sites
- preserves Cloud plus S3 direct icon URLs and preview URLs in other deployments
- leaves the Human Input site projection on its existing behavior
- keeps legacy Agent-compatible mode resolution

Stack: follows #40795.

### DIFF
--- a/api/repositories/app_definition_query_repository.py
+++ b/api/repositories/app_definition_query_repository.py
@@ -9,6 +9,7 @@ from core.agent.publish_visibility import agent_has_workflow_callable_active_sna
 from core.app.apps.agent_app.app_feature_projection import merge_agent_app_features
 from core.app.apps.agent_app.app_variable_projection import agent_app_variables_to_user_input_form
 from core.app.apps.agent_app.errors import AgentAppGeneratorError, AgentAppNotPublishedError
+from models.account import Tenant
 from models.agent import AgentConfigSnapshot
 from models.agent_config_entities import AgentSoulConfig
 from models.model import App, AppMode, AppModelConfig, Site, load_annotation_reply_config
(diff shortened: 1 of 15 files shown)


api/services/web_app_runtime_query_service.py
```diff
@@ -0,0 +1,103 @@
+"""Application service for building the public Web app runtime bootstrap."""
+
+import json
+from collections.abc import Mapping
+from typing import NamedTuple, Protocol, cast
+
+from configs import dify_config
+from services.app_definition_query_service import AppSiteConfiguration
+from services.feature_service import FeatureService
```
41tair: Use inject instead of import.

api/repositories/app_definition_query_repository.py
```diff
@@ -153,23 +175,46 @@ def get_site_configuration(self, app_id: str) -> AppSiteConfiguration | None:
...
+                app_id=app_id,
+                tenant_id=tenant_id,
+                enable_site=enable_site,
+                site=site_configuration,
+                plan=plan,
+                tenant_status=tenant_status,
+                tenant_custom_config_json=tenant_custom_config_json,
             )
 
+    def resolve_compatible_app_mode(self, app_id: str) -> str | None:
```
41tair: maybe we can merge these function to one to avoid open session twice.

41tair (approved):

PR #40955, opened 2026-08-19
feat(workflow): expose bulk execution status as a system variable

### LINKED ISSUE #40021: Add a flag/variable to indicate whether a workflow run is part of a bulk execution
### Self Checks

I was trying to use the workflow's bulk execution feature to process multiple inputs at once, and I ran into a problem when the workflow calls an external API that has strict rate limits.

When running a single input, there's no issue. But when the same workflow is triggered via bulk execution, many runs fire in a short period of time and hit the API's rate limit. To handle this properly, I want to add logic inside the workflow itself — such as throttling requests or inserting a wait/delay — but only when the run is actually part of a bulk execution.

The problem is that, from inside the workflow, there is currently no way to determine whether the current run was triggered as part of a bulk execution or as a single, individual execution. Without this information, I can't conditionally branch the workflow logic (e.g., "if this is a bulk run, wait N seconds before calling the API").

As a workaround, it's technically possible to avoid using the bulk execution feature altogether and instead design the workflow to accept either a single file input or a batch-style input (e.g., a list) manually, then handle rate limiting inside the workflow logic myself.

However, it would be much more convenient if Dify's bulk execution feature natively exposed some kind of system variable or flag (e.g., something like sys.is_bulk_execution or similar) that workflow nodes could reference directly. This would let workflow authors add appropriate throttling/wait logic without having to redesign their input handling just to detect bulk vs. single execution.

This would be especially useful for workflows that call external APIs with rate limits, since bulk execution can easily trigger many concurrent/rapid calls.

lxy123hh:
> [!IMPORTANT]
>

## Summary

- expose a read-only Boolean `sys.is_bulk_execution` variable for top-level Workflow runs
- propagate the bulk execution flag from Batch requests through the workflow generate entity into the system variable pool
- default the flag to `false` for existing clients while allowing Service API callers to set it explicitly
- keep Chatflow and the existing RAG Pipeline `sys.batch` variable unchanged
- add frontend and backend regression coverage for request propagation, serialization, and variable visibility

Fixes #40021

## Screenshots

Image

## Checklist

### DIFF
--- a/api/controllers/common/controller_schemas.py
+++ b/api/controllers/common/controller_schemas.py
@@ -170,6 +170,10 @@ class WorkflowRunPayload(BaseModel):
             "the returned `id` as `upload_file_id` with `transfer_method: local_file`."
         ),
     )
+    is_bulk_execution: bool = Field(
+        default=False,
+        description="Whether this workflow run is part of a bulk execution.",
+    )
 
 
 class WorkflowUpdatePayload(BaseModel):
--- a/api/controllers/console/explore/trial.py
+++ b/api/controllers/console/explore/trial.py
@@ -104,6 +104,10 @@ logger = logging.getLogger(__name__)
 class WorkflowRunRequest(BaseModel):
     inputs: dict
     files: list | None = Field(default=None)
+    is_bulk_execution: bool = Field(
+        default=False,
+        description="Whether this workflow run is part of a bulk execution.",
+    )
 
 
 class ChatRequest(BaseModel):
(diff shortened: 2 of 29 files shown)


crazywoola:
Please attach some screenshots as well.

lxy123hh:
> Please attach some screenshots as well.

Thanks for the feedback. I've added a screenshot showing `sys.is_bulk_execution` in the workflow system variable selector.

PR #40964, opened 2026-08-19
fix(api): prevent dropped workflow_started events in Redis Streams

### LINKED ISSUE #40948: [Bug] workflow_started is dropped for concurrent streaming runs with the streams event transport (regression of #32518 via #34030)
### Self Checks

### Dify version

1.13.3, 1.15.0 and 1.16.1 — measured on all three. The two code paths that cause this are also unchanged on `main`. See "Newer versions" below for the per-version failure rates.

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

This is a regression of #32518, which was closed as completed on 2026-03-04. It reappears when the `streams` event transport is used with concurrent streaming requests.

Prerequisites:

- `EVENT_BUS_REDIS_CHANNEL_TYPE=streams`
- api and worker in separate containers
- `CELERY_WORKER_AMOUNT=8` on the worker, and the script's concurrency set to the same value (`K=8`, its default; override with the `K` env var)

Matching `K` to the worker pool size is deliberate: with a free slot for every request, all tasks start immediately, so queueing cannot contribute to the result. That is what makes the round time meaningful — a round of 8 finishing in about 1.2 s rules out worker starvation and leaves the delivery race as the only explanation. If your worker concurrency differs, set `K` to match it.

**1. Create a workflow app whose Start node is wired straight to End**, publish it, and issue an API key. A short workflow matters: it makes the worker publish as early as possible, which widens the window. DSL for convenience:

<details>
<summary>DSL — import via Studio &rarr; Import DSL file</summary>

```yaml
app:
  description: 'Minimal workflow for reproducing this issue: Start wired straight to End.'
  icon: 🔊
  icon_background: '#EFF1F5'
  mode: workflow
  name: repro-echo
  use_icon_as_answer_icon: false
kind: app
version: 0.4.0
workflow:
  conversation_variables: []
  environment_variables: []
  features:
    file_upload:
      enabled: false
    opening_statement: ''
    retriever_resource:
      enabled: true
    sensitive_word_avoidance:
      enabled: false
    speech_to_text:
      enabled: false
    suggested_questions: []
    suggested_questions_after_answer:
      enabled: false
    text_to_speech:
      enabled: false
  graph:
    edges:
    - data:
        isInIteration: false
        isInLoop: false
        sourceType: start
        targetType: end
      id: start-end
      source: startnode
      sourceHandle: source
      target: endnode
      targetHandle: target
      type: custom
      zIndex: 0
    nodes:
    - data:
        desc: ''
        selected: false
        title: Start
        type: start
        variables:
        - label: message
          max_length: null
          options: []
          required: true
          type: text-input
          variable: message
      height: 90
      id: startnode
      position:
        x: 30
        y: 245
      positionAbsolute:
        x: 30
        y: 245
      selected: false
      sourcePosition: right
      targetPosition: left
      type: custom
      width: 244
    - data:
        desc: ''
        outputs:
        - value_selector:
          - startnode
          - message
          value_type: string
          variable: echoed
        selected: false
        title: End
        type: end
      height: 90
      id: endnode
      position:
        x: 334
        y: 245
      positionAbsolute:
        x: 334
        y: 245
      selected: false
      sourcePosition: right
      targetPosition: left
      type: custom
      width: 244
    viewport:
      x: 0
      y: 0
      zoom: 0.7
```

</details>

**2. Fire K concurrent streaming runs** and record, per request, whether `workflow_started` arrived:

<details>
<summary>repro.mjs — Node 18+, no dependencies</summary>

```js
// Reproduces langgenius/dify#32518 on 1.13.3+ with EVENT_BUS_REDIS_CHANNEL_TYPE=streams.
//
// Fires K concurrent streaming workflow runs and reports, per request, whether
// workflow_started was delivered. The workflow_run_id is captured from whatever
// event does arrive, so a missed request can still be looked up in Redis.
//
//   DIFY_BASE_URL=http://localhost DIFY_API_KEY=app-xxxxxxxx node repro.mjs
//
// Node 18+ (global fetch), no dependencies.

const BASE = process.env.DIFY_BASE_URL ?? "http://localhost";
const KEY = process.env.DIFY_API_KEY;
const K = Number(process.env.K ?? 8);

if (!KEY) {
  console.error("set DIFY_API_KEY to the API key of a workflow app whose Start node is wired straight to End");
  process.exit(2);
}

async function run(i) {
  const res = await fetch(`${BASE}/v1/workflows/run`, {
    method: "POST",
    headers: {
      Authorization: `Bearer ${KEY}`,
      "Content-Type": "application/json",
      Accept: "text/event-stream",
    },
    body: JSON.stringify({ inputs: { message: `r${i}` }, response_mode: "streaming", user: `u${i}` }),
  });
  if (!res.ok) return { i, started: false, runId: null, events: [`HTTP ${res.status}`] };

  const events = [];
  let runId = null;
  let buf = "";
  const decoder = new TextDecoder();
  for await (const chunk of res.body) {
    buf += decoder.decode(chunk, { stream: true });
    const lines = buf.split("\n");
    buf = lines.pop() ?? "";
    for (const line of lines) {
      if (!line.startsWith("data: ")) continue;
      try {
        const evt = JSON.parse(line.slice(6));
        events.push(evt.event);
        runId ??= evt.workflow_run_id ?? evt.data?.id ?? null;
      } catch {
        // ignore partial JSON
      }
    }
  }
  return { i, started: events.includes("workflow_started"), runId, events };
}

const results = await Promise.all(Array.from({ length: K }, (_, i) => run(i)));
for (const r of results.sort((a, b) => a.i - b.i)) {
  console.log(
    `[${r.i}] workflow_started ${r.started ? "OK     " : "MISSING"}  run=${r.runId ?? "unknown"}  events=${r.events.join(",")}`,
  );
}
const ok = results.filter((r) => r.started).length;
console.log(`\n${ok}/${K} received workflow_started`);
```

</details>

```
DIFY_BASE_URL=http://localhost DIFY_API_KEY=app-xxxxxxxx node repro.mjs
```

**3. Cross-check Redis** for any request reported as `MISSING`, using the `workflow_run_id` the script prints.

The failure is non-deterministic, so run it a few times — an occasional round returns 8/8.

### ✔️ Expected Behavior

Every streaming response begins with `workflow_started`, followed by the node events and `workflow_finished`.

### ❌ Actual Behavior

A fraction of concurrent requests never receive `workflow_started` (and often `node_started` too). On 1.13.3, over 10 consecutive rounds of 8 concurrent requests, **44 of 80 requests (55 percent) were missing it**. Each round completed in about 1.2 s, so the workflows themselves all ran — this is not worker starvation.

Sample output — request `[1]` is missing the first two events, and the script still recovers its `workflow_run_id` from a later event:

```
[0] workflow_started OK       run=3be2f2b4-0172-4c81-9154-71344ca4db5f  events=workflow_started,node_started,text_chunk,node_finished,node_started,node_finished,workflow_finished
[1] workflow_started MISSING  run=58f1baf7-af48-4b60-8613-7954e48afba4  events=text_chunk,node_finished,node_started,node_finished,workflow_finished
[2] workflow_started OK       run=f629f931-5d38-47c3-8803-0afb294a1f01  events=workflow_started,node_started,text_chunk,node_finished,node_started,node_finished,workflow_finished
...
7/8 received workflow_started
```

**The events were published — the subscriber just never saw them.** The stream for the request that missed the event still holds it:

```
$ redis-cli XLEN   stream:channel:workflow:58f1baf7-af48-4b60-8613-7954e48afba4
(integer) 7
$ redis-cli XRANGE stream:channel:workflow:58f1baf7-af48-4b60-8613-7954e48afba4 - + COUNT 1
1) 1) "1787..."
   2) 1) "data"
      2) "{\"event\":\"workflow_started\",...}"
```

In one round where the client received only 5 of 8, all eight streams held the complete event set with `workflow_started` as the first entry — and **the three that were missed were the three published earliest**, which is what you would expect if the worker beat the subscriber's first `XREAD`.

#### Cause

| Location | Introduced | Changed since |
|---|---|---|
| `api/services/app_generate_service.py` — for `streams`, `_try_start()` enqueues the Celery task **immediately**, before any subscriber exists | `2f4c740d46` (#32586) | no |
| `api/libs/broadcast_channel/redis/streams_channel.py` — `last_id = "$"` in the lazily started `_listen` thread | `eef13853b2` (#34030) | no |

#32518 was fixed while `StreamsBroadcastChannel` initialised `_last_id = "0-0"`: a late subscriber replayed the stream from the start, so nothing could be lost. Three weeks later #34030 changed it to `$` to stop the frontend replaying a stale `workflow_paused` event and reconnecting in a loop (#34040). That commit does not reference #32518, but it reopens the same race — with `$`, a subscriber only sees messages written after its first `XREAD`.

The docstring in `app_generate_service.py` still reads `With Redis Streams, we can safely start right away; consumers can read past events`, which no longer holds.

#### One-line confirmation

Changing only `last_id = "$"` to `last_id = "0-0"`, on the same 1.13.3 image with everything else identical:

| build | rounds failed | requests missing `workflow_started` |
|---|---|---|
| official 1.13.3 | 10 / 10 | 44 / 80 |
| 1.13.3 with `last_id = "0-0"` | **0 / 10** | **0 / 80** |

So the read position is both necessary and sufficient to explain the loss.

#### Newer versions

Measured on 1.15.0 and 1.16.1 as well as 1.13.3; all three reproduce. Per-version failure rates and the setup are in [this comment](https://github.com/langgenius/dify/issues/40948#issuecomment-5336810469). `origin/main` (2026-07-27) is source-checked only — both locations in the table above are unchanged there.

#### Why the plain revert is not the fix

`0-0` reintroduces #34040, which is what #34030 set out to fix. The requirements only conflict because the initial read position is a property of the channel rather than of the subscription:

- a **fresh** run should replay from the start of its own stream — there is nothing stale to replay
- a **reconnect** after `workflow_paused` should resume from where that client left off

A fix satisfying both would let the caller pass the starting id per subscription — a resume position, or the stream's end id captured **before** the task is enqueued — instead of hardcoding `$` for every subscriber. #35315 / #35321 look like they explored this and were closed unmerged.

zl86790:
#40948

## Summary

When using Redis Streams as the event transport (`EVENT_BUS_REDIS_CHANNEL_TYPE=streams`), the `workflow_started` event (and sometimes the first few node events) can silently disappear on concurrent streaming runs. The events are actually in Redis — the SSE subscriber just never sees them.

The root cause: the workflow execution task gets dispatched to Celery right away, before the SSE response has even started streaming. The Redis Streams subscription only gets set up later, and it starts reading from `$` (only new messages). If the worker publishes `workflow_started` before that subscription exists, it's gone for good. This shows up much more often under concurrent load, which matches what's reported in the issue (55% loss with 8 concurrent requests).

Reverting `$` back to `0-0` isn't an option — that was changed in #34030 specifically to stop replaying stale `workflow_paused` events on reconnect (#34040), and reverting it would bring that bug straight back.

Fix: grab a checkpoint (the current tail of the stream) synchronously, *before* dispatching the task. The subscriber then resumes from that exact point once it actually connects, no matter how late that happens. If reading the checkpoint itself fails, we let it fail loudly instead of quietly falling back to something unsafe.

Pub/sub and sharded transports aren't affected — they already gate task start on a real subscription with a fallback timer, so this failure mode doesn't apply to them.

Fixes #40948

## Tests

- Added unit tests for the new `checkpoint()` / `subscribe_from()` methods on the streams channel: reading the last entry id, empty-stream case, checkpoint errors propagating instead of being swallowed, and that a subscriber resuming from a checkpoint picks up events published before it actually starts listening (while still not replaying anything from before the checkpoint).
- Added a regression test that dispatches the task — letting it publish both `workflow_started` and `workflow_finished` — before the subscriber even calls `retrieve_events()`, matching the real ordering. Also fixed the fake Redis Streams test double, which didn't model `$` semantics at all and had been masking this bug.
- Added a test covering the case where the checkpoint read fails: task must not start, error propagates.
- `ruff check` — clean.
- Full relevant test suite (streams channel, streaming utils, app generate service) — 80 passed.
- Haven't run `make type-check` locally, worth double-checking in CI.

## Checklist


api/services/app_generate_service.py
```diff
@@ -67,14 +78,17 @@ def _try_start() -> bool:
 
         channel_type = dify_config.PUBSUB_REDIS_CHANNEL_TYPE
         if channel_type == "streams":
-            # With Redis Streams, we can safely start right away; consumers can read past events.
+            checkpoint = getattr(topic, "checkpoint", None)
```
wylswz: we could check the type of `topic` instead of using `getattr`

zl86790: Updated to check `StreamsTopic` explicitly instead of using `getattr`. Thanks :P

wylswz:
@QuantumGhost Could you review this stability enhancement on Redis message streaming?

api/services/app_generate_service.py
```diff
@@ -67,14 +79,16 @@ def _try_start() -> bool:
 
         channel_type = dify_config.PUBSUB_REDIS_CHANNEL_TYPE
         if channel_type == "streams":
-            # With Redis Streams, we can safely start right away; consumers can read past events.
+            start_id = topic.checkpoint() if isinstance(topic, StreamsTopic) else None
+
+            # With Redis Streams, we can safely start right away; consumers resume from `start_id`.
             _try_start()
```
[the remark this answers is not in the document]
zl86790: Addressed. In streams mode, we now require topic to be a StreamsTopic and raise TypeError otherwise, instead of silently falling back to start_id = None.

api/core/app/apps/streaming_utils.py
```diff
@@ -27,7 +28,16 @@ def stream_topic_events(
...
-    with topic.subscribe() as sub:
+    # If the caller already captured a checkpoint (e.g. because a background task was
+    # dispatched before this generator started running and may have published events in
+    # the meantime), resume from that point instead of the topic's default `subscribe()`,
+    # which would only pick up messages published after this line actually executes.
+    subscribe_from = getattr(topic, "subscribe_from", None)
+    if start_id is not None and subscribe_from is not None:
+        subscription = subscribe_from(start_id)
+    else:
+        subscription = topic.subscribe()
```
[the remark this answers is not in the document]
zl86790: Addressed. When start_id is provided, we now require subscribe_from to be callable and raise TypeError otherwise, instead of falling back to topic.subscribe().

api/services/app_generate_service.py
wylswz: Please also refactor [this line](https://github.com/langgenius/dify/pull/40964/changes#diff-e992af5aebf0c46a1a44d8bf5baad4aaed125de83361bdee7d775aa4ce827bfaR35), thank you.

QuantumGhost:
**AI disclosure**: This comment was drafted with Codex using GPT-5.6-sol.  I have reviewed it and am responsible for the content.

**UPDATE**: This is NOT enough to prevent race. See https://github.com/langgenius/dify/pull/40964#issuecomment-5360693653.


I agree with capturing the stream position before dispatching the Celery task, but I think the Redis Streams cursor/checkpoint details should remain inside the broadcast-channel abstraction.

After some thought, I would propose capturing the `start_id` inside the method `_StreamsSubscription._start_if_needed`. 

The suggested fix is like the following:

```py
class _StreamsSubscription(Subscription):
    def __init__(self, client: Redis | RedisCluster, key: str):
        self._client = client
        self._key = key
        self._start_id: bytes | str | None = None
        ...

    def _resolve_start_id(self) -> bytes | str:
        entries = self._client.xrevrange(self._key, count=1)
        if not entries:
            return "0-0"

        entry_id, _fields = entries[0]
        return entry_id

    def _start_if_needed(self) -> None:
        """This method must be called with `_lock` held."""
        if self._listener is not None or self._closed:
            return

        # Fix the logical read boundary synchronously before `__enter__`
        # returns and before the producer can be started.
        self._start_id = self._resolve_start_id()

        self._listener = threading.Thread(
            target=self._listen,
            name=f"redis-streams-sub-{self._key}",
            daemon=True,
        )
        self._listener.start()

    def _listen(self) -> None:
        start_id = self._start_id
        assert start_id is not None

        last_id = start_id
        ...
```


This approach minimizes change scope, avoids leaking unnecessary concepts into application services.

The race currently looks like this:

```
API thread                         Listener thread              Celery worker
──────────                         ───────────────              ─────────────
subscription.__enter__()
  start listener thread
return from __enter__()

on_subscribe()
  task.delay()
                                                                publish event
                                   XREAD stream $
                                   "$" resolves after event
                                   event is skipped
```

zl86790:
hi @QuantumGhost,
Thanks for the detailed suggestion, I've updated the code to follow your approach — mind taking another look when you get a chance?

QuantumGhost:
**AI disclosure**: This comment was drafted with Codex. I have reviewed it and am responsible for the content.

@zl86790 I need to correct my previous suggestion: I missed a race caused specifically by the lazy startup of the SSE iterator.

`_build_streaming_task_on_subscribe()` starts the fallback timer while constructing the response, but `stream_topic_events()` is a generator. Returning that generator does not enter the subscription. On the first iteration it yields the initial `PING`, and only when iteration resumes does it enter the subscription and invoke the hook.

This leaves the following race:

1. The fallback timer starts.
2. The iterator has not entered the subscription yet.
3. The timer dispatches the worker, which publishes the workflow events.
4. The iterator resumes and enters the Streams subscription.
5. `XREVRANGE` now observes the already-published events as the current tail, so they are treated as pre-subscription history and skipped.

Because of this iterator-startup gap, moving start-ID resolution into `_StreamsSubscription._start_if_needed()` was not sufficient.

I pushed a follow-up fix that obtains the topic subscriber view and prepares the Streams subscription before returning the iterator. `SupportsPreparedSubscription` now belongs to the `Subscriber` abstraction, and Redis Streams exposes it through a dedicated subscriber rather than `StreamsTopic`. Preparation fixes the delivery boundary synchronously without starting the listener. Streams no longer uses the fallback timer and starts the task from the hook after subscription entry; Pub/Sub and sharded Pub/Sub retain their existing fallback behavior.

QuantumGhost (approved):

fatelei (approved):

snaka:
**Verification on 1.17.0 (reporter of #40948)**

`git tag --contains 9d8dacd0c8` returns only `1.17.0`, so I measured the fix on that release.

Setup: local `docker compose` on the official images, `EVENT_BUS_REDIS_CHANNEL_TYPE=streams`
on `api` and `worker`, `CELERY_WORKER_AMOUNT=8`. Two canary workflow apps driven through
`POST /v1/workflows/run`; K=8 concurrent streaming requests per run, 15 s threshold on
`workflow_started`.

| version | runs with at least one failure | requests without `workflow_started` |
|---|---|---|
| 1.13.3 | 6/10 | 14/80 |
| 1.15.0 | 8/10 | 26/80 |
| 1.16.1 | 9/10 | 34/80 (+1 threshold exceeded) |
| **1.17.0** | **0/30** | **0/240** |

All 240 requests received `workflow_started` (334 ms worst case, ~105 ms warm). A sequential
smoke check of 45 iterations also passed with no failures.

Two caveats. The pre-1.17.0 rates swing with load and warm-up (an earlier session measured
1.13.3 at 10/10 runs, 44/80 requests), so they are not a ranking between releases — only
evidence that every release before 1.17.0 reproduced the loss in most runs. And two runs in
my first 1.17.0 batch failed with all 8 requests hitting the threshold together, which is not
the race signature; `workflow_runs` showed a preceding batch of ~60 s canary workflows still
holding all 8 Celery slots. Batches from a warm stack were 0/10 and 0/20.

Thanks @zl86790 for the fix and @QuantumGhost for catching the iterator-startup gap — that
follow-up is what closed the window in our workload.

PR #40981, opened 2026-08-19
fix: preserve early workflow events in streams transport

### LINKED ISSUE #40948: [Bug] workflow_started is dropped for concurrent streaming runs with the streams event transport (regression of #32518 via #34030)

Zhengcy05:
> [!IMPORTANT]
>

## Summary

Fixes #40948.

Redis Streams subscriptions currently start from `$`, while streaming workflow tasks are enqueued before the listener performs its first `XREAD`. A fast worker can therefore publish `workflow_started` and early node events before `$` is resolved, causing those events to be skipped even though they remain stored in Redis.

This PR captures the stream tip before dispatching a fresh workflow task and uses it as that subscription's starting cursor. Subscriptions without an explicit cursor continue to default to `$`, preserving the live-only behavior introduced to avoid replaying stale `workflow_paused` events in #34040.

Changes include:

- Add a cursor-capable topic protocol and per-subscription starting cursors for Redis Streams.
- Treat a missing or empty stream as cursor `0-0`, since workflow streams are created lazily by the first `XADD`.
- Capture the stream tip before Celery task dispatch for fresh Workflow and Advanced Chat runs.
- Pass the captured topic and cursor through the event generator to the final Streams subscription.
- Correct the Streams test double so `$` only observes entries appended after the initial read.
- Add regression coverage for explicit cursors while retaining coverage for the default `$` behavior.

Validation:

- `uv run --project api pytest -q api/tests/unit_tests/libs/broadcast_channel/redis/test_streams_channel_unit_tests.py api/tests/unit_tests/services/test_app_generate_service_streaming_integration.py api/tests/unit_tests/services/test_app_generate_service.py` — 71 passed.
- Focused Ruff checks passed for all changed files.
- Ruff format checks passed for all changed files.
- Focused Pyrefly checks passed for all changed production files.
- Docker-backed integration tests were not run locally because they are CI-only.

## Screenshots

N/A — backend event-delivery fix.

## Checklist

- [ ] This change requires a documentation update, included: [[Dify Document](https://github.com/langgenius/dify-docs)](https://github.com/langgenius/dify-docs)

### DIFF
--- a/api/core/app/apps/message_based_app_generator.py
+++ b/api/core/app/apps/message_based_app_generator.py
@@ -321,10 +321,14 @@ class MessageBasedAppGenerator(BaseAppGenerator):
         workflow_run_id: str,
         idle_timeout=300,
         on_subscribe: Callable[[], None] | None = None,
+        *,
+        topic: Topic | None = None,
+        cursor: str | None = None,
     ) -> Generator[Mapping | str, None, None]:
-        topic = cls.get_response_topic(app_mode, workflow_run_id)
+        topic = topic if topic is not None else cls.get_response_topic(app_mode, workflow_run_id)
         return stream_topic_events(
             topic=topic,
             idle_timeout=idle_timeout,
             on_subscribe=on_subscribe,
+            cursor=cursor,
         )
--- a/api/core/app/apps/message_generator.py
+++ b/api/core/app/apps/message_generator.py
@@ -28,12 +28,16 @@ class MessageGenerator:
         ping_interval: float = 10.0,
         on_subscribe: Callable[[], None] | None = None,
         terminal_events: Iterable[str | StreamEvent] | None = None,
+        *,
+        topic: Topic | None = None,
+        cursor: str | None = None,
     ) -> Generator[Mapping | str, None, None]:
-        topic = cls.get_response_topic(app_mode, workflow_run_id)
+        topic = topic if topic is not None else cls.get_response_topic(app_mode, workflow_run_id)
         return stream_topic_events(
             topic=topic,
             idle_timeout=idle_timeout,
             ping_interval=ping_interval,
             on_subscribe=on_subscribe,
             terminal_events=terminal_events,
+            cursor=cursor,
         )
--- a/api/core/app/apps/streaming_utils.py
+++ b/api/core/app/apps/streaming_utils.py
@@ -6,7 +6,7 @@ from collections.abc import Callable, Generator, Iterable, Mapping
 from typing import Any
 
 from core.app.entities.task_entities import StreamEvent
-from libs.broadcast_channel.channel import Topic
+from libs.broadcast_channel.channel import CursorTopic, Topic
 from libs.broadcast_channel.exc import SubscriptionClosedError
 
 
@@ -17,6 +17,7 @@ def stream_topic_events(
     ping_interval: float | None = None,
     on_subscribe: Callable[[], None] | None = None,
     terminal_events: Iterable[str | StreamEvent] | None = None,
+    cursor: str | None = None,
 ) -> Generator[Mapping[str, Any] | str, None, None]:
     # send a PING event immediately to prevent the connection staying in pending state for a long time.
     #
@@ -27,7 +28,14 @@ def stream_topic_events(
     terminal_values = _normalize_terminal_events(terminal_events)
     last_msg_time = time.time()
     last_ping_time = last_msg_time
-    with topic.subscribe() as sub:
+    if cursor is None:
+        subscription = topic.subscribe()
+    else:
+        if not isinstance(topic, CursorTopic):
+            raise ValueError("The broadcast topic does not support cursor-based subscriptions")
+        subscription = topic.subscribe(cursor=cursor)
+
+    with subscription as sub:
         # on_subscribe fires only after the Redis subscription is active.
         # This is used to gate task start and reduce pub/sub race for the first event.
         if on_subscribe is not None:
--- a/api/libs/broadcast_channel/channel.py
+++ b/api/libs/broadcast_channel/channel.py
@@ -8,7 +8,7 @@ import types
 from abc import abstractmethod
 from collections.abc import Iterator
 from contextlib import AbstractContextManager
-from typing import Protocol, Self, override
+from typing import Protocol, Self, override, runtime_checkable
 
 
 class Subscription(AbstractContextManager["Subscription"], Protocol):
@@ -115,6 +115,22 @@ class Topic(Producer, Subscriber, Protocol):
     @abstractmethod
     def as_subscriber(self) -> Subscriber:
         """as_subscriber create a read-only view for this topic."""
+        ...
+
+
+@runtime_checkable
+class CursorTopic(Topic, Protocol):
+    """A durable topic that can resume a subscription from a captured cursor."""
+
+    @abstractmethod
+    def latest_cursor(self) -> str:
+        """Return the latest published cursor, or the transport's beginning cursor when empty."""
+        ...
+
+    @override
+    @abstractmethod
+    def subscribe(self, *, cursor: str | None = None) -> Subscription:
+        """Subscribe after ``cursor``; omit it to retain the transport's live-subscription behavior."""
         ...
 
 
--- a/api/libs/broadcast_channel/redis/streams_channel.py
+++ b/api/libs/broadcast_channel/redis/streams_channel.py
@@ -70,16 +70,24 @@ class StreamsTopic:
     def as_subscriber(self) -> Subscriber:
         return self
 
-    def subscribe(self) -> Subscription:
-        return _StreamsSubscription(self._client, self._key)
+    def subscribe(self, *, cursor: str | None = None) -> Subscription:
+        return _StreamsSubscription(self._client, self._key, cursor="$" if cursor is None else cursor)
+
+    def latest_cursor(self) -> str:
+        entries = self._client.xrevrange(self._key, count=1)
+        if not entries:
+            return "0-0"
+        entry_id, _ = entries[0]
+        return entry_id.decode() if isinstance(entry_id, bytes) else str(entry_id)
 
 
 class _StreamsSubscription(Subscription):
     _SENTINEL = object()
 
-    def __init__(self, client: Redis | RedisCluster, key: str):
+    def __init__(self, client: Redis | RedisCluster, key: str, *, cursor: str = "$"):
         self._client = client
         self._key = key
+        self._cursor = cursor
 
         self._queue: queue.Queue[object] = queue.Queue()
 
@@ -104,10 +112,7 @@ class _StreamsSubscription(Subscription):
         # since this method runs in a dedicated thread, acquiring `_lock` inside this method won't cause
         # deadlock.
 
-        # Setting initial last id to `$` to signal redis that we only want new messages.
-        #
-        # ref: https://redis.io/docs/latest/commands/xread/#the-special--id
-        last_id = "$"
+        last_id = self._cursor
         try:
             while True:
                 with self._lock:
--- a/api/services/app_generate_service.py
+++ b/api/services/app_generate_service.py
@@ -23,6 +23,7 @@ from core.app.layers.pause_state_persist_layer import PauseStateLayerConfig
 from core.db import session_factory
 from enums import DeploymentEdition, QuotaType
 from extensions.otel import AppGenerateHandler, trace_span
+from libs.broadcast_channel.channel import CursorTopic, Topic
 from models.model import Account, App, AppMode, EndUser
 from models.workflow import Workflow, WorkflowRun
 from services.errors.app import QuotaExceededError, WorkflowIdFormatError, WorkflowNotFoundError
@@ -45,7 +46,7 @@ class AppGenerateService:
         """
         Build a subscription callback that coordinates when the background task starts.
 
-        - streams transport: start immediately (events are durable; late subscribers can replay).
+        - streams transport: start immediately after the caller captures the topic cursor.
         - pubsub/sharded transport: start on first subscribe, with a short fallback timer so the task
           still runs if the client never connects.
         """
@@ -67,7 +68,6 @@ class AppGenerateService:
 
         channel_type = dify_config.PUBSUB_REDIS_CHANNEL_TYPE
         if channel_type == "streams":
-            # With Redis Streams, we can safely start right away; consumers can read past events.
             _try_start()
 
             # Keep return type Callable[[], None] consistent while allowing an extra (no-op) call.
@@ -86,6 +86,16 @@ class AppGenerateService:
                 timer.cancel()
 
         return _on_subscribe
+
+    @classmethod
+    def _prepare_streaming_task(
+        cls,
+        topic: Topic,
+        start_task: Callable[[], None],
+    ) -> tuple[Callable[[], None], str | None]:
+        """Capture a durable topic's cursor before the task can publish its first event."""
+        cursor = topic.latest_cursor() if isinstance(topic, CursorTopic) else None
+        return cls._build_streaming_task_on_subscribe(start_task), cursor
 
     @classmethod
     @trace_span(AppGenerateHandler)
@@ -255,14 +265,17 @@ class AppGenerateService:
                     def on_subscribe():
                         workflow_based_app_execution_task.delay(payload_json)
 
-                    on_subscribe = cls._build_streaming_task_on_subscribe(on_subscribe)
                     generator = AdvancedChatAppGenerator()
+                    topic = generator.get_response_topic(AppMode.ADVANCED_CHAT, payload.workflow_run_id)
+                    on_subscribe, cursor = cls._prepare_streaming_task(topic, on_subscribe)
                     return rate_limit.generate(
                         generator.convert_to_event_stream(
                             generator.retrieve_events(
                                 AppMode.ADVANCED_CHAT,
                                 payload.workflow_run_id,
                                 on_subscribe=on_subscribe,
+                                topic=topic,
+                                cursor=cursor,
                             ),
                         ),
                         request_id=request_id,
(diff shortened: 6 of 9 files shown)


snaka:
I filed #40948, and this is the fix I hoped for. I verified it on 1.16.1 and I support merging it.

**Result.** I built this branch's production changes into a 1.16.1 image and ran the #40948 reproduction, alternating the patched and stock images inside a single session.

| Image | Rounds | Rounds failed | Requests that never received `workflow_started` |
|---|---|---|---|
| 1.16.1 + this PR | 130 | 0 | 0 / 1040 |
| 1.16.1 stock | 40 | 36 | 127 / 320 |

A *round* is K=8 requests fired simultaneously, and it fails if any of the 8 never receives `workflow_started` within 15 s. `EVENT_BUS_REDIS_CHANNEL_TYPE=streams`, `CELERY_WORKER_AMOUNT=8`, api and worker in separate containers, the Start → End workflow from #40948.

Two findings worth adding to the issue record:

- **The loss is not limited to concurrent bursts.** Sending requests one at a time (15 per run, 5 runs), 1 of 75 sequential requests on the stock image never received `workflow_started`. On the patched image 0 of 75, and all 75 reached `workflow_finished` with no inter-event gap over the silence budget. That last part mattered to check, since a cursor-based subscription now reads the stream from its beginning instead of tailing it.
- **#34030 does not regress.** Every reconnect path still passes `cursor=None` and therefore still starts at `$` (table below).

Two caveats up front.

What I measured is a `git apply -3` backport of the six production files onto the `1.16.1` tag, not this PR's head.

And a note on how this comment was produced. The numbers above and in the collapsed sections come from runs I set up and repeated, so I can stand behind them. Everything that is a claim about the code — the `#34030` bullet above, the request and the suggestion below, and the collapsed sections — came out of an AI-assisted analysis that I have not independently verified myself. Please read those as a starting point for your own review rather than as something I can vouch for, and correct me where I have it wrong.

## One request

`_FakeStreams.xread`'s new `$` handling freezes the snapshot:

```python
if last_id == "$":
    start = self._dollar_snapshots.setdefault(key, len(entries))
```

Because of `setdefault` the snapshot is taken at the *first* XREAD for a key; real Redis re-resolves `$` on every call. That difference is observable, because `_StreamsSubscription._listen` keeps issuing XREAD with `last_id == "$"` until the first entry arrives. Against a real server, anything XADDed between one blocking XREAD returning nil and the next one reaching the server is lost; the double delivers it.

This does not affect the cursor path, where `last_id` is a concrete ID from the start. It does mean the double still cannot exercise that window on the paths that keep using `$`, which is every reconnect endpoint. Would you consider resolving `$` per call, so the double matches the semantics the PR description claims?

## One suggestion

The fix rests on `latest_cursor()` running before `_try_start()`. In `_prepare_streaming_task` those are two adjacent statements, and swapping them would silently disable the fix while every assertion still passes — except `test_streams_full_flow_prepublish_and_replay`, which does catch it. The two new unit tests in `test_app_generate_service.py` do not, since they only assert the cursor value that reached `retrieve_events`. A one-line comment stating that the order is load-bearing would make it harder to lose in a future refactor.

<details>
<summary>What exactly was built and measured</summary>

A plain `git apply` of this PR onto `1.16.1` fails on `api/services/app_generate_service.py`: your base has `from enums import DeploymentEdition, QuotaType` as a context line, and 1.16.1 still has `from enums.quota_type import QuotaType`. I used `git apply -3`.

Of the six production files this PR touches:

- four are byte-identical between `1.16.1` and `main`
- `app_generate_service.py` differs by two unrelated lines, both about billing
- `streams_channel.py` differs by one hunk in `_listen`: on 1.16.1 a `SIG_CLOSE` entry always breaks the loop, whereas on `main` a close signal emitted by another subscription is ignored

The last one is the only place where the tested build could plausibly behave differently from this PR on `main`. Since a cursor-based subscription now reads the stream from its beginning, the tested build is the more fragile of the two, not the less. I mention it so the result is not over-read: this is evidence about 1.16.1 plus these hunks.

Before every block I verified the container image IDs, `EVENT_BUS_REDIS_CHANNEL_TYPE` on both api and worker, and that nginx resolved the api service to the recreated container's actual IP. No block was discarded or re-run.

I also confirmed the transport really was Redis Streams rather than a silent fallback to pub/sub, which would produce zero failures for an unrelated reason (there the task start is gated on subscribe, so the race cannot occur). During a patched run, 40 requests created exactly 40 new `stream:channel:workflow:<uuid>` keys, and the ones I sampled were `TYPE=stream` with `XLEN=8`.

</details>

<details>
<summary>Reconnect paths, i.e. why #34030 does not regress</summary>

I traced every caller of `retrieve_events`. Only the two fresh-dispatch sites pass a cursor:

| Caller | cursor | Start position |
|---|---|---|
| `app_generate_service.py` `_dispatch_generate`, ADVANCED_CHAT | captured | effectively `0-0` |
| `app_generate_service.py` `_dispatch_generate`, WORKFLOW | captured | effectively `0-0` |
| `app_generate_service.py` `get_response_generator` | none | `$` |
| `controllers/console/human_input_form.py` | none | `$` |
| `controllers/web/workflow_events.py` | none | `$` |
| `controllers/service_api/app/workflow_events.py` | none | `$` |
| `controllers/openapi/workflow_events.py` | none | `$` |

A client reconnecting after `workflow_paused` goes through the lower five, so the stale `workflow_paused` is not replayed. On the fresh-dispatch side the cursor is effectively `0-0`, but the key is new, so there is no past event to replay.

`workflow_based_app_execution_task.delay` is called from only the two places this PR patches, so both enqueue-then-subscribe entry points are covered.

</details>

<details>
<summary>What the numbers do and do not support</summary>

The 8 requests in a round share one load spike, so they are not independent trials; treating 1040 requests as 1040 trials would overstate the precision. I report rounds.

At the stock per-round failure rate of 0.9, 130 consecutive passing rounds has probability on the order of 1e-130, so "the patch changed nothing" is excluded. For the positive claim, 0 failures in 130 rounds bounds the residual per-round failure rate at **2.3%** (95%, rule of three); over the five interleaved blocks alone the bound is 3.0%. So the honest statement is "not observed in 130 rounds under conditions where the stock image fails 90% of rounds", not "proven to be zero".

The first round after a container recreate is a warm-up round that always passes on both images and runs 3 to 8 times slower than the rest. There were 4 such rounds on each side. Dropping them from both, the stock image failed 36 of 36 informative rounds and the patched image failed 0 of 126.

One qualitative point I found more convincing than the arithmetic: on the stock image the rounds that pass are exactly the slow ones. Failing rounds complete in roughly 200 to 300 ms, while the warm-up rounds that pass take over a second. Every patched round after warm-up completed in that same fast band, i.e. in the regime where the stock image fails. The result is therefore not an artifact of the patched blocks running under lighter load.

Limits I would rather name than have assumed away:

- The K=8 harness aborts each SSE connection as soon as `workflow_started` arrives, so the 130-round figure says nothing about what happens after that point. The sequential check covers the end-to-end path, but only over 75 requests.
- I did not systematically check for duplicated events. I read one patched SSE stream to completion as a spot check and saw exactly one `workflow_started` and one `workflow_finished`, but that is n=1 with no stock-image comparison.
- I have not run the API test suite in this environment, so I cannot report on the unit tests. Everything above about tests is a source read.

</details>

<details>
<summary>Two minor observations, no action needed from me</summary>

`stream_topic_events` raises `ValueError` when it receives a cursor for a non-`CursorTopic`. Both call sites capture the cursor from the very topic instance they then pass in, so the branch is unreachable today. If it ever became reachable the failure would not be a 500: `stream_topic_events` is a generator whose first statement is `yield StreamEvent.PING.value`, so the exception surfaces on the second `next()`, after the status line, headers and a PING have been written, and the client sees a truncated SSE stream. `retrieve_events(app_mode, run_id, cursor="...")` with `topic=None` is accepted by the signature and would land there under a non-streams transport.

At both call sites `workflow_run_id` is a fresh `uuid.uuid4()`, so the stream key cannot exist yet and `latest_cursor()` always returns `"0-0"` — `subscribe(cursor="0-0")` would behave identically today. If the generality is there for a future in which a run ID can be reused, or for a caller that captures a cursor mid-stream, a comment saying so would help, since the current call sites do not motivate the extra XREVRANGE on the request hot path.

</details>

Zhengcy05:
Thanks for the thorough verification and for catching this. 
You were right: `setdefault` froze `$` across calls. I updated both test doubles to resolve `$` per `XREAD` while preserving blocking behavior, added regression coverage for the between-call window, and documented the load-bearing cursor-before-dispatch order.

Validation: 73 focused tests passed, along with Ruff and Pyrefly checks.

QuantumGhost:
@Zhengcy05 Thank you for your contribution.

This proposal seems quite similar to #40964. However, it breaks the `BroadcastChannel` abstraction by introducing an incompatible `subscribe` method signature.

I’ve reviewed #40964 and pushed my adjustments there. Could you take a look at that PR?

I’ll close this PR in favor of #40964.

PR #40982, opened 2026-08-19
feat(cli): add intent-based command search

### LINKED ISSUE #41035: Add intent-based command discovery to difyctl
### Self Checks

Agents currently load the complete `difyctl help -o json` command tree to discover a single command. This sends unnecessary CLI output into the agent context and makes command discovery less targeted.

Add `difyctl help -o json --compact`: the complete command map carrying only `command`, `description`, and `effect` per entry — about 4 KB, against ~45 KB for the full `difyctl help -o json` tree.

The Agent Skill then reads that map, selects one command, and inspects it with `difyctl help <path> -o json`, which remains the only surface that carries args and flags. Two calls per task, and no path where an agent names a flag it never fetched.

An intent-ranked `difyctl search` was prototyped and measured against this approach: 240 paired agent runs, 4 models, 18 tasks (direct hits, paraphrase, morphology, typos, ambiguous, unsupported). Every strategy was equally correct; they differ in cost per task.

| discovery strategy | correct | input footprint |
| --- | ---: | ---: |
| compact map, then per-command help | 60/60 | 116,963 |
| search first, compact fallback | 60/60 | 125,922 |
| search first, full-tree fallback | 60/60 | 136,967 |
| full tree (current behaviour) | 60/60 | 160,450 |

At the current 28-command surface the extra search round trip costs more than the 4 KB map it avoids, so this issue is now scoped to the compact map. Ranked search is deferred until the command surface is large enough to pay for the extra hop; extrapolating from the numbers above that is on the order of a few hundred commands, and it also depends on ranking precision rather than command count alone.

samzong:
## Summary

- add `difyctl search "<intent>"` over live command help metadata
- update the Agent Skill to discover candidates with search and verify them with per-command help
- cover ranking, generated registry wiring, Skill guidance, and agent workflow behavior

Issue: none; this draft is opened for maintainer discussion.

## Screenshots

Not applicable — CLI-only change.

## Checklist

- [x] Documentation is updated in `cli/README.md`.
- [x] I understand this PR may be closed without prior discussion or an assigned issue.
- [x] Tests cover each introduced behavior and the change is atomic.
- [x] `pnpm tree:check`, `pnpm test`, local agent-workflow E2E, `pnpm build`, and staged-path `vp check` passed.

Fixes #41035

### DIFF
--- a/cli/src/help/skill-template.ts
+++ b/cli/src/help/skill-template.ts
@@ -1,15 +1,3 @@
-// The difyctl agent skill, in full — one hand-authored, pure-delegation file.
-//
-// It inlines NO command list, NO flag list, and ships no reference files. The
-// command surface is discovered at runtime via `difyctl help -o json`, so this
-// template has nothing to derive from the binary and nothing that can drift
-// from it. `{{VERSION}}` is the only substitution; it is filled at emit time by
-// renderSkill() in ./skill.
-//
-// RED LINE: keep this pure delegation. The moment a command or flag listing is
-// added here, the embedded static copy can fall out of sync with the command
-// surface — at which point it must instead be generated from the command model
-// with a snapshot test (see SKILL-SPEC.md §10, decision D2).
 export const SKILL_TEMPLATE = `---
 name: difyctl
 description: Drive the difyctl CLI to manage Dify apps, workspaces, members and runs. Use when the task involves difyctl or operating a Dify instance from the command line.
@@ -20,10 +8,21 @@ description: Drive the difyctl CLI to manage Dify apps, workspaces, members and 
 difyctl is self-describing — do not guess commands.
 
 ## Discover the command surface
-Run \`difyctl help -o json\` for the version-current map: every command
-(args, flags, examples, \`effect\`) plus the global \`contract\` (exit codes,
-output formats, error envelope, HITL protocol). Treat that JSON as the
-source of truth; this file only bootstraps you into it.
+1. Find candidates:
+   \`difyctl search "<intent>" -o json\`
+   Treat hits as candidates only — do not execute from this output.
+2. Inspect the chosen command:
+   \`difyctl help <path> -o json\`
+   Expand a multi-token path as separate shell arguments; never quote the
+   entire returned path.
+   That JSON is the source of truth for args, flags, examples, \`effect\`,
+   and \`agentGuide\`.
+
+When the task depends on global exit codes, error envelopes, HITL, or retry
+semantics, read \`difyctl help agent\`.
+
+Only when search returns no viable candidate, fall back to
+\`difyctl help -o json\`. Do not run the full sitemap after useful search results.
 
 ## The one non-obvious thing: HITL pauses are not failures
 A run can pause for human input. It exits with **code 0** and emits a
@@ -31,8 +30,8 @@ A run can pause for human input. It exits with **code 0** and emits a
 Resume as the payload instructs (see \`difyctl resume app --help\`).
 
 ## Before any write/destructive action
-Check the command's \`effect\` (\`read\` / \`write\` / \`destructive\`) in
-\`difyctl help -o json\` before running it.
+Never execute a write/destructive candidate from search output. Treat the
+inspected command's \`effect\` as the confirmation gate before running it.
 
 ---
 difyctl skill v{{VERSION}} — if \`difyctl version\` differs, re-run
--- a/cli/test/e2e/suites/agent/agent-skill-workflow.e2e.ts
+++ b/cli/test/e2e/suites/agent/agent-skill-workflow.e2e.ts
@@ -3,7 +3,7 @@
  *
  * Scenario: an AI agent has loaded difyctl SKILL.md and drives difyctl to:
  *   1. Bootstrap  - read SKILL.md via `skills install --stdout`
- *   2. Discover   - `help -o json` for full command surface + contract
+ *   2. Discover   - `search` for candidates, then per-command help
  *   3. Auth check - no token → exit 4 + JSON error envelope
  *   4. Discover apps    - `get app -o json`
  *   5. Describe app     - `describe app <id> -o json`
@@ -14,7 +14,7 @@
  *  10. Pipeline safety  - no ANSI/spinner, stdout/stderr separation
  *
  * PRD: §3 Agent-Driven, §5 Agent onboarding, Req 1.3/2.1/2.3/3.1-3.3
- * Agent Skills PRD: §4/§5.2 SKILL.md → help -o json discovery pattern
+ * Agent Skills PRD: §4/§5.2 SKILL.md → live command discovery
  *
  * Groups 1-3, 9: no auth required (local mode compatible)
  * Groups 4-8, 10: require DIFY_E2E_TOKEN / staging — wrapped with optionalIt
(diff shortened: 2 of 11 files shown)


samzong:
## Four-model agent workflow A/B report

### Recommendation

Keep the `search` command and the final search-first Agent Skill workflow.

The evaluation covered four models, six command-discovery tasks, and two workflows for 48 isolated Cursor Agent runs. Both workflows selected the expected command path in every run. The measurable benefit is lower context and discovery overhead rather than higher accuracy on this already-saturated task set.

### Aggregate results

| Metric | Help-only baseline | Search loop | Change |
| --- | ---: | ---: | ---: |
| Correct command path | 24/24 | 24/24 | unchanged |
| Runs loading full `help -o json` | 24/24 | 0/24 | eliminated |
| Runs loading per-command help | 23/24 | 24/24 | complete verification |
| Average CLI output | 45,510 B | 4,344 B | -90.5% |
| Average input footprint | 188,893 tokens | 137,959 tokens | -27.0% |
| Median input footprint | 162,870 tokens | 119,242 tokens | -26.8% |
| Average duration | 29.5 s | 24.8 s | -15.9% |
| Median duration | 29.1 s | 23.4 s | -19.4% |
| Average model output | 1,208 tokens | 833 tokens | -31.0% |
| Average tool calls | 5.54 | 4.00 | -27.8% |
| Average CLI calls | 2.38 | 2.38 | unchanged |

`Input footprint` is `inputTokens + cacheReadTokens + cacheWriteTokens` from Cursor Agent usage. It is useful for within-evaluation comparisons and is not a billing-token measurement.

### Per-model results

| Model | Accuracy | Input footprint | CLI output | Average duration |
| --- | ---: | ---: | ---: | ---: |
| Claude Sonnet 5 High | 6/6 → 6/6 | -23.6% | -87.7% | -10.6% |
| GPT-5.6 Sol High Fast | 6/6 → 6/6 | -24.1% | -90.7% | -25.0% |
| Cursor Grok 4.6 High Fast | 6/6 → 6/6 | -32.9% | -91.3% | -26.6% |
| GPT-5.6 Luna High Fast | 6/6 → 6/6 | -29.8% | -91.2% | +3.1% |

Luna's median duration improved by 6.9% despite its average increasing by 3.1%. With one sample per model/task/workflow combination, latency is not statistically significant; CLI bytes and token footprint are the stronger signals.

### Agent-workflow test cases

Each case was run once per model against both the help-only baseline and the search loop. Agents were required to return the command path, effect, required arguments, and relevant flags without reading source code or executing the business command.

| Case | Prompt | Expected path |
| --- | --- | --- |
| Export definition | Export the editable Dify Studio app definition as a YAML backup. | `export studio-app` |
| Import definition | Create or update an editable Studio app from a Dify DSL YAML file. | `import studio-app` |
| Inspect input schema | Obtain a published app's mode and complete input schema before running it. | `describe app` |
| Resume human input | Continue a workflow app paused for a human-input form submission. | `resume app` |
| Revoke devices | Revoke one session device, with an option to revoke all session devices. | `auth devices revoke` |
| Switch account | Switch the active account when several Dify accounts are authenticated locally. | `use account` |

The final search-loop matrix used `search` in 24/24 runs, exact per-command help in 24/24, full sitemap help in 0/24, `help agent` in 0/24, and business-command execution in 0/24.

### Automated test cases and verification

`cli/src/commands/search/search.test.ts` covers:

- path matches ranking above description matches, including simple plural folding;
- flag and agent-guide token matches without repeated-word inflation;
- deterministic path ordering for equal scores;
- empty results for intents with no searchable tokens.

`cli/test/e2e/suites/agent/agent-skill-workflow.e2e.ts` covers:

- the installed Skill's search → per-command-help workflow;
- zero command-tree enumeration in the Skill;
- `search` ranking `export studio-app` first for `export app`;
- candidate output containing path, description, effect, and score while excluding flags and agent guides;
- machine-readable help and effect metadata used by the confirmation gate.

Fresh verification on the committed diff:

- `pnpm tree:check` — passed;
- `pnpm test` — 1,252 passed, 26 skipped;
- local agent-workflow E2E — 62 passed, 29 credential-dependent cases skipped;
- `pnpm build` — passed;
- `vp check` on all staged CLI paths — passed with no formatting, lint, or type errors.

### Interpretation limits

These tasks are correctly solvable from the existing sitemap, so the evaluation demonstrates that search preserves command-selection accuracy while reducing context cost. It does not establish an accuracy improvement for ambiguous, synonym-heavy, no-match, or noun-collision intents. Those require a larger blind ranking set with top-1, top-3, and MRR measurements.

wylswz:
Since the search is based on token matching, will stemming help improving the searching performance?

samzong:
@wylswz Good point! we can improve this with broader stemming later.

lin-snow:
> 由于搜索基于令牌匹配，词干提取是否有助于提升搜索表现？

@samzong @wylswz 

I looked through this PR, and it feels like the core idea is essentially still progressive disclosure: controlling the amount of context the Agent receives at each step to avoid context growth as the number of CLI commands increases.

A couple of points I’d like to discuss:

1. The current eval mainly demonstrates that search performs better than the full sitemap approach. However, when the command surface is relatively small (e.g. only a few dozen commands), would a compact sitemap already be sufficient? In that case, the context reduction benefit from search may be limited, while it also introduces an additional retrieval/ranking layer. It would be interesting to validate whether that tradeoff is still worthwhile.

2. The current skill design uses a search-first strategy, but should we consider the uncertainty of the search results themselves? If search returns a candidate that looks plausible but is actually incorrect, and prevents the Agent from seeing the full command list, it may reduce the model’s own semantic reasoning capability. Would it make sense to fallback to a compact sitemap when confidence is low, no results are returned, or when the command surface is still small?

It might be useful to add a compact sitemap baseline, along with some adversarial cases (e.g. synonyms, paraphrased intents, queries with low vocabulary overlap) to verify whether search still provides a net benefit at the current scale.

lin-snow:
@samzong @wylswz 

I think the search direction is valuable. At its core, it is solving the progressive disclosure and context control problem. Beyond that, it may also be worth strengthening the security semantics of the commands themselves. In an Agent scenario, after finding the right command, accurately understanding the execution risk is also an important part of the workflow.

Currently, the effect metadata seems to have some coverage gaps. For example, some commands that actually modify server-side state do not declare an effect and may be interpreted as read operations by default, while semantically similar commands are already explicitly marked as write. Cases like this could potentially affect the Agent's confirmation flow.

Additionally, for some high-risk commands, adding dry-run support could be another direction: allowing the Agent to preview the potential changes before execution, and then complete the workflow with user confirmation.

So besides continuing to improve search/retrieval, it may also be worth considering command metadata (effect) and dry-run capabilities together. These improvements could further make the CLI more Agent-friendly. 🤔

cli/test/e2e/suites/agent/agent-skill-workflow.e2e.ts
```diff
@@ -14,7 +14,7 @@
  *  10. Pipeline safety  - no ANSI/spinner, stdout/stderr separation
  *
  * PRD: §3 Agent-Driven, §5 Agent onboarding, Req 1.3/2.1/2.3/3.1-3.3
- * Agent Skills PRD: §4/§5.2 SKILL.md → help -o json discovery pattern
+ * Agent Skills PRD: §4/§5.2 SKILL.md → live command discovery
```
lin-snow, line 17: BTW, this comment seems to include some details from the PRD that are unrelated to the code context.

cli/src/help/skill-template.ts
```diff
@@ -20,19 +8,30 @@ description: Drive the difyctl CLI to manage Dify apps, workspaces, members and
...
+   Expand a multi-token path as separate shell arguments; never quote the
+   entire returned path.
+   That JSON is the source of truth for args, flags, examples, \`effect\`,
+   and \`agentGuide\`.
+
+When the task depends on global exit codes, error envelopes, HITL, or retry
+semantics, read \`difyctl help agent\`.
+
+Only when search returns no viable candidate, fall back to
+\`difyctl help -o json\`. Do not run the full sitemap after useful search results.
```
lin-snow, line 25: > Do not run the full sitemap after useful search results.

This is the strategy I was referring to in my previous comment.

PR #40986, opened 2026-08-20
fix(ci): run Playwright E2E on GitHub Ubuntu

WH-2099:
## Summary

- Run the Playwright full-stack E2E job on the GitHub Ubuntu 24.04 image.
- Keep the standard Playwright system-dependency installation and the existing 15-minute step bound.
- Avoid the Depot apt mirror that prevented browsers from being installed before any scenario ran.

## Evidence

- Depot timed out once at 15 minutes and again at 30 minutes while fetching Ubuntu packages at roughly 22 KB/s.
- Skipping system dependencies installed both browser shells in 7 seconds and the Chromium suite passed, proving the test harness itself is healthy; WebKit then correctly reported its required GTK and GStreamer libraries.
- The GitHub Ubuntu runner provides the supported apt environment without maintaining a repository-owned package list or mirror rewrite.

## Validation

- The final diff is one workflow line.
- git diff --check passes.
- Full-Stack E2E CI is the behavioral gate for Chromium and WebKit.


hjlarry (dismissed):

lyzno1:
We should ask the upstream Depot team for a proper fix, or switch back to the native runner. Increasing the timeout would only mask the underlying issue. Taking 30 minutes to download Playwright is clearly unacceptable.

hjlarry:
30m  also not work...

crazywoola (approved):

PR #41018, opened 2026-08-20
chore: when no tool_call response raise exception

### LINKED ISSUE #40672: FC agent runner: detect zero-tool-call terminal answers when a tool call was expected, instead of silently accepting a fabricated response
### Self Checks

- [X] I have read the [Contributing Guide](<https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md>) and [Language Policy](<https://github.com/langgenius/dify/issues/1542>).
- [X] I have searched for existing issues [search for existing issues](<https://github.com/langgenius/dify/issues>), including closed ones.
- [X] I confirm that I am using English to submit this report, otherwise it will be closed.

### 1\. Is this request related to a challenge you're experiencing? Tell me about your story.

Follow-up to langgenius/dify#40671, where the root cause of a tool-call fabrication bug <br>was diagnosed down to the runner logic, confirmed by @dosu against the <br>actual fc_agent_runner.py source.

In the FC agent runner's main loop, function_call_state defaults to False <br>at the top of each iteration and only flips to True if check_tool_calls() <br>or check_blocking_tool_calls() detects a tool call in the model's <br>response. If the model returns plain text instead, the loop exits after <br>one iteration, and that text becomes final_answer with no validation that <br>a tool call was actually expected or required.

The signal that this happened is already present — tool_responses stays <br>empty and function_call_state stays False — it just isn't acted on. <br>There's currently no way to distinguish "the agent had nothing to do" <br>from "the agent was supposed to call a tool and didn't."

We hit this in production: an agent configured to delete records via a <br>tool call sometimes fabricated a "deleted successfully" message with zero <br>actual tool invocation in the trace. We've had to permanently disable the <br>affected tools rather than risk this — which is exactly the kind of gap <br>this request is meant to close.

**What I'd like to see:** a way to mark specific tools (or agent <br>configurations generally) as requiring a tool call when invoked in a <br>context that expects one, and detect when a terminal answer is produced <br>with zero tool invocations in that context. When detected, expose this as <br>a distinct error or retry path, rather than silently returning the <br>model's fabricated text as a successful answer. A structural fix at the <br>runner level would let us (and presumably others) re-enable delete/write-<br>action tools with confidence, rather than relying entirely on the <br>underlying model's tool-calling reliability.

### 2\. Additional context or comments

Alternatives already tried, none of which address the underlying gap:

* System prompt enforcement ("You MUST call the tool") — reduces but does <br>not eliminate the failure, since it's a model-level probabilistic <br>behavior, not something promptable away.
* Renaming tool parameters to avoid vocabulary overlap — currently <br>testing this separately; will share results in langgenius/dify#40671 if useful.
* Switching agent strategy to ReAct — a different mechanism, but a real <br>behavioral trade-off, not evaluated yet for our use case.

Full diagnosis and six rounds of trial data (showing the failure is <br>probabilistic and correlated with semantic overlap between tool-adjacent <br>vocabulary and processed data, not fixable via prompting alone) are in <br>langgenius/dify#40671.

### 3\. Can you help us with this feature?

fatelei:
> [!IMPORTANT]
>

## Summary

fix #40672 no tool_response but use the tool need raise exception

## Screenshots

| Before | After |
| ------ | ----- |
| ...    | ...   |

## Checklist

### DIFF
--- a/api/core/agent/errors.py
+++ b/api/core/agent/errors.py
@@ -7,3 +7,10 @@ class AgentMaxIterationError(Exception):
             f"Agent exceeded the maximum iteration limit of {max_iteration}. "
             f"The agent was unable to complete the task within the allowed number of iterations."
         )
+
+
+class AgentToolCallRequiredError(Exception):
+    """Raised when a required tool call was not produced by the model."""
+
+    def __init__(self):
+        super().__init__("Agent expected a tool call, but the model returned a final answer without calling a tool.")
--- a/api/core/agent/fc_agent_runner.py
+++ b/api/core/agent/fc_agent_runner.py
@@ -9,7 +9,7 @@ from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from core.agent.base_agent_runner import BaseAgentRunner
-from core.agent.errors import AgentMaxIterationError
+from core.agent.errors import AgentMaxIterationError, AgentToolCallRequiredError
 from core.app.apps.base_app_queue_manager import PublishFrom
 from core.app.entities.queue_entities import QueueAgentThoughtEvent, QueueMessageEndEvent, QueueMessageFileEvent
 from core.app.file_access import grant_upload_file_access
@@ -126,6 +126,7 @@ class FunctionCallAgentRunner(BaseAgentRunner):
         llm_usage: dict[str, LLMUsage | None] = {"usage": None}
         final_answer = ""
         prompt_messages: list = []  # Initialize prompt_messages
+        has_tool_response = False
 
         # get tracing instance
         trace_manager = app_generate_entity.trace_manager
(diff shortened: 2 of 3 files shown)


david-yang-pixel (approved):

CryoThrust (commented):
I traced the new guard in `fc_agent_runner.run`: `tool_call_required = bool(prompt_messages_tools and not has_tool_response)` is computed before each model invocation, and `has_tool_response` only becomes true after a tool has already executed. That makes the condition equivalent to “tools were offered and this is before the first tool result,” not “the caller explicitly required a tool call.”

As a result, any first-turn agent that has tools available but is allowed to answer directly will raise `AgentToolCallRequiredError` when the model returns a legitimate text answer. This is broader than #40672 and would be a compatibility regression for normal tool-capable agents. The guard needs a caller-owned expectation (for example an opt-in agent/run configuration or an explicit required tool/effect contract), and should validate the matching tool/result receipt rather than infer requirement from non-empty `prompt_messages_tools`.

Please add regression coverage for the allowed no-tool first turn, alongside the strict-required case, before merging.

CryoThrust:
I checked the failing Web Full-Stack E2E job (run 32953133860). The only failed scenario is `Switch a Completion app to Workflow Orchestrate`; setup reaches the app-options step, but the expected link for the newly created E2E app is not visible within 5s (`duplicate-app.steps.ts:21`). API unit/integration/coverage, type checks, and style checks are green. This looks unrelated to the FC runner change, but the required E2E gate still needs a maintainer retry or an explicit project-side quarantine decision before merge.

PR #41039, opened 2026-08-20
feat(rag): add native knowledge graph (GraphRAG) for built-in knowledge base

### LINKED ISSUE #40755: [Feature Request] Native Knowledge Graph / GraphRAG support for built‑in knowledge base
### Self Checks

### Self‑check
- [x] I have searched existing issues and discussions, no exactly matching feature request.

### Is your feature request related to a problem? Please describe.
Our team is currently using Dify for internal knowledge‑base applications.
The current built‑in knowledge base only supports vector/keyword hybrid retrieval.
When facing complex business documents, multi‑hop reasoning, entity‑relationship correlation questions, traditional vector‑based RAG often fails, and brings hallucinations.

Solutions like RAGFlow, KAG have native GraphRAG / knowledge‑graph enhanced RAG, while Dify does not have built‑in support.
We can integrate external GraphRAG services via workflow HTTP node or external‑knowledge‑base API, but this brings extra maintenance overhead, extra service deployment, and cannot reuse Dify’s existing document upload, parsing, citation, permission management.

### Describe the solution you'd like
I would like to ask: are there official plans to add native knowledge‑graph / GraphRAG capability inside Dify’s knowledge‑base pipeline?

Expected features:
1. Entity & relation extraction when importing documents inside Dify.
2. Hybrid retrieval: vector search + knowledge‑graph multi‑hop traversal.
3. Keep original citation capability, reuse Dify’s document management & permission.
4. Optional: support popular graph databases(Neo4j etc.) as backend.

### Describe alternatives you've considered
1. Call external GraphRAG service through workflow HTTP node, maintain another independent service.
2. Use Dify external knowledge‑base API to connect third‑party GraphRAG system.
Both options increase operation complexity for our production deployment.

### Additional context
Many enterprise scenarios require multi‑hop reasoning over business documents. This feature will greatly improve complex‑question answering quality.

_No response_

mesrop2008:
Fixes #40755

## Summary

Adds native GraphRAG support to Dify's built-in knowledge base, as requested in #40755. During indexing, an LLM extracts entities and relations from each chunk; retrieval can then traverse those connections across chunks and documents, in addition to the existing vector/keyword index — answering multi-hop questions that plain similarity search misses (e.g. "what does Company A ultimately own?" when the answer is two acquisitions away).

Design choices, matching the issue's four requirements:

* **Entity/relation extraction on import** — one LLM call per chunk during indexing, gated behind a per-dataset `graph_index_setting` so datasets that don't opt in see zero behavior or cost change.
* **Hybrid retrieval (vector + graph traversal)** — the graph runs as an additional leg inside Hybrid Search (and Keyword Search, for Economy-indexed datasets), merged and reranked alongside vector/full-text results. Plain Vector Search is untouched by design.
* **Citations reused as-is** — every extracted fact carries a provenance link back to a `DocumentSegment.index_node_id`, so graph hits resolve to ordinary segments and go through Dify's existing citation, document management, and permission paths unchanged.
* **Pluggable graph store** — Postgres by default (no new infrastructure required, reuses the metadata database); optional Neo4j via `GRAPH_STORE=neo4j` (requires `uv sync --group graph-neo4j`).

New surface area: a Knowledge Graph settings panel (toggle, extraction model, traversal depth, query fallback) and a graph explorer tab in the dataset sidebar (force-directed visualization, entity search, stats).

## Impact

* **Existing datasets: zero behavior change.** The graph is entirely opt-in per dataset (`graph_index_setting.enabled`, off by default). Datasets that don't enable it see no change to indexing time, retrieval results, cost, or UI — no new sidebar tab, no extra queries, no extra LLM calls.
* **Database migration required.** Adds three tables (`dataset_graph_entities`, `dataset_graph_relations`, `dataset_graph_chunk_links`) and one column (`datasets.graph_index_setting`, JSONB). Runs automatically via the existing `flask upgrade-db` / `MIGRATION_ENABLED` startup path; no manual steps, no backfill of existing data.
* **Cost/latency, only for datasets that opt in:**
  * *Indexing*: one additional LLM call per chunk (entity/relation extraction), run concurrently, capped by `KNOWLEDGE_GRAPH_EXTRACTION_WORKERS` (default 5). Extraction failures are caught and logged, never fail the underlying vector/keyword index for that chunk.
  * *Retrieval*: Hybrid Search and Keyword Search gain one extra parallel leg (a lexical entity lookup plus a bounded graph walk, capped by `max_depth`/`max_neighbors_per_hop`). No new embedding calls; only an LLM call as a fallback when lexical entity matching finds nothing — and that fallback can be switched off per dataset.
* **No new required dependencies.** The default Postgres graph store reuses the existing metadata database. `echarts`/`echarts-for-react` (used for the graph visualization) were already project dependencies; the graph view imports them via the modular entry point (only the graph series, tooltip, and canvas renderer) rather than the full package, to keep bundle/build cost close to zero. `neo4j` is a genuinely optional dependency, only needed if a tenant sets `GRAPH_STORE=neo4j`.
* **No breaking API changes.** All new endpoints (`GET /datasets/{id}/graph`, `GET /datasets/{id}/graph/stats`) and fields are additive; existing console/service API responses gained one new optional field (`graph_index_setting`) on the dataset detail response.

## Review feedback addressed

All findings from the previous review round are resolved. Summary of what changed:

### Blocking

* **A graph failure could fail the whole retrieval.** `graph_search` joined the shared `exceptions` list that `_retrieve` raises on, so a graph backend outage took vector and full-text results down with it — the opposite of an enhancement layer. The leg no longer receives that list at all; the parameter was removed from the signature so it cannot regress, and a test asserts its absence. Failures are logged and the leg is skipped.

### High

* **Graph rows were stranded when a dataset was deleted.** Deletion reconstructs a bare `Dataset` whose settings are absent, so the opt-in guard skipped cleanup and left rows behind forever. Added `GraphIndexService.purge_dataset()`, which drops every graph row unconditionally, and wired it into `clean_dataset_task` and `clean_when_dataset_deleted`. Graph cleanup in the index processors also moved to the *first* step of `clean()`, because the vector/keyword legs below it can raise and callers swallow that failure while continuing to delete the dataset.
* **Neo4j store correctness.** Schema constraints and indexes are now created once per process (including an index on `SUPPORTS.relation_id`); merges run inside `execute_write` under the same distributed lease as Postgres; orphan pruning deletes unsupported relations before entities, via `OPTIONAL MATCH` + `count()`, so zero-support edges cannot survive. Counters are recomputed rather than incremented (see below).
* **`document_ids_filter` was applied as a result post-filter.** That let an excluded document seed the walk or act as an intermediate hop, leaking its existence through the returned path even when its chunks were filtered from the output. The filter is now an authorization boundary threaded through seed selection, traversal, and chunk-link lookup in both the retrieval service and the Postgres store. Four tests cover it: an excluded document cannot seed the walk, cannot bridge a multi-hop path, and never appears in a returned path.

### Medium

* **Unbounded extraction concurrency.** `KNOWLEDGE_GRAPH_EXTRACTION_WORKERS` bounded one call's thread pool, but bulk indexing runs many celery tasks at once, so real concurrency was workers × concurrent tasks. Added a process-wide `BoundedSemaphore` budget. *Partial by design and documented:* this is per process, not deployment-global — a true global budget needs Redis coordination, and total spend still scales with the number of deployed workers, which is the number operators actually control.
* **Counter drift.** `frequency` and `weight` were incremented on every merge, so re-indexing a document inflated them permanently. Both are now recomputed from the chunk links that actually exist, in both backends, with tests covering the drift case.
* **Unbounded prompt and per-segment lookups.** Added `MAX_EXTRACT_PROMPT_LENGTH` (20000) so an oversized chunk cannot produce a runaway prompt, and segment resolution now happens before scoring, batched, rather than one query per chunk.
* **Settings were not validated at the save boundary.** Enabling the graph without an extraction model silently produced a dataset that could never build one. `validate_graph_index_setting` now merges incoming values over stored ones, validates the result, and verifies the extraction model exists for the tenant via `check_graph_extraction_model_setting`. Validation deliberately lives at the save boundary, not in the domain model used to *read* stored settings — otherwise stored-invalid data becomes a hard parse failure and blocks retrieval over an existing graph. The dataset response also gained the full set of 11 fields, so the console can round-trip what it saved.
* **Concurrent merges across workers.** Added a Redis distributed lease (`graph_index_lock`) held per dataset, renewed between merge phases with `replace_ttl` so a long merge cannot lose it silently, and released in a `finally`. Losing the lease mid-write raises rather than letting two writers interleave.
* **Console kept serving graph data after the feature was switched off.** `get_stats` and `explore` are now gated on `is_enabled(dataset)`. Rows survive so re-enabling does not mean re-indexing, but the API stops serving them the moment the UI stops offering the feature.
* **Untrusted input was undocumented.** Document content reaches the extraction prompt verbatim and the resulting facts are persisted without a confidence gate, so a crafted document can plant edges that later act as false multi-hop bridges. Documented in the extractor's docstring and in the README's Limitations section, along with the mitigations that do exist: facts never cross knowledge bases, and each one is traceable back to its source chunk.

### Minor

* Neo4j's description merge used a literal `\n` instead of a real newline, so merged descriptions differed from Postgres — fixed for parity.
* Added an index on `(dataset_id, frequency)` to support the ordering the explorer uses.
* External knowledge bases now reject `graph_index_setting` explicitly and hide the control, since they have no documents of their own to extract from.
* The graph view passed `relation.predicate` directly as an ECharts label formatter, which treats a string as a *template* — a predicate containing `{a}` or `{b}` would have been expanded as a placeholder. It is now returned from a callback so it renders verbatim.
* Added an `aria-label` to the Query Fallback switch, moved long internal rationale out of the generated OpenAPI/zod docstrings, and added operator documentation (`api/core/rag/graph/README.md`) covering cost, backend choice, settings, and limitations.

### Deferred, with reviewer's agreement

* **Per-chunk failed state and a "rebuild graph" console action.** Extraction failures are logged and swallowed by design, so the vector index survives — but a user on a rate-limited or misconfigured model sees a sparse graph with no explanation in the UI, and the only rebuild path today is re-indexing the document. Documented under Limitations; worth a follow-up.

### Known limitations, documented rather than fixed

* Entity name lookup uses `LIKE '%term%'`, which no btree index can serve. A trigram index would require the `pg_trgm` extension, which cannot be assumed to be available.
* `graph_search` exists as an enum member for callers building a retrieval config directly, but is not offered as a standalone mode in the console — the value is the merged result set.

## Testing & Validation

* **87 tests** across `api/tests/unit_tests/core/rag/graph/`, `api/tests/unit_tests/core/rag/datasource/graph/` and `test_retrieval_methods_graph.py` — entity normalization, JSON extraction parsing, graph traversal/scoring, null-tolerant settings, Postgres store merge/dedup/orphan-pruning, counter drift, document scoping, the distributed lease, and which retrieval methods trigger the graph leg.
* **218 tests** passing across those suites plus every suite this change touches (`test_datasource_retrieval`, `test_external`, `test_dataset_service_dataset`, `test_clean_dataset_task`).
* Matching frontend tests for the graph chart color/sizing logic and the dataset-settings form state; `tsc --noEmit` clean across the web app.
* `ruff format`, `ruff check`, response-contract lint, import-linter, dotenv-linter and `mypy` run clean; `pyrefly` reports no new errors.
* Branch merged up to current `main` and re-verified afterwards.
* **Manually verified end-to-end** against a live stack: enabled the graph on a real dataset, confirmed extraction ran per chunk during indexing (32 entities / 35 relations from two documents), confirmed the explorer rendered and filtered the extracted subgraph, and confirmed that a Hybrid Search query whose answer spans two documents returns the bridging chunks — and ranks them above the shallow title matches the same query returns with the graph switched off. See the before/after below.

### Dataset settings

The Knowledge Graph panel in **Knowledge → Settings**. Off by default; enabling it requires choosing an extraction model, and the cost of doing so is stated at the point of choice.

graph_settings

### Knowledge graph explorer

| Extracted graph | Entity detail |
| --- | --- |
| full graph | entity detail |
| 32 entities and 35 relations extracted from two documents, typed and counted. | Hovering a node shows the description the extractor produced, not just a label. |

| Focused view | Entity search |
| --- | --- |
| focused view | entity search |
| Clicking an entity narrows the graph to its neighbourhood. | Searching filters both the graph and the entity list. |

**Empty state** — the graph is built during indexing, so a dataset that enabled it after its documents were indexed sees this until they are re-indexed.

### Retrieval: Hybrid Search with the graph off vs. on

Two documents connected only through a shared person, as a two-hop path with **no direct edge** between the endpoints:

```
Northwind Robotics --founded_by--> Dr. Elena Vasquez --leads--> Project Meridian
```

Asking *"What is the connection between Northwind Robotics and Project Meridian?"* under Hybrid Search, Top K 3:

| Before — graph disabled | After — graph enabled |
| --- | --- |
| retrieval_graph_disabled | retrieval_graph_enabled |
| `0.67` "Project Meridian: Research Programme Overview"<br>`0.61` "Northwind Robotics: Company Profile"<br>`0.56` "In 2021 Northwind Robotics was acquired by Halberd Industrial Group…" | `0.75` "Project Meridian was established in 2020 and is **led by Dr. Elena Vasquez**…"<br>`0.74` "Northwind Robotics is an industrial automation company…"<br>`0.69` "Northwind Robotics was **founded in 2011 by Dr. Elena Vasquez**…" |
| Both documents are reached — the query names both entities, so full-text matches them. But the hits are two document **titles** and an unrelated acquisition fact. None of them contains the link, and the person joining the two never appears. The question is not answerable from these chunks. | The graph leg seeds on the two named entities, walks the two-hop path through the shared person, and pulls back the chunks that actually **carry** the connection — both naming Dr. Elena Vasquez. Every score rises, and the top two chunks together answer the question. Each hit resolves to an ordinary segment and is cited like any other. |

## Checklist

- [x] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs) — new dataset setting, new retrieval behavior, and optional `GRAPH_STORE`/Neo4j env vars need to be documented there.
- [ ] I've updated the documentation accordingly. — not yet done; the dify-docs update is a separate follow-up PR in that repo.


QianJinGuo (commented):
# Review: native knowledge graph (GraphRAG) for built-in KB

First — this is a serious, well-executed feature: the store abstraction mirrors the keyword factory, every fact carries provenance back to a citable `DocumentSegment`, extraction is best-effort and never aborts indexing, the query path keeps cost near zero for lexical matches, normalization edge cases (unicode folding, self-loops, predicate casing, endpoint backfill) are all unit-tested, and the multi-hop chain-graph test is exactly the right behavioral test for this feature. That said, I found one availability bug, three correctness/lifecycle bugs, and a set of smaller issues — details below, all verified against the diff and the base repo.

## Blocking

**B1. A graph-store failure takes down hybrid/keyword retrieval entirely.**
`RetrievalService.graph_search` appends any exception to the shared `exceptions` list, and `_retrieve` raises when that list is non-empty (`retrieval_service.py:877-878`). So a Neo4j outage or a graph-query error fails every hybrid/keyword retrieval for graph-enabled datasets even when the vector/full-text legs succeeded. This contradicts the PR's own design note ("a failure here is logged and swallowed"), which holds at index time but not at retrieval time. Fix: log and skip the leg in `graph_search` — don't touch the shared `exceptions` list (it's shared with the primary legs, so a graph failure can also mask or compound other failures).

## High

**H1. Dataset deletion can strand graph data — two failure modes.**
(a) `clean_dataset_task` reconstructs a minimal `Dataset` without `graph_index_setting` (`clean_dataset_task.py:60-66`), so `GraphIndexService.delete_all()` returns immediately on its `if not dataset.graph_index_setting` guard. (b) Graph cleanup is appended at the *tail* of each processor's `clean()`; the base task catches cleanup exceptions and continues deleting the dataset (`clean_dataset_task.py:95-104`), so if vector/keyword cleanup throws first, the graph cleanup line never runs. The three new tables also have no DB-level cascade. Net effect: Postgres rows and Neo4j nodes survive for deleted datasets (untracked cost + data retention issue). Fix: load `graph_index_setting` in the deletion task (or add explicit DELETEs for the three tables) and make graph cleanup first-class/final-`finally` rather than last-line.

**H2. Neo4j keeps traversing deleted relations.**
Orphan pruning on deletion removes entities and chunk nodes, but `DIFY_RELATION` edges that have lost all of their supporting chunks are never removed (`neo4j_graph_store.py:166-203`). Retrieval walks relations by endpoints (`neo4j_graph_store.py:256-279`), so deleted facts keep bridging live entities and influencing multi-hop results. The Postgres backend doesn't have this variant (relations are rows deleted by document/chunk). Fix: after pruning, delete relations with zero remaining chunk links.

**H3. Traversal ignores `document_ids_filter` and segment state until the very end.**
Seeds, the BFS walk, and chunk scoring all run unfiltered; `document_ids_filter` and the `enabled`/`COMPLETED` checks are applied only when final segments are resolved (`graph_retrieval.py:357-407`). Consequences: excluded documents contribute seed entities and bridge paths; disabled/indexing segments influence scores; and `graph_path` metadata can expose entity names/predicates originating from excluded documents. If document filters are an authorization boundary this is security-sensitive; either way it's a retrieval-correctness defect. Fix: push the filter into seed selection and traversal, or state explicitly that filters are best-effort.

## Medium

**M1 (ops risk).** Per-call extraction threads are capped (`KNOWLEDGE_GRAPH_EXTRACTION_WORKERS`), but bulk indexing runs through multiple celery tasks, so real extraction concurrency is deployments × workers with no graph-specific budget, and the settings form shows no cost estimate. "1 LLM call per chunk" on a 100k-chunk backfill is a big surprise bill. Suggest a cost hint in the settings UI and optionally a global semaphore.

**M2 (broadened).** Counter drift affects entity frequency **and** relation weight, on re-indexing *and* delete/prune paths: counters only ever increase, and orphan pruning never recomputes the surviving counts (Postgres `postgres_graph_store.py:938,1017`; Neo4j `:596-624`). Since `search_entities` ranks by frequency, drift distorts retrieval over time. Fix: recompute from chunk links after replacement/pruning.

**M3.** Query-time LLM seed fallback runs on every lexical miss with no separate toggle or warning. Consider a per-dataset switch, and document that a miss costs one LLM call.

**M4.** Settings round-trip can erase server-side config: the response contract exposes only five fields while `GraphIndexSetting` carries more (`extract_prompt`, `max_seed_entities`, `hop_decay`, …); the frontend sends the returned object back wholesale and the service replaces the stored setting — a later console save silently drops fields it doesn't know. The update payload also accepts an unvalidated `dict[str, Any]` (no provider/model existence or prompt-size checks).

**M5.** Neo4j `add_chunk_graphs` has no serialization (Postgres uses the Redis lock) and no single transaction spanning delete + entity + relation writes — concurrent re-indexing can interleave, and a mid-flight failure leaves a partial graph while `GraphIndexService` swallows the error.

**M6.** Disabling graph indexing doesn't disable the console graph APIs: `get_stats()`/`explore()` only check that `graph_index_setting` exists, not `enabled` — stale graph data stays queryable through the API after the UI hides the feature.

**M7.** Untrusted document text flows into the extraction prompt, and parsed output is persisted without confidence gates or schema constraints — a crafted document can fabricate entities/relations that later act as false multi-hop bridges. This is an integrity/quality risk worth at least a documented limitation (and later: schema-guided extraction or confidence thresholds).

## Minor

- `Neo4jGraphStore.__init__` runs `_ensure_constraints()` (two DDL round-trips) on every construction, including retrieval paths. Constraints are global — create once, module-level.
- The Redis lock lease is a hard-coded 600s with no renewal; a >10 min merge releases it mid-write. Renew or fail fast with a structured warning.
- Entity lookup scales poorly: Postgres `LIKE '%term%'` + frequency ordering and Neo4j `CONTAINS` have no supporting index (`list_entities` also sorts by frequency unindexed).
- The graph toggle can be enabled with no extraction model selected; `get_setting()` then silently no-ops while the UI shows "enabled". Show a validation error instead.
- The echarts edge-label formatter passes `relation.predicate` unescaped — safe with the canvas renderer today, but worth escaping defensively in case the renderer changes.
- External datasets render/save graph settings in the form, but `_update_external_dataset()` ignores the field — a dead control; hide or reject it.

## Follow-ups (fine to defer)

- Reconciliation semantics for failed/empty extractions (a per-chunk failed state or a "rebuild graph" console action) — today the docs tell users to re-index documents, which works but is heavy-handed.
- `GRAPH_SEARCH` is in the public enum and generated contracts, but console retrieval config still offers only semantic/full-text/hybrid — state the policy (internal-only?) explicitly.
- Are `packages/contracts/generated/*` regenerated by CI on merge, or should reviewers diff them by hand?
- An operator docs page for the Neo4j extra (`GRAPH_STORE`, `KNOWLEDGE_GRAPH_NEO4J_*`, `uv sync --group graph-neo4j`).

## Verified non-issues (checked, no finding)

- Cypher and SQL injection: all queries are parameterized; the LIKE probe escapes wildcards.
- Graph-viewer XSS: tooltip HTML is escaped; other values render as React text nodes.
- Tenant isolation: dataset access is RBAC-gated, dataset IDs are global PKs, graph rows carry `tenant_id`/`dataset_id`, and model resolution is tenant-scoped via `ModelManager.for_tenant`. Provider/store selection is deployment-global — a rollout trade-off worth documenting, not an isolation defect.

---

Happy to help with fixes for any of the above — the multi-hop retrieval and the lifecycle pieces (H1/H2) in particular would be fun to dig into. Thanks for building this properly.

mesrop2008:
@QianJinGuo 
Thank you for this — an unusually thorough review, and several findings were things I would not have caught myself. B1 and H3 especially: I had convinced myself the "logged and swallowed" note covered retrieval too, and it only ever held at index time.

All of it is addressed — blocking, high, medium and minor. Rather than repeat it here, I've written it up point by point under **"Review feedback addressed"** in the PR description, using your numbering so you can check items off against your own list.

Three things worth calling out that you shouldn't have to go looking for:

* **Entity lookup scaling is the one I did not fix.** `LIKE '%term%'` is still unindexable; a trigram index needs `pg_trgm`, which I don't think can be assumed available. Documented under Limitations. If you'd accept a conditional `CREATE EXTENSION`, I'm glad to add it.
* **The extraction budget is per process, not deployment-global.** A true global budget needs Redis coordination, and total spend still scales with the number of deployed workers. Documented as such rather than papered over — happy to do the Redis version if you'd rather it landed here.
* **Per-chunk failed state / "rebuild graph"** stays in the follow-up bucket, as you flagged. One data point for it: while testing I hit a rate-limited API key, which produces a sparse graph with no explanation anywhere in the UI, and re-indexing is the only way back. It's the first thing I'd pick up next.

On your contracts question — `autofix.ci` regenerates them on the PR and commits the result, so reviewers shouldn't need to diff them by hand. I regenerate locally as well, mostly so the diff stays small: running `openapi-ts` without the `vp fmt generated/api` step afterwards turns a ~40-line diff into tens of thousands of formatting-only lines.

The branch is also merged up to current `main` and re-verified. That wasn't free: upstream changed the `rbac_permission_required` signature, removed the `useModelList` hook, and moved `consoleQuery` to `@/service/console` — none of which conflicted textually, all of which broke at build time. Fixed, and the knowledge-graph migration is repointed onto the current head so the chain keeps a single head.

Re-review whenever you have time would be much appreciated. And thank you for offering to help with the fixes — I've had a go at all of them here, but H1/H2 and the traversal changes are the parts I'd most value your eyes on, since you know this area far better than I do.

PR #41055, opened 2026-08-21
refactor(api): standardize console billing portal errors

hjlarry:
## Summary

part of https://github.com/langgenius/dify/issues/39993

- isolate Console subscription and invoice portal failures behind typed service and HTTP error boundaries
- authorize with the role already loaded for the current account instead of querying workspace membership again
- validate the documented `{url}` responses, mark the legacy model-provider checkout route as deprecated, and remove unused notification write wrappers

## Behavior changes

- permission failures now return `403 billing_access_denied`
- invalid or terminal upstream responses return `502 billing_operation_failed`
- timeout, rate-limit, 5xx, and transport failures return `503 billing_unavailable`
- successful subscription and invoice responses are projected to the documented `{url}` shape

### DIFF
--- a/api/controllers/console/billing/billing.py
+++ b/api/controllers/console/billing/billing.py
@@ -12,7 +12,6 @@ from controllers.console.billing.error import (
     BillingOperationFailedErrorResponse,
     BillingUnavailableErrorResponse,
     BillingUnprocessableEntityErrorResponse,
-    dump_billing_response,
     to_billing_request_error,
 )
 from controllers.console.wraps import (
@@ -25,6 +24,7 @@ from controllers.console.wraps import (
 )
 from enums import CloudPlan
 from fields.base import ResponseModel
+from libs.helper import dump_response
 from libs.login import login_required
 from models import Account
 from services.billing_service import BillingService
@@ -103,7 +103,7 @@ class Subscription(Resource):
             )
         except BillingError as error:
             raise to_billing_request_error(error) from error
-        return dump_billing_response(BillingSubscriptionResponse, data)
+        return dump_response(BillingSubscriptionResponse, data)
 
 
 @console_ns.route("/billing/invoices")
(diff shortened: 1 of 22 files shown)


crazywoola (dismissed):

api/controllers/console/billing/billing.py
```diff
@@ -16,11 +23,12 @@
     with_current_user,
 )
 from enums import CloudPlan
-from extensions.ext_database import db
 from fields.base import ResponseModel
+from libs.helper import dump_response
 from libs.login import login_required
 from models import Account
 from services.billing_service import BillingService
```
41tair: Still import BillingService

api/controllers/console/workspace/model_providers.py
```diff
@@ -403,6 +403,7 @@ def post(self, tenant_id: str, provider: str):
 
 @console_ns.route("/workspaces/current/model-providers/<path:provider>/checkout-url")
 class ModelProviderPaymentCheckoutUrlApi(Resource):
+    @console_ns.doc(deprecated=True)
```
41tair, line 405: If we change this interface to deprecated, maybe we need update docs at the same time

api/controllers/console/billing/billing.py
```diff
@@ -45,13 +53,38 @@ class BillingSubscriptionResponse(ResponseModel):
...
+        502,
+        "Billing operation failed",
+        console_ns.models[BillingOperationFailedErrorResponse.__name__],
+    )
+    @console_ns.response(
+        503,
+        "Billing unavailable",
+        console_ns.models[BillingUnavailableErrorResponse.__name__],
+    )
     @setup_required
```
41tair: Those decoratores need change to.

api/controllers/console/workspace/model_providers.py
hjlarry: yep,  I've already add a inner doc before submit this PR

api/controllers/console/billing/billing.py
hjlarry: I've concern to use `console_account_admission`, which already get `email` and `role` in its intern,  but only inject `account_id` to the `RequestContext`.  
the `email` is in the `AccountRepository`,   the `role` is in the Membership repository.  
So I haven't change the decoratores for now

crazywoola (approved):

PR #41062, opened 2026-08-21
refactor(agent): simplify Agent v2 output contract

BeautyyuYanli:
> [!IMPORTANT]
>

## Summary

- Simplify Agent v2's default output contract from preset `text` / `files` / `json` declarations to one derived system `text` output.
- Omit the Agent backend output layer when no custom outputs are configured, then adapt the backend's plain string result to the existing workflow-facing `{ "text": ... }` shape.
- When custom outputs exist, add optional `text: string` to the structured schema while applying required flags, type checks, retry/default handling, and file normalization only to custom declarations.
- Persist only custom declarations, allow `files` and `json` to be reused as custom names, and reserve system-owned `text` plus future `switch` and `_session` fields.
- Migrate draft and published Workflow Agent bindings away from the three legacy preset declarations without changing downstream `node.text` selectors.

Issue: N/A — scoped Agent v2 output-contract simplification.

## Root cause

Agent v2 treated `text`, `files`, and `json` as a default structured-output schema even when users had not configured any outputs. That forced every run through the object-output contract, persisted system-owned declarations alongside user configuration, and duplicated default-output projection across backend services and the frontend.

The preset `files` and `json` names also conflated a legacy UI default with the supported file and object output types. File normalization is driven by the declared type rather than the field name, so permanently reserving those retired names was unnecessary.

## Impact

With no custom declarations, Agent v2 now uses the backend's plain string contract and exposes the result to workflows as `text`. Existing downstream `node.text` selectors keep the same name and string type.

With custom declarations, the backend receives an object schema containing optional system `text` followed by the custom fields. Only custom fields are persisted and participate in API-side output validation or failure handling. Users may define custom fields named `files` or `json`; `text` remains immutable and system-owned, while `switch` and `_session` are reserved for future system use.

Alembic revision `925e75620b69` removes the legacy `text`, `files`, and `json` preset declarations from every persisted `workflow_agent_node_bindings.node_job_config`. Other custom declarations retain their order and configuration. The migration is intentionally one-way because removed declarations cannot be reconstructed precisely. Deployments should apply the migration before starting the new API code; the effective-output projection reintroduces `text` after persistence cleanup, so workflows that only depend on the standard `text` output require no changes.

No Agent v1, `dify-agent`, generated contract, or Workflow graph-blob format is changed.

## Validation

- Backend model, Composer, Workflow binding sync, and migration suites: 166 passed.
- Frontend Agent v2 Panel, output editor, and downstream variable projection suites: 60 passed.
- Focused frontend `vp check`: formatting, lint, and type checks passed.
- Commit hooks: Ruff checks and staged frontend checks passed.
- `git diff --check`: passed.
- All five staged implementation review areas passed: code/logic hygiene, requirement completeness, test hygiene, test sufficiency, and documentation accuracy.

The full backend/frontend suites, full repository type checks, and CI-only backend integration tests were not run locally. The affected integration fixture was updated for the plain-string output contract.

## Screenshots

| Before | After |
| ------ | ----- |
| Agent v2 showed preset `text`, `files`, and `json` outputs and always requested structured output. | Agent v2 shows derived `text` by default; structured output is enabled only after adding a custom field. |

## Checklist

****This PR is made with Codex, while I'm responsible for all the changes.****

### DIFF
--- a/api/models/agent_config_entities.py
+++ b/api/models/agent_config_entities.py
@@ -1070,7 +1070,8 @@ SYSTEM_DECLARED_OUTPUTS: Final[tuple[DeclaredOutputConfig, ...]] = (
         description="Free-form text answer.",
     ),
 )
-RESERVED_DECLARED_OUTPUT_NAMES: Final[frozenset[str]] = frozenset({"text"})
+# ``switch`` and ``_session`` are reserved for future system output contracts.
+RESERVED_DECLARED_OUTPUT_NAMES: Final[frozenset[str]] = frozenset({"text", "switch", "_session"})
 
 
 def effective_declared_outputs(
--- a/api/models/agent_config_entities.py
+++ b/api/models/agent_config_entities.py
@@ -4,9 +4,19 @@ import re
 from enum import StrEnum
 from typing import Annotated, Any, Final, Literal, Self
 
-from pydantic import BaseModel, ConfigDict, Field, WithJsonSchema, field_validator, model_validator
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    Field,
+    WithJsonSchema,
+    field_validator,
+    model_validator,
+)
 
-from core.rag.entities.metadata_entities import ConditionValue, SupportedComparisonOperator
+from core.rag.entities.metadata_entities import (
+    ConditionValue,
+    SupportedComparisonOperator,
+)
 from core.tools.entities.tool_entities import ToolProviderType
 from core.workflow.file_reference import is_canonical_file_reference
 from graphon.file import FileTransferMethod, FileType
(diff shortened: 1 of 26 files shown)


api/models/agent_config_entities.py
```diff
@@ -1050,45 +1060,25 @@ def _is_valid_file_default_value(value: Any) -> bool:
...
-        array_item=DeclaredArrayItem(type=DeclaredOutputType.FILE),
-    ),
-    DeclaredOutputConfig(
-        name="json",
-        type=DeclaredOutputType.OBJECT,
-        required=False,
-        description="Free-form JSON object.",
-    ),
 )
+RESERVED_DECLARED_OUTPUT_NAMES: Final[frozenset[str]] = frozenset({"text"})
```
wylswz: Adding new reserved output names later would be a breaking change for users already using those names. If additional system-owned outputs are planned, even for a future release, consider reserving their names here now.

lyzno1 (commented):
LGTM

wylswz:
LGTM

api/models/agent_config_entities.py
BeautyyuYanli: I reserved two more item `switch` and `_session`

wylswz (dismissed):

MRZHUH (approved):

PR #41063, opened 2026-08-21
fix(api): re-sign bare and backticked file URLs

### LINKED ISSUE #40788: re_sign_file_url_answer only re-signs markdown-formatted links, so bare/backticked tool file URLs keep the INTERNAL_FILES_URL host
### Self Checks

### Dify version

1.16.1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Self-hosted Docker, `INTERNAL_FILES_URL=http://api:5001`, `FILES_URL=http://<lan-ip>`.

1. Create an Agent app (app mode `agent`) and give it a task that makes it write a file, e.g. "write a poem and save it as a .txt". The agent uploads the file via `dify-agent file upload` -> `POST /files/upload/for-plugin`.

2. That endpoint returns `preview_url`, built by `ToolFileManager.sign_file` (`api/core/tools/tool_file_manager.py:43`). Unlike `core/tools/signature.py::sign_tool_file`, it has no `for_external` switch and hardcodes the internal base:

```python
# Use internal URL for plugin/tool file access in Docker environments
base_url = dify_config.INTERNAL_FILES_URL or dify_config.FILES_URL
```

The response also carries `source_url` / `original_url`, but `create_file_by_raw` sets `original_url=None` (`tool_file_manager.py:94`), so `preview_url` is the only URL the agent has.

3. The agent pastes that URL into its answer. Whether it comes out as a markdown link is entirely up to the model. In my runs it frequently wraps it in backticks instead:

```
Download:
`http://api:5001/files/tools/5c465079-...-d1eec17ef8fd.txt?timestamp=1786703155&nonce=...&sign=...`
```

4. Read that message back through the Service API (`GET /messages`) or the console. `MessageListItem.answer` is bound to `Message.re_sign_file_url_answer`, whose regex (`api/models/model.py:1655`) is:

```
\[!?.*?\]\((((http|https)://.+)?/files/(tools/)?[\w-]+.*?timestamp=.*&nonce=.*&sign=.*)\)
```

It requires the `[...](...)` wrapper, so a bare or backticked URL produces no match and the method returns `self.answer` unchanged.

5. Re-run until the model happens to write `[download](http://api:5001/...)` instead. That answer is re-signed correctly. Same file, same config - the only variable is how the model formatted the link.

### ✔️ Expected Behavior

Re-signing should key off the tool file URL itself, not off the markdown syntax that happens to surround it. Any `/files/tools/{id}?timestamp=...&nonce=...&sign=...` occurrence in `answer` should be rewritten to `FILES_URL` with a fresh timestamp - whether it is bare, inside backticks, inside a fenced block, or wrapped in `[...](...)`.

### ❌ Actual Behavior

Only `[...](...)` occurrences are re-signed. Everything else reaches the client with an internal Docker hostname that no browser outside the compose network can resolve, and with the original timestamp, so it is already counting down against `FILES_ACCESS_TIMEOUT`.

Notes:

- Related and already merged: #39952 (`feat(agent): separate CLI file URL audiences`) splits signing into origin-neutral URIs and gives the Agent CLI a `public_download_url`. That removes the source of the internal URL for new Agent CLI traffic, but it does not close this gap: it is not in any release yet (latest is 1.16.1), and stored messages are still rendered through this regex on every read, so existing conversations keep serving internal URLs even after upgrading.
- Earlier report of the same symptom from a different producer: #23118 (tool output URLs using `INTERNAL_FILES_URL` instead of `FILES_URL`).
- Secondary observation: `ToolFileManager.sign_file` and `core/tools/signature.py::sign_tool_file` are the same signing routine except that the former lacks the `for_external` parameter. That looks like copy-paste drift rather than intent.
- Client-side workaround, in case it helps others: the signature does not cover the hostname (`data_to_sign = f"file-preview|{tool_file_id}|{timestamp}|{nonce}"`), so a downstream consumer can rewrite the origin and the link still verifies. It does not help with the stale timestamp.

gozzz8410:
## Summary

Fixes #40788.

`Message.re_sign_file_url_answer` currently only detects signed file URLs when they are wrapped in Markdown link syntax. If an Agent returns the same signed URL as plain text, inside backticks, or in a fenced block, the URL is not re-signed and an internal Docker hostname such as `http://api:5001` can leak to the client.

This change matches the signed file URL itself instead of requiring the surrounding Markdown wrapper. It keeps the existing re-signing logic intact while allowing bare, backticked, fenced, and Markdown-linked file URLs to be refreshed and rebound to the externally configured file origin.

### Changes

- Update `Message.re_sign_file_url_answer` to match signed `/files/...` URLs independently of surrounding Markdown syntax.
- Bound URL matching at common answer delimiters so multiple signed URLs in one answer are handled independently.
- Add regression coverage for:
  - a bare `file-preview` URL with an internal host;
  - a backticked tool-file URL;
  - multiple bare signed URLs in the same answer.

### Reproduction

With a self-hosted Docker deployment:

```env
FILES_URL=https://example.com
INTERNAL_FILES_URL=http://api:5001
```

an Agent may return:

```text
Download:
`http://api:5001/files/tools/<id>.py?timestamp=...&nonce=...&sign=...`
```

Before this change, the backticked URL is not recognized by `re_sign_file_url_answer` and reaches the client unchanged. After this change, it is re-signed using the existing external file URL path.

### Scope

This PR intentionally does not change `ToolFileManager.sign_file()`. That method is used for internal plugin/tool access and `INTERNAL_FILES_URL` remains valid for that audience. The fix is on the message read/display path where URLs exposed to clients must be re-signed for the external audience.

### Testing

Regression tests are included in `api/tests/unit_tests/models/test_re_sign_file_url_answer.py`.

I was able to validate the change structurally against the current `main` branch and keep the diff isolated to the URL matcher plus regression tests. Full repository test execution was not available in this ChatGPT environment.

## Checklist

- [x] I've added regression tests for the introduced behavior.

### DIFF
--- a/api/models/model.py
+++ b/api/models/model.py
@@ -1662,16 +1662,21 @@ class Message(Base):
         if not self.answer:
             return self.answer
 
-        pattern = r"\[!?.*?\]\((((http|https):\/\/.+)?\/files\/(tools\/)?[\w-]+.*?timestamp=.*&nonce=.*&sign=.*)\)"
+        # Match the signed file URL itself rather than the markdown that may surround it, so bare,
+        # backticked and fenced occurrences are re-signed too. `url_char` stops the match at the
+        # delimiters that can terminate a URL in an answer (whitespace, markdown link/quote chars).
+        url_char = r"[^\s)\]`\"'<>]"
+        pattern = (
+            rf"(?:https?://{url_char}+)?/files/(?:tools/)?"
+            rf"{url_char}*?timestamp={url_char}*?nonce={url_char}*?sign={url_char}*"
+        )
         matches = re.findall(pattern, self.answer)
 
         if not matches:
             return self.answer
 
-        urls = [match[0] for match in matches]
-
         # remove duplicate urls
-        urls = list(set(urls))
+        urls = list(set(matches))
 
         if not urls:
             return self.answer
--- a/api/tests/unit_tests/models/test_re_sign_file_url_answer.py
+++ b/api/tests/unit_tests/models/test_re_sign_file_url_answer.py
@@ -1,0 +1,54 @@
+import pytest
+
+from models import model as model_module
+from models.model import Message
+
+
+@pytest.fixture
+def patch_file_signers(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setattr(
+        model_module.file_helpers,
+        "get_signed_file_url",
+        lambda file_id: f"https://files.example/{file_id}",
+    )
+    monkeypatch.setattr(
+        model_module,
+        "sign_tool_file",
+        lambda tool_file_id, extension: (
+            f"https://files.example/tools/{tool_file_id}{extension}?timestamp=999&nonce=new&sign=new"
+        ),
+    )
+
+
+def test_bare_file_preview_url_is_re_signed(patch_file_signers: None) -> None:
+    upload_id = "bare-1"
+    url = f"http://api:5001/files/{upload_id}/file-preview?timestamp=111&nonce=222&sign=333"
+    message = Message(answer=f"Download:\n{url}")
+
+    result = message.re_sign_file_url_answer
+
+    assert result == f"Download:\nhttps://files.example/{upload_id}"
+    assert "http://api:5001" not in result
+
+
+def test_backticked_tool_file_url_is_re_signed(patch_file_signers: None) -> None:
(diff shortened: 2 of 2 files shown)


crazywoola:
Closing as a duplicate of #40799, which is being retained for #40788 under the issue-author-first, then FIFO triage rule. Thanks for your contribution.

PR #41067, opened 2026-08-21
fix(api): attach exc_info=True to 19 logger.warning('...: %s', str(e)) sites — preserve WARNING level, add traceback

### LINKED ISSUE #41066: fix(api): attach exc_info=True to 19 logger.warning('...: %s', str(e)) sites — preserve WARNING level, add traceback
### Problem

Nineteen `except Exception as e:` (or `except BaseException as e:`) blocks in `api/` log the exception with `str(e)` but no `exc_info=True`, so the traceback is dropped on the floor. The log line shows the exception message, but not the call chain that produced it. The same block is then immediately followed by `continue` or `pass`, so the exception is gone after the log line — only the str remains in the log.

Same observability shape as the PEP 3134 sweep landed in cycles 13 (#40738) and 16 (#40810), but in `logger.*` calls instead of `raise` statements.

### Sites (all under `api/`, production code only — no tests, no vdb/, no .venv/)

| File | Line | Current |
| --- | --- | --- |
| `tasks/enable_segments_to_index_task.py` | 121 | `logger.warning("Failed to enable summaries for segments: %s", str(e))` |
| `tasks/disable_segment_from_index_task.py` | 77 | `logger.warning("Failed to disable summary for segment %s: %s", segment.id, str(e))` |
| `tasks/enable_segment_to_index_task.py` | 121 | `logger.warning("Failed to enable summary for segment %s: %s", segment.id, str(e))` |
| `tasks/add_document_to_index_task.py` | 132 | `logger.warning("Failed to enable summaries for document %s: %s", dataset_document.id, str(e))` |
| `tasks/remove_document_from_index_task.py` | 62 | `logger.warning("Failed to disable summaries for document %s: %s", document.id, str(e))` |
| `tasks/disable_segments_from_index_task.py` | 85 | `logger.warning("Failed to disable summaries for segments: %s", str(e))` |
| `core/app/features/annotation_reply/annotation_reply.py` | 94 | `logger.warning("Query annotation failed, exception: %s.", str(e))` |
| `core/app/apps/workflow/app_generator.py` | 690 | `logger.warning("Task stopped: %s", str(e))` |
| `core/model_manager.py` | 927 | `logger.warning("Load balancing config %s failed policy compliance check: %s", config.id, str(e))` |
| `core/rag/index_processor/processor/paragraph_index_processor.py` | 472 | `logger.warning("Failed to convert image file to prompt message content: %s", str(e))` |
| `core/rag/index_processor/processor/paragraph_index_processor.py` | 576 | `logger.warning("Failed to create File object from UploadFile %s: %s", upload_file.id, str(e))` |
| `core/rag/index_processor/processor/paragraph_index_processor.py` | 633 | `logger.warning("Failed to create File object from UploadFile %s: %s", upload_file.id, str(e))` |
| `core/rag/extractor/word_extractor.py` | 135 | `logger.warning("Failed to download image from URL: %s: %s", url, str(e))` |
| `core/entities/provider_configuration.py` | 490 | `logger.warning("Error generating next credential name: %s", str(e))` |
| `providers/trace/trace-aliyun/src/dify_trace_aliyun/data_exporter/traceclient.py` | 89 | `logger.warning("AliyunTrace API check failed: %s", str(e))` |
| `services/summary_index_service.py` | 247 | `logger.warning("Failed to calculate embedding tokens for summary: %s", str(e))` |
| `services/summary_index_service.py` | 918 | `logger.warning("Failed to remove summary vectors: %s", str(e))` |
| `configs/remote_settings_sources/apollo/client.py` | 171 | `logger.warning(str(e))` |
| `configs/remote_settings_sources/apollo/client.py` | 239 | `logger.warning(str(e))` |

Four other sites in `providers/trace/trace-arize-phoenix/...`, `extensions/ext_redis.py`, and `services/enterprise/workspace_sync.py` already pass `exc_info=True`; they are not in scope.

### Why it matters

When the segment-indexing / dataset-cleanup / annotation-reply / model-config load paths fail in production, the on-call engineer has only the exception message (e.g. `KeyError: 'documents'`) to debug with. The full traceback — which would identify the offending code path in seconds — is silently dropped. This pattern recurs across at least 6 modules and 19 sites; a sweep at the call-site level closes the loop in one PR.

### Fix shape

For each site, the most conservative fix is to keep the WARNING level (preserves the original "non-fatal, continue" intent) and replace `str(e)` with `exc_info=True`. The traceback then becomes part of the log record automatically. Format-string args other than `str(e)` are preserved.

```python
# Before
try:
    ...
except Exception as e:
    logger.warning("Failed to enable summaries for segments: %s", str(e))

# After
try:
    ...
except Exception:                          # `as e` no longer used
    logger.warning("Failed to enable summaries for segments", exc_info=True)
```

For the two `logger.warning(str(e))` sites in `apollo/client.py` (no message at all), the fix is `logger.exception("...")` with a short context phrase, since the traceback's first line already carries the exception message.

The `as e` binding is dropped because `str(e)` is the only use of `e` in each of these 19 blocks. This avoids an unused-name ruff finding (the project enables `F841` via its baseline; new sites that bind unused names may be flagged in stricter configs).

### Why WARNING, not ERROR or `logger.exception`

The original call is `logger.warning`, and the surrounding code typically continues execution (`continue`, `pass`, or returns a fallback). Bumping to `logger.exception` (ERROR) would change the level for downstream consumers — alerts, dashboards, log-aggregator severity mappings. The conservative fix preserves the level and only adds the missing traceback, which is a strict observability improvement with no semantic change.

The existing pattern of `logger.exception("...")` (used in many of the same files for outer-level errors) remains in place; this PR only touches the 19 sites where the current call is `logger.warning(...)` with `str(e)` and no `exc_info`.

### Scope

- 19 sites across 13 files. One mechanical change per site.
- A small unit test that exercises one of the modified sites and asserts the captured log record contains a non-empty traceback (smoke test, not a comprehensive test of all 19 sites).
- No API contract change.
- No new dependency.
- Backward compatible: log messages keep the same string format; only the traceback is added.

### Acceptance criteria

- Each of the 19 sites in the table above uses `exc_info=True` and no longer passes `str(e)` to the logger.
- `ruff check` clean on all modified files.
- `ruff format --check` clean on all modified files.
- The new unit test passes and demonstrates the traceback is captured.
- Existing tests in the touched modules still pass.
- `pyrefly check` 0 diagnostics on the modified files (where covered).

### Out of scope (potential follow-ups)

- `logger.error(..., str(e))` without `exc_info` in the vdb/ tree — same bug class but a different sweep.
- `logger.exception(..., str(e))` patterns — already include the traceback; only the redundant `str(e)` would be removable.
- The `except BaseException as e:` in `apollo/client.py:170` — covers a broader exception class than `except Exception`; that's a separate concern (catches `KeyboardInterrupt` / `SystemExit`) tracked by upstream #40801 for some files. This PR keeps the existing `except BaseException` and only fixes the log call.
- Bumping any of the WARNING sites to `logger.exception` — the user-facing log level stays WARNING here.

Harsh23Kashyap:
## Summary

Attached `exc_info=True` to 19 `logger.warning("...: %s", str(e))` sites across 15 files in `api/`, so the traceback is now captured at WARNING level instead of being silently dropped. Same observability shape as the PEP 3134 sweep (cycles 13 #40738 and 16 #40810) but in `logger.*` calls instead of `raise` statements.

## Problem

Nineteen `except Exception as e:` (or `except BaseException as e:`) blocks logged the exception with `str(e)` but no `exc_info=True`. The log line showed the exception message (e.g. `KeyError: 'documents'`) but not the call chain that produced it. The exception is then immediately followed by `continue`, `pass`, or a return — so the traceback is gone after the log line, and the on-call engineer is left with only the str in the log.

This pattern recurs across at least 6 modules and 19 sites. A sweep at the call-site level closes the loop in one PR.

## Fix shape

For each site:

```python
# Before
try:
    ...
except Exception as e:
    logger.warning("Failed to enable summaries for segments: %s", str(e))

# After
try:
    ...
except Exception:                              # `as e` no longer used
    logger.warning("Failed to enable summaries for segments", exc_info=True)
```

- WARNING level preserved (the original "non-fatal, continue" intent).
- The `as e` binding is dropped where `str(e)` was the only use.
- The two bare `logger.warning(str(e))` sites in `apollo/client.py` (no message) are converted to `logger.exception("...")` — the traceback's first line already carries the exception message, and `logger.exception` is the idiomatic Python way to log an exception.

## Sites

| File | Sites |
| --- | --- |
| `tasks/enable_segments_to_index_task.py` | 1 |
| `tasks/disable_segment_from_index_task.py` | 1 |
| `tasks/enable_segment_to_index_task.py` | 1 |
| `tasks/add_document_to_index_task.py` | 1 |
| `tasks/remove_document_from_index_task.py` | 1 |
| `tasks/disable_segments_from_index_task.py` | 1 |
| `core/app/features/annotation_reply/annotation_reply.py` | 1 |
| `core/app/apps/workflow/app_generator.py` | 1 |
| `core/model_manager.py` | 1 |
| `core/rag/index_processor/processor/paragraph_index_processor.py` | 3 |
| `core/rag/extractor/word_extractor.py` | 1 |
| `core/entities/provider_configuration.py` | 1 |
| `providers/trace/trace-aliyun/src/dify_trace_aliyun/data_exporter/traceclient.py` | 1 |
| `services/summary_index_service.py` | 2 |
| `configs/remote_settings_sources/apollo/client.py` | 2 (bare `str(e)` → `logger.exception`) |

## Tests

Extended one existing test that already exercised the error path:

- `tests/unit_tests/core/entities/test_entities_provider_configuration.py::test_generate_next_api_key_name_falls_back_to_default_on_error` — now also asserts:
  - `logger.warning` is called exactly once
  - `exc_info=True` is passed in the kwargs
  - The original exception message is no longer interpolated into the format string (the `str(e)` arg is gone)

The 1 extended test passes (`--noconftest`). The adjacent Redis-dependent tests in the same file fail with the pre-existing `RuntimeError: Redis client is not initialized` (reproduced on clean main, not my changes; same env issue noted in prior cycles).

## Files changed

- 15 source files: one or two-line change per site (`except Exception as e: logger.warning("...: %s", str(e))` → `except Exception: logger.warning("...", exc_info=True)`).
- 1 test file: extended `test_generate_next_api_key_name_falls_back_to_default_on_error` with `exc_info` assertions.

## Verification

- `pytest tests/unit_tests/core/entities/test_entities_provider_configuration.py::test_generate_next_api_key_name_falls_back_to_default_on_error --noconftest -v` → 1/1 pass.
- `ruff check` on all 16 files → All checks passed.
- `ruff format --check` on all 16 files → already formatted.
- `pyrefly check` on the 15 source files → 0 diagnostics.

## Scope

- No API contract change.
- No new dependency.
- Backward compatible: log messages keep the same string format; only the traceback is added. Where WARNING was used, WARNING is preserved; the two bare `logger.warning(str(e))` sites in `apollo/client.py` bump to ERROR via `logger.exception` (the only level change in the sweep).

## Out of scope (potential follow-ups)

- `logger.debug` and `logger.info` sites in `providers/trace/trace-{langsmith,langfuse,weave}/...` that pass `str(e)` — same observability bug class but DEBUG/INFO level. Not in this PR's scope.
- `logger.error(..., str(e))` without `exc_info` in the vdb/ tree — same bug class but a different sweep.
- The `except BaseException as e:` in `apollo/client.py:170` — the broader `BaseException` catch is a separate concern (catches `KeyboardInterrupt`/`SystemExit`), tracked by upstream #40801 for some files. This PR keeps the `except BaseException` and only fixes the log call.
- Bumping any of the WARNING sites to `logger.exception` — the user-facing log level stays WARNING here.

Fixes #41066

### DIFF
--- a/api/configs/remote_settings_sources/apollo/client.py
+++ b/api/configs/remote_settings_sources/apollo/client.py
@@ -167,8 +167,8 @@ class ApolloClient:
                 old_value = old_kv.get(key)
                 if old_value is None:
                     self._change_listener("add", namespace, key, new_value)
-        except BaseException as e:
-            logger.warning(str(e))
+        except BaseException:
+            logger.exception("Apollo change listener raised")
 
     def _path_checker(self) -> None:
         if not os.path.isdir(self._cache_file_path):
@@ -235,8 +235,8 @@ class ApolloClient:
                     return
             else:
                 logger.warning("Sleep...")
-        except Exception as e:
-            logger.warning(str(e))
+        except Exception:
+            logger.exception("Apollo long poll failed")
 
     def _get_net_and_set_local(self, namespace: str, n_id: int, call_change: bool = False) -> None:
         namespace_data = self.get_json_from_net(namespace)
--- a/api/core/app/apps/workflow/app_generator.py
+++ b/api/core/app/apps/workflow/app_generator.py
@@ -686,8 +686,8 @@ class WorkflowAppGenerator(BaseAppGenerator):
             try:
                 with active_workflow_task(application_generate_entity.task_id):
                     runner.run()
-            except GenerateTaskStoppedError as e:
-                logger.warning("Task stopped: %s", str(e))
+            except GenerateTaskStoppedError:
+                logger.warning("Task stopped", exc_info=True)
                 pass
             except InvokeAuthorizationError:
                 queue_manager.publish_error(
--- a/api/core/app/features/annotation_reply/annotation_reply.py
+++ b/api/core/app/features/annotation_reply/annotation_reply.py
@@ -90,8 +90,8 @@ class AnnotationReplyFeature:
                     )
 
                     return annotation
-        except Exception as e:
-            logger.warning("Query annotation failed, exception: %s.", str(e))
+        except Exception:
+            logger.warning("Query annotation failed", exc_info=True)
             return None
 
         return None
--- a/api/core/entities/provider_configuration.py
+++ b/api/core/entities/provider_configuration.py
@@ -486,8 +486,8 @@ class ProviderConfiguration(BaseModel):
             next_number = max(numbers, default=0) + 1
             return f"API KEY {next_number}"
 
-        except Exception as e:
-            logger.warning("Error generating next credential name: %s", str(e))
+        except Exception:
+            logger.warning("Error generating next credential name", exc_info=True)
             return "API KEY 1"
 
     def _get_provider_names(self):
--- a/api/core/model_manager.py
+++ b/api/core/model_manager.py
@@ -923,8 +923,8 @@ class LBModelManager:
                         provider=self._provider,
                         credential_type=PluginCredentialType.MODEL,
                     )
-            except Exception as e:
-                logger.warning("Load balancing config %s failed policy compliance check: %s", config.id, str(e))
+            except Exception:
+                logger.warning("Load balancing config %s failed policy compliance check", config.id, exc_info=True)
                 cooldown_load_balancing_configs.append(config)
                 if len(cooldown_load_balancing_configs) >= len(self._load_balancing_configs):
                     # all configs are in cooldown or failed policy compliance
--- a/api/core/rag/extractor/word_extractor.py
+++ b/api/core/rag/extractor/word_extractor.py
@@ -131,8 +131,8 @@ class WordExtractor(BaseExtractor):
                         continue
                     try:
                         response = remote_fetcher.make_request("GET", url)
-                    except Exception as e:
-                        logger.warning("Failed to download image from URL: %s: %s", url, str(e))
+                    except Exception:
+                        logger.warning("Failed to download image from URL: %s", url, exc_info=True)
                         continue
                     if response.status_code == 200:
                         image_ext = mimetypes.guess_extension(response.headers.get("Content-Type", ""))
--- a/api/core/rag/index_processor/processor/paragraph_index_processor.py
+++ b/api/core/rag/index_processor/processor/paragraph_index_processor.py
@@ -468,8 +468,8 @@ class ParagraphIndexProcessor(BaseIndexProcessor):
                         file, image_detail_config=ImagePromptMessageContent.DETAIL.LOW
                     )
                     prompt_message_contents.append(file_content)
-                except Exception as e:
-                    logger.warning("Failed to convert image file to prompt message content: %s", str(e))
+                except Exception:
+                    logger.warning("Failed to convert image file to prompt message content", exc_info=True)
                     continue
 
             # Add text content
@@ -572,8 +572,8 @@ class ParagraphIndexProcessor(BaseIndexProcessor):
                     access_controller=_file_access_controller,
                 )
                 file_objects.append(file_obj)
-            except Exception as e:
-                logger.warning("Failed to create File object from UploadFile %s: %s", upload_file.id, str(e))
+            except Exception:
+                logger.warning("Failed to create File object from UploadFile %s", upload_file.id, exc_info=True)
                 continue
 
         return file_objects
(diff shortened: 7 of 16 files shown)


crazywoola:

### Why this is being closed

### Next steps

PR #41076, opened 2026-08-21
feat(dify-agent): add OpenShell runtime backend

### LINKED ISSUE #40880: Supporting Nvidia OpenShell as shell provider
### Self Checks

https://github.com/NVIDIA/openshell

_No response_

Jinsong-Zhou:
## Summary

Fixes #40880

- Add `openshell` as a fourth Dify Agent runtime backend (`DIFY_AGENT_RUNTIME_BACKEND=openshell`): one sandbox per Execution Binding on a self-hosted [NVIDIA OpenShell](https://github.com/NVIDIA/openshell) gateway (≥ 0.0.106), reached only through the gateway gRPC API.
- Bootstrap shellctl through idempotent gateway exec on acquire and reach it through an authenticated `ForwardTcp` tunnel; exec scripts travel via stdin so tokens never enter gateway-logged argv, and session tokens auto-renew ahead of expiry.
- Store Home Snapshots as directory copies on an operator-provided shared volume mounted into every sandbox; restore fails rather than falling back when a snapshot is missing.
- Enforce an opt-in egress allowlist (`DIFY_AGENT_OPENSHELL_EGRESS_ALLOW`, comma-separated `host:port`); empty sends no network policy, so sandbox egress keeps following the gateway/driver default.
- Stop the sandbox on release and restart + re-bootstrap it on the next acquire; treat destroy/delete as idempotent on NOT_FOUND.
- Add `iproute2` to the shared sandbox runtime image (an OpenShell supervisor requirement, harmless for other backends), a branch-validation compose overlay, the env reference block, and operator documentation.
- Keep Local, Enterprise, and E2B backends and the backend-neutral runtime contracts unchanged.

## Notes for reviewers

- The Home Snapshot volume is operator-created (two documented commands), not compose-managed, so `docker compose down -v` can never delete snapshots. Production multi-tenant deployments use one OpenShell workspace and one volume per tenant; sandbox names and labels carry deterministic digests, keeping product identifiers out of gateway metadata.
- The egress allowlist is opt-in rather than derived from the stub/files URLs: a spec-supplied allowlist is default-deny for all other egress and would silently cut off agent-run tools (package installs, web access) on deployments whose gateway/driver default is open.
- `dify-agent/Dockerfile` base moves to `python:3.12-slim-trixie`: the `openshell` wheel only ships `manylinux_2_39` binaries. Only the agent-backend image is affected — dify-api installs the base `dify-agent` dependency without the `server` extra.

## Validation

- `uv sync --extra server --dev && uv run pytest` (dify-agent, Linux CI-parity container): 885 passed, 35 skipped.
- `make check`: passed. `pyrefly check`: 0 errors; no new warnings in changed files.
- `go test ./...` (dify-agent-runtime): passed.
- Opt-in integration contract against a real gateway (0.0.106 macOS, 0.0.110 Linux; Docker driver + mTLS): full Binding lifecycle — create → acquire → job via tmux → snapshot save → release → re-acquire → restore into a new Binding → cleanup — passed in ~47–50 s. Run instructions are in the operator guide.
- Compose overlay against a real gateway: Binding create/destroy 201/204; with `DIFY_AGENT_OPENSHELL_EGRESS_ALLOW` set, the create is accepted and both `NetworkPolicyRule`s are persisted on the sandbox.
- UI end-to-end (console Agent app, real LLM): build-mode session writes a file through `shell_run` → applying the build saves a Home Snapshot to the shared volume → a brand-new session restores it on its first message. See screenshots.

## Screenshots
agent-stub
snapshot
workspace

## Limitation
- The `release` does not reclaim sandbox resources, it simply closes connections instead. The sandbox created via OpenShell currently do not have a TTL, which means resource leak could potentially happen if the binding is not destroyed for some reason. Related to https://github.com/NVIDIA/OpenShell/issues/2591. The reason we use light weight release is to reduce the latency and avoid race conditions, otherwise each command waits for the container start/stop, and the latency becomes unacceptable.

## Checklist

(Operator documentation is included in-repo under `dify-agent/docs/`; happy to mirror it to dify-docs if wanted. The backend/frontend lint commands don't apply — this PR changes no `api/` or `web/` code; the dify-agent equivalents `make check && make typecheck` were run instead.)


docker/.env.example
```diff
@@ -287,6 +287,25 @@ DIFY_AGENT_E2B_TEMPLATE=difys-default-team/dify-agent-local-sandbox
 # RuntimeLease active limit; its default matches the independently configurable Agent run deadline.
 DIFY_AGENT_E2B_ACTIVE_TIMEOUT_SECONDS=3600
 DIFY_AGENT_E2B_SHELLCTL_PORT=5004
+# OpenShell backend (DIFY_AGENT_RUNTIME_BACKEND=openshell): variable reference
```
wylswz: This file only contains minimal env vars the system require to run. I think it's sufficient to only declare these variables in `dify-agent.env.example`

docker/envs/core-services/dify-agent.env.example
```diff
@@ -34,6 +34,32 @@ DIFY_AGENT_E2B_TEMPLATE=difys-default-team/dify-agent-local-sandbox
...
+DIFY_AGENT_OPENSHELL_SANDBOX_IMAGE=langgenius/dify-agent-local-sandbox:latest
+# Required JSON driver_config mounting the operator-owned shared volume;
+# one-time volume initialization and driver examples live in the guide.
+DIFY_AGENT_OPENSHELL_DRIVER_CONFIG=
+DIFY_AGENT_OPENSHELL_SHARED_MOUNT_PATH=/mnt/dify-agent-shared
+# Required. shellctl path isolation stays off (sandbox Landlock is authoritative).
+DIFY_AGENT_OPENSHELL_SHELLCTL_AUTH_TOKEN=
+DIFY_AGENT_OPENSHELL_SHELLCTL_PORT=5004
+DIFY_AGENT_OPENSHELL_READY_TIMEOUT_SECONDS=300
+DIFY_AGENT_OPENSHELL_EXEC_TIMEOUT_SECONDS=120
```
wylswz, line 62: Accessing to agent-stub and dify-api's file endpoint from sandbox may get rejected by OpenShell's policy, consider adding a white list of permitted domains along with the **client binaries** that send the requests when starting the sandbox.

For my local setup, I did

```python
        for endpoint in self.egress_allow:
            host, _, port_str = endpoint.rpartition(":")
            if not host or not port_str.isdigit():
                raise BindingCreateError(f"invalid OpenShell egress endpoint (want host:port): {endpoint!r}")
            rule_name = "allow_" + "".join(c if c.isalnum() else "_" for c in endpoint)
            policy.network_policies[rule_name].CopyFrom(
                sandbox_pb2.NetworkPolicyRule(
                    name=rule_name,
                    endpoints=[
                        sandbox_pb2.NetworkEndpoint(
                            host=host,
                            port=int(port_str),
                            protocol="rest",
                            enforcement="enforce",
                            rules=[sandbox_pb2.L7Rule(allow=sandbox_pb2.L7Allow(method="*", path="/**"))],
                        )
                    ],
                    # OpenShell also gates which in-sandbox binaries may open the
                    # connection; "/**" allows any (the Agent Stub client, plus
                    # tools the agent runs). Endpoints stay the actual allowlist.
                    binaries=[sandbox_pb2.NetworkBinary(path="/**")],
                )
            )
```

dify-agent/docs/dify-agent/guide/index.md
```diff
@@ -222,6 +236,200 @@ provider `RuntimeError` observed first becomes a tool observation. In contrast,
 run-deadline cancellation propagates through the Shell boundary; only the Dify
 Agent run deadline owns the terminal `agent_run_limit_exceeded` failure.
 
+## Deploy with the OpenShell backend
+
+The OpenShell backend runs each Binding in its own sandbox on a self-hosted
```
wylswz, line 241: Consider documenting a minimum version required for openshell.

wylswz:
@Jinsong-Zhou Thanks for the contribution! Could you please attach some screenshots of your local tests, illustrating that agent-stub, snapshot and workspaces are working as expected?

docker/.env.example
Jinsong-Zhou: Done — and largely resolved by the rebase itself: #41065 removed the per-variable passthrough and the E2B block this one was sitting next to, so after rebasing onto current main the `DIFY_AGENT_OPENSHELL_*` block is gone from `docker/.env.example` (and from the compose template). The variable reference now lives only in `docker/envs/core-services/dify-agent.env.example`, and the operator guide was updated to point there. Values set in `docker/.env` still reach `agent_backend` through its `env_file` list and override the service env file.

docker/envs/core-services/dify-agent.env.example
Jinsong-Zhou: Implemented in 488d9cd20a as an opt-in allowlist: `DIFY_AGENT_OPENSHELL_EGRESS_ALLOW` (comma-separated `host:port`) sends one enforced `NetworkPolicyRule` per endpoint with the create spec, same shape as your snippet (`protocol="rest"`, `enforcement="enforce"`, `L7Allow(method="*", path="/**")`, `binaries=[NetworkBinary(path="/**")]`); an index prefix keeps rule names unique when sanitized endpoints would collide.

Kept it opt-in rather than derived from the stub/files URLs: a spec-supplied allowlist is default-deny for all other egress, so deriving it implicitly would silently cut off agent-run tools (package installs, web access) on deployments whose gateway/driver default is open. Empty (the default) sends no `network_policies` and preserves current behavior.

Malformed entries are rejected at startup validation; unit tests pin the outgoing create-spec shape both ways (allowlist set / empty). Documented in the operator guide (deployment prerequisite 2 + env table) and `dify-agent.env.example`. Validated against a real gateway (0.0.110, Docker driver + mTLS): Binding create with the allowlist configured returns 201 and `openshell sandbox get` shows both rules persisted on the sandbox.

dify-agent/docs/dify-agent/guide/index.md
Jinsong-Zhou: Done in 488d9cd20a — deployment prerequisite 1 now states gateway ≥ 0.0.106, matching the SDK pin (`>=0.0.106,<0.1.0`; validated against 0.0.106 and 0.0.110).

Jinsong-Zhou:
> @Jinsong-Zhou Thanks for the contribution! Could you please attach some screenshots of your local tests, illustrating that agent-stub, snapshot and workspaces are working as expected?

@wylswz Thanks for the review! Screenshots are now in the PR body (Screenshots section), taken from the local compose deployment (Docker driver + mTLS gateway, `DIFY_AGENT_RUNTIME_BACKEND=openshell`), covering the three paths:

- **agent-stub** — build-mode session: the agent's `shell_run` round-trips through the authenticated ForwardTcp tunnel (request script, `exit_code: 0` metadata, output echoed back to the chat).
- **snapshot** — after applying the build, a brand-new session (new Binding, new sandbox) reads `~/pr41076-test.txt` on its first message — the home was restored from the Home Snapshot saved to the shared volume at apply time.
- **workspaces** — gateway side: `openshell sandbox list` shows the per-Binding `dify-<digest>` sandbox (Stopped between runs per the release/acquire lifecycle), the snapshot directories on the operator-owned shared volume, and the agent-backend lifecycle log (`/home-snapshots/from-binding` 201 on apply, Binding create/destroy).

The branch is also rebased onto current main, and the inline comments are addressed in 488d9cd20a — see the replies on each thread (opt-in `DIFY_AGENT_OPENSHELL_EGRESS_ALLOW` allowlist, `docker/.env.example` cleanup via the rebase, documented gateway minimum version).

wylswz:
Hi @Jinsong-Zhou, thanks for the work, I will accept this PR with known limitation, which is described in PR's summary. To fully resolve it, we can either wait for OpenShell to provide a native expiration mechanism, or manage this in Dify.

wylswz (approved):

Issue #41092, opened 2026-08-22
Claude extended thinking: signed/encrypted thinking blocks are dropped between Agent tool-use rounds, causing 400 errors


hpiclaranet:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.16.1

### Cloud or Self Hosted

- [x] Cloud
- [x] Self Hosted (Docker)
- [x] Self Hosted (Source)

*Expected to apply to all deployments running the plugin daemon (code-path inference; no per-deployment reproduction performed).*

### Steps to reproduce

*Scope note: this report is based on code-path analysis rather than a live-deployment reproduction. It was verified against the cited sources — `dify_plugin` SDK 0.10.0, the official `anthropic` plugin v0.3.27, current Dify core / graphon sources, and the official Anthropic API documentation — and applies to releases built from the same code paths.*

1. Use an official Anthropic plugin model with thinking. No user configuration is needed to reach this state on the current model generation — per the [official docs](https://platform.claude.com/docs/en/build-with-claude/thinking#configuring-thinking), "On Claude Opus 5, Claude Sonnet 5, Claude Fable 5, Claude Mythos 5, and Claude Mythos Preview, thinking is already on: no configuration needed" (the failure still only manifests on turns where the model actually emits a thinking block — adaptive thinking may skip it on simple requests).
2. Create an **Agent** app using the function-calling strategy with at least one tool — or a workflow **Agent** node — such that a request can require more than one LLM round (assistant `tool_use` → tool result → next LLM call).
3. Run a query that forces a tool call.
4. First LLM call: the API returns `thinking` (and possibly `redacted_thinking`) + `tool_use` blocks. This round succeeds.
[...]

PR #41095, opened 2026-08-22
fix(agent): carry provider opaque_body state across function-calling agent iterations

hpiclaranet:
> [!IMPORTANT]
>

## Summary

Part of #41092 (Phase 1 — core-side, in-memory agent loop).

`FunctionCallAgentRunner` rebuilt each assistant turn from the flattened text plus tool calls only, so provider round-trip state exposed by plugins via `AssistantPromptMessage.opaque_body` (e.g. signed/encrypted Claude extended-thinking blocks) was lost between LLM iterations. With Anthropic extended thinking + multi-round tool use this causes HTTP 400 on the next round: `thinking or redacted_thinking blocks in the latest assistant message cannot be modified`.

This PR captures the provider state and replays it:

- `api/core/agent/fc_agent_runner.py`: accumulate `opaque_body` from the LLM result — streaming chunks (complete-snapshot semantics, last non-`None` wins) and blocking results — and attach it to the assistant message appended to `_current_thoughts`, so the next iteration's prompt carries it back to the provider.
- `api/tests/unit_tests/core/agent/test_fc_agent_runner.py`: 5 new tests (`TestOpaqueBodyRoundTrip`) covering blocking capture, streaming capture with no-clobber, default `None`, and end-to-end in-memory round trips for both blocking and streaming tool-call turns (second `invoke_llm` call receives the assistant message carrying `opaque_body`).

Dependencies (see commits):

1. `chore(deps)`: declares `pypandoc` / `pypandoc-binary` directly — `api/core/rag/extractor/unstructured/unstructured_epub_extractor.py` imports pypandoc, but it was only satisfied transitively via graphon's PyPI wheel.
2. **TEMPORARY** `graphon` source override (rev-pinned at the cherry-picked `v0.7.0`-based commit, zero drift from the current `graphon==0.7.0` pin) so CI can validate against the opaque_body entity before the PyPI release exists. **This draft cannot merge as-is**: once [langgenius/graphon#261](https://github.com/langgenius/graphon/pull/261) is released, the source override is removed and the pin becomes `graphon==0.7.1`.

Scope and safety notes:

- Fixes the in-run (in-memory) agent loop only. Cross-session history reconstruction (`organize_agent_history`, `TokenBufferMemory`) still drops the field — Phase 2 follow-up.
- No new data exposure: agent-thought/message queue events and DB prompt saving project only content/tool-call fields; the signed state stays in the in-memory prompt path (verified against the queue-to-HTTP and prompt-saving code paths).
- Deployment ordering is safe in either direction: core without the plugin = no behavior change; plugin without core = field silently dropped (today's behavior). End-to-end Anthropic support additionally requires the plugin-side counterpart [langgenius/dify-official-plugins#3715](https://github.com/langgenius/dify-official-plugins/pull/3715).

Validation: full api unit suite green (15,000+ tests), `make lint` and `make type-check` (pyrefly + mypy) green, agent suite 247 passed.

## Screenshots

Not applicable — backend-only change.

## Checklist


hpiclaranet:
Heads-up to maintainers — this draft needs two things from the core side:

1. **Workflow approval** for the fork PR (several checks are `action_required`).
2. **Review of the companion PR [langgenius/graphon#261](https://github.com/langgenius/graphon/pull/261)** (entity/accumulator change, small and additive). This draft pins graphon at the cherry-picked commit of that PR (zero drift from the current `0.7.0` pin) so its CI is green. It is a **draft because it cannot merge** until the graphon release containing #261 is published — then the temporary `[tool.uv.sources] graphon` override comes out and the pin becomes `graphon==0.7.1`.

One heads-up for that eventual pin bump: graphon `main` has drifted since the `0.7.0` release (added `odfdo`, removed the `unstructured[epub]` extra that was transitively providing `pypandoc`, LLM node changes). The `chore(deps)` commit here already declares `pypandoc`/`pypandoc-binary` directly (they were an undeclared direct import of `core/rag/extractor/unstructured`), but the `0.7.1` bump will likely need a few dify test updates for the graphon main behavioral changes.

Tracking: #41092 · graphon #260/#261 · plugin side dify-official-plugins#3715.

PR #41102, opened 2026-08-23
fix(security): constant-time secret compares; stop echoing webhook errors

simpleqt:
## What changed
1. Secret-material comparisons used plain `!=`: the inner-api key checks (`X-Inner-Api-Key`, plugin inner key), the admin API key check, and both HMAC file-signature verifications. `libs/token.py` already uses `hmac.compare_digest` for the same admin key; this PR aligns the remaining call sites.
2. `controllers/trigger/webhook.py` returned `{message: str(e)}` on 500, echoing internal exception text to an **unauthenticated** caller (the debug endpoint in the same file deliberately avoids this). The exception stays in the server log only.

## Verification
- `python -m py_compile` on all touched files; `compare_digest` args are both ASCII strings (header value / configured key / base64 signature).

### DIFF
--- a/api/controllers/console/admin.py
+++ b/api/controllers/console/admin.py
@@ -1,3 +1,4 @@
+from hmac import compare_digest
 from collections.abc import Callable
 from functools import wraps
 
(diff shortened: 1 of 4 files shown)


simpleqt:
Rebased onto current `main` to re-run CI. Gentle ping for review — happy to make any changes if needed. Thanks!

simpleqt:
Two CI gates were red on the previous head; both fixed and squashed into a single commit (ab010e0b):

- **Validate PR title** – retitled to `fix(security): …` (the bare `security:` type isn't in the allowed set).
- **Python Style (pyrefly)** – `ADMIN_API_KEY` / `INNER_API_KEY` are `str | None`, so `compare_digest` no longer matched the `(Buffer, Buffer)` overload. Both are now bound to locals with the Optional guard before the constant-time compare — same behavior as the previous `!=` (unset config still rejects), and it type-checks. Verified with the repo-pinned pyrefly 1.3.1 (`no-matching-overload` reproduces on the old code, 0 diagnostics after the fix) and ruff 0.16.6 clean.

PR #41103, opened 2026-08-23
fix(security): require login on trial explore read APIs

simpleqt:
## What changed
Five trial explore read resources were mounted **without** `login_required`:

- `TrialSitApi` (site info)
- `TrialAppParameterApi` (app parameters)
- `AppApi` (app detail incl. full app model config)
- `AppWorkflowApi` (full workflow graph incl. `environment_variables` plaintext values)
- `DatasetListApi` (template tenant dataset enumeration)

They only had `@with_session` + `@get_previewable_app_model`, neither of which authenticates. For comparison, `RecommendedAppListApi` — which exposes only catalog summaries — requires `@login_required @account_initialization_required`, and the trial execution endpoints inherit `TrialAppResource` whose `method_decorators` include `login_required`. The read side looks like an oversight rather than an intentional public surface.

An unauthenticated caller holding a previewable app UUID (Dify Cloud / deployments with catalog apps) could read the full workflow graph including plaintext environment variables.

This PR adds `@login_required` + `@account_initialization_required` to the five classes, aligned with `recommended_app.py`.

## Verification
- `python -m py_compile` passes; decorator stack mirrors `RecommendedAppListApi`.

### DIFF
--- a/api/controllers/console/explore/trial.py
+++ b/api/controllers/console/explore/trial.py
@@ -48,7 +48,13 @@ from controllers.console.explore.error import (
 from controllers.console.explore.wraps import TrialAppResource
 from controllers.console.files import FILE_UPLOAD_PARAMS, upload_file_from_request
 from controllers.console.remote_files import RemoteFileUploadPayload, upload_remote_file_from_request
-from controllers.console.wraps import cloud_edition_billing_resource_check, model_validate, with_current_user
+from controllers.console.wraps import (
+    account_initialization_required,
+    cloud_edition_billing_resource_check,
+    model_validate,
+    with_current_user,
+)
+from libs.login import login_required
 from controllers.web.error import InvokeRateLimitError as InvokeRateLimitHttpError
 from core.app.apps.base_app_queue_manager import AppQueueManager
 from core.app.entities.app_invoke_entities import InvokeFrom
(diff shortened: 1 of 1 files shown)


simpleqt:
Rebased onto current `main`: main's import refactor changed this file's imports (and dropped `account_initialization_required` from it), so the PR now adds only `@login_required` on the five unauthenticated read endpoints — the minimal fix for the reported hole.

simpleqt:
Rebased onto current `main` (single commit 6e07836, same 6-line change: `@login_required` on the five trial explore read endpoints) to clear the stale 09-18 E2E shard failure and the cancelled title check; that Chromium Full (1/3) shard passed on re-run before, expecting infra flake. Happy to address anything from review.

simpleqt:
The branch went CONFLICTING because upstream rewrote this file (per-resource `@with_session`/`@get_previewable_app_model` → `@get_preview_app` + `application_services()` query services, and `@console_account_admission()` on the execution endpoints). Force-pushed a rebuilt patch on current main instead of a merge:

- `@login_required` now lands on the four resources that still only have `@get_preview_app`: `TrialSitApi`, `TrialAppParameterApi`, `AppWorkflowApi`, `DatasetListApi`.
- `AppApi` (app detail) is **dropped from scope**: the refactor gave it `@console_account_admission()`, and that stack already wraps `login_required` + `account_initialization_required` (see `controllers/console/flask_admission.py`).

Net diff is back to 5 lines (1 import + 4 decorators). `py_compile`, `ruff check` and `ruff format --check` green on the file.

PR #41131, opened 2026-08-24
fix(api): scope hosted credits by tenant plan

linw1995:
## Summary

- Filter hosted credit providers by the current workspace subscription plan.
- Keep Trial providers available to Sandbox tenants while hiding PAID-only providers such as xAI.
- Add regression coverage across the feature service, gateway, query service, and controller paths.

Related issue: [SNP-697](https://linear.app/dify/issue/SNP-697/支持-xai-仅对-paid-租户启用-credits)

## Screenshots

Not applicable (backend-only change).

## Checklist

- [x] This change requires no documentation update.
- [x] I've added tests for the introduced behavior and kept the change atomic.

### DIFF
--- a/api/controllers/console/feature.py
+++ b/api/controllers/console/feature.py
@@ -75,18 +75,18 @@ class FeatureVectorSpaceApi(Resource):
 @console_ns.route("/trial-models")
 class TrialModelsApi(Resource):
     @console_ns.doc("get_trial_models")
-    @console_ns.doc(description="Get hosted trial model provider configuration")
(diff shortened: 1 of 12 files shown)


linw1995:
Addressed in commit `3374a09d69`.

`get_trial_models` now requires the current workspace context. The static `_trial_models` constructor state and the no-context fallback were removed, so hosted credit provider discovery always uses the workspace subscription plan.

crazywoola (approved):

Issue #17065, opened 2025-03-28
Confusion about weighted rerank


bohaocheung:
_Originally posted by @bohaocheung in [#17048](https://github.com/langgenius/dify/issues/17048#issuecomment-2760779994)_

> I am still confused.
> There are 2 rerank ways: One is using specific reranking model whose input is query full text and output is  `document.metadata["score"]`.
> Another is keyword and vector weighted hybrid reranking, which I talked in this issue exactly. In this condition, I think the code confused the meaning of `document.metadata["score"]`.

Issue #13426, opened 2025-02-08
Knowledge APIs: impossible to setup a KB to use hybrid search in weighted score mode


DavideDelbianco:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

0.15.3

### Cloud or Self Hosted

Cloud

### Steps to reproduce

Use the knowledge API to create an empty dataset, without specify"indexing_technique" to avoid locking it to the default "vector_search"
Upload a document by API and specify a "search_method": "hybrid_search"  -> ???

### ✔️ Expected Behavior

The first uploaded document that specify the retrieval_model with "search_method": "hybrid_search"
Should be allowed to setup the "weighted_score" mode and its weights

### ❌ Actual Behavior

It is impossible to specify the hybrid_search + weighted_score mode by API

DavideDelbianco:
@dosu
a) reranking_mode is a property not documented
b) the weights configuration is not documented
c) even when using the proper configuration will result in the values not being applied, this is especially noticeable when setting a value other than the UI default 0.7 / 0.3 (try with 0.5 / 0.5)
d) also the UI is bugged and will not retain the configuration setup during the KB creation until manually updated through the UI in the KB settings

lucasvalentee:
Any news?

DavideDelbianco:
Yes the First doc decides

PR #41152, opened 2026-08-24
fix: Weaviate full-text search results missing score in metadata

### LINKED ISSUE #41151: [Bug]Weaviate full-text search results carry no score, causing Hybrid Search to drop valid full-text-only matches
### Self Checks

### Dify version

1.14.1 (confirmed still present, unchanged, on current `main` as of this report)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Create a dataset with indexing technique "High Quality" using Weaviate as the vector store, containing many near-duplicate/similar entries (e.g. a schema dictionary with hundreds of tables sharing similar naming and boilerplate columns).
2. Set the dataset's (or a workflow Knowledge Retrieval node's) search method to "Hybrid Search" with reranking disabled (`reranking_enable: false`), using the default weighted-score blending of `vector_weight` + `keyword_weight`.
3. Query for something a full-text/keyword match would find well but that has many semantically-similar competing documents (e.g. an exact table/identifier name that shares a lot of vocabulary with many similarly-named records).
4. Compare against running the same query with "Full-text search" alone (semantic off) — the correct document is found. Compare against "Semantic search" alone — often found too, just ranked lower. In "Hybrid Search", the same document is frequently missing entirely from the returned `top_k` results.

Root cause (traced in source): `WeaviateVector.search_by_full_text()` in
`api/providers/vdb/vdb-weaviate/src/dify_vdb_weaviate/weaviate_vector.py`
builds `Document` objects from the BM25 query results without ever setting
`metadata["score"]`:

```python
docs.append(Document(page_content=text, vector=vec, metadata=properties))
```

Compare to `search_by_vector()` right above it, which explicitly computes and
attaches a score:

```python
score = 1.0 - distance
if score > score_threshold:
    properties["score"] = score
    docs.append(Document(page_content=text, metadata=properties))
```

Any code path that later reads `document.metadata.get("score", 0.0)` —
most notably when merging/ranking the combined vector + full-text result
set in hybrid retrieval — treats every full-text/BM25 hit as score `0`,
regardless of how relevant it actually was. Full-text-only matches
therefore always lose to vector hits once the two are combined and
truncated to `top_k`, even though plain "Full-text search" mode (which
never needs to compare against vector scores) works fine and confirms the
underlying BM25 retrieval itself is correct.

I believe this is the same root cause behind the "fewer results than
configured Top K" symptom reported in #32421, and is related to the
confusion described in #13426 and #17065.

I have a fix ready and will open a PR referencing this issue.

### ✔️ Expected Behavior

Full-text/BM25 search results should carry a real relevance score in
`metadata["score"]`, so hybrid search can rank/merge them fairly against
vector search results instead of always treating them as the lowest
possible score.

### ❌ Actual Behavior

In Hybrid Search mode, documents that are correctly found by full-text
search (and confirmed present via standalone "Full-text search" mode) are
missing from the final results, because they are always scored 0 and get
outranked/truncated by vector hits.

shinyj0:
> [!IMPORTANT]
>

## Summary

Fixes #41151

`WeaviateVector.search_by_full_text()` built `Document` objects from BM25
query results without ever setting `metadata["score"]`, unlike
`search_by_vector()` which explicitly computes and attaches one. Any caller
that ranks/filters results by `metadata.get("score", 0.0)` — most notably
hybrid search, which merges vector and full-text hits and sorts the
combined list by score — treats every full-text/BM25 match as a flat `0`
regardless of how relevant it actually was. In practice, a document that
only the keyword search finds (e.g. an exact identifier with little
semantic overlap to the query) gets outranked by unrelated vector hits and
silently dropped before reaching `top_k`, especially in knowledge bases
with many similar documents. Standalone "Full-text search" mode is
unaffected since it never needs to compare its scores against anything
else, which is why this was easy to miss.

Fix: request `return_metadata=MetadataQuery(score=True)` on the BM25 query
and copy `obj.metadata.score` into `properties["score"]`, mirroring what
`search_by_vector()` already does for vector hits.

Related: #32421, #13426, #17065 describe symptoms (fewer/wrong results in
hybrid + weighted-score mode) consistent with this root cause.

## Screenshots

Not applicable — backend retrieval scoring change, no UI difference.

## Checklist

From Claude Code

### DIFF
--- a/api/providers/vdb/vdb-weaviate/src/dify_vdb_weaviate/weaviate_vector.py
+++ b/api/providers/vdb/vdb-weaviate/src/dify_vdb_weaviate/weaviate_vector.py
@@ -480,6 +480,7 @@ class WeaviateVector(BaseVector):
                 query_properties=[Field.TEXT_KEY.value],
                 limit=top_k,
                 return_properties=props,
+                return_metadata=MetadataQuery(score=True),
                 include_vector=True,
                 filters=where,
             )
@@ -490,6 +491,7 @@ class WeaviateVector(BaseVector):
                 query_properties=[Field.TEXT_KEY.value],
                 limit=top_k,
                 return_properties=props,
+                return_metadata=MetadataQuery(score=True),
                 include_vector=True,
                 filters=where,
             )
@@ -503,6 +505,12 @@ class WeaviateVector(BaseVector):
             if isinstance(vec, dict):
                 vec = vec.get("default") or next(iter(vec.values()), None)
 
+            if obj.metadata and obj.metadata.score is not None:
+                # Without this, BM25 hits carry no score at all, so callers that
+                # rank documents by metadata["score"] (e.g. hybrid search merging
+                # vector + full-text results) treat every full-text match as 0 and
+                # it gets outranked/dropped regardless of how relevant it actually is.
+                properties["score"] = obj.metadata.score
             docs.append(Document(page_content=text, vector=vec, metadata=properties))
         return docs
 
--- a/api/providers/vdb/vdb-weaviate/tests/unit_tests/test_weaviate_vector.py
+++ b/api/providers/vdb/vdb-weaviate/tests/unit_tests/test_weaviate_vector.py
@@ -561,6 +561,7 @@ class TestWeaviateVector(unittest.TestCase):
             "doc_type": "image",
         }
         mock_obj.vector = {"default": [0.1] * 128}
+        mock_obj.metadata = SimpleNamespace(score=None)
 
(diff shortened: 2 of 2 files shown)


crazywoola:
Hi @shinyj0, thanks for opening this pull request.

### Why this is being closed

The linked issue https://github.com/langgenius/dify/issues/41151 was closed because its reported Dify version is 1.15.0 or older.

### Next steps

If the change is still needed on the latest release, please retest there and open a current issue and pull request with the updated reproduction details.

PR #41176, opened 2026-08-24
fix(agent): memoize and fail-open LOG provider icon enrichment

### LINKED ISSUE #41143: Agent LOG icon enrichment performs full plugin and built-in provider scans in the execution path
### Self Checks

### Dify version

`main` at commit `0024a845f7098d0bdf0878fd1a2db5c870f8d3f0` (2026-08-24)

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

1. Run an Agent node whose strategy emits one or more `LOG` messages with a non-empty `metadata.provider` value.
2. Instrument the plugin daemon management endpoint, or place counters/breakpoints on `PluginInstaller.list_plugins()` and `BuiltinToolManageService.list_builtin_tools()`.
3. Execute the Agent and inspect each `LOG` message before its `AgentLogEvent` is emitted.
4. Observe that every matching `LOG` message synchronously performs both full-list operations, even though their results are used only to add display icons.

The current call chain is:

- [`AgentMessageTransformer.transform()`](https://github.com/langgenius/dify/blob/0024a845f7098d0bdf0878fd1a2db5c870f8d3f0/api/core/workflow/nodes/agent/message_transformer.py#L194-L228) handles a `LOG` message and calls both list operations before yielding the log event.
- [`PluginInstaller.list_plugins()`](https://github.com/langgenius/dify/blob/0024a845f7098d0bdf0878fd1a2db5c870f8d3f0/api/core/plugin/impl/plugin.py#L63-L70) requests page 1 with `page_size=256` from the plugin daemon, after which the caller performs a linear search for one provider.
- [`BuiltinToolManageService.list_builtin_tools()`](https://github.com/langgenius/dify/blob/0024a845f7098d0bdf0878fd1a2db5c870f8d3f0/api/services/tools/builtin_tools_manage_service.py#L625-L680) enumerates all built-in provider controllers, loads default provider configurations, converts providers with `decrypt_credentials=True`, repacks icons, and converts every tool schema.
- The resulting values are used only for `metadata.icon` and `metadata.icon_dark`.

This is a distinct Agent LOG enrichment path from #37334, which reported redundant plugin-daemon model-list calls during concurrent workflow execution.

### ✔️ Expected Behavior

Agent execution and log delivery should not synchronously enumerate all installed plugins and all built-in providers/tools solely to resolve two presentation fields.

Icon enrichment should be outside the execution-critical path, use a targeted or safely cached lookup, or otherwise fail open. A slow or unavailable plugin daemon, credential decryption failure, or unrelated provider conversion failure should not delay or fail an Agent run when only `icon` and `icon_dark` are missing.

Repeated `LOG` messages for the same provider in one execution should not repeat the same full-list work.

### ❌ Actual Behavior

For every `LOG` message with `metadata.provider`, the Agent message transformer currently:

1. Makes a synchronous plugin-daemon request that returns up to 256 plugins and linearly scans the result.
2. Enumerates and fully converts all built-in providers and their tools, including default configuration loading and credential decryption.
3. Performs both operations before emitting the log event and only uses the result to populate `icon` and `icon_dark`.

There is no per-execution memoization in this path. The work is therefore repeated across matching log messages. The plugin-daemon error path is also handled as an Agent message transformation failure in [`AgentNode._run()`](https://github.com/langgenius/dify/blob/0024a845f7098d0bdf0878fd1a2db5c870f8d3f0/api/core/workflow/nodes/agent/agent_node.py#L168-L195), allowing decorative metadata enrichment to affect execution success.

Harsh23Kashyap:
## Summary

The Agent message transformer synchronously called
`PluginInstaller.list_plugins(tenant_id)` (which fetches page 1 of the
plugin daemon list, up to 256 entries) and
`BuiltinToolManageService.list_builtin_tools(user_id, tenant_id)`
(which enumerates every built-in provider controller, loads default
configurations, decrypts credentials, and converts every tool schema)
for every `LOG` message whose `metadata.provider` was set, only to
populate two presentation fields (`icon` and `icon_dark`).

This affected Agent execution on the hot path in three ways:

1. **Repeated work.** The same provider typically emits multiple `LOG`
   messages during one execution, and each one repeated both full
   scans. There was no per-execution memoization.

2. **Slow / unavailable plugin daemon, credential decryption error, or
   unrelated schema conversion failure in the built-in path all
   surfaced as `PluginDaemonClientSideError` at the AgentNode level,
   which `agent_node.py:_run` converted into a failed execution. A
   decorative icon lookup should never fail an Agent run.

3. **`icon` and `icon_dark` are display metadata only.** Forcing them
   to be resolvable before yielding the log event couples execution
   success to a presentation path.

## Fix

Extract icon resolution into a static helper
`_resolve_provider_icons` on `AgentMessageTransformer` that:

- Uses a **function-local cache** (`provider_icon_cache`) in
  `transform()`, not a class-level one, because the AgentNode factory
  returns a shared `AgentMessageTransformer` instance and a class-level
  cache would leak icons across concurrent runs.

- Catches any exception from `PluginInstaller.list_plugins` and from
  `BuiltinToolManageService.list_builtin_tools`, returning an empty
  list so the lookup degrades to "not found" rather than propagating.

- Caches the final `(icon, icon_dark)` tuple for each provider. Hits
  reuse the cached value without touching the plugin daemon or the
  built-in provider list.

- Is wrapped in an outer try/except in the LOG message branch as
  defense in depth: any uncaught failure in the helper (cache bug,
  unexpected return type, etc.) leaves the metadata with the seed
  icon from `tool_info` and `icon_dark=None` rather than raising.

## Behavior preserved

- When the provider is found in the plugin list, `icon` is set from
  the plugin declaration and `icon_dark` stays `None` (matches the
  pre-fix behavior).
- When the provider is found in the built-in list, both `icon` and
  `icon_dark` are set from the built-in entity.
- When the provider is not found in either list, `icon` keeps the
  agent-strategy seed icon from `tool_info["icon"]` and `icon_dark`
  stays `None`.
- The `LOG` event is still emitted with the resolved metadata; only
  the lookup cost and the failure surface change.

## Measured impact

Per the issue reporter's instrumentation, a single Agent `LOG`
emission for a non-empty `metadata.provider` triggered both
`list_plugins` and `list_builtin_tools` end-to-end. With this fix:

- The first emission for a given provider still pays the full lookup
  cost (same as pre-fix).
- Every subsequent emission for the same provider within one
  execution is a dict lookup, no I/O.
- A slow or unavailable plugin daemon no longer fails the run; the
  built-in path (or the seed icon from `tool_info`) covers it.
- A `decrypt_credentials=True` failure for one built-in provider no
  longer aborts the rest of the built-in list build and the Agent
  execution.

## Test coverage (7 new tests in `test_message_transformer.py`)

- `test_log_icon_enrichment_repeated_provider_uses_cache` - three
  LOG messages for the same provider hit `list_plugins` and
  `list_builtin_tools` only once.
- `test_log_icon_enrichment_fails_open_when_plugin_daemon_unavailable`
  - the built-in path still resolves when `list_plugins` raises.
- `test_log_icon_enrichment_fails_open_when_builtin_service_unavailable`
  - the plugin path still resolves when `list_builtin_tools` raises.
- `test_log_icon_enrichment_fails_open_when_both_lookups_raise` -
  both raise -> metadata falls back to the seed icon, no dark.
- `test_log_icon_enrichment_uses_plugin_icon_when_not_in_builtins`
  - provider in plugin only -> plugin icon, no dark.
- `test_log_icon_enrichment_uses_seed_when_provider_unknown` -
  provider unknown to both -> seed icon, no dark, plugin daemon
  queried once.
- `test_log_icon_enrichment_handles_uncaught_helper_failure` - a
  bug in the helper itself (uncaught exception) -> outer try/except
  recovers with the seed icon, log event still emitted.

## Verification

| step | tool | result |
|---|---|---|
| `ruff format --check` on the 2 touched files | ruff | 2 files already formatted |
| `ruff check` on the 2 touched files | ruff | All checks passed |
| `pytest tests/unit_tests/core/workflow/nodes/agent/test_message_transformer.py` | pytest 9 | 15 passed (8 existing + 7 new) |
| `pytest tests/unit_tests/core/workflow/` (regression) | pytest 9 | 1236 passed, 13 warnings, no new failures |
| `pyrefly check core/workflow/nodes/agent/message_transformer.py` | pyrefly 1.2.0 | 0 diagnostics |

## /consult-kiro note

Per repo policy I ran `/consult-kiro` on the diff before opening this
PR. The local OpenCode-Zen panel returned a `meta.degraded: true`
envelope with `ok_count: 0` of 3 models: two failed with
`No payment method` (workspace billing page required) and the third
hung in the model-build phase. I did not block on the missing second
opinion. The implementation is independently reviewed by the seven
new tests above and the existing 1236-workflow-test regression run.

## Out of scope (called out for the reviewer)

- A future hardening could move the icon enrichment into a
  post-execution `AgentLogEvent` transformer that runs in a
  background task. That would require a larger API change to the
  event shape and is a separate concern from the per-execution
  memoization + fail-open covered here.
- The same N+1 pattern may exist for other node types (tool nodes,
  knowledge retrieval nodes). This PR is scoped to the Agent LOG
  path that the issue flagged.

Fixes #41143

### DIFF
--- a/api/core/workflow/nodes/agent/message_transformer.py
+++ b/api/core/workflow/nodes/agent/message_transformer.py
@@ -45,8 +45,6 @@ class AgentMessageTransformer:
         node_id: str,
         node_execution_id: str,
     ) -> Generator[NodeEventBase, None, None]:
-        from core.plugin.impl.plugin import PluginInstaller
-
         message_stream = ToolFileMessageTransformer.transform_tool_invoke_messages(
             messages=messages,
             user_id=user_id,
@@ -62,6 +60,11 @@ class AgentMessageTransformer:
         agent_execution_metadata: Mapping[WorkflowNodeExecutionMetadataKey, Any] = {}
         llm_usage = LLMUsage.empty_usage()
         variables: dict[str, Any] = {}
+        # Per-execution cache for LOG provider icon enrichment. The AgentNode
+        # factory returns a shared AgentMessageTransformer instance, so a
+        # class-level cache would leak icons across concurrent runs; a
+        # function-local cache scopes the memoization to one transform() call.
+        provider_icon_cache: dict[str, tuple[str | Mapping[str, str], str | Mapping[str, str] | None]] = {}
 
         for message in message_stream:
             if message.type in {
@@ -194,35 +197,24 @@ class AgentMessageTransformer:
             elif message.type == ToolInvokeMessage.MessageType.LOG:
                 assert isinstance(message.message, ToolInvokeMessage.LogMessage)
                 if message.message.metadata:
-                    icon = tool_info.get("icon", "")
                     dict_metadata = dict(message.message.metadata)
-                    if dict_metadata.get("provider"):
-                        manager = PluginInstaller()
-                        plugins = manager.list_plugins(tenant_id)
+                    provider = dict_metadata.get("provider")
+                    if provider:
+                        # Fail open: any lookup failure (slow plugin daemon,
+                        # credential decryption error, schema conversion error)
+                        # must not affect Agent execution. A missing icon is
+                        # purely cosmetic, so we always fall through to the
+                        # seed icon from tool_info.
                         try:
-                            current_plugin = next(
-                                plugin
-                                for plugin in plugins
-                                if f"{plugin.plugin_id}/{plugin.name}" == dict_metadata["provider"]
+                            icon, icon_dark = self._resolve_provider_icons(
+                                provider=provider,
+                                user_id=user_id,
+                                tenant_id=tenant_id,
+                                seed_icon=tool_info.get("icon", ""),
+                                cache=provider_icon_cache,
                             )
-                            icon = current_plugin.declaration.icon
-                        except StopIteration:
-                            pass
-                        icon_dark = None
-                        try:
-                            builtin_tool = next(
-                                provider
-                                for provider in BuiltinToolManageService.list_builtin_tools(
-                                    user_id,
-                                    tenant_id,
-                                )
-                                if provider.name == dict_metadata["provider"]
-                            )
-                            icon = builtin_tool.icon
-                            icon_dark = builtin_tool.icon_dark
-                        except StopIteration:
-                            pass
-
+                        except Exception:
+                            icon, icon_dark = tool_info.get("icon", ""), None
                         dict_metadata["icon"] = icon
                         dict_metadata["icon_dark"] = icon_dark
                         message.message.metadata = dict_metadata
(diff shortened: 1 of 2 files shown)


crazywoola:

### Why this is being closed

### Next steps

PR #41178, opened 2026-08-24
fix(api): persist trigger subscription's real `expires_at` lease

### LINKED ISSUE #41162: [Bug] Gmail trigger silently stops after 7 days: subscription expires_at is persisted as -1
### Self Checks

### Dify version

1.16.1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Gmail trigger subscriptions stop firing after 7 days because the `expires_at` returned by
the plugin is not persisted. Dify stores `-1`, which per the trigger plugin specification
means "never expires", so `_refresh_subscription()` is never called and the Gmail
`users.watch` lease silently lapses.

**1.** Install `langgenius/gmail_trigger` (0.1.1) and create a subscription, then bind it
to a trigger node in a published workflow. Confirm the workflow fires on incoming mail.

**2.** Inspect the stored subscription immediately after creation:

```sql
SELECT name, expires_at,
       to_timestamp(expires_at) AS watch_expires,
       to_timestamp(credential_expires_at) AS cred_expires
FROM trigger_subscriptions;
```

```
       name        | expires_at |     watch_expires      |      cred_expires
-------------------+------------+------------------------+------------------------
 SES Mail Intake   |         -1 | 1969-12-31 23:59:59+00 | 2026-08-24 08:42:00+00
```

`expires_at` is already `-1`, although `_create_subscription()` returned a real timestamp
(see below). `credential_expires_at` is stored correctly.

**3.** Wait 7 days (the maximum lifetime of a Gmail `users.watch` lease).

**4.** Observe that the trigger no longer fires. Everything downstream is still healthy:

```sql
SELECT count(*) FROM workflow_plugin_triggers;   -- 1 (binding intact)
```

- The Pub/Sub push subscription is still `push`, with the correct endpoint, ack deadline 60s.
- Its metrics show no data at all (no messages ever arrive).
- nginx receives zero POSTs to `/triggers/plugin/<endpoint_id>`.
- OAuth credentials are still being refreshed successfully.

**5.** Check the worker log. `trigger_subscription_refresh` runs every minute but only
refreshes the OAuth token. `_refresh_subscription()` (which re-issues `users.watch`) is
never invoked — no such call appears in 24 hours of logs:

```
worker-1 | Begin subscription refresh: tenant=... id=f5f027d0-...
worker-1 | Refreshing OAuth token: subscription_id=f5f027d0-... expires_at=1787560681 now=1787557201
worker-1 | OAuth token refreshed: result={'result': 'success', 'expires_at': 1787560741}
worker-1 | Task tasks.trigger_subscription_refresh_tasks.trigger_subscription_refresh[...] succeeded
```

Note the OAuth token is refreshed every 60 seconds even though it is valid for another
58 minutes, which appears to be a side effect of the same `expires_at = -1` value.

**6.** For reference, the plugin returns a correct value.
`triggers/gmail_trigger/provider/gmail_trigger.py`:

```python
data: dict[str, Any] = resp.json() or {}
expiration_ms: int = int(data.get("expiration") or 0)
# Gmail watch is time-limited; if expiration is not provided, use 6 days as a safe default
expires_at: int = int(expiration_ms / 1000) if expiration_ms else int(time.time()) + 6 * 24 * 60 * 60
...
return Subscription(
    expires_at=expires_at,
    endpoint=endpoint,
    parameters=parameters,
    properties=properties,
)
```

`_refresh_subscription()` is also implemented and re-issues `users.watch`, so the value
appears to be dropped between the plugin response and persistence in Dify.

Recreating the subscription restores service, but `expires_at` is `-1` again, so the
problem recurs after another 7 days.

### ✔️ Expected Behavior

`trigger_subscriptions.expires_at` should hold the timestamp returned by
`_create_subscription()`, so that Dify calls `_refresh_subscription()` before the lease
expires and the provider can renew it. The Gmail trigger should keep firing indefinitely
without manual intervention.

### ❌ Actual Behavior

`expires_at` is persisted as `-1`. According to the trigger plugin documentation, `-1`
means the subscription never expires and Dify will never call `_refresh_subscription()`
again. The Gmail `users.watch` lease therefore expires after 7 days and the trigger stops
firing permanently.

The failure is completely silent: no error, no log entry, no indication in the UI. The
subscription still shows as active in Dify and the Pub/Sub subscription still shows as
active in Google Cloud. The only symptom is that the workflow stops running.

This affects any trigger provider that relies on a lease-based subscription
(Gmail `users.watch`, Microsoft Graph subscriptions, and similar), not only the Gmail trigger.

Harsh23Kashyap:
## Summary

`update_and_build_builder` calls `TriggerManager.subscribe_trigger` to register the plugin's trigger (e.g. Gmail `users.watch`) and receives a `Subscription` entity whose `expires_at` carries the real lease timestamp. The autosubscribe path then called

```python
TriggerProviderService.add_trigger_subscription(
    ...
    expires_at=subscription_builder.expires_at,   # wrong: builder's -1 placeholder
)
```

but the builder is initialised to `expires_at=-1` (`api/services/trigger/trigger_subscription_builder_service.py:105`) and is never updated with the real value. The manual / UNAUTHORIZED path on line 242 legitimately reads the builder value, so that path is left unchanged.

**Result:** every OAuth/API-key trigger subscription is persisted with `expires_at=-1`. Per the trigger plugin spec `-1` means "never expires", so the `trigger_subscription_refresh` worker never calls `_refresh_subscription()` to re-issue `users.watch`. After the provider-side lease lapses (7 days for Gmail `users.watch`) the channel stops firing; the worker log shows the OAuth token refresh succeeding and zero `users.watch` re-issues.

## Repro from the issue reporter (Dify 1.16.1, `langgenius/gmail_trigger` 0.1.1)

```sql
SELECT name, expires_at, credential_expires_at FROM trigger_subscriptions;

       name        | expires_at | credential_expires_at
-------------------+------------+------------------------
 SES Mail Intake   |         -1 | 2026-08-24 08:42:00+00
```

`credential_expires_at` is correct because the OAuth flow fills it in via `SubscriptionBuilderUpdater` before the build step. Only `expires_at` is broken.

## Fix

```python
TriggerProviderService.add_trigger_subscription(
    ...
    expires_at=subscription.expires_at,   # the lease from subscribe_trigger()
)
```

Plus an in-memory builder sync so any diagnostic / logging in the rest of the request (or in the cache, if a later step re-reads it) sees the same value the refresh task will use to schedule renewal:

```python
subscription: Subscription = TriggerManager.subscribe_trigger(...)
# Keep the in-memory builder in sync with the real lease.
subscription_builder.expires_at = subscription.expires_at
```

The `Subscription` entity's `expires_at` is a required `int` (`api/core/trigger/entities/entities.py:162`), so no `None`-guard is needed at the call site.

## Out of scope (documented for the reviewer)

- **Existing rows in `trigger_subscriptions` that already have `expires_at=-1` will not self-heal from this PR.** They need a one-time data fix (resubscribe via the existing endpoint, or a targeted UPDATE) plus a small change to the refresh task to treat `<= 0` as "expired, enqueue renewal" for finite-lease providers. That is a separate concern from this PR, which only fixes the *create* path. A follow-up PR is recommended: detect `expires_at <= 0` in the refresh task and enqueue an immediate renewal; for finite-lease providers (the plugin manifest declares one), the renewal path will then write the correct lease. For genuine "never expires" providers that return `-1` deliberately, the plugin should also signal that in the manifest so the refresh task can skip the renewal.
- **The `Subscription` entity exposes `parameters` and `properties` separately.** `properties` is what the autosubscribe path persists (it is the plugin's own subscription metadata), while `parameters` is the builder-validated user input. This PR does not change that split.

## Measured impact

For a Gmail trigger subscription:

- Pre-fix: `expires_at = -1`, refresh task interprets as "never expires", no `users.watch` re-issue, channel lapses on day 7.
- Post-fix: `expires_at = <real lease epoch>`, refresh task enqueues a renewal shortly before the lease expires, the `users.watch` lease is re-issued, the channel keeps firing.

## Verification

| step | tool | result |
|---|---|---|
| `ruff format --check` on the 2 touched files | ruff | 2 files already formatted |
| `ruff check` on the 2 touched files | ruff | All checks passed |
| `pytest tests/unit_tests/services/test_trigger_subscription_builder_service.py` | pytest 9 | 18 passed (16 existing + 2 new) |
| `pytest tests/unit_tests/services/test_trigger_subscription_builder_service.py tests/unit_tests/services/test_trigger_provider_service.py` | pytest 9 | 41 passed, no regressions |
| `pyrefly check services/trigger/trigger_subscription_builder_service.py` | pyrefly 1.2.0 | 0 diagnostics |

## /consult-kiro (per the standing rule)

Ran `consult-kiro.sh --model openai/gpt-5 --variant low` on both the plan and the final diff. Reviewer's notes (folded in):

- Confirmed the one-line fix is correct and minimal; `Subscription.expires_at` is a required `int`, so a `None` guard is unnecessary.
- Suggested the in-memory builder sync for observability, applied (3 added lines in the autosubscribe path).
- Called out the existing-`-1`-rows risk as a separate concern; documented in "Out of scope" above.
- Suggested a small `None`-handling test, declined: `Subscription.expires_at: int = Field(...)` (required), no test value would reach that branch with a Pydantic validation failure upstream.

## Tests added

- `test_update_and_build_persists_subscription_expires_at_not_builder_placeholder` - regression for #41162. Mocks `TriggerManager.subscribe_trigger` to return a `Subscription(expires_at=1787560681)`, calls `update_and_build_builder` with `CredentialType.OAUTH2`, asserts `add_trigger_subscription` was called with the real lease (not `-1`), and that the other fields still flow through unchanged. Pre-fix: fails (`-1 != 1787560681`). Post-fix: passes.
- `test_update_and_build_uses_builder_expires_at_for_unauthorized_path` - guard against accidentally rewriting the line 242 (UNAUTHORIZED) path. The manual create legitimately reads the builder's value because no `Subscription` is produced.

Fixes #41162

### DIFF
--- a/api/services/trigger/trigger_subscription_builder_service.py
+++ b/api/services/trigger/trigger_subscription_builder_service.py
@@ -254,6 +254,11 @@ class TriggerSubscriptionBuilderService:
                     credentials=subscription_builder.credentials,
                     credential_type=credential_type,
                 )
+                # Keep the in-memory builder in sync with the real lease so
+                # any diagnostic / logging in the rest of this request (or
+                # in the cache, if a later step re-reads it) sees the same
+                # value the refresh task will use to schedule renewal.
+                subscription_builder.expires_at = subscription.expires_at
 
                 TriggerProviderService.add_trigger_subscription(
                     subscription_id=subscription_builder.id,
@@ -267,7 +272,7 @@ class TriggerSubscriptionBuilderService:
                     credentials=subscription_builder.credentials,
                     credential_type=credential_type,
                     credential_expires_at=subscription_builder.credential_expires_at or -1,
-                    expires_at=subscription_builder.expires_at,
+                    expires_at=subscription.expires_at,
                 )
 
             # Delete the builder after successful subscription creation
--- a/api/tests/unit_tests/services/test_trigger_subscription_builder_service.py
+++ b/api/tests/unit_tests/services/test_trigger_subscription_builder_service.py
@@ -5,7 +5,7 @@ from unittest.mock import Mock, patch
 import pytest
 
 from core.plugin.entities.plugin_daemon import CredentialType
-from core.trigger.entities.entities import SubscriptionBuilder, SubscriptionBuilderUpdater
+from core.trigger.entities.entities import Subscription, SubscriptionBuilder, SubscriptionBuilderUpdater
 from core.trigger.trigger_manager import TriggerManager
 from models.provider_ids import TriggerProviderID
 from services.trigger.trigger_subscription_builder_service import TriggerSubscriptionBuilderService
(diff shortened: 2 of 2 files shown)


crazywoola:

### Why this is being closed

### Next steps

PR #41203, opened 2026-08-24
fix(api): reserve the app when starting an annotation reply job

### LINKED ISSUE #40595: Annotation reply jobs can overlap per app and remain stuck in waiting
### Self Checks

### Dify version

main (7aba539e82, 2026-08-12)

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

1. Run the API and Celery worker from current `main`.
2. For the same app, call `AppAnnotationService.enable_app_annotation(...)` twice before the first task finishes.
3. Observe that both requests receive different job IDs and both enqueue `enable_annotation_reply_task`.
4. While an enable task is in progress, call `disable_app_annotation(...)` for the same app. Enable and disable use separate app-level keys, so the two operations are not mutually exclusive.
5. Let a queued task run after the app was deleted, or disable annotation reply when its setting is already absent. These early-return paths do not write a terminal job status.

The service checks `enable_app_annotation_{app_id}` and `disable_app_annotation_{app_id}`, but it never writes either key when enqueueing a task. It only creates the per-job status key with `setnx`, without an expiration. The tasks delete the app-level keys even though those keys were never acquired.

Related: #37646 reported the missing app-level key write and was closed as stale after its proposed fixes were not merged. The same code path is still present on current `main`; the shared enable/disable exclusion and terminal-status paths are also affected.

### ✔️ Expected Behavior

Annotation reply mutations should have one application-scoped job lifecycle:

- only one enable or disable job may mutate an app at a time;
- repeated requests for the same action should reuse the in-flight job;
- a conflicting action should not start concurrently;
- enqueue failures and all worker exit paths should release the application reservation;
- every accepted job should transition from waiting to processing and then to completed or error.

### ❌ Actual Behavior

Multiple jobs can be queued for one app and can concurrently rebuild/delete the same annotation vector data while updating or deleting the same annotation setting. Some early-return paths leave the job status at waiting indefinitely, so the caller can continue polling a job that will never finish.

alanhuangyoo:
Closes #40595

## Problem

`AppAnnotationService.enable_app_annotation` reads the app-level key that is supposed to serialize annotation reply jobs, but nothing ever writes it:

```python
enable_app_annotation_key = f"enable_app_annotation_{app_id}"
cache_result = redis_client.get(enable_app_annotation_key)      # read
if cache_result is not None:
    return {"job_id": cache_result, "job_status": "processing"}

job_id = str(uuid.uuid4())
redis_client.setnx(f"enable_app_annotation_job_{job_id}", "waiting")   # only the per-job key
enable_annotation_reply_task.delay(...)                                # app key never written
```

The guard can therefore never fire. Every call mints a new job id and enqueues another task, and those tasks concurrently rebuild or delete the same annotation vector data and mutate the same `AppAnnotationSetting`. The worker's `finally` block then deletes a key that was never acquired. `disable_app_annotation` has the same shape.

Two smaller defects come from the same code:

- The per-job status key is written with `setnx` and no expiry, so it survives forever whenever the task does not reach a terminal path.
- The workers return early when the app or the annotation setting is missing, before any status is written, so a caller polling that job id sees `waiting` indefinitely.

## Change

- Reserve the app with `SET key job_id NX EX` before enqueueing. A second request for the same action gets the in-flight job id back with `processing` instead of starting a competing task.
- Give the per-job status key the same TTL the workers already use for terminal statuses (600s), so nothing outlives the job it describes.
- Release the reservation if `.delay()` raises, so a broker failure does not lock the app for the full TTL.
- Record a terminal `error` status (plus the existing `*_error_*` key) on the worker early-return paths and drop the reservation there.

`_running_job_id()` decodes the reserved value because the client is configured with `decode_responses=False`; the previous code would have returned raw bytes as the job id, which was masked by the guard never firing.

Enable and disable still use separate reservations. The issue also proposes making them mutually exclusive with each other, which changes behaviour beyond the broken guard, so I left it out of this PR — happy to follow up if you want it.

## Tests

`TestAppAnnotationServiceEnableDisable` previously mocked the whole redis client and asserted the exact `setnx` call, so it described the implementation rather than the behaviour and could not observe the missing reservation. It is rewritten against a small fake that implements `SET NX EX` semantics, and covers: the reservation is written with a TTL, a second request reuses the in-flight job without enqueueing, and a failed enqueue releases the reservation.

Against the current `main` sources, all five fail:

```
FAILED ...::test_enable_enqueues_and_reserves_the_app
FAILED ...::test_enable_reuses_the_in_flight_job
FAILED ...::test_enable_releases_the_reservation_when_enqueue_fails
FAILED ...::test_disable_enqueues_and_reserves_the_app
FAILED ...::test_disable_reuses_the_in_flight_job
```

with, for the reuse cases:

```
E  AssertionError: {'job_status': 'waiting'} != {'job_status': 'processing'}
E                  {'job_id': 'eb28b41d-...'} != {'job_id': '1e812dc9-...'}
```

With this change:

```
api/tests/unit_tests/services/test_annotation_service.py   ->  47 passed
api/tests/unit_tests/controllers -k annotation             ->  74 passed
```

`ruff check` and `ruff format --check` are clean on the four touched files.

### DIFF
--- a/api/services/annotation_service.py
+++ b/api/services/annotation_service.py
@@ -89,6 +89,20 @@ class UpdateAnnotationSettingArgs(TypedDict):
     score_threshold: float
 
 
+# Annotation reply jobs reserve their app for this long. It matches the TTL the
+# workers use for terminal job statuses, so a reservation cannot outlive the job
+# it guards even if a worker dies before releasing it.
+ANNOTATION_JOB_TTL = 600
+
+
+def _running_job_id(reservation_key: str) -> str:
+    """Job id currently holding *reservation_key*, or "" if it just expired."""
+    reserved = redis_client.get(reservation_key)
+    if reserved is None:
+        return ""
+    return reserved.decode() if isinstance(reserved, bytes) else str(reserved)
+
+
 class AppAnnotationService:
     @staticmethod
     def _get_annotation_by_ref(annotation_ref: AnnotationRef, session: Session) -> MessageAnnotation | None:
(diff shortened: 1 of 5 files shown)


crazywoola:
Thanks for the contribution. Closing this as a duplicate pull request for https://github.com/langgenius/dify/issues/40595. We are keeping langgenius/dify#40591 under the author-first, then FIFO policy for duplicate PRs in the same repository.

PR #41207, opened 2026-08-24
fix(agent): seal the run event stream at the first terminal event

### LINKED ISSUE #40765: [Bug] Dify Agent can append non-terminal events after a terminal transition

alanhuangyoo:
Closes #40765

## Problem

`RedisRunStore` uses a Lua compare-and-set to commit a terminal event and its run status together, but `append_event` wrote through a plain pipeline that never looked at the durable run status:

```python
async with self.redis.pipeline(transaction=True) as pipeline:
    _ = pipeline.xadd(events_key, {"payload": payload})
    _ = pipeline.expire(events_key, self.run_retention_seconds)
    _ = pipeline.expire(run_record_key(self.prefix, event.run_id), self.run_retention_seconds)
```

There is no read of the run record anywhere on that path, so the check the caller performs (`is_cancelled()`) and the write are in different processes' hands:

1. Process A passes its local check and prepares a stream event.
2. Process B atomically commits `run_cancelled` and flips the run record.
3. Process A's `XADD` lands after the terminal cursor.

The stream then reads `run_started -> run_cancelled -> pydantic_ai_event`, which breaks the contract that a consumer may stop at the first terminal event: an SSE reader and a later polling reader see different histories. The losing write also refreshes the stream and run-record TTL for a run that is already finished.

## Change

`append_event` now goes through `_APPEND_EVENT_SCRIPT`, so the status check and the `XADD` happen in the same script the same way the terminal writers already do it. A sealed run returns an empty event id and nothing is written — no entry, no cursor, no TTL refresh.

The terminal status list lives in one place (`_TERMINAL_RUN_STATUSES`) and the Lua condition is built from it, so the script and the Python side cannot drift.

A missing run record is treated as sealed rather than raising. `append_event` did not raise before, and a run whose record has expired is long finished, so this keeps the change to the behaviour the issue describes.

## Tests

`FakeRedis.eval` previously required every test to hand it a canned result, which meant nothing could observe what the script actually did. It now interprets `_APPEND_EVENT_SCRIPT` against its own stored record, so the seal is visible in tests and the existing `iter_events` tests can keep using `append_event` to build their fixtures.

- `test_append_event_serializes_typed_event_without_id_and_expires_run_keys` — rewritten against observable state: the entry lands and both keys get their TTL refreshed.
- `test_append_event_is_a_no_op_once_the_run_is_sealed` — parametrized over `succeeded`, `failed` and `cancelled`: no stream entry and no `EXPIRE` after the seal.

```
tests/local/dify_agent/storage/test_redis_run_store.py   ->  19 passed
tests/ (excluding pre-existing collection errors)        ->  748 passed, 32 skipped
```

The five collection errors under `layers/config`, `layers/dify_core_tools` and `protocol/test_working_environment` are present on an unmodified checkout too — missing optional dependencies in my environment, unrelated to this change.

`ruff check` and `ruff format --check` are clean on both files.

## Not included

The issue notes the cancellation observer narrows but cannot close the race; that observer is unchanged here since the seal now holds regardless of how wide the window is.

### DIFF
--- a/dify-agent/src/dify_agent/storage/redis_run_store.py
+++ b/dify-agent/src/dify_agent/storage/redis_run_store.py
@@ -82,6 +82,28 @@ return {1, ARGV[1], event_id}
 """
 
 
+_TERMINAL_RUN_STATUSES = ("succeeded", "failed", "cancelled")
+_LUA_TERMINAL_STATUS_TEST = " or ".join(f'record.status == "{status}"' for status in _TERMINAL_RUN_STATUSES)
+
+_APPEND_EVENT_SCRIPT = f"""
+local record_json = redis.call("GET", KEYS[1])
+if not record_json then
+    return {{0, ""}}
+end
+
+local record = cjson.decode(record_json)
+if {_LUA_TERMINAL_STATUS_TEST} then
+    return {{0, ""}}
+end
+
+local ttl = tonumber(ARGV[2])
+local event_id = redis.call("XADD", KEYS[2], "*", "payload", ARGV[1])
+redis.call("EXPIRE", KEYS[2], ttl)
+redis.call("EXPIRE", KEYS[1], ttl)
+return {{1, event_id}}
+"""
+
+
 _REQUEST_CANCELLATION_SCRIPT = """
 local record_json = redis.call("GET", KEYS[1])
 if not record_json then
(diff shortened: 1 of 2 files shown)


crazywoola:
Thanks for the contribution. Closing this as a duplicate pull request for https://github.com/langgenius/dify/issues/40765. We are keeping langgenius/dify#41084 under the author-first, then FIFO policy for duplicate PRs in the same repository.

PR #41225, opened 2026-08-25
fix(cli): declare write effect for import and context-switch commands

samzong:
## Summary

`difyctl` publishes an `effect` field (`read` / `write` / `destructive`) on every command through `difyctl help -o json`, and the agent skill installed by `difyctl skills install` tells agents to treat that field as the confirmation gate before running a command.

Three commands declared no `effect`, so `describeCommand` fell back to the `read` default and agents following the published contract ran them without asking:

- `import studio-app` creates an app, and with `--app-id` overwrites an existing workflow or advanced-chat app
- `use account` and `use host` rewrite the local active-context registry, which the sibling `use workspace` and `config set` already report as `write`

This declares `effect: write` on all three and adds a `[P0]` e2e assertion next to the existing `effect=destructive` guard so the metadata cannot silently regress.

Fixes #41223

## Screenshots

| Before | After |
| ------ | ----- |
| `difyctl help import studio-app -o json \| jq .effect` → `"read"` | `"write"` |
| `difyctl help use account -o json \| jq .effect` → `"read"` | `"write"` |
| `difyctl help use host -o json \| jq .effect` → `"read"` | `"write"` |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [x] I ran `make lint && make type-check` (backend) and `vp staged` (frontend) to appease the lint gods

From Claude Code

Issue #41223, opened 2026-08-25
difyctl reports effect=read for commands that write remote or local state


samzong:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

main (`cli/difyctl`), commit 679514d4

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

`difyctl` publishes an `effect` field (`read` / `write` / `destructive`) on every command through `difyctl help -o json`. The agent skill installed by `difyctl skills install` instructs agents to treat that field as the confirmation gate before running a command.

Three commands do not declare `effect`, so `describeCommand` falls back to the `read` default:

```
difyctl help import studio-app -o json | jq .effect   # "read"
difyctl help use account -o json       | jq .effect   # "read"
difyctl help use host -o json          | jq .effect   # "read"
```

`import studio-app` creates an app, and with `--app-id` it overwrites an existing workflow or advanced-chat app. `use account` and `use host` rewrite the local active-context registry; the sibling `use workspace` and `config set` already declare `effect: write`.

### ✔️ Expected Behavior

Every command that mutates remote or local state reports `effect: write` (or `destructive`), so an agent following the published contract asks for confirmation before running it.

### ❌ Actual Behavior

All three report `effect: read`, so an agent following the skill runs them without confirmation. In the worst case `difyctl import studio-app --app-id <id>` overwrites an existing app with no confirmation step.

PR #41226, opened 2026-08-25
feat(cli): add a compact command map for agent discovery

### LINKED ISSUE #41035: Add intent-based command discovery to difyctl

samzong:
## Summary

Agents discovering a single command had to load the whole `difyctl help -o json` tree: args, flags, examples and the global contract for every command, about 45 KB, to pick one path.

This adds `--compact` on top-level help. It emits the complete command map with only `command`, `description` and `effect` per entry — about 4 KB at the current 28-command surface:

```console
$ difyctl help -o json | wc -c
45579
$ difyctl help -o json --compact | wc -c
3923
```

The agent skill installed by `difyctl skills install` now reads that map, selects one command, and inspects it with `difyctl help <path> -o json`, which stays the only surface carrying args and flags. Two calls per task instead of one oversized one, and no path where an agent names a flag it never fetched.

Details:

- `--compact` is rejected with exit `2` and the standard error envelope when combined with per-command help or a non-structured output format
- it is listed under `GLOBAL FLAGS`, so it is discoverable from `difyctl --help`
- `summarizeCommand` now owns the `description` / `effect` defaults shared by the full descriptor and the compact map, so the two cannot drift
- `help agent` and the README describe the same two-step flow

The skill tells agents to treat a command's `effect` as the confirmation gate, and three commands still fall back to the `read` default while declaring no effect of their own. That is pre-existing on `main` — the skill on `main` names `effect` as the gate too — and is fixed separately in #41225 (issue #41223). The two branches do not conflict and can merge in either order.

An intent-ranked `difyctl search` was prototyped and measured against this approach before settling on the compact map; the numbers are in #41035.

Fixes #41035

## Screenshots

| Before | After |
| ------ | ----- |
| `difyctl help -o json` — 45,579 B, the only machine-readable map | `difyctl help -o json --compact` — 3,923 B, `command` / `description` / `effect` only |
| SKILL.md pointed agents at the full tree | SKILL.md reads the compact map, then per-command help |

## Checklist

From Claude Code

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -48,11 +48,13 @@ Background docs: `difyctl help account`, `difyctl help external`, `difyctl help 
 Run `difyctl --help` for the full list of commands.
 Run `difyctl <cmd> --help` for per-command reference.
 
-For agents (and scripting), start with `difyctl help agent` — the cross-command operating guide (output, discovery, auth, exit codes, errors, HITL, retry). Every help surface is also machine-readable: `difyctl help -o json` dumps the whole command tree plus the global contract (exit codes, output formats, error envelope, HITL protocol), and `difyctl <cmd> --help -o json` returns one command's descriptor.
+Use `difyctl help -o json --compact` to read the complete command map, then inspect the selected command with `difyctl help <path> -o json` before running it — that per-command descriptor is the only surface that carries args and flags.
+
+For agents (and scripting), start with `difyctl help agent` — the cross-command operating guide (output, discovery, auth, exit codes, errors, HITL, retry). Every help surface is also machine-readable: `difyctl help -o json` dumps the whole command tree plus the global contract (exit codes, output formats, error envelope, HITL protocol), and `difyctl help <path> -o json` returns one command's descriptor.
 
 ## Agent skill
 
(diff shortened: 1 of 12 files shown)


GareArc (approved):
LGTM

PR #41251, opened 2026-08-25
fix: separate WebApp analytics from console page tracking

iamjoel:
## Summary

- Route analytics by Next.js layout group so `/agents` captures UTM registration and app-creation attribution without conflicting with published WebApp routes.
- Add a consent-gated WebApp Amplitude instance that disables automatic tracking, page views, remote configuration, and session replay while preserving explicit `webapp_run` events.
- Keep Google Analytics exclusive to console routes and add regression coverage for analytics boundaries, consent changes, and WebApp event entry points.

Fixes DIFY-2948

## Checklist


web/app/components/base/amplitude/web-app-client.ts
```diff
@@ -0,0 +1,56 @@
...
+async function initializeWebAppAmplitude(): Promise<WebAppAmplitudeClient | null> {
+  try {
+    const { createInstance } = await import('@amplitude/analytics-browser')
+    const client = createInstance()
+
+    client.init(AMPLITUDE_API_KEY, {
+      autocapture: false,
+      defaultTracking: false,
+      fetchRemoteConfig: false,
+      instanceName: 'webapp',
```
CodingOnStar, line 49: **[bug]** The dedicated WebApp client is not identity-isolated from the console Amplitude instance. Both SDKs use the same `AMPLITUDE_API_KEY`. Amplitude Browser 2.x names the identity cookie `AMP_{apiKey[:10]}` and the unsent-event key `AMP_unsent_{apiKey[:10]}` with no `instanceName` suffix (`getCookieName` / Destination `storageKey`). Console `setUserId(userProfile.email)` therefore persists the builder’s email (and device/session ids) into that cookie; WebApp `client.init()` reads `previousCookies.userId` and attaches it to consented `webapp_run` events. The same shared unsent queue can also flush leftover console events from the WebApp instance. `instanceName: 'webapp'` only namespaces the analytics connector (session-replay / Experiment), which is why autocapture/replay can look isolated while user identity still leaks.

**Suggestion:** Initialize the WebApp instance with storage that cannot read or write the console cookie/queue, for example `identityStorage: 'none'` plus a WebApp-prefixed `storageProvider` (or in-memory storage if events are flushed immediately). Pass `optOut: shouldWebAppAmplitudeOptOut` into `init`, and do not call `setUserId(undefined)` on a shared cookie (that would wipe console identity). Extend `web-app-client.spec.ts` so init options include the isolated storage settings and a case where an existing `AMP_*` cookie user id is not applied to `track()`.

CodingOnStar (approved):

PR #41261, opened 2026-08-25
fix(dataset): delete child chunks the index processor leaves behind

shaba:
Extracted from #38519 so it can be reviewed independently. Refs #38518.

### The leak

Nothing in `clean_dataset_task` deletes rows from `child_chunks`. The only dataset-wide delete lives in `ParentChildIndexProcessor.clean()`, and the entire block that performs it sits inside:

```python
if dataset.indexing_technique == IndexTechniqueType.HIGH_QUALITY:
    ...
    if delete_child_chunks:
        # Use existing compound index: (tenant_id, dataset_id, ...)
        session.execute(
            delete(ChildChunk).where(
                ChildChunk.tenant_id == dataset.tenant_id, ChildChunk.dataset_id == dataset.id
            )
        )
```

A dataset whose `indexing_technique` is anything other than high quality therefore keeps every child chunk after it is deleted. `ParagraphIndexProcessor.clean()` and `QAIndexProcessor.clean()` never touch the table at all.

The same rows are stranded on a second path, one the task tolerates on purpose. `clean_dataset_task` wraps the index cleanup in its own `try`/`except` and continues deliberately:

```python
except Exception:
    logger.exception(...)
```

Documents, segments, attachments and metadata are then deleted, and the child chunks that referenced them are not. So whenever the vector store is unreachable at delete time — the case that `except` exists for — the child chunks are orphaned even for a high-quality parent-child dataset.

Both leaks are silent. The dataset disappears from the UI, the rows stay in the table indefinitely, and because the dataset row is already gone there is nothing left to retry the cleanup.

### The change

One `delete` in the task, after the index processor has had its turn:

```python
session.execute(
    delete(ChildChunk).where(ChildChunk.tenant_id == tenant_id, ChildChunk.dataset_id == dataset_id)
)
```

The predicate matches both columns so it can use the existing compound index `child_chunk_dataset_id_idx` on `(tenant_id, dataset_id, document_id, segment_id, index_node_id)` — the same access path the parent-child processor already relies on for its own delete.

Placing it after the index cleanup makes it a no-op for the high-quality parent-child datasets that are already handled: those rows are gone by the time it runs.

### Tests

Two cases in `TestChildChunkCleanup`, using the suite's existing SQLite session factory and fixtures:

- a child chunk that the index processor did not delete (the mocked processor stands in for both leak paths) is gone after the task runs — this fails on `main`
- another dataset's child chunks survive, covering the scoping of the predicate

`tests/unit_tests/tasks/test_clean_dataset_task.py` passes: 10 tests. `ruff check` and `ruff format --check` are clean.

### DIFF
--- a/api/tasks/clean_dataset_task.py
+++ b/api/tasks/clean_dataset_task.py
@@ -12,6 +12,7 @@ from extensions.ext_storage import storage
 from models import WorkflowType
 from models.dataset import (
     AppDatasetJoin,
+    ChildChunk,
     Dataset,
     DatasetMetadata,
     DatasetMetadataBinding,
@@ -154,6 +155,14 @@ def clean_dataset_task(
                     SegmentAttachmentBinding.id.in_(binding_ids)
                 )
                 session.execute(binding_delete_stmt)
+
+            # The index processor only deletes child chunks for high-quality parent-child
+            # datasets, so economy-indexed datasets -- and datasets whose index cleanup
+            # above failed -- keep theirs forever. Both columns are matched so the delete
+            # can use the compound (tenant_id, dataset_id, ...) index.
+            session.execute(
+                delete(ChildChunk).where(ChildChunk.tenant_id == tenant_id, ChildChunk.dataset_id == dataset_id)
+            )
 
             session.execute(delete(DatasetProcessRule).where(DatasetProcessRule.dataset_id == dataset_id))
             session.execute(delete(DatasetQuery).where(DatasetQuery.dataset_id == dataset_id))
--- a/api/tests/unit_tests/tasks/test_clean_dataset_task.py
+++ b/api/tests/unit_tests/tasks/test_clean_dataset_task.py
@@ -26,6 +26,7 @@ from extensions.storage.storage_type import StorageType
 from models.base import TypeBase
 from models.dataset import (
     AppDatasetJoin,
+    ChildChunk,
     DatasetMetadata,
     DatasetMetadataBinding,
     DatasetProcessRule,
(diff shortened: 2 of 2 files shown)


shaba:
The Web Full-Stack E2E failure here is unrelated to this change.

This PR touches two files, both Python: `api/tasks/clean_dataset_task.py` and its unit test. The failing scenario is a Playwright test of the Agent v2 Access Point UI — `features/agent-v2/access-point.feature:69`, "Web app access can be disabled and restored from Access Point" — which times out waiting for an `In service` badge to render. 56 of the 57 scenarios passed, and nothing in that path reads dataset cleanup code.

For comparison, #38519 changes the same task far more extensively and its run of the same E2E job passed a few minutes later, so the suite was exercising that scenario successfully at the time.

API Unit Tests and API Integration Tests both pass here. I do not have permission to re-run the job; a maintainer re-run should clear it.

shaba:
Correction to my previous comment: I cited #38519 as having passed this same E2E job. That is no longer true — its next run failed the same scenario, so the comparison I drew there does not hold, and I withdraw it.

The conclusion is unchanged, but here is the actual evidence. `features/agent-v2/access-point.feature:69` is failing intermittently across unrelated PRs right now. For example run 32860893991 on branch `agent/sqlite-orm-app-pipelines` — an ORM migration touching nothing in this area — fails on exactly the same step, same locator, same result line:

```
1) Web app access can be disabled and restored from Access Point # features/agent-v2/access-point.feature:69
     Then Agent v2 Web app access should be in service
         Locator: getByRole("region", { name: "Access Point" }).getByRole("article", { name: "Web app" }).first().getByText("In service")
         Timeout: 30000ms
57 scenarios (56 passed, 1 failed)
```

The same scenario also passed and then failed on an unchanged branch here, which is the signature of a flaky test rather than a regression.

This PR adds one `delete` statement to `api/tasks/clean_dataset_task.py` plus unit tests; no frontend code and no code path reachable from the Access Point UI. API Unit Tests and API Integration Tests pass.

shaba:
Filed the E2E flake as #41263, with the sampling across recent CI runs and an analysis of where the race appears to be.

PR #41293, opened 2026-08-26
refactor: remove unused workflow run export endpoint

hjlarry:
## Summary

- remove the unused per-run workflow archive export endpoint
- remove its endpoint-specific response model and storage/database dependencies
- regenerate Console OpenAPI and TypeScript/Zod/ORPC contracts

The current Console archive download UI uses the workspace-level workflow-run-archives download flow. The removed endpoint has no caller in the current web or e2e code.

## Breaking change

This removes an undocumented Console endpoint: GET /apps/{app_id}/workflow-runs/{run_id}/export. Repository code has no active consumer

PR #41298, opened 2026-08-26
refactor(api): standardize console workflow run endpoints

hjlarry:
## Summary

- migrate Console workflow-run routes to console_account_admission and the shared application-services composition root
- require an explicit database session factory for WorkflowRunService while retaining explicit Engine injection compatibility
- move pause-detail repository and form-token orchestration out of the controller
- use RequestContext workspace identity for tenant-scoped workflow-run and node-execution queries
- standardize successful responses with dump_response without changing the HTTP response contract

## Stack

Depends on #41293, which removes the unused per-run export endpoint. The branch is based directly on that PR commit; because its head lives in a fork, GitHub cannot select it as an upstream base branch. This PR therefore targets main and will be refreshed after #41293 lands.

## Testing

- 103 focused controller, service, composition, and Swagger tests passed
- make type-check-core
- import-linter: 20 contracts kept
- response contract lint
- no-new-getattr check
- API contract regeneration produced no diff; contracts type-check passed
- 16 container integration tests collect successfully; execution is left to CI because the local Docker socket is unavailable

### DIFF
--- a/api/controllers/console/app/workflow_run.py
+++ b/api/controllers/console/app/workflow_run.py
@@ -3,26 +3,19 @@ from uuid import UUID
 
 from flask_restx import Resource
 from pydantic import BaseModel, Field, field_validator
-from sqlalchemy.orm import sessionmaker
 
 from configs import dify_config
 from controllers.common.errors import NotFoundError
 from controllers.common.schema import query_params_from_model, register_response_schema_models, register_schema_models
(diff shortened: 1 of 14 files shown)


hjlarry:
Closing this PR so the refactor can be reviewed locally first. A correctly based stacked PR will be opened after review.

PR #41310, opened 2026-08-26
fix: improve home page accessibility

iamjoel:
## Summary

- Add accessible pause, previous, and next controls for the home banner and stop automatic rotation during user interaction or reduced-motion preferences.
- Align accessible names for workspace, account, search, app navigation, app actions, credits, and Continue Work cards with their visible labels.
- Add focused accessibility regression coverage for the affected home navigation and card interactions.

Fixes SNP-627

## Checklist

### DIFF
--- a/web/app/__tests__/home-banner.browser.spec.tsx
+++ b/web/app/__tests__/home-banner.browser.spec.tsx
@@ -58,7 +58,7 @@ async function renderBanner() {
   return render(
     <>
       <button type="button">Before carousel</button>
-      <div data-testid="banner-host" style={{ width: 300 }}>
+      <div style={{ width: 300 }}>
         <Banner banners={banners} />
       </div>
       <button type="button">After carousel</button>
@@ -78,23 +78,35 @@ describe('Home banner browser interactions', () => {
   })
 
   it('puts the rotation control first in the native tab order and keeps focus-triggered pauses', async () => {
-    // Chromium owns native tab order and focus transitions; happy-dom cannot prove either contract.
+    // Chromium owns native tab order, layout geometry, and focus transitions; happy-dom cannot prove these contracts.
     const screen = await renderBanner()
-    const pauseControl = screen.getByRole('button', { name: 'common.operation.pause' })
-    await expect.element(pauseControl).toBeVisible()
-    const pauseControlElement = pauseControl.element()
+    const stopRotationControl = screen.getByRole('button', {
+      name: 'explore.banner.stopRotation',
+    })
+    const firstPicker = screen.getByRole('button', { name: '01 First banner' })
+    await expect.element(stopRotationControl).toBeVisible()
+    const rotationControlElement = stopRotationControl.element()
+    expect(rotationControlElement.getBoundingClientRect().left).toBeLessThan(
+      firstPicker.element().getBoundingClientRect().left,
+    )
 
     await screen.getByRole('button', { name: 'Before carousel' }).click()
     await userEvent.tab()
 
-    expect(pauseControlElement).toHaveFocus()
+    expect(rotationControlElement).toHaveFocus()
     await expect
-      .element(screen.getByRole('button', { name: 'common.operation.play' }))
+      .element(screen.getByRole('button', { name: 'explore.banner.startRotation' }))
       .toBeVisible()
 
-    await screen.getByRole('button', { name: 'After carousel' }).click()
+    await userEvent.keyboard('{Space}')
     await expect
-      .element(screen.getByRole('button', { name: 'common.operation.play' }))
+      .element(screen.getByRole('button', { name: 'explore.banner.stopRotation' }))
+      .toBeVisible()
+
+    await userEvent.tab()
+    expect(firstPicker.element()).toHaveFocus()
+    await expect
+      .element(screen.getByRole('button', { name: 'explore.banner.stopRotation' }))
       .toBeVisible()
   })
 
@@ -102,7 +114,7 @@ describe('Home banner browser interactions', () => {
     // Chromium hit testing is required to prove that moving onto the overlaid control stays in one hover boundary.
     const screen = await renderBanner()
     const carousel = screen.getByRole('region', { name: 'explore.banner.carouselLabel' })
-    const pauseControl = screen.getByRole('button', { name: 'common.operation.pause' })
+    const pauseControl = screen.getByRole('button', { name: 'explore.banner.stopRotation' })
     const slidesContainer = getSlidesContainer(carousel.element())
     await expect.element(pauseControl).toBeVisible()
 
@@ -117,50 +129,42 @@ describe('Home banner browser interactions', () => {
     await expect.poll(() => slidesContainer.getAttribute('aria-live')).toBe('off')
   })
 
-  it('resumes after a real pointer drag without desynchronizing the pause action', async () => {
-    // Embla's pointer lifecycle depends on browser mouse events and cannot be reproduced faithfully in happy-dom.
+  it('keeps an explicit pause after a real pointer drag released inside the carousel', async () => {
+    // Embla's pointer lifecycle and owner-window timer depend on browser behavior that happy-dom cannot reproduce faithfully.
     const screen = await renderBanner()
     const carousel = screen.getByRole('region', { name: 'explore.banner.carouselLabel' })
-    const pauseControl = screen.getByRole('button', { name: 'common.operation.pause' })
+    const stopRotationControl = screen.getByRole('button', {
+      name: 'explore.banner.stopRotation',
+    })
     const slidesContainer = getSlidesContainer(carousel.element())
-    await expect.element(pauseControl).toBeVisible()
-
-    await userEvent.dragAndDrop(
-      slidesContainer,
-      screen.getByRole('button', { name: 'After carousel' }),
-    )
-
-    await expect.poll(() => slidesContainer.getAttribute('aria-live')).toBe('off')
-    await expect.element(pauseControl).toBeVisible()
-  })
-
-  it('preserves an explicit pause after ResizeObserver reinitializes Embla', async () => {
-    // A real ResizeObserver-driven Embla reInit is a browser lifecycle that happy-dom does not implement.
-    const screen = await renderBanner()
-    const carousel = screen.getByRole('region', { name: 'explore.banner.carouselLabel' })
-    const slidesContainer = getSlidesContainer(carousel.element())
-    const host = screen.getByTestId('banner-host').element()
-    await screen.getByRole('button', { name: 'common.operation.pause' }).click()
+    const slidesRect = slidesContainer.getBoundingClientRect()
+    const firstSlide = screen.getByRole('group', { name: 'First banner' })
+    const firstSlideElement = firstSlide.element()
+    await stopRotationControl.click()
     await userEvent.unhover(carousel)
     await expect
-      .element(screen.getByRole('button', { name: 'common.operation.play' }))
+      .element(screen.getByRole('button', { name: 'explore.banner.startRotation' }))
       .toBeVisible()
 
-    const resized = new Promise<void>((resolve) => {
-      const observer = new ResizeObserver(([entry]) => {
-        if (!entry || entry.contentRect.width >= 260) return
-        observer.disconnect()
-        resolve()
+    vi.useFakeTimers()
+    try {
+      await userEvent.dragAndDrop(slidesContainer, slidesContainer, {
+        sourcePosition: { x: slidesRect.width / 2 - 4, y: slidesRect.height / 2 },
+        targetPosition: { x: slidesRect.width / 2 + 4, y: slidesRect.height / 2 },
+        steps: 2,
       })
-      observer.observe(carousel.element())
-    })
-    host.style.width = '240px'
-    await resized
+      await vi.advanceTimersByTimeAsync(1000)
+      const firstSlideStateAfterDrag = firstSlideElement.getAttribute('aria-hidden')
+      await vi.advanceTimersByTimeAsync(5001)
+      expect(firstSlideElement.getAttribute('aria-hidden')).toBe(firstSlideStateAfterDrag)
+    } finally {
+      vi.useRealTimers()
+    }
 
-    await expect
-      .element(screen.getByRole('button', { name: 'common.operation.play' }))
-      .toBeVisible()
     expect(slidesContainer).toHaveAttribute('aria-live', 'polite')
+    await expect
+      .element(screen.getByRole('button', { name: 'explore.banner.startRotation' }))
+      .toBeVisible()
   })
 
   it('keeps autoplay initialized but stopped for reduced motion so Play remains safe', async () => {
@@ -185,12 +189,12 @@ describe('Home banner browser interactions', () => {
     const screen = await renderBanner()
     const carousel = screen.getByRole('region', { name: 'explore.banner.carouselLabel' })
     const slidesContainer = getSlidesContainer(carousel.element())
-    const playControl = screen.getByRole('button', { name: 'common.operation.play' })
+    const playControl = screen.getByRole('button', { name: 'explore.banner.startRotation' })
     await expect.element(playControl).toBeVisible()
 
     await playControl.click()
     await expect
-      .element(screen.getByRole('button', { name: 'common.operation.pause' }))
+      .element(screen.getByRole('button', { name: 'explore.banner.stopRotation' }))
       .toBeVisible()
     await userEvent.unhover(carousel)
 
--- a/web/app/__tests__/home-banner.browser.spec.tsx
+++ b/web/app/__tests__/home-banner.browser.spec.tsx
@@ -1,0 +1,199 @@
+import type { BannerResponse } from '@dify/contracts/api/console/explore/types.gen'
+import { userEvent } from 'vite-plus/test/browser'
+import { render } from 'vitest-browser-react'
+import { Banner } from '@/features/home/banner/banner'
+
+const REDUCED_MOTION_QUERY = '(prefers-reduced-motion: reduce)'
+
+vi.mock('@tanstack/react-query', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@tanstack/react-query')>()
+  return {
+    ...actual,
+    useSuspenseQuery: () => ({ data: { id: 'account-123' } }),
+  }
+})
+
+vi.mock('@/features/account-profile/client', () => ({
+  userProfileQueryOptions: () => ({}),
+}))
+
+vi.mock('@/context/i18n', () => ({
+  useLocale: () => 'en-US',
+}))
+
+vi.mock('@/app/components/base/amplitude', () => ({
+  trackEvent: vi.fn(),
+}))
+
+const banners: BannerResponse[] = [
+  {
+    id: 'banner-1',
+    status: 'enabled',
+    link: '',
+    created_at: '2024-01-01T00:00:00Z',
+    sort: 1,
+    content: {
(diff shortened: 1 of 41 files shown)


web/features/home/banner/banner.tsx
```diff
@@ -105,9 +143,13 @@ function BannerCarouselContent({ banners, accountId, language }: BannerCarouselC
 
     const handleAutoplayPlay = () => setIsPlaying(true)
     const handleAutoplayStop = () => setIsPlaying(false)
+    const autoplayIsPlaying = api.plugins().autoplay?.isPlaying() ?? false
 
     // oxlint-disable-next-line eslint-react/set-state-in-effect -- Embla owns this external playback state.
-    setIsPlaying(api.plugins().autoplay?.isPlaying() ?? false)
+    setIsPlaying(autoplayIsPlaying)
+    // oxlint-disable-next-line eslint-react/set-state-in-effect -- Embla owns the initial rotation-control state.
+    setIsRotationEnabled(autoplayIsPlaying)
```
lyzno1: [P2] Preserve the user's explicit pause across Embla reinitialization. Embla destroys and initializes its plugins again during `reInit()` (including resize-driven reinitialization), and Autoplay starts because `playOnInit` defaults to `true`. The new rotation intent remains disabled while `handleAutoplayPlay` only updates `isPlaying`, so Pause → resize leaves the carousel rotating while the control still says Play. I reproduced `stop()` → `reInit()` → `isPlaying() === true` with the installed Embla 8.6.0. Please make the user intent the playback owner and add a real-Embla browser regression test. Upstream lifecycle: https://github.com/davidjerleke/embla-carousel/blob/v8.6.0/packages/embla-carousel-autoplay/src/components/Autoplay.ts#L95

web/features/home/banner/banner.tsx
```diff
@@ -56,26 +59,61 @@ function BannerCarouselContent({ banners, accountId, language }: BannerCarouselC
...
+    }
+
+    isRotationEnabledRef.current = true
+    setIsRotationEnabled(true)
+    if (temporaryPauseReasonsRef.current.size > 0) {
+      shouldResumeAfterInteractionRef.current = true
+      return
+    }
+
+    autoplay.play()
```
lyzno1: [P2] Do not call `play()` on an Autoplay plugin disabled with `active: false`. Under `prefers-reduced-motion: reduce`, Embla skips this plugin's `init`, but `api.plugins()` still exposes the plugin object, so this null check passes and `play()` throws `TypeError: Cannot read properties of undefined (reading 'internalEngine')`. Keep the plugin initialized with `playOnInit: false` if explicit playback should remain available, or omit the Play control in this mode. Please cover this with browser-emulated reduced motion. Upstream behavior: https://github.com/davidjerleke/embla-carousel/blob/v8.6.0/packages/embla-carousel/src/components/PluginsHandler.ts#L22-L29

web/features/home/banner/banner.tsx
```diff
@@ -121,13 +163,17 @@ function BannerCarouselContent({ banners, accountId, language }: BannerCarouselC
...
       <div
         data-carousel-control
-        role="group"
-        aria-label={t(($) => $['pagination.pageNumber'], { ns: 'common' })}
         className="pointer-events-auto flex h-7 min-w-0 shrink-0 items-center gap-2 @min-[996px]/banner:max-w-150 @min-[996px]/banner:min-w-60 @min-[996px]/banner:flex-[1_0_0] @min-[996px]/banner:pr-10"
-        onFocusCapture={pauseRotationForFocus}
-        onBlurCapture={resumeRotationAfterFocus}
       >
-        <div className="flex items-center gap-0.5">
+        <div
```
lyzno1, line 135: [P2] The pause boundary excludes the sibling rotation button. When focus moves from an indicator to Pause, this group's blur handler calls `play()`, while the newly focused button has no matching focus handler; the same restart occurs when the pointer moves from the content or indicators to Pause. The carousel therefore rotates while focus or hover is still inside it. Please put all carousel controls under the same interaction owner. The WAI-ARIA carousel pattern requires rotation to stop whenever an element in the carousel has keyboard focus and while the pointer is hovering it: https://www.w3.org/WAI/ARIA/apg/patterns/carousel/

web/features/home/banner/banner.tsx
```diff
@@ -197,26 +265,22 @@ export function Banner({ banners }: BannerProps) {
     Carousel.Plugin.Fade(),
     Carousel.Plugin.Autoplay({
       delay: AUTOPLAY_DELAY,
-      stopOnFocusIn: true,
-      stopOnInteraction: false,
-      stopOnMouseEnter: true,
+      stopOnFocusIn: false,
+      stopOnInteraction: true,
```
lyzno1, line 227: [P2] Keep the rotation intent and the actual plugin state synchronized after pointer interaction. With `stopOnInteraction: true`, Embla stops permanently on `pointerDown` and does not register the pointer-up restart. The `autoplay:stop` listener only sets `isPlaying`, leaving `isRotationEnabledRef` true, so the stopped carousel still exposes Pause; the first click merely changes the control to Play and a second click is needed to start it. If dragging is a temporary pause, retain `stopOnInteraction: false`; if it is a durable stop, update the intent and accessible action at the same owner. Upstream behavior: https://github.com/davidjerleke/embla-carousel/blob/v8.6.0/packages/embla-carousel-autoplay/src/components/Autoplay.ts#L71-L77

web/features/home/banner/banner.tsx
```diff
@@ -141,14 +187,35 @@ function BannerCarouselContent({ banners, accountId, language }: BannerCarouselC
             />
           ))}
         </div>
+        <IconButton
```
lyzno1: [P2] Place the rotation control first in the carousel's DOM tab sequence. The current order is the active slide link, every slide picker, and then Pause/Play, so keyboard users encounter automatically changing content before the control that stops it. The WAI-ARIA carousel pattern explicitly makes the rotation control the first element in the carousel's Tab sequence: https://www.w3.org/WAI/ARIA/apg/patterns/carousel/. The current test mocks `BannerItem` as a non-interactive article, so please verify the real DOM order rather than only the button count.

web/app/components/main-nav/components/search-button.tsx
```diff
@@ -27,6 +27,7 @@ function useDisplayPlatform() {
 export function MainNavSearchButton() {
   const { t } = useTranslation()
   const displayPlatform = useDisplayPlatform()
+  const ariaKeyShortcuts = `${GOTO_ANYTHING_HOTKEY.replace('Mod', 'Control')} ${GOTO_ANYTHING_HOTKEY.replace('Mod', 'Meta')}`
```
lyzno1: [P2] Expose only the shortcut that is actually implemented on the current platform. `Mod+K` resolves to `Meta+K` on macOS and `Control+K` on Windows/Linux; it does not register both combinations. Declaring `Control+K Meta+K` therefore tells assistive technology about one shortcut that will not work. Derive a single ARIA value from the existing `displayPlatform` so it hydrates alongside the visible `<kbd>`. `aria-keyshortcuts` represents implemented shortcuts: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-keyshortcuts

web/app/components/explore/item-operation/index.tsx
```diff
@@ -36,6 +38,7 @@ function ItemOperation({
   return (
     <DropdownMenu modal={false}>
       <DropdownMenuTrigger
+        aria-label={tCommon(($) => $['operation.moreActionsFor'], { name: itemName })}
```
lyzno1, line 41: [P2] Keep this contextual accessible name, but update the affected E2E locator in the same change. The app now legitimately has a `More actions for <name>` button in both the Web apps and Studio regions, while `duplicate-app.steps.ts` queries the whole page. The current Full-Stack E2E run consequently fails four scenarios with a strict-mode match of two buttons. Scope the locator to the Studio region or the hovered app card; changing this label back would weaken the production semantics instead of fixing the test boundary. Failing job: https://github.com/langgenius/dify/actions/runs/32957914618/job/98143893796

lyzno1 (changes requested):
Requesting changes because the Banner update introduces deterministic autoplay lifecycle and accessibility regressions, and the current Full-Stack E2E failure is caused by an unscoped locator after the accessible-name improvement.

Recommendation: do not merge the current Banner slice as-is. If automatic rotation is not a hard product requirement, the safest and simplest solution is to remove Autoplay and keep manual slide selection. If automatic rotation must remain, use one owner for rotation intent and Embla playback, preserve an explicit pause across reinitialization, and cover reduced motion, resize/reInit, pointer interaction, focus boundaries, and tab order with the real Carousel/Embla implementation in browser tests.

The workspace, account, app-navigation, and Continue Work naming changes otherwise look sound. Keep the contextual app-action name and fix the E2E locator by scoping it to the intended region/card rather than weakening the production semantics.

web/features/home/banner/banner.tsx
```diff
@@ -179,47 +167,179 @@ function BannerCarouselContent({ banners, accountId, language }: BannerCarouselC
...
+  const isPointerActivationRef = useRef(false)
+  const hasExplicitRotationPreferenceRef = useRef(false)
   const [carouselPlugins] = useState(() => [
     Carousel.Plugin.Fade(),
     Carousel.Plugin.Autoplay({
       delay: AUTOPLAY_DELAY,
-      stopOnFocusIn: true,
+      playOnInit: false,
+      stopOnFocusIn: false,
       stopOnInteraction: false,
```
lyzno1: [P1] Preserve pause intent across Embla pointer-up

With `stopOnInteraction: false`, Embla 8.6.0 registers a `pointerUp` handler that calls `startAutoplay()`. Its implementation [emits `autoplay:play` before setting `autoplayActive = true`](https://github.com/davidjerleke/embla-carousel/blob/v8.6.0/packages/embla-carousel-autoplay/src/components/Autoplay.ts#L125-L135), while the public `stop()` [only acts when that flag is already true](https://github.com/davidjerleke/embla-carousel/blob/v8.6.0/packages/embla-carousel-autoplay/src/components/Autoplay.ts#L182-L184). Consequently, `handleAutoplayPlay` calls `stop()` too early and Embla installs the timer afterward. I reproduced this in Chromium: explicitly Pause, release a drag inside the carousel, and the first slide becomes hidden after one delay even though the control still says Play and `aria-live` remains `polite`. This makes the required pause control unreliable under [WCAG 2.2.2](https://www.w3.org/WAI/WCAG22/Understanding/pause-stop-hide.html). Prefer `stopOnInteraction: true` and let the Banner pointer boundary be the sole resume owner, or reconcile only after Embla finishes pointer-up. Replace the current external-drop test with an in-carousel release that asserts the active slide stays unchanged past the delay.

web/features/home/banner/banner.tsx
```diff
@@ -179,47 +167,179 @@ function BannerCarouselContent({ banners, accountId, language }: BannerCarouselC
...
-  if (!firstBanner) return null
+  const toggleRotation = () => {
+    const nextRotationEnabled = !isRotationEnabledRef.current
+    hasExplicitRotationPreferenceRef.current = true
+    isRotationEnabledRef.current = nextRotationEnabled
+    setIsRotationEnabled(nextRotationEnabled)
+    setAutoplayPlaying(api, nextRotationEnabled && !isPointerInsideRef.current)
+  }
+
+  const stopRotationForFocus = () => {
```
lyzno1: [P2] Stop only when focus enters the carousel

This capture handler runs for every descendant focus, not only when focus crosses into the carousel. In Chromium: Tab into the rotation control (it stops), press Space to explicitly Play, then Tab to the first slide picker; the second focus event immediately sets the user's explicit rotation preference back to false. The carousel APG says rotation stops when focus *enters* and does not restart until the user explicitly requests it; moving focus inside the same carousel must not revoke that request. Accept the focus event and return when `event.relatedTarget` is still contained by `event.currentTarget`, then cover the Play -> internal Tab transition in the existing native-focus browser test.

web/features/home/banner/banner.tsx
```diff
@@ -100,34 +103,32 @@ function BannerCarouselContent({ banners, accountId, language }: BannerCarouselC
...
-        onFocusCapture={pauseRotationForFocus}
-        onBlurCapture={resumeRotationAfterFocus}
       >
-        <div className="flex items-center gap-0.5">
+        <IconButton
+          size="md"
+          aria-label={t(($) => $[isRotationEnabled ? 'operation.pause' : 'operation.play'], {
+            ns: 'common',
+          })}
+          className="order-2 shrink-0"
```
lyzno1: [P2] Keep visual and keyboard order aligned

The DOM correctly places the rotation control first, but `order-2` puts it visually after the `order-1` picker group. The browser test currently proves only DOM Tab order. A Chromium geometry check gives the focused rotation control at x=122 and the first picker at x=64, so focus starts on the right and then jumps back left. That disconnect is problematic for keyboard users with low vision and conflicts with the WCAG 2.4.3 technique of [matching DOM and visual order](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html). Keep the rotation control first in both DOM and visual order; the existing browser test can assert the relative bounding boxes because this is a real layout contract.

web/features/home/banner/banner.tsx
```diff
@@ -100,34 +103,32 @@ function BannerCarouselContent({ banners, accountId, language }: BannerCarouselC
...
-        role="group"
-        aria-label={t(($) => $['pagination.pageNumber'], { ns: 'common' })}
         className="pointer-events-auto flex h-7 min-w-0 shrink-0 items-center gap-2 @min-[996px]/banner:max-w-150 @min-[996px]/banner:min-w-60 @min-[996px]/banner:flex-[1_0_0] @min-[996px]/banner:pr-10"
-        onFocusCapture={pauseRotationForFocus}
-        onBlurCapture={resumeRotationAfterFocus}
       >
-        <div className="flex items-center gap-0.5">
+        <IconButton
+          size="md"
+          aria-label={t(($) => $[isRotationEnabled ? 'operation.pause' : 'operation.play'], {
```
lyzno1: [P2] Name the carousel-specific rotation action

These shared keys resolve to only `Pause` and `Play`. The carousel APG requires the changing label to communicate that slide content rotates automatically, with names such as `Stop slide rotation` and `Start slide rotation`; a generic media-style label does not identify what will be paused when the control is encountered. Use Banner-specific localized labels and keep the name based on rotation intent rather than the temporary hover pause. This semantic contract is sufficient to verify in the owner-level component test; it does not need another browser test.

web/app/__tests__/home-banner.browser.spec.tsx
```diff
@@ -0,0 +1,199 @@
...
+    const carousel = screen.getByRole('region', { name: 'explore.banner.carouselLabel' })
+    const slidesContainer = getSlidesContainer(carousel.element())
+    const host = screen.getByTestId('banner-host').element()
+    await screen.getByRole('button', { name: 'common.operation.pause' }).click()
+    await userEvent.unhover(carousel)
+    await expect
+      .element(screen.getByRole('button', { name: 'common.operation.play' }))
+      .toBeVisible()
+
+    const resized = new Promise<void>((resolve) => {
```
lyzno1: [P3] Remove this browser case or prove Embla reinitialized

This promise proves only that the test's own `ResizeObserver` saw the width change; it never observes an Embla `reInit`. The assertions before and after resize are identical, and deleting the production `reInit` handler still leaves this test green. If the ResizeObserver-to-Embla integration is a verified regression boundary worth the browser cost, assert behavior that requires reinitialization (for example, enabled autoplay still advances afterward). Otherwise keep the intent/reconciliation assertion at the owner-level unit boundary and remove this browser test rather than retaining a false-positive integration check.

lyzno1 (changes requested):
Requesting changes after re-reviewing the complete updated diff. The non-Banner accessible-name changes, the Studio-scoped E2E locator, and the platform-specific search shortcut are sound. The Banner still has deterministic pause/focus regressions in real Chromium, and two browser tests currently give false confidence. Browser coverage should remain limited to the real Embla/pointer, native focus, and layout boundaries that owner-level unit tests cannot faithfully prove.

web/features/home/banner/banner.tsx
```diff
@@ -179,47 +172,181 @@ function BannerCarouselContent({ banners, accountId, language }: BannerCarouselC
...
+    hasExplicitRotationPreferenceRef.current = true
+    isRotationEnabledRef.current = nextRotationEnabled
+    setIsRotationEnabled(nextRotationEnabled)
+    setAutoplayPlaying(api, nextRotationEnabled && !isPointerInsideRef.current)
+  }
+
+  const stopRotationForFocus = (event: ReactFocusEvent<HTMLDivElement>) => {
+    if (event.relatedTarget instanceof Node && event.currentTarget.contains(event.relatedTarget))
+      return
+    if (isPointerActivationRef.current) return
```
lyzno1, line 243: [P2] Stop rotation when input switches from pointer to keyboard

`isPointerActivationRef` suppresses focus-based stopping for every carousel control, not only the rotation control. In Chromium, click a slide picker, move the pointer outside, then press Tab to another picker: `resumeRotationAfterPointer` restarts autoplay because the rotation intent is still true, and the internal-focus guard prevents the subsequent Tab transition from stopping it. `aria-live` remains `off` and slides keep rotating even though the user never activated “Start slide rotation”. The [WAI carousel pattern](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/) requires rotation to stop when an element receives keyboard focus and resume only when the rotation control is activated. Limit the pointer-focus exception to the rotation control, or otherwise track the modality/explicit-start intent. This native pointer/focus/Tab sequence is browser-owned, so it is reasonable to extend the existing focus Browser Mode case to cover it.

web/features/home/banner/__tests__/banner.spec.tsx
```diff
@@ -331,13 +395,35 @@ describe('Banner', () => {
...
+    render(
+      <Banner
+        banners={[
+          createMockBanner('1', 'enabled', 'First banner'),
+          createMockBanner('2', 'enabled', 'Second banner'),
+        ]}
+      />,
+    )
+
+    await user.click(screen.getByRole('button', { name: 'banner.stopRotation' }))
```
lyzno1: [P2] Exercise the reInit behavior this handler actually owns

This scenario explicitly stops rotation before emitting `reInit`, so the intent ref is already false. I removed the production `handleReInit` function and its subscription locally, and this test still passed unchanged; it therefore does not protect the resize/reinitialization behavior. With `playOnInit: false`, the Dify-owned contract is the opposite path: rotation intent remains enabled while Embla comes back stopped, and the handler must call `play()` to restore it. Set `mockAutoplayPlaying = false` without clicking Stop, emit `reInit`, and assert `mockAutoplay.play()` plus the stop-rotation UI. That proves the owner contract at unit level without bringing back the false-positive ResizeObserver browser test.

lyzno1 (changes requested):
Two issues remain after the latest accessibility update. The earlier production findings are fixed, and the new drag Browser Mode coverage is justified because it detects the real Embla pointer-up regression.

lyzno1 (approved):
LGTM. Re-reviewed head `2ed5050676ad6d2a6aac70b8051fa2454caaa6eb`. The pointer-to-keyboard focus gap is fixed by scoping pointer activation to the rotation control, and the `reInit` unit case now exercises the handler-owned restart path. Browser coverage remains limited to native pointer/focus ordering, geometry, hit testing, real Embla pointer behavior, and reduced-motion initialization.

Verified locally: Banner unit tests 18/18, Banner Browser Mode 5/5, targeted `vp check`, accessibility lint, and the live `localhost:3000` interaction flow.

PR #41311, opened 2026-08-26
refactor(openapi): delete the dead bearer→handler auth adapter

GareArc:
## Summary

Deletes the dead parallel bearer→handler auth adapter under `/openapi/v1`, so the seam has exactly one implementation.

## Checklist

**Notes on the checklist:** this PR only deletes code, so no tests were added — 4 test files targeting deleted code were removed, and the remaining suite passes unchanged (**5384 passed, 1 skipped**). `ruff check` and `ruff format --check` are clean. `make type-check` aborts inside pyrefly before mypy runs, which is pre-existing and unrelated to this branch, so mypy was run directly against the touched modules: its 232 findings are all `import-untyped` (missing `flask_restx` / `flask_login` stubs), with zero real type errors. This branch only removes imports, so it cannot have introduced them.


api/libs/oauth_bearer.py
lin-snow, line 428: ENABLE_OAUTH_BEARER=false should return 500 instead of 503？

The corresponding validate_bearer block is removed in this PR. Please make sure the disabled-bearer path now maps to 500, rather than exposing it as a 503 Service Unavailable.

api/controllers/openapi/_models.py
```diff
@@ -50,20 +50,6 @@ class MessageMetadata(BaseModel):
     retriever_resources: list[dict[str, Any]] = []
 
 
-class PaginationEnvelope[T](BaseModel):
```
lin-snow, line 53: Removing PaginationEnvelope leaves pagination without a single source of truth 🤔

The PR head now has four duplicated declarations in _models.py at :63, :71, :154, and :357, plus three hand-written has_more expressions。

api/libs/oauth_bearer.py
GareArc: The deleted 503 branch was unreachable, since no route used `validate_bearer`.

wylswz (dismissed):

wylswz:
There's a conflict

wylswz (approved):

PR #41318, opened 2026-08-26
refactor(api): move web authentication into application services

41tair:
> [!IMPORTANT]
>

## Summary
part of #39993

## Screenshots

| Before | After |
| ------ | ----- |
| ...    | ...   |

## Checklist

- [X] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [X] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [X] I've updated the documentation accordingly.
- [X] I ran `make lint && make type-check` (backend) and `vp staged` (frontend) to appease the lint gods


api/services/web_authentication_adapters.py
```diff
@@ -0,0 +1,236 @@
...
+    WebAuthenticationAuditGateway,
+    WebAuthenticationSecurityGateway,
+    WebAuthenticationTokenGateway,
+    WebEmailDeliveryRateLimitError,
+)
+from tasks.mail_email_code_login import send_email_code_login_mail_task
+from tasks.mail_reset_password_task import send_reset_password_mail_task
+
+
+class TokenManagerWebAuthenticationGateway(WebAuthenticationTokenGateway):
```
Stream29, line 27: What't the meaning of these gateways?
XxxGateway is not a good practice of naming.

41tair: XXXGateway is meaning that is an abstract of external service.The implementation can be changed so we don't need to patch for test and we can use different impl for different edition.

api/repositories/account_repository.py
```diff
@@ -63,6 +64,22 @@ def get_credentials(self, account_id: str) -> AccountCredentials | None:
                 return None
             return AccountCredentials(password_hash=account.password, password_salt=account.password_salt)
 
+    @override
+    def find_for_authentication(self, email: str) -> AccountAuthenticationSnapshot | None:
+        with self._session_factory() as session:
+            account = session.scalar(select(Account).where(Account.email == email).limit(1))
```
Stream29: Incorrect and inconsistent behavior.
We have `normalized_email` field for this usage.

api/repositories/webapp_access_query_repository.py
```diff
@@ -22,3 +24,22 @@ def find_app_id_by_code(self, app_code: str) -> str | None:
                 return app_id if app_id is not None else None
         except (DBAPIError, TimeoutError) as e:
             raise WebAppAccessUnavailableError from e
+
+    @override
+    def find_active_session(
+        self,
+        *,
+        app_id: str,
+        app_code: str,
```
wylswz: `app_id` is derivable from `app_code`, so could we accept just `app_code` and resolve the app from the site? If we retain both, we should verify that `site.app_id == app_id` and that the end user belongs to the same app (`end_user.app_id == app_id`). The current checks only establish that these records exist independently, so mismatched app/site/end-user combinations could pass.

PR #41319, opened 2026-08-26
fix(web): improve a11y of dataset clickable divs with native buttons

LuckTerence:
Replace clickable `<div>` elements with native `<button>` elements in two dataset components to fix suppressed jsx-a11y violations (`no-static-element-interactions`, `click-events-have-key-events`), following the same approach as #41301.

## Changes

- `ExternalApiSelect.tsx`: convert the select trigger, dropdown options, and the add-new-API row to `<button type="button">`; add `aria-haspopup="listbox"` and `aria-expanded` on the trigger
- `template-card/operations.tsx`: convert the edit / export / delete menu items to `<button type="button">`
- Remove the corresponding entries from `oxlint-suppressions.json`
- Add accessibility tests to `ExternalApiSelect.spec.tsx` (trigger semantics, expanded state, button roles)

## Verification

- `pnpm lint:a11y` on both files: 0 warnings, 0 errors
- `pnpm test ExternalApiSelect.spec`: 12/12 passed
- `pnpm type-check`: 0 errors

### DIFF
--- a/oxlint-suppressions.json
+++ b/oxlint-suppressions.json
@@ -1870,14 +1870,6 @@
       "count": 1
     }
   },
-  "web/app/components/datasets/create-from-pipeline/list/template-card/operations.tsx": {
-    "jsx-a11y/click-events-have-key-events": {
-      "count": 3
-    },
-    "jsx-a11y/no-static-element-interactions": {
-      "count": 3
-    }
-  },
   "web/app/components/datasets/create/file-preview/index.tsx": {
     "eslint-react/set-state-in-effect": {
       "count": 1
@@ -2252,12 +2244,6 @@
   "web/app/components/datasets/external-knowledge-base/create/ExternalApiSelect.tsx": {
     "eslint-react/set-state-in-effect": {
       "count": 1
-    },
-    "jsx-a11y/click-events-have-key-events": {
-      "count": 3
-    },
-    "jsx-a11y/no-static-element-interactions": {
-      "count": 3
     }
   },
   "web/app/components/datasets/external-knowledge-base/create/ExternalApiSelection.tsx": {
--- a/web/app/components/datasets/create-from-pipeline/list/template-card/operations.tsx
+++ b/web/app/components/datasets/create-from-pipeline/list/template-card/operations.tsx
@@ -12,21 +12,21 @@ type OperationsProps = {
 const Operations = ({ openEditModal, onDelete, onExport, onClose }: OperationsProps) => {
   const { t } = useTranslation()
 
-  const onClickEdit = (e: React.MouseEvent<HTMLDivElement>) => {
+  const onClickEdit = (e: React.MouseEvent<HTMLButtonElement>) => {
     e.stopPropagation()
     e.preventDefault()
     onClose?.()
     openEditModal()
   }
 
-  const onClickExport = (e: React.MouseEvent<HTMLDivElement>) => {
+  const onClickExport = (e: React.MouseEvent<HTMLButtonElement>) => {
     e.stopPropagation()
     e.preventDefault()
     onClose?.()
     onExport()
   }
 
-  const onClickDelete = (e: React.MouseEvent<HTMLDivElement>) => {
+  const onClickDelete = (e: React.MouseEvent<HTMLButtonElement>) => {
     e.stopPropagation()
     e.preventDefault()
     onClose?.()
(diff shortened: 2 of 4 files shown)


asukaminato0721 (dismissed):

LuckTerence:
@asukaminato0721 the formatting fix for `ExternalApiSelect.spec.ts` is pushed and CI is green. The new commit dismissed your earlier approval — could you take another look when you have a moment?

asukaminato0721 (approved):

lyzno1:
@LuckTerence @asukaminato0721 I completed a post-merge accessibility review of the full production diff, including the final DOM owners, Base UI contracts, MDN guidance, the WAI-ARIA APG patterns, and WCAG 2.2. There are two P2 accessibility defects in the shipped change, plus one relevant labeling issue that remains unresolved.

### 1. aria-haspopup="listbox" declares a widget that does not exist

In ExternalApiSelect, the trigger declares aria-haspopup="listbox", but the popup is a generic div and every selectable row is a native button. There is no element with role="listbox", no role="option", no selected state, and no listbox focus/arrow-key model. The popup also mixes a separate "Create new External API" command into the same surface.

MDN is explicit that the popup role must match the aria-haspopup value, and that ARIA only conveys intended behavior; it does not add the behavior. The APG listbox pattern requires a listbox owner, option children, selection state, and composite keyboard navigation. This implementation therefore announces a listbox to assistive technology while exposing a different widget in the DOM. That is an incorrect role/state contract under WCAG 4.1.2, not an accessibility improvement.

Source: https://github.com/langgenius/dify/blob/9d396524b95e4bb4a3057846554ccacb91b03aed/web/app/components/datasets/external-knowledge-base/create/ExternalApiSelect.tsx#L60-L118

Guidance:
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-haspopup
- https://www.w3.org/WAI/ARIA/apg/patterns/listbox/
- https://www.w3.org/WAI/WCAG22/Understanding/name-role-value

Fix direction: use the Dify UI Select owner and move the create command outside the option list, or model this as a DropdownMenu with radio items plus a separate menu item if command-menu semantics are the intended product contract. Do not hand-author aria-haspopup="listbox" onto unrelated buttons.

### 2. Native buttons inside Menu.Popup are not Menu.Items

The edit/export/delete controls are rendered inside Dify UI DropdownMenuPopup, whose Base UI owner exposes role="menu". Replacing the inner divs with raw buttons does not register them as Menu.Item elements. The popup therefore has no menuitem descendants and Base UI cannot provide its item registry, roving focus, Up/Down Arrow, Home/End, or typeahead behavior.

Source:
- https://github.com/langgenius/dify/blob/9d396524b95e4bb4a3057846554ccacb91b03aed/web/app/components/datasets/create-from-pipeline/list/template-card/actions.tsx#L50-L70
- https://github.com/langgenius/dify/blob/9d396524b95e4bb4a3057846554ccacb91b03aed/web/app/components/datasets/create-from-pipeline/list/template-card/operations.tsx#L39-L68

Guidance:
- https://base-ui.com/react/components/menu
- https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menu_role

Fix direction: render these actions through DropdownMenuItem at the existing Menu owner. Native button semantics alone are not a substitute for the composite menu contract.

### 3. The visible field label is still not associated with the new trigger

The surrounding "External Knowledge API" label has no htmlFor/id relationship and the new trigger has no aria-labelledby relationship. Its accessible name is instead the placeholder or selected API name, so the field name changes with the value and does not include the visible label. This issue predates the PR, but it means the modified control still does not have a complete accessible form-control contract. A Select.Label owner or an explicit label relationship should provide the stable name.

Source: https://github.com/langgenius/dify/blob/9d396524b95e4bb4a3057846554ccacb91b03aed/web/app/components/datasets/external-knowledge-base/create/ExternalApiSelection.tsx#L69-L99

### Test and review gap

The added tests assert the aria-haspopup string and native button roles, but never assert that a listbox/options exist, never exercise selected state or arrow-key/focus behavior, and never render Operations through its actual Menu.Popup owner. The test named "listbox semantics" therefore locks in the invalid contract instead of detecting it:

https://github.com/langgenius/dify/blob/9d396524b95e4bb4a3057846554ccacb91b03aed/web/app/components/datasets/external-knowledge-base/create/__tests__/ExternalApiSelect.spec.tsx#L131-L171

The final commit cc34942a only formats this test; it makes no production or accessibility correction. The green lint, unit, Browser, and E2E checks do not exercise either broken composite-widget contract.

I did not find a normal-state spacing, width, typography, or alignment regression from the div-to-button conversion: the project preflight resets button margin/padding/background and inherits typography, and the added width/text-alignment classes preserve the prior geometry. The native keyboard focus outline is expected accessibility feedback, not a visual regression.

Please do not approve accessibility changes based only on removed lint suppressions, a div-to-button replacement, green CI, or tests that assert ARIA strings. Review the final rendered primitive owner, role hierarchy, accessible name/state, focus movement, and keyboard interaction first. These issues need a follow-up fix.

PR #41327, opened 2026-08-26
ci: run the dify-agent test suite

alanhuangyoo:
## Summary

`dify-agent` has 78 test files and 777 tests. Nothing in `.github/workflows` runs pytest against them — `autofix.yml` is the only workflow that looks at the directory, and it only runs ruff.

The sibling component `dify-agent-runtime` already has its own paths filter and its own `sandbox-runtime-tests.yml`. This PR gives `dify-agent` the same arrangement.

## Why it matters right now

The suite does not collect on `main`:

```
$ cd dify-agent && uv run pytest -q
Interrupted: 5 errors during collection
5 errors in 1.94s
```

That is what #41208 fixes, and the absent CI job is why it went unnoticed. This branch is stacked on #41208, so with both:

```
777 passed, 32 skipped in 22.37s
```

Both numbers are from `origin/main` as of `0d05eae5`.

If you would rather land them as one change, I am happy to close #41208 and fold its four lines in here.

## What it does

- `.github/workflows/dify-agent-tests.yml` — mirrors `sandbox-runtime-tests.yml`, with the uv/Python setup taken from `api-tests.yml` since this component is Python rather than Go.
- `main-ci.yml` — a `dify-agent` paths filter next to `sandbox-runtime`, plus the run / skip / finalize trio, so the required check stays stable when no `dify-agent` file changed.

The workflow is `workflow_call`-only, like the other component workflows, and the filter includes the workflow file itself so a change to it re-runs the job.

### DIFF
--- a/.github/workflows/dify-agent-tests.yml
+++ b/.github/workflows/dify-agent-tests.yml
@@ -41,7 +41,7 @@ jobs:
         run: uv lock --check
 
       - name: Install dependencies
-        run: uv sync --dev
+        run: uv sync --extra server --dev
 
       - name: Run unit tests
-        run: uv run pytest --timeout "${PYTEST_TIMEOUT:-20}"
+        run: uv run pytest
--- a/.github/workflows/dify-agent-tests.yml
+++ b/.github/workflows/dify-agent-tests.yml
@@ -1,0 +1,47 @@
+name: Dify Agent Tests
+
+on:
+  workflow_call:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: dify-agent-tests-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  dify-agent-unit:
+    name: Dify Agent Unit Tests
+    runs-on: depot-ubuntu-24.04
+    defaults:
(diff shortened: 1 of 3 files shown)


BeautyyuYanli (dismissed):

alanhuangyoo:
The workflow was red on its own first run — two things, both mine:

`--timeout` came from `api-tests.yml`, but `pytest-timeout` is a dependency of `api/` and not of `dify-agent`, so pytest bailed during argument parsing:

```
pytest: error: unrecognized arguments: --timeout
```

Dropping it alone still left the run red. `uv sync --dev` installs only the six packages in `[project].dependencies`, and every server/storage test imports `fastapi`, `graphon`, `jsonschema`, `jwcrypto`, `pydantic-settings` or `redis` — all in the `server` extra.

Now `uv sync --extra server --dev` and a bare `uv run pytest`, which picks up the projects own `addopts` and `testpaths`:

```
777 passed, 32 skipped in 21.20s
```

Sorry for the churn @BeautyyuYanli — re-approval needed since the workflow changed.

wylswz (approved):

PR #41247, opened 2026-08-25
chore: bump version to 1.17.0

wylswz:
## Summary
Bump the project version from `1.16.1` to `1.17.0` across all version-carrying files:

- `api/pyproject.toml` and `api/uv.lock` (`dify-api`, `dify-agent`)
- `dify-agent/pyproject.toml` and `dify-agent/uv.lock`
- `web/package.json`
- `cli/package.json` — `difyctl` compat `maxDify` bumped to `1.17.0` (keeps `minDify` at `1.16.0`, so the CLI still supports the 1.16.x range)
- `docker/docker-compose-template.yaml` and `docker/docker-compose.yaml` image tags

#### Test plan
- [ ] Verify no lingering `1.16.1` references remain in version files
- [ ] CI passes

Generated with [Devin](https://devin.ai)

PR #41355, opened 2026-08-27
fix(ci): correct difyctl release and e2e workflow gates

GareArc:
Fixes three CI gates that were passing while doing nothing, and one that could not fail.

### `cli-e2e.yml` suite toggles never worked

The six suite guards read `if: ${{ inputs.suite_x != 'false' }}`. The inputs are declared `type: boolean`, so this compares a boolean to the string `'false'` — never equal, always true. Every suite ran on every dispatch regardless of which boxes were ticked. Now `if: ${{ inputs.suite_x }}`.

### The provenance-tag step was silently green

`Create provenance tag` reports success on every release, but `git/matching-refs/tags/difyctl` returns zero refs — no tag has ever been created. Repository ruleset `1715221` (`target=tag`, `enforcement=active`) blocks tag creation, the POST comes back `422`, and the step's `case` treats `422` as "tag already exists; skipping (immutable)" and exits 0. Three releases shipped on that false message. Step removed.

### Compat bounds did not fail closed

`release-build.sh` read `difyctl.compat.minDify`/`maxDify` straight into the binary. A malformed bound — including the literal string `undefined` when the field is missing — was baked in and silently disabled the runtime version gate. Both `release-build.sh` and `release-naming.mjs validate` now reject anything that is not a plain `X.Y.Z`, and reject `minDify > maxDify`.

### Two manifest gates, two comparators

`release-validate-manifest.sh` carried its own weaker copy of the semver comparison. Folded into `release-naming.mjs validate` and deleted (41 lines); `cli-tests.yml` and `cli-release.yml` both call the single gate.

### Also

- `dry_run` input on `cli-release.yml`, guarding asset upload and the stale-asset prune.
- Running from source (`bin/dev.js`, `vite.config.ts`) defaults to `DIFYCTL_CHANNEL=dev` instead of self-reporting the release channel; `pnpm build:bin:local` does the same for local binaries.
- `cli/package.json` moves to `1.17.0` on the `stable` channel, aligning difyctl with the current Dify release line. The version stays hand-maintained and subsequent bumps belong to the release SOP — `cli/AGENTS.md` documents that, including the constraint that the version must match its channel's form or `validate` fails the manifest gate.

## Testing

Exercised end to end against real GitHub releases on a fork, 6/6:

| Case | Result |
|---|---|
| out-of-window release | red X at the compat gate, `release` job skipped, 0 assets |
| release staged as prerelease | no run; fires on promotion to full release |
| `dry_run=true` | builds + checksums, upload and prune skipped, 0 assets |
| in-window release | 6 assets, sha256 matched the published manifest, binary runs and loads the keyring |
| hotfix-shaped tag (`1.16.9-fix.1`) | attaches — the tag's shape no longer gates, only its precedence |
| re-run at a bumped difyctl version | 6 stale assets pruned, 6 new uploaded, other releases untouched |


cli/AGENTS.md
```diff
@@ -10,6 +10,11 @@ This package is the Node 22+, ESM TypeScript implementation of `difyctl`. Develo
 - Preserve flags, output, and exit codes during refactors. Do not add dependencies or compatibility shims unless the task explicitly requires them.
 - `ARD.md` owns CLI code structure. Keep wire behavior aligned with typed API clients and the real mock-server behavior tests.
 
+## Version and Channel
+
+- `cli/package.json` `version` is an inert placeholder pinned to `0.0.0-private`. Do not bump it; `scripts/release-naming.mjs validate` fails the manifest gate if it changes.
```
wylswz: Manually editing the version before release is currently a part of our SOP, see #41247
I think we can keep the pipeline simple for now.

cli/AGENTS.md
```diff
@@ -10,6 +10,11 @@ This package is the Node 22+, ESM TypeScript implementation of `difyctl`. Develo
 - Preserve flags, output, and exit codes during refactors. Do not add dependencies or compatibility shims unless the task explicitly requires them.
 - `ARD.md` owns CLI code structure. Keep wire behavior aligned with typed API clients and the real mock-server behavior tests.
 
+## Version and Channel
+
+- `cli/package.json` `version` is bumped by hand before a release. It must match the form of `difyctl.channel` — `scripts/release-naming.mjs validate` fails the manifest gate otherwise, so moving to a prerelease version means moving the channel with it.
```
wylswz: Maybe move this backtick backward?

cli/scripts/install-cli.test.ts
```diff
@@ -149,6 +149,21 @@ describe('install-cli pick_asset', () => {
   })
 })
 
+describe('install-cli pick_asset with a +r2 rebuild alongside the original', () => {
+  const BOTH = JSON.stringify({
+    assets: [{ name: 'difyctl-v1.16.1-linux-x64' }, { name: 'difyctl-v1.16.1+r2-linux-x64' }],
+  })
+
+  it('picks the superseded plain asset, not the rebuild', () => {
+    expect(pickAsset('linux-x64', BOTH)).toBe('difyctl-v1.16.1-linux-x64')
```
wylswz: Why do we prefer `difyctl-v1.16.1-linux-x64` here?

cli/AGENTS.md
GareArc: After thinking about it, I've decided to drop this part since it's already covered in the SOP.

cli/scripts/install-cli.test.ts
GareArc: It's over-protection for the case where a release needs to be overwritten but the old
executables failed to be deleted. In that case we sort and take the last one.  Unlikely to happend so I'll drop the test.

wylswz (approved):

PR #41380, opened 2026-08-27
fix(plugin): make uninstall credential cleanup failure-safe

### LINKED ISSUE #39893: [Refactor/Chore] Make plugin uninstall credential cleanup failure-safe

### Description

After #39764, `PluginService.uninstall` performs the plugin-daemon uninstall **before** deleting `TenantPreferredModelProvider` / `ProviderCredential` records and invalidating provider caches (see `api/core/plugin/plugin_service.py`, `uninstall`). If the credential cleanup raises (DB failure, transient lock, etc.), the endpoint reports an error but the plugin is already uninstalled, leaving orphaned credentials and stale caches. The previous implementation deleted credentials first and only uninstalled after a successful cleanup, so a failure left the installation and its credentials consistent.

### Motivation

The current ordering makes an ordinary uninstall non-atomic across two systems (plugin daemon and database). For a multi-tenant SaaS/self-hosted deployment, a transient DB error during uninstall can silently orphan model-provider credentials; the next install with the same plugin may reuse credentials that were supposed to be removed, or the operator sees an error although the plugin is gone.

### Additional Context

`preserve_credentials=True` intentionally skips cleanup for local replacement flows, so the fix only needs to cover the ordinary path. Suggested directions (open to maintainer preference):

1. Keep the current order but make cleanup resilient: on cleanup failure, log a tenant-scoped warning and queue a bounded retry (Celery task) so credentials are eventually removed; the API still reports the uninstall result accurately.
2. Restore cleanup-before-uninstall for the non-preserve path and accept the smaller failure window (uninstall fails → credentials already removed).
3. Treat credential cleanup failure as an uninstall failure and record the plugin as "uninstalled with pending cleanup" for reconciliation.

No PR is attached yet; this is a behavior/consistency question that deserves maintainer input before implementing.

ddddajie:
## Summary

- keep plugin-daemon uninstall as the authoritative first step
- make post-uninstall credential cleanup retryable after transient failures
- keep credential cleanup idempotent across partial DB/cache completion
- preserve existing `preserve_credentials=True` replacement behavior

## Details

After the plugin daemon successfully uninstalls a plugin, credential cleanup can
still fail because of a transient database or cache error. In that case the
plugin is already gone, but the uninstall request currently fails and may leave
orphaned credentials or stale provider caches.

This change keeps the existing daemon-first ordering, performs cleanup
synchronously when possible, and schedules a bounded Celery retry on cleanup
failure.

Provider IDs are discovered independently from surviving credential rows so a
retry can still invalidate provider caches when the database cleanup completed
before a cache failure.

The retry task only performs credential/cache cleanup and never retries the
plugin daemon uninstall.

## Tests

- daemon uninstall failure preserves credentials and does not schedule cleanup
- `preserve_credentials=True` skips cleanup and retry scheduling
- cleanup failure after daemon success returns the successful uninstall result
- retry dispatch failure does not change the successful uninstall result
- partial DB-success/cache-failure cleanup is idempotent on retry
- Celery task uses the plugin queue, bounded retries, shared cleanup logic, and
  never calls daemon uninstall

Fixes #39893

### DIFF
--- a/api/core/plugin/plugin_service.py
+++ b/api/core/plugin/plugin_service.py
@@ -1243,6 +1243,72 @@ class PluginService:
         return result
 
     @staticmethod
+    def _cleanup_plugin_credentials(tenant_id: str, plugin_id: str) -> None:
+        """Delete plugin-owned credentials and invalidate their provider caches."""
+        provider_name_pattern = f"{plugin_id}/%"
+
+        with Session(db.engine) as session, session.begin():
+            session.execute(
+                delete(TenantPreferredModelProvider).where(
+                    TenantPreferredModelProvider.tenant_id == tenant_id,
+                    TenantPreferredModelProvider.provider_name.like(provider_name_pattern),
+                )
+            )
+
+            provider_ids = session.scalars(
+                select(Provider.id).where(
+                    Provider.tenant_id == tenant_id,
+                    Provider.provider_name.like(provider_name_pattern),
+                )
+            ).all()
+            credential_ids = session.scalars(
+                select(ProviderCredential.id).where(
+                    ProviderCredential.tenant_id == tenant_id,
+                    ProviderCredential.provider_name.like(provider_name_pattern),
+                )
+            ).all()
+
+            if not credential_ids:
+                logger.info("No credentials found for plugin: %s", plugin_id)
+            else:
+                session.execute(
+                    update(Provider)
+                    .where(
+                        Provider.tenant_id == tenant_id,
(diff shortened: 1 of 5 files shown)


ddddajie:
Hi maintainers, gentle follow-up on this PR. The main CI and Pyrefly checks are green. The change keeps plugin-daemon uninstall as the authoritative step and only makes post-uninstall credential cleanup failure-safe and retryable, with focused regression coverage for the failure and retry paths.

I'd appreciate a review when convenient. Happy to make any changes you think are needed. Thanks!

PR #41393, opened 2026-08-27
fix: improve accessibility across the apps list

iamjoel:
## Summary

- Add accessible list and heading semantics with contextual labels for app cards and tag controls.
- Improve creator filter and marketplace modal keyboard and screen-reader behavior while preserving the existing visual layout.
- Announce empty, loading, and error states and mark decorative app icons appropriately.
- Keep the Studio AA audit running for all non-color rules while explicitly excluding `color-contrast` from this page's scoped check.

Fixes SNP-637

## Checklist


web/app/components/apps/creators-filter.tsx
```diff
@@ -92,155 +89,169 @@ const CreatorsFilter = ({ value, onChange }: CreatorsFilterProps) => {
...
-              >
-                <Checkbox id={creator.id} checked={checked} className="shrink-0" />
-                <div className="flex min-w-0 grow items-center gap-2 px-1">
-                  <Avatar
-                    avatar={creator.avatarUrl}
-                    name={creator.name}
-                    size="xs"
-                    className="border-[0.5px] border-divider-regular"
+              return (
+                <button
```
lyzno1: **[P2] Use Combobox as the selection owner**

This surface owns a persistent multiple selection from a predefined, filterable creator collection, which is exactly the Combobox contract in Dify UI and Base UI. Keeping it as a dialog of hand-built checkbox buttons gives every creator a separate Tab stop instead of the listbox active-option/arrow-key model. It also nests Dify's Base UI `Checkbox` (including its checkbox root/hidden input) inside another button exposed as a checkbox; `aria-hidden` and `tabIndex={-1}` do not make that interactive composition valid.

Please migrate this now with the current API: use `Combobox<CreatorOption, true>`, `multiple`, object `items`/selected values, `isItemEqualToValue` by `id`, and map `onValueChange` back to the public `string[]` boundary while preserving any selected IDs missing from the current member collection. Compose `ComboboxTrigger`, the supported input-inside-popup pattern, `ComboboxList`, `ComboboxItem`, and `ComboboxItemIndicator`; keep Reset/Clear outside `ComboboxList` and do not add ARIA roles manually. Base UI should own `aria-expanded`, `aria-controls`, `aria-activedescendant`, option selection, highlight, and keyboard behavior. The future third `Item` generic/`createItems` can later remove the object-to-ID adapter; it is not a prerequisite for the correct structure. See [Base UI usage guidance and input-inside-popup example](https://base-ui.com/react/components/combobox#input-inside-popup) and the [WAI-ARIA combobox relationships](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#wai-ariaroles_states_and_properties).

web/app/components/apps/creators-filter.tsx
```diff
@@ -92,155 +89,169 @@ const CreatorsFilter = ({ value, onChange }: CreatorsFilterProps) => {
...
+                )}
+                value={keywords}
+                onChange={(e) => setKeywords(e.target.value)}
+                placeholder={t(($) => $['studio.filters.searchCreators'], { ns: 'app' })}
+              />
+              {!!keywords && (
+                <button
+                  type="button"
+                  aria-label={t(($) => $['operation.clear'], { ns: 'common' })}
+                  className="absolute top-1/2 right-2 flex size-4 -translate-y-1/2 items-center justify-center rounded-sm text-components-input-text-placeholder outline-hidden hover:text-components-input-text-filled focus-visible:ring-2 focus-visible:ring-state-accent-solid"
```
lyzno1: **[P2] Enlarge the search clear target**

This button's pointer target is only 16×16 CSS px and overlaps the input, which is another target. That is the specific overlapping-small-target case that fails WCAG 2.2 SC 2.5.8's spacing exception. Keep the glyph at 16px if desired, but make the button target at least 24×24 and reserve matching end padding in the input. See [WCAG 2.5.8 Target Size (Minimum)](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html#spacing).

web/app/components/apps/creators-filter.tsx
```diff
@@ -92,155 +89,169 @@ const CreatorsFilter = ({ value, onChange }: CreatorsFilterProps) => {
...
+              )
+            })}
+          </div>
+        </PopoverContent>
+      </Popover>
+      {isSelected && (
+        <button
+          type="button"
+          aria-label={t(($) => $['studio.filters.reset'], { ns: 'app' })}
+          className="absolute top-1/2 right-2 flex h-4 w-4 -translate-y-1/2 items-center justify-center rounded-xs text-text-quaternary outline-hidden hover:text-text-tertiary focus-visible:ring-2 focus-visible:ring-state-accent-solid"
```
lyzno1: **[P2] Do not overlay a 16px reset target on the trigger**

The reset button is 16×16 and sits directly on top of the adjacent Popover trigger target, so a 24px spacing circle centered on it intersects another target. Make the reset target at least 24×24 and reserve that width in the trigger so the two controls have unambiguous hit areas. The popup Reset can remain a separate full-size action. See [WCAG 2.2 SC 2.5.8](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html#spacing).

web/app/components/apps/app-card/index.tsx
```diff
@@ -187,12 +194,11 @@ export const AppCard = memo(
     )
 
     return (
-      <div className="group relative col-span-1 h-41.5">
+      <div role="listitem" aria-labelledby={appNameId} className="group relative col-span-1 h-41.5">
```
lyzno1: **[P2] Keep the app control's name concise**

This change makes the accessibility tree expose a named list item (`Test App`) followed by a link/button named `Test App chatbot <full description> <author> <edited time>`; the updated unit/E2E locators then codify the extra list-item name as a test hook. That duplicates the app name and turns supplementary, frequently changing metadata into the control's name, making link-list and card-by-card navigation unnecessarily verbose. MDN explicitly recommends that `aria-labelledby` reference brief label text and that longer supporting content use `aria-describedby`.

Keep the new `role=listitem`/parent `role=list` structure, but remove `aria-labelledby` from the list item, name the link/button with `appNameId` only, and retain the description through `aria-describedby` when present. Mode/author/timestamp can remain visible content without being forced into the accessible name. The E2E steps can then use the exact app link again and scope the sibling action within the card through structure rather than requiring a named list item. See [MDN aria-labelledby guidance](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby#description).

web/app/components/apps/starred-app-card.tsx
```diff
@@ -94,11 +100,11 @@ export const StarredAppCard = memo(
     )
 
     return (
-      <div className="group relative">
+      <div role="listitem" aria-labelledby={appNameId} className="group relative">
```
lyzno1: **[P2] Apply the same concise-name contract to starred cards**

The wrapper is now named with the app name and the inner link/button is named with both the app name and all author/time metadata, so screen-reader navigation repeats the name and receives volatile metadata as part of the control label. Keep `role=listitem`, remove its author-provided name, label the interactive card with `appNameId`, and expose genuinely useful supporting text as a description rather than adding it to `aria-labelledby`. [MDN distinguishes brief labels from longer descriptions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby#description).

web/app/components/apps/empty.tsx
```diff
@@ -12,7 +12,11 @@ const Empty = ({ message, stepByStepTourTarget }: EmptyProps) => {
 
   return (
     <FilterEmptyState
-      title={message ?? t(($) => $['filterEmpty.noApps'], { ns: 'app' })}
+      title={
+        <span role="status" aria-live="polite" aria-atomic="true">
```
lyzno1: **[P2] Put result announcements on a stable owner**

`Empty` is mounted only after the result set becomes empty, so the live-region node and its populated text are created in the same render. Status announcements are most reliable when the live region already exists and its content changes after the query settles; this component also has no access to the catalog's `isFetching` state, so it cannot coordinate `aria-busy` with stale/placeholder results. Keep this component visual, and place one stable `role="status"` at `AppListCatalogContent`, update its text only when `!isFetching && !isPlaceholderData`, and mark the result owner busy while fetching. `role=status` already implies `aria-live=polite` and `aria-atomic=true`. See [MDN status role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/status_role) and [WCAG 4.1.3 status messages](https://www.w3.org/WAI/WCAG22/Understanding/status-messages.html).

web/app/components/main-nav/components/web-apps-section.tsx
```diff
@@ -221,7 +221,12 @@ const WebAppsSectionContent = () => {
                 </div>
               )}
               {!installedAppsQuery.isError && installedApps.length === 0 && (
-                <div className="px-2 py-1 system-xs-regular">
+                <div
+                  className="px-2 py-1 system-xs-regular"
+                  role="status"
```
lyzno1: **[P2] Coordinate the no-results status with refetching**

This status node is also created only when the list is already empty, while the region's current `aria-busy` tracks only `isFetchingNextPage`, not search/refetch activity. Keep a stable status element in the region, update its content after the search query settles, and set the result region's busy state from the query's actual fetching state. This avoids announcing `No results` from stale/transitioning data and gives assistive technology one consistent update owner. `role=status` already supplies polite/atomic defaults. See [MDN status role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/status_role).

web/app/components/apps/import-from-marketplace-template-modal.tsx
```diff
@@ -73,22 +73,32 @@ const ImportFromMarketplaceTemplateModal = ({
...
+              </button>
+            }
+          />
         </div>
 
-        <div className="px-6 py-4">
+        <div className="px-6 py-4" aria-busy={isLoading}>
           {isLoading && (
-            <div className="flex items-center justify-center py-8">
+            <div className="flex items-center justify-center py-8" role="status" aria-live="polite">
```
lyzno1: **[P3] Localize the newly announced loading status**

Adding `role=status` makes the existing hard-coded English `Loading...` an assistive-technology announcement in every locale. Use the existing common loading translation instead, and use the single ellipsis character in any new copy. `role=status` already has implicit `aria-live=polite`, so the explicit duplicate is unnecessary.

lyzno1 (changes requested):
I reviewed the full PR against the Dify UI selection/overlay contracts, the current Base UI Combobox anatomy, WAI-ARIA APG, MDN ARIA guidance, and WCAG 2.2. The semantic list/headings, decorative app icons, dialog title/close control, contextual tag labels, and sort-label spacing are valid improvements. The inline findings below are the changes needed before merge; in particular, the Creators filter can adopt the correct Combobox structure now and does not need to wait for Base UI's future third Item generic.

lyzno1 (dismissed):
LGTM

lyzno1 (approved):

lyzno1 (approved):
LGTM

PR #41397, opened 2026-08-27
perf(web): virtualize the Web Apps sidebar list again

### LINKED ISSUE #41396: Web Apps sidebar keeps every loaded row in the DOM since #39739, degrading navigation on large workspaces
### Self Checks

### Dify version

1.17.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Use a workspace with a large number of installed apps in the Web Apps sidebar (ours has 427).
2. Scroll the `WEB APPS` section in the left navigation so infinite scroll loads several pages.
3. Click through the top-level nav entries (Home / Studio / Agents / Knowledge / Skills / Integrations / Marketplace).

`web/app/components/main-nav/components/web-apps-section.tsx` rendered this list through `useVirtualizer` with a `virtualizationThreshold` of 50 rows until #39739 ("chore: paginate installed apps") replaced it with cursor pagination and dropped the virtualizer. Every loaded page now stays in the DOM. Because the section lives in the main nav, those rows are not unmounted on navigation, so they are re-rendered and repainted on every route change.

This is a regression against 1.16.x, which does not reproduce for us.

Measurements from a Safari Web Inspector timeline recorded while clicking through the top-level menus (3.06s):

| Metric | Value |
| --- | --- |
| Rendering frames | 23 in 3.06s (~7.5 fps), worst frame 976ms |
| `composite` | 1067.7ms total, 35% of the recording |
| `paint` | 12,112 records totaling 34.5ms (highly fragmented) |
| All script categories | ~830ms combined |
| Navigation RSC payloads | 340 B – 3.2 KB each, 12–58ms |

So the cost is in painting and compositing, not in JavaScript or in loading.

Supporting in-page A/B/A experiments on the same deployment (frame pacing under a fixed scroll workload, baseline re-measured first and last to detect drift):

| Change | fps |
| --- | --- |
| baseline | 14.2 |
| `mask-image: none` on all elements | 20.2 (+42%) |
| `content-visibility: auto` on cards | 14.9 (noise) |
| `box-shadow: none` | 15.0 (noise) |
| `border-radius: 0` | 14.9 (noise) |
| `filter: none` | 14.2 (noise) |
| `transition/animation: none` | 13.9 (noise) |
| baseline again | 14.2 |

To be precise about which icon that is, since it matters for any follow-up: the app icon on each row is an `<em-emoji>` or an `` rendered by `AppIcon`, not a mask icon. The `mask-image` icon on every row is the `ItemOperation` trigger, `i-ri-more-fill`, which is mounted for each row and kept at `opacity-0` until the row is hovered. So each un-virtualized row paints a mask icon that is never visible. Cutting the row count addresses this; not painting a permanently invisible icon per row would be a separate, independent improvement.

One caution for anyone trying `backdrop-filter` as a lever: disabling it made things about 2x worse in the same harness, because it removes beneficial layer promotion.

### ✔️ Expected Behavior

The Web Apps sidebar keeps only the visible rows (plus overscan) in the DOM once the list grows past a threshold, as it did before #39739, so navigation cost does not grow with the number of installed apps.

### ❌ Actual Behavior

Every loaded page of installed apps stays mounted in the main nav and is repainted on each route change. Switching between top-level menu entries drops to ~7.5 fps on a workspace with 427 installed apps.

BenjaminX:
Fixes #41396

## Summary

`web/app/components/main-nav/components/web-apps-section.tsx` rendered the installed-app list through `useVirtualizer` with a `virtualizationThreshold` of 50 rows. #39739 ("chore: paginate installed apps") moved the section to cursor pagination and dropped the virtualizer along the way, so every loaded page now stays in the DOM. The section lives in the main nav, so those rows survive route changes and are repainted on every navigation.

On a workspace with 427 installed apps, a Safari Web Inspector timeline recorded while clicking through the top-level nav entries spends 1067.7ms of 3.06s in `composite` (35%) across 12,112 `paint` records, and renders 23 frames — about 7.5 fps. All script categories together account for only ~830ms, and navigation RSC payloads are 340 B – 3.2 KB, so the cost is in painting and compositing rather than in JavaScript or loading. Controlled A/B/A measurements on that deployment put CSS `mask-image` icons as the largest single paint factor (+42% fps when masks are disabled, with `box-shadow`, `border-radius`, `filter`, `transition` and `content-visibility` all within noise), and every row here carries one. The un-virtualized rows multiply exactly that cost. Full numbers are in the issue.

This change restores the row model and virtualizer that existed before #39739, on top of the current infinite-query data source:

- The pinned/unpinned `Divider` is folded back into a `WebAppListRow` union, so the separator is a virtual row with its own `estimateSize` rather than something rendered inline between rows.
- The list is now always virtualized. The old threshold is not restored, so there is a single render path instead of a branch between flow layout and virtual rows — one fewer branch to lose next time. Worth noting the threshold protected the common case rather than the expensive one: pagination loads 20 rows at a time, so a list only crossed 50 after the user scrolled the sidebar. The reason to drop it is the simpler single path, not the row count.
- `InfiniteScrollSentinel` is untouched. The virtual container is sized to the full list, so the sentinel still sits below it and preloads exactly as before.

### Which icon, precisely

Since it matters for any follow-up: the app icon on each row is an `<em-emoji>` or an `` rendered by `AppIcon`, not a mask icon. The `mask-image` icon on every row is the `ItemOperation` trigger, `i-ri-more-fill`, which is mounted per row and kept at `opacity-0` until the row is hovered. Each un-virtualized row therefore paints a mask icon that is never visible. This PR cuts the row count; not painting a permanently invisible icon per row is a separate improvement and is not attempted here.

On the suggestion of moving these icons from `mask-image` to `background-image`: the generated stylesheet already splits the two, 426 monochrome icons on `mask-image` and 16 multi-color icons on `background-image`. The `mask-image` path is what lets monochrome icons inherit `currentColor` — `i-ri-more-fill` is coloured through `text-*` tokens here — so moving them to `background-image` would lose theming. The alternative that keeps theming is inline `<svg fill="currentColor">`, which runs against `web/CLAUDE.md`'s instruction not to add generated React icons under `app/components/base/icons/src/`. I also have not measured `background-image` to be cheaper to paint than `mask-image`, so I would not want to assume that without a benchmark. Happy to open a separate issue for it.

## Screenshots

| Before | After |
| ------ | ----- |
| N/A | N/A |

This change is visually identical — it only affects how many rows are mounted and painted. I have not attached a capture of the reporting workspace because the sidebar shows internal application names. The measurements above and in the issue stand in for a visual diff.

## Checklist

Notes on the checklist: this is a frontend-only change, so the backend commands do not apply. Two cases were added to `web/app/components/main-nav/__tests__/index.spec.tsx`, one covering that the list is virtualized and one pinning the separator row's own measurement, and both were verified to fail without the component change. The file needs `vi.mock('@tanstack/react-virtual')` — matching the four existing specs that do the same — because the real virtualizer measures a zero-height scroll element under jsdom and renders no rows. `app/components/main-nav` is green at 5 files / 100 tests. Four spec files unrelated to this change are already failing on `main` (for example `features/agent-v2/agent-detail/access/components/__tests__/access-surface-cards.spec.tsx`); I confirmed they fail identically with this change stashed.

From Claude Code

### DIFF
--- a/web/app/components/main-nav/__tests__/index.spec.tsx
+++ b/web/app/components/main-nav/__tests__/index.spec.tsx
@@ -1373,25 +1373,17 @@ describe('MainNav', () => {
     expect(screen.queryByText('common.mainNav.workspace.inviteMembers')).not.toBeInTheDocument()
   })
 
-  const createInstalledAppList = (count: number) =>
+  const createInstalledAppList = (count: number, pinnedCount = 0) =>
     Array.from({ length: count }, (_, index) =>
       createInstalledApp({
         id: `installed-${index}`,
+        is_pinned: index < pinnedCount,
         app: { ...createInstalledApp().app, id: `app-${index}`, name: `App ${index}` },
       }),
     )
 
-  it('renders the installed web app list in flow layout below the virtualization threshold', async () => {
-    mockInstalledApps = createInstalledAppList(50)
-
-    renderMainNav()
-
-    const firstRow = await screen.findByText('App 0')
-    expect(firstRow.closest('div.absolute')).toBeNull()
-  })
-
-  it('virtualizes the installed web app list above the virtualization threshold', async () => {
-    mockInstalledApps = createInstalledAppList(51)
+  it('virtualizes the installed web app list', async () => {
+    mockInstalledApps = createInstalledAppList(3)
 
     renderMainNav()
 
@@ -1399,10 +1391,21 @@ describe('MainNav', () => {
     const virtualRow = firstRow.closest('div.absolute')
     expect(virtualRow).not.toBeNull()
 
-    // Rows are absolutely positioned inside a container sized to the full list, so
-    // off-screen entries stay out of the layout and paint path.
-    const virtualContainer = virtualRow!.parentElement
-    expect(virtualContainer).toHaveStyle({ height: '1632px' })
+    // Rows are absolutely positioned inside a container sized to the whole list, so
+    // entries outside the viewport stay out of the layout and paint path.
+    expect(virtualRow!.parentElement).toHaveStyle({ height: '96px' })
+  })
+
+  it('measures the pinned separator as its own virtual row', async () => {
+    mockInstalledApps = createInstalledAppList(4, 2)
+
+    renderMainNav()
+
+    const firstRow = await screen.findByText('App 0')
+    expect(screen.getByTestId('divider')).toBeInTheDocument()
+
+    // Four app rows at 32px plus one 17px separator row.
+    expect(firstRow.closest('div.absolute')!.parentElement).toHaveStyle({ height: '145px' })
   })
 
   it('searches installed web apps and renders the matching navigation link', async () => {
--- a/web/app/components/main-nav/__tests__/index.spec.tsx
+++ b/web/app/components/main-nav/__tests__/index.spec.tsx
@@ -1373,41 +1373,6 @@ describe('MainNav', () => {
     expect(screen.queryByText('common.mainNav.workspace.inviteMembers')).not.toBeInTheDocument()
   })
 
-  const createInstalledAppList = (count: number, pinnedCount = 0) =>
-    Array.from({ length: count }, (_, index) =>
-      createInstalledApp({
-        id: `installed-${index}`,
-        is_pinned: index < pinnedCount,
-        app: { ...createInstalledApp().app, id: `app-${index}`, name: `App ${index}` },
-      }),
-    )
-
-  it('virtualizes the installed web app list', async () => {
-    mockInstalledApps = createInstalledAppList(3)
-
-    renderMainNav()
-
-    const firstRow = await screen.findByText('App 0')
-    const virtualRow = firstRow.closest('div.absolute')
-    expect(virtualRow).not.toBeNull()
-
-    // Rows are absolutely positioned inside a container sized to the whole list, so
-    // entries outside the viewport stay out of the layout and paint path.
-    expect(virtualRow!.parentElement).toHaveStyle({ height: '96px' })
-  })
-
-  it('measures the pinned separator as its own virtual row', async () => {
-    mockInstalledApps = createInstalledAppList(4, 2)
-
-    renderMainNav()
-
-    const firstRow = await screen.findByText('App 0')
-    expect(screen.getByTestId('divider')).toBeInTheDocument()
-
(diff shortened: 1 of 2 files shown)


BenjaminX:
Thanks for the analysis — pushed `204b096871` in response. Point by point:

**Separator row.** Already handled in the first commit, in the form you suggested: `WebAppListRow` is a union of `'app'` and `'separator'` entries and `estimateSize` returns `appNavSeparatorHeight` for the latter, rather than the inline `index === pinnedAppsCount - 1` that this PR removes. I've added a test that pins the measurement (four 32px rows plus one 17px separator = 145px), since it is the easy part to get wrong.

**Virtualization threshold.** Dropped, so there is now a single render path. One correction to the reasoning though: loading 20 at a time means a list only crossed the old threshold of 50 *after* the user scrolled the sidebar, so the threshold was protecting the common case rather than the expensive one. The good reason to drop it is the simpler single path — one fewer branch to lose next time — which is what I've gone with.

**`mask-image` → `background-image`.** I don't think this works as stated. The generated stylesheet already splits the two: 426 monochrome icons on `mask-image`, 16 multi-color ones on `background-image`. The mask path is what lets monochrome icons inherit `currentColor`, and the icon in question here is coloured through `text-*` tokens, so moving it to `background-image` would lose theming. Inline `<svg fill="currentColor">` would keep theming but runs against `web/CLAUDE.md`'s instruction not to add generated React icons under `app/components/base/icons/src/`. I also haven't measured `background-image` to actually be cheaper to paint, so I'd rather not assume it. Happy to open a separate issue if you want it explored.

While checking that, I found I'd been imprecise in the description and have corrected it in both the issue and this PR: the per-row `mask-image` icon is not the app icon (that's `<em-emoji>`/`` via `AppIcon`) — it's the `ItemOperation` trigger `i-ri-more-fill`, mounted for every row and held at `opacity-0` until hover. So each un-virtualized row paints an icon that is never visible. Cutting the row count addresses that here; not painting an invisible icon per row is a separate improvement I haven't attempted.

**`content-visibility: auto`.** Agreed, and your framing is better than mine — it skips rendering while keeping the node and its layout box, so it was never going to substitute for not mounting the rows. That also resolves a caveat I had, since I couldn't rule out that the null result was just Safari version support.

From Claude Code

lyzno1 (approved):

lyzno1:
Thanks!

PR #41400, opened 2026-08-27
fix(dataset): delete segment attachment blobs from storage

### LINKED ISSUE #41399: Deleting a single multimodal knowledge segment leaves its attachment object in storage
### Self Checks

### Dify version

1.17.0

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

1. Configure a self-hosted instance with local storage, for example:
   ```env
   STORAGE_TYPE=local
   STORAGE_LOCAL_PATH=/tmp/dify-storage
   ```
2. Start the API and the dataset Celery worker.
3. Create a multimodal knowledge dataset.
4. Add a document and wait until indexing is completed.
5. In the document's segment editor, add an image attachment to one segment and save it. Wait until the segment is indexed successfully.
6. Before deleting the segment, record its attachment ID and storage key from the database:
   ```sql
   SELECT sab.segment_id, sab.attachment_id, uf.key
   FROM segment_attachment_bindings AS sab
   JOIN upload_files AS uf ON uf.id = sab.attachment_id
   WHERE sab.segment_id = '<segment-id>';
   ```
7. Delete only that segment from the Console, or call:
   ```http
   DELETE /console/api/datasets/<dataset-id>/documents/<document-id>/segments/<segment-id>
   ```
8. The request returns `204 No Content`. Wait for `delete_segment_from_index_task` to finish.
9. Verify the database and storage separately:
   ```sql
   SELECT COUNT(*)
   FROM segment_attachment_bindings
   WHERE attachment_id = '<attachment-id>';

   SELECT COUNT(*)
   FROM upload_files
   WHERE id = '<attachment-id>';
   ```
   Both counts are `0`.

   Then check the previously recorded object key:
   ```bash
   test -e "/tmp/dify-storage/<recorded-key>" \
     && echo "attachment object still exists"
   ```
   For S3/OSS or another object store, use the provider's equivalent `head`/`stat` command with the recorded key.

### ✔️ Expected Behavior

Deleting a single indexed segment from a multimodal dataset should remove:

- the segment attachment binding;
- the corresponding `UploadFile` record;
- the attachment vector entry; and
- the physical attachment object from the configured storage backend.

The deletion should use the same storage cleanup behavior as document and batch-document deletion. A storage backend failure should be logged without undoing database cleanup.

### ❌ Actual Behavior

There is no traceback or user-visible error. The delete request returns `204 No Content`, and the Celery task logs normal completion.

However, after the task finishes:

- the `SegmentAttachmentBinding` row is gone;
- the `UploadFile` row is gone; but
- the physical object at the recorded storage key still exists.

The multimodal branch of `delete_segment_from_index_task` deletes the database rows but never calls the configured storage backend. Because the database row is removed first, Dify no longer has a normal application-level reference to clean up the orphaned object.

This occurs on the normal single-segment deletion path for indexed segments with image attachments. It is separate from #41255/#41256, which covers batch document deletion.

CoralGarden52:
## Summary

Fixes #41399.

Deleting an indexed segment from a multimodal dataset removed its `SegmentAttachmentBinding` and `UploadFile` rows, but left the physical attachment object in storage. This change:

- captures attachment storage keys before deleting the database rows;
- deletes the objects through the configured storage abstraction after the database commit;
- scopes attachment reads and deletes to the dataset tenant, dataset, and document;
- logs storage deletion failures without undoing successful database cleanup.

No dependency or documentation changes are required.

## Screenshots

Not applicable (backend-only change).

## Checklist

- [ ] This change requires a documentation update.
- [x] I've added tests for the introduced behavior and kept the change atomic.
- [ ] I ran the full `make lint && make type-check` suite.

Targeted verification:
- `api/.venv/bin/pytest -o addopts='' -q api/tests/unit_tests/tasks/test_segment_index_cleanup_tasks.py` (5 passed)
- `api/.venv/bin/pytest -o addopts='' -q api/tests/unit_tests/services/test_dataset_service_segment.py` (42 passed)
- `ruff check`, `ruff format --check`, and `git diff --check`

From Codex

### DIFF
--- a/api/tasks/delete_segment_from_index_task.py
+++ b/api/tasks/delete_segment_from_index_task.py
@@ -7,6 +7,7 @@ from sqlalchemy import delete, select
 
 from core.db.session_factory import session_factory
 from core.rag.index_processor.index_processor_factory import IndexProcessorFactory
+from extensions.ext_storage import storage
 from models.dataset import Dataset, Document, SegmentAttachmentBinding
 from models.model import UploadFile
 
@@ -63,10 +64,25 @@ def delete_segment_from_index_task(
             if dataset.is_multimodal:
                 # delete segment attachment binding
                 segment_attachment_bindings = session.scalars(
-                    select(SegmentAttachmentBinding).where(SegmentAttachmentBinding.segment_id.in_(segment_ids))
+                    select(SegmentAttachmentBinding).where(
+                        SegmentAttachmentBinding.tenant_id == dataset.tenant_id,
+                        SegmentAttachmentBinding.dataset_id == dataset.id,
+                        SegmentAttachmentBinding.document_id == document_id,
+                        SegmentAttachmentBinding.segment_id.in_(segment_ids),
+                    )
                 ).all()
                 if segment_attachment_bindings:
                     attachment_ids = [binding.attachment_id for binding in segment_attachment_bindings]
+                    attachment_storage_keys = list(
+                        dict.fromkeys(
+                            session.scalars(
+                                select(UploadFile.key).where(
+                                    UploadFile.tenant_id == dataset.tenant_id,
+                                    UploadFile.id.in_(attachment_ids),
+                                )
+                            ).all()
+                        )
+                    )
                     index_processor.clean(
                         session=session, dataset=dataset, node_ids=attachment_ids, with_keywords=False
                     )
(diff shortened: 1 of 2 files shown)


FFXN:
Hi, @CoralGarden52 . Thanks for your fix. 
`attachment_id` is not unique in `segment_attachment_bindings`, so the same `UploadFile` can be bound to multiple segments. This can also happen during normal document indexing when the same image reference appears in multiple chunks, for example because of chunk overlap.

The current implementation collects every attachment referenced by the deleted segment and then deletes its vector, `UploadFile` row, and storage object without checking for bindings from other segments. Deleting one segment could therefore break image retrieval for another segment that still references the same attachment.

Please delete the target segment's bindings first, but only remove the attachment vector, `UploadFile`, and storage object for attachment IDs that have no remaining `SegmentAttachmentBinding` references. It would also be good to add a regression test where two segments share one attachment and deleting one segment preserves the shared attachment.

CoralGarden52:
Hi @FFXN, thanks for catching this.

Fixed in commit 397e072662. The cleanup now deletes the target segment bindings first, then checks whether each attachment ID still has any SegmentAttachmentBinding references. Vector entries, UploadFile rows, and storage objects are removed only for attachments with no remaining bindings.

I also added a regression test covering two segments sharing one attachment. Deleting one segment now preserves the shared attachment and skips its vector and storage cleanup.

Targeted tests pass.

FFXN (approved):

PR #41413, opened 2026-08-27
perf(web): defer card link prefetching until the user shows intent

### LINKED ISSUE #41412: Card lists prefetch one app route per visible card, fanning out into server renders
### Self Checks

### Dify version

1.17.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Use a workspace with a large number of apps. Ours has 432.
2. Open Studio.
3. Watch the reverse proxy access log, or the Network panel.

App Router `<Link>` prefetches every link that enters the viewport, and each prefetch of an app route costs a server render on the Next server. The Studio, Skills and Snippets cards never opted out, so opening one of these lists prefetches one app route per visible card at once.

From our proxy access log, in a single second on Studio:

| Requests in one second | Count |
| --- | --- |
| RSC prefetch `/app/<id>/workflow` | 42 |
| RSC prefetch `/app/<id>/configuration` | 14 |
| RSC prefetch `/app/<id>/access-point` | 8 |
| `GET /console/api/system-features` (User-Agent `node`, i.e. server-side) | 59 |

The `system-features` calls are the fan-out: each prefetched route is rendered on the server, and each of those renders calls it.

#40133 already fixed exactly this for the installed-app sidebar by deferring prefetch until hover or focus, but the card links were not covered.

One caveat for anyone fixing this: `prefetch={false}` alone is not the right lever, because in the App Router it disables the hover prefetch as well as the viewport one, so the route the user is actually heading to would no longer warm up. The sidebar's approach — start at `false`, switch back to the default on hover or focus — keeps that.

### ✔️ Expected Behavior

Opening a card list prefetches the route a user shows intent toward, not one route per visible card, so the number of server renders does not scale with how many cards happen to be on screen.

### ❌ Actual Behavior

Every visible card prefetches its app route as soon as it enters the viewport. On a workspace with 432 apps this produces bursts of ~60 RSC prefetches and ~60 server-side `system-features` calls per second while browsing Studio.

We also see menu clicks that appear not to register until clicked a second time on the same lists, which would be consistent with a navigation queued behind this burst, but we have not established that causally and are not claiming it here.

BenjaminX:
Fixes #41412

## Summary

App Router links prefetch as soon as they enter the viewport, and each prefetch of an app route costs a server render. The Studio, Skills and Snippets cards never opted out, so a list of cards prefetches one route per visible card at once.

Production logs from a workspace with 432 apps show what that costs. In a single second on Studio: 42 RSC prefetches for `/app/<id>/workflow`, 14 for `/configuration`, 8 for `/access-point`, and 59 server-side `GET /console/api/system-features` calls that those renders fan out into.

#40133 already solved this for the installed-app sidebar and the Continue Work home cards, but the card lists were missed. This extracts that fix into `usePrefetchOnIntent` and applies it to every card link, including both places that had hand-written copies of it, so there is one way to do this rather than a pattern to remember at each new call site. A sweep for `prefetch={` now returns no remaining copies.

`prefetch={false}` alone would be the wrong lever here. In the App Router it disables the hover prefetch as well as the viewport one, so the route a user is actually heading to would no longer warm up. The hook starts at `false` and flips back to the default (`null`) on hover or focus, which is what the sidebar already did — the tests assert both halves of that so the distinction cannot be lost in a later refactor.

Knowledge cards are not included: they navigate without a `<Link>`, so they never prefetched.

## Screenshots

| Before | After |
| ------ | ----- |
| N/A | N/A |

No visual change — this only affects when a link decides to prefetch. The request counts above stand in for a visual diff.

## Checklist

Notes on the checklist: this is a frontend-only change, so the backend commands do not apply. `use-prefetch-on-intent.spec.ts` covers the hook, including that hover restores the default rather than leaving prefetching off, and `app-card.spec.tsx` gains a case asserting the card link does not prefetch until it is hovered; both were verified to fail without the change. That card spec needed a pass-through `@/next/link` mock to observe `prefetch`, since the real component does not surface it in the DOM — I confirmed the 75 existing cases in that file still pass with the mock in place before adding the new one. The full unit run matches `main`: the same four spec files fail there before and after this change (for example `features/agent-v2/agent-detail/access/components/__tests__/access-surface-cards.spec.tsx`), with the pass count moving 23372 → 23377.

From Claude Code

### DIFF
--- a/web/app/components/apps/__tests__/app-card.spec.tsx
+++ b/web/app/components/apps/__tests__/app-card.spec.tsx
@@ -26,13 +26,14 @@ const mockWorkflowAppDslExport = vi.hoisted(() => ({
   isExporting: false,
 }))
 const mockCopyApp = vi.hoisted(() =>
-  vi.fn((_variables: unknown): Promise<unknown> =>
-    Promise.resolve({
-      id: 'new-app-id',
-      mode: 'chat',
-      maintainer: 'user-1',
-      permission_keys: [],
-    }),
+  vi.fn(
+    (_variables: unknown): Promise<unknown> =>
+      Promise.resolve({
+        id: 'new-app-id',
+        mode: 'chat',
+        maintainer: 'user-1',
+        permission_keys: [],
+      }),
   ),
 )
 const mockUpdateAppMutation = vi.hoisted(() =>
@@ -47,6 +48,23 @@ const mockStarAppMutation = vi.hoisted(() =>
 const mockUnstarAppMutation = vi.hoisted(() =>
   vi.fn((_variables: unknown): Promise<unknown> => Promise.resolve()),
 )
+
+vi.mock('@/next/link', () => ({
+  default: ({
+    children,
+    href,
+    prefetch,
+    ...rest
+  }: {
+    children?: React.ReactNode
+    href: string
+    prefetch?: boolean | 'auto' | null
+  } & Record<string, unknown>) => (
+    <a href={href} data-prefetch={String(prefetch)} {...rest}>
+      {children}
+    </a>
+  ),
+}))
 
 vi.mock('@/service/client', async (importOriginal) => {
   const actual = await importOriginal<typeof import('@/service/client')>()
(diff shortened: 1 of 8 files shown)


lyzno1:
Thanks for the contribution. Closing this together with #41412.

The current evidence is limited to request counts observed in Studio and does not establish a performance regression that justifies changing the default behavior across multiple card surfaces. The PR also broadens the change to Skills, Snippets, Continue Work, and Installed Apps and extracts a shared hook before a product-wide contract has been established.

If we revisit this optimization with a defined performance target and measurable impact, the implementation should remain scoped to the surface where the problem is demonstrated, without introducing a shared abstraction unless there is a proven cross-surface requirement.

PR #41415, opened 2026-08-27
fix(web): stop feature flags reverting to disabled while unknown

### LINKED ISSUE #41414: Skills nav entry disappears and reappears because an unresolved feature flag reads as disabled
### Self Checks

### Dify version

1.17.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Run with `ENABLE_SKILL` on, so the Skills entry is present in the main navigation.
2. Click the Skills entry in the left navigation.
3. Watch that entry, not the page.

The whole row — icon and label — disappears for a moment and then comes back. Neighbouring entries do not do this.

The nav entry is gated on `skillEnabled && !isCurrentWorkspaceDatasetOperator` (`web/app/components/main-nav/routes.ts`), and `skillEnabled` comes from the provider context:

```ts
// web/context/provider-context-provider.tsx
const enableSkill = features?.enable_skill ?? false
```

`features` is `useQuery(consoleQuery.features.get…).data`, so the fallback collapses "not known yet" into "the feature is off". Any moment that query is without data, the entry is filtered out of the nav, and it returns once the data is back.

This is specific to Skills because of an asymmetry in where the flags come from:

| Nav entry | Flag source | While unresolved |
| --- | --- | --- |
| Marketplace | `systemFeatures.enable_marketplace`, a `useSuspenseQuery` that is prefetched on the server | Always resolved before the nav renders |
| Skills | `enableSkill`, from a client-side `useQuery` in the provider context | Reads as disabled, so the row is hidden |

Skills is the only nav entry whose flag can be in an unknown state, and it is the only one that flickers. Every other flag derived in that provider (`enableBilling`, `enableEducationPlan`, `enableReplaceWebAppLogo`, …) has the same fallback, so anything gated on them can flicker the same way.

For what it is worth, `enable_skill` is set from `dify_config.ENABLE_SKILL` (`api/services/feature_service.py`), i.e. it is a system-level value served on the per-workspace features endpoint. Moving it to `system-features` would put it on the prefetched path and side-step this entirely, but that is an API contract change and not what this issue is asking for.

### ✔️ Expected Behavior

A nav entry for an enabled feature stays visible. A flag that is not yet known should not be presented as a flag that is known to be off.

### ❌ Actual Behavior

The Skills nav entry disappears and reappears whenever the workspace features query is momentarily without data.

BenjaminX:
Fixes #41414

## Summary

Every flag the provider context derives falls back to a disabled default:

```ts
const enableSkill = features?.enable_skill ?? false
```

`features` is a client-side `useQuery`, so that fallback collapses "not known yet" into "the feature is off", and anything gated on one of these flags disappears until the query resolves again.

The visible case is the Skills entry in the main nav. It is gated on `enableSkill`, while its neighbours are gated on `system-features` — a suspense query prefetched on the server, so always resolved by the time the nav renders. Skills is the only entry whose flag can be in an unknown state, and it is the only one that flickers. Details and the comparison table are in the issue.

This holds on to the last answer instead, so a feature known to be enabled never silently un-knows itself. It covers every flag derived in that provider rather than only the one that surfaced the problem, and it does not depend on enumerating every way the query can end up without data — which matters, because I was not able to pin down the specific trigger. What I could establish is that the usual suspects are not it: `refreshFeatures` uses `invalidateQueries`, which keeps data; nothing calls `removeQueries` or `resetQueries` against this key; there is a single `ProviderContextProvider` instance, so it is not a second provider mounting empty. The one signal I do have from production logs is a `401` on `/console/api/features` with a `/skills` referer.

Guarding the derivation is the right layer regardless of the trigger: a fallback that cannot distinguish "off" from "unknown" will keep producing this class of bug whatever opens the gap.

Trade-off worth calling out: once a feature set is known, a later gap now shows the previous values rather than defaults. That is the intent, and workspace switches reload the console, so the retained values do not outlive the workspace they belong to.

## Screenshots

| Before | After |
| ------ | ----- |
| N/A | N/A |

The change is a flicker fix — a still frame does not show it, and the reproduction is described in the issue.

## Checklist

Notes on the checklist: this is a frontend-only change, so the backend commands do not apply. The new `provider-context-provider.spec.tsx` covers both halves — a flag still reads as disabled before anything is known, and a known flag survives the query going empty — and the second case was verified to fail without the change. The full unit run matches `main`: the same four spec files fail there before and after (for example `features/agent-v2/agent-detail/access/components/__tests__/access-surface-cards.spec.tsx`), with the pass count moving 23372 → 23374.

From Claude Code

### DIFF
--- a/web/context/__tests__/provider-context-provider.spec.tsx
+++ b/web/context/__tests__/provider-context-provider.spec.tsx
@@ -1,0 +1,102 @@
+import { render, screen } from '@testing-library/react'
+import { useProviderContextSelector } from '../provider-context'
+import { ProviderContextProvider } from '../provider-context-provider'
+
+const mockFeaturesQuery = vi.hoisted(() => vi.fn())
+
+vi.mock('@tanstack/react-query', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@tanstack/react-query')>()
+  return {
+    ...actual,
+    useQuery: (options: { queryKey?: unknown[] }) =>
+      String(options.queryKey?.[0]).includes('features')
+        ? mockFeaturesQuery()
+        : { data: undefined, isLoading: false, isSuccess: false, isFetched: false },
+    useQueryClient: () => ({ invalidateQueries: vi.fn() }),
+  }
+})
+
+vi.mock('@/service/client', () => ({
+  consoleQuery: {
+    features: {
+      get: { queryOptions: () => ({ queryKey: ['features'] }), key: () => ['features'] },
+    },
+    workspaces: {
+      current: {
+        modelProviders: { summary: { get: { queryOptions: () => ({ queryKey: ['providers'] }) } } },
+      },
+    },
+  },
+}))
+
+vi.mock('@/service/use-common', () => ({
+  commonQueryKeys: { modelList: () => ['model-list'] },
+  useModelListByType: () => ({ data: [] }),
+  useSupportRetrievalMethods: () => ({ data: undefined }),
+}))
+
+vi.mock('@/app/components/base/zendesk/utils', () => ({ setZendeskConversationFields: vi.fn() }))
+
+vi.mock('jotai', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('jotai')>()
+  return { ...actual, useAtomValue: () => undefined }
+})
+
+const SkillFlag = () => {
+  const enableSkill = useProviderContextSelector((state) => state.enableSkill)
+
+  return <span data-testid="enable-skill">{String(enableSkill)}</span>
+}
+
+const renderProvider = () =>
+  render(
+    <ProviderContextProvider>
+      <SkillFlag />
+    </ProviderContextProvider>,
+  )
+
+const featuresResult = (data: unknown) => ({
+  data,
+  isLoading: false,
+  isSuccess: !!data,
+  isFetched: !!data,
+})
+
+const enabledFeatures = {
+  billing: { enabled: false },
+  education: { enabled: false },
+  enable_skill: true,
+  can_replace_logo: false,
+  model_load_balancing_enabled: false,
+  webapp_copyright_enabled: false,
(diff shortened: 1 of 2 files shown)


BenjaminX:
Closing this: the mechanism I proposed it on does not hold, so the change would not fix the reported flicker.

I had assumed the workspace features query can end up without data mid-session, which combined with `features?.enable_skill ?? false` would hide the Skills entry until the data returned. Testing that assumption with a control disproves it — React Query retains data for a mounted observer, so `data` does not go back to `undefined`:

| | after `removeQueries` | after `resetQueries` |
| --- | --- | --- |
| without `placeholderData: keepPreviousData` | `true` / `success` | `true` / `success` |
| with `placeholderData: keepPreviousData` | `true` / `success` | `true` / `success` |

Both identical, so neither this PR's approach nor `keepPreviousData` (suggested on the issue) changes anything here. I also found the test I added passes against `main` unmodified, which makes it a false guard rather than a regression test — my mistake for not checking that before opening this.

What does still stand, and is worth someone picking up separately: `?? false` genuinely cannot distinguish "disabled" from "not known yet", so on a cold load the Skills entry is hidden and then appears once the query resolves. That is a real, if milder, flicker, and every flag derived in `ProviderContextProvider` shares it. It is just not the reported symptom, which is an entry that is already visible disappearing and coming back.

The reported flicker therefore has a different cause that I have not identified. #41414 is left open with the analysis; the asymmetry it documents between `system-features` (suspense, server-prefetched) and the workspace features query is still accurate, and I have corrected the conclusion there. The next thing to establish is whether only the Skills row vanishes or the whole nav blinks — the latter would point at `MainNav` re-suspending rather than at any feature flag.

CI on this branch also failed, but on an unrelated flake: one case in `features/skills/__tests__/detail-page.spec.tsx` hit the 5s per-test timeout. That file runs 136 passing tests in ~129s locally and took ~360s on CI. Noting it here in case it shows up elsewhere.

From Claude Code

PR #41420, opened 2026-08-27
refactor(openapi): route /openapi/v1 auth by subject

GareArc:
## Subject-open auth for `/openapi/v1`

PR #41311 argued from a security contradiction. This one argues from **ownership**.

Subject knowledge under `/openapi/v1` was smeared across roughly fifteen sites: two enums in
1:1 correspondence (`SubjectType` and `TokenType`), three copies of the token-prefix literals,
`allowed_token_types=` repeated on 27 handlers, and an audit line whose scope list disagrees
with what the mint profile actually issues. This PR replaces the generic step-DSL auth pipeline
with a router, one pipeline per subject, and self-contained requirement classes, and collapses
most of that fan-out onto one owner.

**This is a consolidation win, not an extensibility win.** PAT (`dfp_`) is deferred post-v1.0,
so the subject axis is N=2 and stays N=2 for the foreseeable future. Nobody should merge this
because it makes adding a third subject easy. Merge it because the request path stops repeating
itself: auth becomes `subject × requirements`, evaluated by one guard.

In one sentence: **`@endpoint` declares what a route requires, `Rank` decides what order the
checks run in, and the subject's pipeline supplies the rest.**

### What moved

| | Before | After |
|---|---|---|
| `controllers/openapi/auth/` | 8 files, 751 LOC — `pipeline` / `prepare` / `verify` / `composition` / `data` / `conditions` / `flow` | 7 files, 797 LOC — `subjects` / `context` / `requirements` / `spec` / `pipelines` / `router`, plus `data.py` reduced to `CallerKind` + `ExternalIdentity` |
| Guard declaration | `guard(...)` / `guard_workspace(...)` + `allowed_token_types=` on 27 handlers | one `@endpoint(requirements=(...))` on all 25 routes |
| `TokenType` production references | 49, across 11 files | **5, all in `libs/oauth_bearer.py`** |
| `allowed_token_types=` | 27 sites | **0** |
| Test seam | `__wrapped__` (39 sites), `inspect.unwrap`, `bypass_pipeline` (38) | `view.__handler__`, one documented depth |
| Ordering | implicit in `composition.py:48-55` | `Rank` on each requirement |


api/controllers/openapi/_contract.py
```diff
@@ -84,3 +82,64 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+
+
+def endpoint(
+    *,
+    requirements: Sequence[Requirement] = (),
+    query: type[BaseModel] | None = None,
+    body: type[BaseModel] | None = None,
+    returns: ReturnSpec | Sequence[ReturnSpec] | None = None,
+    edition: frozenset[DeploymentEdition] | None = None,
+    write: bool = True,
```
wylswz: Controller-managed session is a deprecated pattern, so it may be better to leave them out of the contract for now.

Session will be managed and only accessible to repository layer in the future.

api/controllers/openapi/auth/context.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def _loaded[T](value: T | None, name: str) -> T:
+    if value is None:
+        raise _missing(name)
+    return value
+
+
+def _missing(name: str) -> LookupError:
+    return LookupError(f"{name} was not loaded: no requirement on this route asked for it")
```
wylswz: Would it be clearer to use a custom exception here to explicitly represent the contract violation—accessing context that the route’s requirements haven’t loaded?

api/controllers/openapi/auth/spec.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    requirements: tuple[Requirement, ...]
+    edition: frozenset[DeploymentEdition] | None = None
+
+    def __post_init__(self) -> None:
+        """Coerce at construction, not at the first request: a list survives
+        `__init__` untouched and only fails where `Pipeline.run` concatenates.
+        """
+        object.__setattr__(self, "requirements", tuple(self.requirements))
```
wylswz: Could we keep EndpointSpec tuple-only and leave type correctness to its callers? This silently accepts and coerces values that don’t match the declared type, making the constructor’s contract less explicit. The existing callers already pass tuples, and endpoint() normalizes its input before constructing the spec, so this extra coercion seems unnecessary.

wylswz (approved):

PR #41433, opened 2026-08-28
fix(web): give the favicon a single owner in root metadata

### LINKED ISSUE #41432: Branding favicon detaches React's icon link, so the first click on a nav entry is lost
### Self Checks

### Dify version

1.17.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Configure a workspace branding favicon, so `system-features` returns a non-empty `branding.favicon`.
2. Open the console and click any entry in the left navigation.

The first click does nothing. Clicking the same entry again navigates, and that second navigation is a full page load rather than a client-side transition. The console shows, continuously and on each failed click:

```
Uncaught TypeError: can't access property "removeChild", t.stateNode.parentNode is null
```

Reproduced in Safari 26.6.2, Firefox 154 and Chrome 151, so it is not engine-specific.

**Cause.** `web/hooks/use-document-title.ts` removes every icon link from the document before appending its own:

```ts
if (branding.favicon) {
  document
    .querySelectorAll("link[rel='icon'], link[rel='shortcut icon'], link[rel='apple-touch-icon'], link[rel='mask-icon']")
    .forEach((n) => n.parentNode?.removeChild(n))
  ...
}
```

The `<link rel="icon">` on the page is rendered by React — it is present in the RSC flight payload, and React 19 hoists it into the head and keeps a fiber pointing at that node. Detaching it here leaves that fiber holding an element with no parent, so React's next deletion pass throws. Decoding the reported offset in the deployed chunk lands exactly on React's `HostHoistable` (fiber tag 26) deletion branch:

```js
case 26: ... t.stateNode && (t = t.stateNode).parentNode.removeChild(t)
```

The throw happens inside a commit, so it takes down the render that commit belongs to. On a navigation that is the render the router is waiting on, which is why the click appears to do nothing and the retry falls back to a document load.

**Why this is new in 1.17.0.** The removal has been there since at least 1.16.1, and React is 19.2.8 in both, so neither is the change. 1.16.1 also rendered React-owned icon links — written directly in the persistent root layout — so the removal was already detaching React's nodes there. What changed is that moving the asset into `app/` made the link request-keyed file metadata that is replaced during navigation, so the detach now lands on the render the router is waiting on rather than on a link that simply stays put.

Two conditions therefore have to hold to hit this, which is presumably why it has not surfaced more widely: 1.17.0 or later, **and** a configured branding favicon.

### ✔️ Expected Behavior

Clicking a navigation entry navigates on the first click, whether or not a branding favicon is configured.

### ❌ Actual Behavior

With a branding favicon configured, React throws during commit whenever it deletes the icon link it owns. The click that render belonged to is lost, so the entry only responds to a second click, which then does a full page load.

BenjaminX:
Fixes #41432

## Summary

`useDocumentTitle` removed every icon link from the document before appending its own whenever a workspace configured a branding favicon. That detaches a node React still holds a `HostHoistable` fiber for, so React throws while deleting it, inside a commit:

```
Uncaught TypeError: can't access property "removeChild", t.stateNode.parentNode is null
```

The throw takes down the render that commit belongs to. On a navigation that is the render the router is waiting on, so the first click on a nav entry was lost and the retry fell back to a full document load.

Per review, this takes the ownership fix rather than patching the detach:

1. `web/app/favicon.ico` moves to `web/public/favicon.ico`, so it remains the default static asset without outranking `generateMetadata` as file-based metadata.
2. Root `generateMetadata()` returns `Metadata.icons` — the configured favicon when branding is enabled and non-empty, otherwise `${basePath}/favicon.ico`. It already awaits `prefetchSystemFeatures()` for the application title, and branding is deployment-level bootstrap data, so this needs no new request, workspace lookup, or cache change.
3. `useDocumentTitle` loses the favicon calculation, `useFavicon`, and the icon DOM effect. It owns only the client-side title again.

The `apple-touch-icon` the hook used to append is preserved as `icons.apple` on the branded path.

## Correction to the earlier explanation

My first write-up said 1.16.1 matched nothing because the asset lived in `public/`. That was wrong, as @lyzno1 pointed out: 1.16.1 rendered React-owned icon links directly in the persistent root layout, so the removal was already detaching React's nodes there. What changed in 1.17.0 is that moving the asset into `app/` made the link request-keyed file metadata that is replaced during navigation, which is what turns the detach into a crash on the render the router is waiting on. The issue description has been corrected.

## Screenshots

| Before | After |
| ------ | ----- |
| N/A | N/A |

No visual change: the branded favicon is still applied, now server-rendered from metadata rather than swapped in on the client, so it is also correct on first paint instead of after hydration.

## Checklist

Notes on the checklist: frontend-only, so the backend commands do not apply. The icon assertions moved from the hook spec to `app/__tests__/layout.spec.tsx`, covering the branded path, the unbranded fallback, and branding enabled with an empty favicon; all three were verified to fail against `main`. The existing one-fetch/dehydration case in that file is untouched and still asserts `getSystemFeatures` is called once. `vp check` is clean and the unit suite is green at 2110 files / 23384 tests.

One consequence worth naming: `public/favicon.ico` is served without the content hash `app/favicon.ico` produced, so the default icon no longer cache-busts on change. That is the trade for keeping metadata authoritative; say the word if you would rather express the default through `icons` differently.

From Claude Code


web/hooks/use-document-title.ts
```diff
@@ -17,12 +17,11 @@ export default function useDocumentTitle(title: string | null) {
...
-        .querySelectorAll(
-          "link[rel='icon'], link[rel='shortcut icon'], link[rel='apple-touch-icon'], link[rel='mask-icon']",
-        )
-        .forEach((n) => n.parentNode?.removeChild(n))
-
+      // The icon link the framework renders is React's: it hoists it into the head
+      // and keeps a fiber pointing at that node, so detaching it here makes React
+      // throw while deleting an element that no longer has a parent. `useFavicon`
+      // below already repoints that link at the branded URL, so nothing needs
+      // removing for the branded icon to win.
```
lyzno1: [P2] Please move favicon ownership to root metadata.

Removing the selector stops the immediate crash, but this still relies on `ahooks/useFavicon`, whose effect only reruns when `href` changes. Next replaces request-keyed metadata during client navigation, while persistent layouts such as App Detail keep this hook mounted with the same branding URL. The replacement link can therefore remain `/favicon.ico`; I reproduced that branded-to-default transition with this repository React 19.2.8 and ahooks 3.9.7.

Root `generateMetadata()` already has `branding.favicon`. Please emit `Metadata.icons` there, move `app/favicon.ico` to the public fallback, and remove favicon/apple/`useFavicon` ownership from this hook.

web/hooks/use-document-title.spec.ts
```diff
@@ -25,6 +25,30 @@ describe('useDocumentTitle', () => {
...
+    icon.href = '/favicon.ico?build-hash'
+    document.head.appendChild(icon)
+
+    renderHookWithConsoleQuery(() => useDocumentTitle('Settings'), {
+      systemFeatures: {
+        branding: { enabled: true, favicon: 'https://cdn.example.com/brand.png' },
+      },
+    })
+
+    // React renders this link and keeps a fiber pointing at the node. Detaching it
```
lyzno1: This link is created with `document.createElement`, so it is not framework-rendered and has no React HostHoistable fiber. The comment therefore describes behavior this test does not exercise, and the test also misses request-keyed metadata replacement while a layout stays mounted.

After moving ownership to root metadata, please cover the public contract in `app/__tests__/layout.spec.tsx`: configured branding returns the branded icon, missing/disabled branding returns the default icon, and the existing request-local System Features reuse remains intact. The detailed Fiber explanation belongs in the PR or commit message.

lyzno1 (changes requested):
The reported React commit failure is valid, but I am requesting changes because this patch fixes only the immediate detach and leaves favicon ownership split between Next metadata and a client hook.

Root `generateMetadata()` already awaits `prefetchSystemFeatures()` and receives `branding.favicon`; it currently uses the same branding snapshot for the application title. This branding is deployment-level, pre-auth bootstrap data, so the fix does not require a new API, workspace lookup, or request/cache change.

Please keep a single owner:

1. Move `web/app/favicon.ico` to `web/public/favicon.ico` so it remains the default static asset without acting as higher-priority file-based metadata.
2. Return `Metadata.icons` from root `generateMetadata()`: use the configured favicon when branding is enabled and non-empty, otherwise fall back to `${basePath}/favicon.ico`; preserve the existing apple-touch behavior there if required.
3. Remove favicon calculation, `useFavicon`, and the apple/icon DOM effect from `useDocumentTitle`; this hook should only own the client-side title.
4. Replace the hook-level icon test with root metadata tests for branded and default favicon output, while retaining the existing one-fetch/dehydration contract.

There is also an accuracy issue in the explanation: 1.16.1 already rendered React-owned icon links in the persistent root layout. The relevant 1.17 change is that moving the asset into `app/` made it request-keyed file metadata that is replaced during navigation. Please keep the detailed React Fiber explanation in the issue/PR description or commit message rather than production and test comments.

BenjaminX:
Thanks — reworked along all five points, and the correction was needed.

**On the accuracy issue.** You are right and I was wrong. I checked only where the asset lived and concluded 1.16.1 matched nothing; in fact `web/app/layout.tsx` there renders `link[rel=icon]` and `link[rel=apple-touch-icon]` directly in the persistent root layout, so the removal was already detaching React-owned nodes. The distinction is the one you drew: those links stay put, whereas moving the asset into `app/` made it request-keyed file metadata that is replaced during navigation, which is what puts the detach on the render the router is waiting on. Both the issue and the PR description now say that instead.

**What changed:**

1. `web/app/favicon.ico` → `web/public/favicon.ico`.
2. Root `generateMetadata()` returns `Metadata.icons`: the configured favicon when branding is enabled and non-empty, otherwise `${basePath}/favicon.ico`. The apple-touch link the hook used to append is preserved as `icons.apple` on the branded path.
3. `useDocumentTitle` loses the favicon calculation, `useFavicon` and the icon DOM effect — it owns only the client title again.
4. The icon assertions moved to `app/__tests__/layout.spec.tsx`: branded, unbranded fallback, and branding enabled with an empty favicon. All three fail against `main`. The existing one-fetch/dehydration case is untouched and still asserts `getSystemFeatures` is called once.
5. The Fiber detail is out of the code and lives in the commit message and the descriptions.

**One consequence I want to put in front of you rather than bury.** `public/favicon.ico` is served without the content hash `app/favicon.ico` produced, so the default icon no longer cache-busts when it changes — and browsers are stubborn about favicons. It is the same position 1.16.1 was in, and it is the trade for keeping metadata authoritative, but if you would rather express the default some other way so the hash survives, I am happy to redo that part.

From Claude Code

lyzno1 (approved):

PR #41438, opened 2026-08-28
chore: enforce titles on CSS-truncated elements

iamjoel:
## Summary

- Add an autofixable Oxlint rule requiring CSS-truncated JSX content to provide a non-empty title unless covered by an existing Tooltip.
- Apply the rule across Web, including shared controls, inline styles, dynamic class expressions, and CSS Modules.
- Cache CSS Module analysis and keep the full frontend rule runtime to approximately 80.7 ms.

Fixes FPRD-124

## Checklist

- [x] I have added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.


lyzno1:
## Review findings

The product goal is valid: important truncated text should have a reliable full-content path. The current implementation is not ready for LGTM because it turns one possible presentation technique, native `title`, into a repository-wide correctness invariant. The audit found the following concrete problems.

### Blocking rule-design findings

1. **The rule cannot prove that truncation happened.** A static `truncate` or `line-clamp-*` class only means that overflow may occur at runtime. The rule reports even when the rendered text fits, as demonstrated by inputs and short static labels in this diff.

2. **Missing `title` is not an accessibility failure.** CSS truncation does not remove the full text from the DOM or accessibility tree. Native title tooltips are not reliably available to keyboard, touch, or assistive-technology users, so this rule cannot be described as enforcing an accessible full-content path.

3. **The empty-title error rejects valid HTML semantics.** An explicit `title=""` can intentionally prevent an element from inheriting advisory information from an ancestor. Treating it as universally invalid, or deleting it through autofix, can change hover behavior. This is defined by the [HTML Standard](https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute) and documented by [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/title).

4. **The autofix cannot safely choose the disclosure content.** The correct owner may be a Tooltip, PreviewCard, Popover, detail panel, copy action, or the visible element itself. Duplicating a child expression into `title` cannot determine that product contract.

Because the rule is enabled as an error, these are CI-blocking false invariants rather than optional UX suggestions. The current migration spans roughly 390 files and adds more than 600 title props.

### Confirmed consumer regressions

5. **A raw API secret is copied into an additional native tooltip surface.**  
   `web/features/agent-v2/agent-detail/access/components/agent-api-key-modal.tsx:329-337` adds `title={apiKey?.token}` to the newly generated key. This value is intentionally shown once with an explicit copy action; the global lint should not propagate sensitive values into additional advisory attributes.

6. **Unbounded user content produces an unbounded browser tooltip.**  
   `web/app/components/workflow/panel/comments-panel/index.tsx:203-209` adds `title={c.content}` to a three-line comment summary. Comment input has no maximum length, and clicking the card already opens the full thread. Long or multiline comments therefore create a large duplicate tooltip instead of using the existing detail owner.

7. **Native title competes with an existing PreviewCard owner.**  
   `web/app/components/workflow/block-selector/blocks.tsx:151-170` adds a title inside `PreviewCardTrigger`. Hovering the same item can now activate both the browser tooltip and the PreviewCard.

8. **Several added titles are not reachable by pointer hit testing.**  
   `web/features/home/template-card.tsx:65-72,92-146` adds titles to inner text while an absolute `inset-0 z-10` button covers the card. The button owns the interaction and already receives the full name and description through `aria-labelledby` and `aria-describedby`; the covered inner titles do not provide a usable disclosure path.

9. **The PromptEditor title is attached to a non-interactive, non-hit-testable node.**  
   `web/app/components/base/prompt-editor/plugins/placeholder.tsx:20-31` adds a title to an element whose class includes `pointer-events-none`. The browser tooltip cannot reliably be triggered from that element.

10. **Search inputs gain tooltip noise without any overflow condition.**  
    `web/app/components/plugins/marketplace/search-box/index.tsx:122-138` adds `title={search}`. The title changes on every keystroke and is present for short values where no truncation occurs.

11. **Shared date/time controls receive contradictory empty-title state.**  
    `web/app/components/base/date-and-time-picker/date-picker/index.tsx:259-266` sets `title={state.open ? "" : displayValue}`. Opening the picker deliberately creates the empty title that the rule claims is an error, while changing hover behavior for every DatePicker consumer. TimePicker has the same class of change.

12. **The ReactNode-to-string helper does not model rendered output.**  
    `web/utils/react.ts` recursively reads `props.children`, but custom components can render text without children, transform children, or include screen-reader-only content. The helper also misses other valid iterable ReactNode shapes. Titles created through this helper can therefore be missing or disagree with the visible content. It is used by PromptEditor, Markdown directive cards, configuration headings, citations, and publisher actions.

13. **Structured content is flattened into advisory strings.**  
    Examples such as `web/app/components/base/markdown-with-directive/components/with-icon-card-item.tsx:27-32` and `web/app/components/base/chat/chat/citation/popup.tsx:68-82` apply the helper to React content and can also place nested titles on both a container and an interactive child. This creates duplicate or incorrect disclosure owners.

14. **The test changes frequently lock the implementation instead of the user contract.**  
    Added `getByTitle` and `toHaveAttribute("title", ...)` assertions verify that the migration happened, but do not verify actual overflow, hover reachability, keyboard or touch access, or overlay competition. A behavior test should keep assertions such as visible labels, overflow summaries, accessible roles/names, and the actual detail interaction; a title assertion is only appropriate when that title is itself an explicitly accepted product contract.

### Required direction before LGTM

- Do not enable the current missing/empty-title rule as a repository-wide error, and do not keep the mass autofix migration in its current form.
- Define the FPRD-124 acceptance criterion at the product level, then fix only reproduced, feature-owned truncation problems using the existing disclosure owner and behavior-focused tests.

I moved the PR back to draft so these blocking findings can be addressed without changing the owner branch scope. `packages/dify-ui` remains out of scope.

lyzno1 (approved):

PR #41295, opened 2026-08-26
chore: clean away unnecessary-type-conversion (#39947)

Tarulatapriya:
### Description
This PR resolves #39947.

- Cleaned away remaining instances of redundant `bool()` type conversions in `api/repositories/workflow_collaboration_repository.py`. (Note: The instances in `dataset_retriever_tool.py`, `account.py`, and `runtime_feature_manifest.py` were already resolved by other recent commits).
- Promoted the `unnecessary-type-conversion` Pyrefly rule from `"info"` to `"error"` in `api/pyproject.toml` so that CI enforces it going forward.

PR #41454, opened 2026-08-28
chore: remove unnecessary type conversions

lin597410-design:
> [!NOTE]
> Partial cleanup for #39947 — plenty of remaining flagged sites are left for other contributors; I only touched sites that are provably redundant both statically and at runtime.

## Summary

Removes redundant type conversions flagged by pyrefly's `unnecessary-type-conversion` rule (currently demoted to `info` in `api/pyproject.toml`) across 6 files, where the argument's type is already the target type:

- `api/controllers/console/auth/error.py` — drop `int()` around the `minutes: int` parameter in rate-limit error constructors
- `api/controllers/console/workspace/snippets.py` — drop `str()` around `snippet_id: str` route parameters. Also pass `str(snippet_id)` explicitly in `SnippetDetailApi.get`: its route parameter is typed `UUID` while `SnippetService.get_snippet_by_id` expects `str`, and the previously bare argument surfaced as a `bad-argument-type` error once the redundant casts were removed from the sibling endpoints
- `api/controllers/console/app/app.py` — drop `str()` around `current_tenant_id: str` parameters, one `permission_keys_map.get(str(item.id), ...)` (`App.id` is `Mapped[str]`), and one already-`str` OAuth client id
- `api/services/annotation_service.py` — drop `str()` around `app_id: str`, around `job_id = str(uuid.uuid4())` results (f-string cache keys and `.delay(...)` args), and one truthiness-narrowed `raw_message_id`
- `api/services/dataset_service.py` — drop `str()` around `Sequence[str]` values; `[str(x) for x in seq]` becomes `list(seq)`
- `api/core/workflow/generator/runner.py` — drop one `str()` around a `GraphNodeDict.id` lookup

Deliberately **not** touched (load-bearing at runtime despite the annotations):

- defensive coercions of unvalidated LLM output in `runner.py` (`str(plan.get(...) or "")`, `float(viewport.get(...))` — the planner dict is `cast` from raw LLM JSON, so these guards against non-`str`/non-`float` values are load-bearing)
- the pydantic `mode="before"` validators in `app.py`, whose `str()` casts guard against non-`str` runtime input
- `int(dify_config.BATCH_UPLOAD_LIMIT)` in `DocumentService.check_document_creation_limits` — unit tests patch the attribute with strings, so the runtime value is not guaranteed to be `int` despite the `NonNegativeInt` annotation (initially removed, restored in 7b73afee after CI caught it)
- `str(upload_file_id)` in `DocumentService._get_upload_file_id_for_upload_file_document` — `data_source_info` is unvalidated JSON where `upload_file_id` may be an `int`, and the string-coercion contract is pinned by `test_get_upload_file_id_for_upload_file_document_returns_string_id` (initially removed, restored in 7b73afee after CI caught it)

## Verification

With `unnecessary-type-conversion` temporarily promoted to `error` in `api/pyproject.toml` (not part of this PR — that promotion should land once all sites are clean, as attempted in #41295):

- `pyrefly check .` resolves exactly these 31 diagnostics with **0 new errors** introduced (verified by diffing full-run JSON output before/after; the one newly surfaced `bad-argument-type` in `snippets.py` is fixed by the explicit `str(snippet_id)` noted above)
- `ruff check` and `ruff format --check` pass on all changed files (the identity comprehensions left behind by the `str()` removal are rewritten as `list(...)` to satisfy C416)

## Screenshots

N/A (no UI change)

## Checklist

- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. (pure refactor, no observable behavior change)
- [x] I've updated the documentation accordingly. (not required)
- [x] I ran `make lint && make type-check` (backend) and `vp staged` (frontend) to appease the lint gods (ran `ruff check` / `ruff format --check` / `pyrefly check` directly on the changed files)

### DIFF
--- a/api/controllers/console/app/app.py
+++ b/api/controllers/console/app/app.py
@@ -642,13 +642,13 @@ class AppListApi(Resource):
         )
 
         permissions = enterprise_rbac_service.RBACService.MyPermissions.get(
-            str(current_tenant_id),
+            current_tenant_id,
             current_user_id,
             session=session,
         )
         if dify_config.RBAC_ENABLED:
             access_filter = resolve_app_access_filter(
-                str(current_tenant_id),
+                current_tenant_id,
                 current_user_id,
                 session=session,
                 permissions=permissions,
@@ -675,7 +675,7 @@ class AppListApi(Resource):
             pagination_model = pagination_model.model_copy(
                 update={
                     "data": [
-                        item.model_copy(update={"permission_keys": permission_keys_map.get(str(item.id), [])})
+                        item.model_copy(update={"permission_keys": permission_keys_map.get(item.id, [])})
                         for item in pagination_model.data
                     ]
                 }
@@ -712,7 +712,7 @@ class AppListApi(Resource):
         app_service = AppService()
         app = app_service.create_app(current_tenant_id, params, current_user, session=session)
         permission_keys_map = enterprise_rbac_service.RBACService.AppPermissions.batch_get(
-            str(current_tenant_id),
+            current_tenant_id,
             current_user.id,
             [str(app.id)],
             session=session,
@@ -882,7 +882,7 @@ class AppApi(Resource):
             app_model.access_mode = app_setting.access_mode
 
         permissions = enterprise_rbac_service.RBACService.MyPermissions.get(
-            str(current_tenant_id),
+            current_tenant_id,
             current_user.id,
             app_id=str(app_model.id),
             session=session,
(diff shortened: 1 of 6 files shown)


asukaminato0721 (dismissed):

lin597410-design:
Restored two conversions that turned out to be load-bearing at runtime (7b73afee):

- `int(dify_config.BATCH_UPLOAD_LIMIT)` — unit tests patch the config attribute with strings, so the `NonNegativeInt` annotation does not hold at runtime
- `str(upload_file_id)` in `_get_upload_file_id_for_upload_file_document` — `data_source_info` is unvalidated JSON where the id may be an int; the string-coercion contract is pinned by `test_get_upload_file_id_for_upload_file_document_returns_string_id`

The remaining 31 removals stand; pyrefly full-repo diff still reports 0 new errors. PR description updated accordingly.

lin597410-design:
The workflow runs for the new commit (7b73afee) are awaiting approval — could a maintainer approve them so CI can re-run? 🙏

asukaminato0721 (approved):

PR #41463, opened 2026-08-29
fix(api): resolve aliases in remote settings sources

GYJ99:
Fixes #41786

Original report #41439 was closed during triage because its top-level version field referenced an older release. #41786 revalidates the same issue against current `main`; this PR is the existing implementation for that current-version report.

## Summary

- Preserve exact field-name precedence for remote settings.
- Fall back to `validation_alias` and ordered string entries in `AliasChoices` when the field name is absent.
- Apply the same resolution behavior to both Nacos and Apollo sources, returning the matched alias key so aliased internal fields are populated correctly.
- Keep the implementation scoped to flat remote configuration keys; remove unneeded `AliasPath` handling that was not required by the issue or covered by regression tests.
- Add focused regression tests covering direct precedence, a single validation alias, and multiple alias choices for both sources.

## Screenshots

N/A (backend configuration resolution only).

## Validation

- Focused resolver coverage previously exercised the six direct-name / single-alias / `AliasChoices` cases successfully in an isolated harness.
- On current head `ab210b0f`, Main CI Pipeline, Pyrefly Diff Check, Pyrefly Type Coverage, Semantic Pull Request, and `autofix.ci` have passed.
- `Style Check / Python Style` passed on the current head, including `Run Type Checks`.
- The latest style fix removes the banned `typing.Any` annotation from the parameterized regression tests and uses the precise Nacos/Apollo source class union instead.
- Before the latest code edit, `GYJ99/dify:main` was fast-forward synced to upstream `langgenius/dify:main` at `a233b2f5`.
- Full repository pytest, lint, and type-check have not been run locally.

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [ ] I ran `make lint && make type-check` (backend) and `vp staged` (frontend) to appease the lint gods

From ChatGPT

PR #41465, opened 2026-08-29
fix(config): resolve pydantic validation aliases in remote settings sources

### LINKED ISSUE #41439: Nacos remote settings source doesn't match config keys against pydantic field aliases (AliasChoices)
### Self Checks

### Dify version

1.14.2

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Summary
The Nacos remote settings source only looks up config keys by the exact pydantic field name, so any field that defines a validation_alias / AliasChoices never matches the corresponding key stored in Nacos. This silently leaves the config at its default value.

Environment
Dify version: 1.17.0
Remote settings source: REMOTE_SETTINGS_SOURCE_NAME=nacos
Code location
api/configs/remote_settings_sources/nacos/__init__.py → NacosSettingsSource.get_field_value

@override
def get_field_value(self, field: FieldInfo, field_name: str) -> tuple[Any, str, bool]:
    field_value = self.remote_configs.get(field_name)
    if field_value is None:
        return None, field_name, False

    return field_value, field_name, False

How it works today
RemoteSettingsSourceFactory.__call__ (api/configs/app_config.py) iterates over self.settings_cls.model_fields.items() and passes only field_name into get_field_value:

for field_name, field in self.settings_cls.model_fields.items():
    field_value, field_key, value_is_complex = remote_source.get_field_value(field, field_name)
    ...

Because get_field_value does a plain self.remote_configs.get(field_name), it only ever matches keys equal to the field name. It never inspects field.alias, field.validation_alias, or AliasChoices/AliasPath.

### ✔️ Expected Behavior

Expected behavior
When a config field declares aliases, the value should be resolved by matching the Nacos key against the field's AliasChoices (and the declared aliases) in addition to the field name — mirroring how the standard EnvSettingsSource handles aliases.

Proposed fix (for reference)
Fall back to the field's aliases when the primary field_name lookup misses, e.g.:

from pydantic import AliasChoices
from pydantic.fields import FieldInfo

@override
def get_field_value(self, field: FieldInfo, field_name: str) -> tuple[Any, str, bool]:
    field_value = self.remote_configs.get(field_name)
    if field_value is None and field.validation_alias is not None:
        aliases = (
            field.validation_alias.choices
            if isinstance(field.validation_alias, AliasChoices)
            else [str(field.validation_alias)]
        )
        for alias in aliases:
            field_value = self.remote_configs.get(alias)
            if field_value is not None:
                return field_value, field_name, False
    return field_value, field_name, False

Impact
Any Nacos-managed setting whose pydantic field carries an alias is currently ignored (falls back to default). This makes the remote config source behave inconsistently with the env/dotenv sources, which do honor aliases.

### ❌ Actual Behavior

_No response_

dparkmit24:
## Summary

Fixes #41439

`NacosSettingsSource.get_field_value` looked up remote config keys only by the exact pydantic field name, never consulting `field.validation_alias` / `AliasChoices`. Any aliased field stored in Nacos under one of its documented alias keys was therefore never found, and silently kept its default — no error, no warning. This is inconsistent with the env/dotenv sources, which do honor aliases, so the same key works via env and fails via Nacos.

`ApolloSettingsSource` had the identical defect, so the resolution now lives on the shared `RemoteSettingsSource` base as a small helper that both sources delegate to. That framing is deliberate: the bug wasn't that one source forgot aliases, it was that every remote source had to remember, and both of the two that exist forgot. A future third source now inherits the correct behavior instead of re-introducing it.

Two details worth review attention:

- **The matched alias is returned as the key, not the field name.** `DifyConfig` does not set `populate_by_name`, so validation only accepts keys present in a field's `AliasChoices`; keying by field name would let a value be silently dropped by `extra="ignore"` for any field whose own name is absent from its choices. `EnvSettingsSource` keys by the alias for the same reason.
- **Field-name lookup keeps priority.** `EnvSettingsSource` ignores the field name entirely when a `validation_alias` is set; keeping the name first is the smaller change and preserves behavior for deployments whose remote keys are already stored by field name.

Non-string alias choices (`AliasPath`) are skipped rather than raising — no config field uses one today, and a flat properties payload has no way to address a path. Apollo's existing `ValueError` guard for a non-dict payload is unchanged and still runs ahead of the lookup.

**Scope.** 12 of 772 `DifyConfig` fields declare a `validation_alias`. Five of them (`inner_UPLOAD_FILE_EXTENSION_BLACKLIST`, `inner_CONSOLE_CORS_ALLOW_ORIGINS`, `inner_WEB_API_CORS_ALLOW_ORIGINS`, `inner_OPENAPI_CORS_ALLOW_ORIGINS`, `inner_OPENAPI_KNOWN_CLIENT_IDS`) do not list their own field name among their choices, so a remote source could never populate them at all. The other seven — including `FILES_URL`, `INTERNAL_FILES_URL` and the `PUBSUB_*` family — worked when set under the primary name but silently ignored the secondary or legacy alias.

**Verification.** `api/tests/unit_tests/configs/` → 88 passed. With both source files reverted to `main` and the tests kept, exactly four tests fail — the alias tests for each source, each returning `None` where the alias-resolved value is expected — and all 88 pass again once the sources are restored. `make lint` and `make type-check` both exit 0, the latter including `mypy: Success: no issues found in 1734 source files`.

These are unit tests against `get_field_value` with a mocked client; they prove the method resolves aliases. They do not exercise end-to-end `DifyConfig` construction against a live Nacos or Apollo server, and I haven't claimed to have tested that.

**Note on scope and coordination.** @GYJ99 offered to take this issue on 2026-08-28 and I don't want to cut across that — I had a tested fix ready and am posting it in case it's useful. Happy to close this if @GYJ99 is proceeding, or to fold in any review feedback.

## Screenshots

| Before | After |
| ------ | ----- |
| n/a — no user-visible surface. The change affects config resolution at startup; the observable difference is an aliased setting resolving from its alias key instead of silently falling back to its default. The revert→fail→restore→pass evidence is in the Summary. | n/a |

## Checklist

From Claude Code

### DIFF
--- a/api/configs/remote_settings_sources/apollo/__init__.py
+++ b/api/configs/remote_settings_sources/apollo/__init__.py
@@ -52,5 +52,5 @@ class ApolloSettingsSource(RemoteSettingsSource):
     def get_field_value(self, field: FieldInfo, field_name: str) -> tuple[Any, str, bool]:
         if not isinstance(self.remote_configs, dict):
             raise ValueError(f"remote configs is not dict, but {type(self.remote_configs)}")
-        field_value = self.remote_configs.get(field_name)
-        return field_value, field_name, False
+        field_value, field_key = self._resolve_field_value(self.remote_configs, field, field_name)
+        return field_value, field_key, False
--- a/api/configs/remote_settings_sources/base.py
+++ b/api/configs/remote_settings_sources/base.py
@@ -1,6 +1,7 @@
 from collections.abc import Mapping
 from typing import Any
 
+from pydantic import AliasChoices
 from pydantic.fields import FieldInfo
 
 
@@ -13,3 +14,32 @@ class RemoteSettingsSource:
 
     def prepare_field_value(self, field_name: str, field: FieldInfo, value: Any, value_is_complex: bool):
         return value
+
+    @staticmethod
+    def _resolve_field_value(remote_configs: Mapping[str, Any], field: FieldInfo, field_name: str) -> tuple[Any, str]:
+        """
+        Look up ``field`` in ``remote_configs`` by its field name first, then fall back to its
+        ``validation_alias`` (a plain string or any ``str`` member of an ``AliasChoices``),
+        mirroring how ``EnvSettingsSource`` resolves keys for aliased fields.
+
+        Returns ``(value, key)`` where ``key`` is the name that actually matched, so the caller
+        emits a key that ``DifyConfig`` validation accepts. Returns ``(None, field_name)`` on a miss.
+        """
+        field_value = remote_configs.get(field_name)
+        if field_value is not None:
+            return field_value, field_name
+
+        validation_alias = field.validation_alias
+        if isinstance(validation_alias, str):
+            alias_names = [validation_alias]
+        elif isinstance(validation_alias, AliasChoices):
+            alias_names = [choice for choice in validation_alias.choices if isinstance(choice, str)]
+        else:
+            alias_names = []
+
+        for alias_name in alias_names:
+            alias_value = remote_configs.get(alias_name)
+            if alias_value is not None:
+                return alias_value, alias_name
+
+        return None, field_name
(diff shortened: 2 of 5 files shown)


crazywoola:
Thanks for the contribution. Closing this as a duplicate pull request for https://github.com/langgenius/dify/issues/41439. We are keeping langgenius/dify#41463 under the author-first, then FIFO policy for duplicate PRs in the same repository.

PR #41485, opened 2026-08-30
fix(mcp): keep receiver alive on orphaned responses

### LINKED ISSUE #41482: MCP client session receiver exits on orphaned responses
## Description

The MCP `BaseSession` routes responses that cannot be matched to an active request through `_handle_incoming(RuntimeError(...))`. `ClientSession` uses the default message handler, which converts incoming exceptions to `ValueError`. That exception escapes `_receive_loop`, terminates the receiver future, and prevents subsequent valid MCP messages from being processed.

This affects two paths:

1. a JSON-RPC response/error whose request ID is no longer present in `_response_streams`;
2. a bare `HTTPStatusError` when there is no waiting response queue.

A late, duplicate, or otherwise orphaned response should be observable, but should not terminate the entire MCP session.

## Steps to reproduce

1. Create a `BaseSession` using a handler with the same exception behavior as `ClientSession._default_message_handler`.
2. Queue a JSON-RPC response for an unknown request ID, followed by a valid notification.
3. Run `_receive_loop()`.

The same result occurs with an unmatched `HTTPStatusError`.

## Actual behavior

Latest `main` raises and exits before processing the notification:

```text
ValueError: Server Error: SessionMessage(... JSONRPCResponse(id=999) ...)
ValueError: Received response with an unknown request ID: Unauthorized
```

The receiver logs `Error in message processing loop` and stops.

## Expected behavior

Log a warning for the orphaned response and continue processing later messages. Transport exceptions that are not orphaned responses should retain their current error behavior.

## Reproduction evidence

A test-only patch against current `main` produces two deterministic failures. A local minimal patch replacing only the two `_handle_incoming(RuntimeError(...))` calls with warning logs passes the complete BaseSession test module (`26 passed`) plus Ruff and `git diff --check`.

I can submit the focused two-file fix and regression tests if maintainers would like to assign this issue.

mesrop2008:
## Summary

Fixes #41482

The MCP client session's background receiver died permanently the first time it saw a response it could not match to an in-flight request.

`BaseSession._receive_loop` routed unmatched messages through `_handle_incoming(RuntimeError(...))`. `ClientSession`'s default message handler re-raises any `Exception` it is handed (`_default_message_handler` in `api/core/mcp/session/client_session.py`), so that `RuntimeError` propagated back out of the loop and hit the `except Exception: logger.exception(...); raise` at the bottom. The receiver thread ended, and every subsequent message on that session — notifications, responses to other in-flight requests — was silently dropped. Callers blocked in `send_request` only unblocked because `check_receiver_status()` re-raised the stored failure, so a single stray message failed the whole session rather than one request.

Two paths reached it:

- a JSON-RPC response/error whose `id` is no longer in `_response_streams` — a duplicate, a late reply, or a request that already timed out
- a bare `HTTPStatusError` arriving when no request is awaiting a response

Both are now logged at `WARNING` and skipped, so one orphaned message can no longer tear down the session.

**Deliberately unchanged:** the `not isinstance(response_root, (JSONRPCResponse, JSONRPCError))` branch still reports through `_handle_incoming`. A malformed message is a protocol failure, not an orphaned response, and it should keep surfacing to the message handler rather than being downgraded to a log line.

**Known follow-up, out of scope here:** `send_request` waits in a `while True` loop with no absolute deadline — its only exit besides a real response is `check_receiver_status()` raising. Now that the receiver survives, a caller whose response was mis-`id`ed waits indefinitely instead of failing fast. That is still strictly better than killing every other request on the session, but the loop needs a bounded deadline. Happy to open a separate issue for it.

Related: #41483 addresses the same issue. This PR keeps the two orphan paths from that change, restores the protocol-error branch that was flagged in review there, and adds a regression test that actually fails on the unpatched code.

## Screenshots

N/A — backend-only change, no user-facing UI.

## Testing

`api/tests/unit_tests/core/mcp/session/test_base_session.py`

- Rewrote `test_receive_loop_unknown_request_id` and `test_receive_loop_http_error_unknown_id`, which asserted the old `RuntimeError` routing. Each now queues the orphan, queues a valid notification behind it, and asserts the notification is still received — i.e. the receiver survived.
- Added `test_receive_loop_survives_orphan_when_handler_raises`. This is the actual regression test: it subclasses `MockSession` with a `_handle_incoming` that re-raises, mirroring `_default_message_handler`, which is the real mechanism that killed the receiver. The stock `MockSession` only records exceptions, so a test built on it passes even against the buggy code.

Results:

- `tests/unit_tests/core/mcp/session/test_base_session.py` — 27 passed
- `tests/unit_tests/core/mcp/session/test_client_session.py` — 29 passed
- `ruff check` and `ruff format --check` clean on both changed files

Red/green verified: reverting only the `base_session.py` change and rerunning the suite gives 3 failed / 24 passed, with `ValueError: Server Error: ...` escaping `_receive_loop`. The new tests fail without the fix and pass with it.

## Checklist

### DIFF
--- a/api/core/mcp/session/base_session.py
+++ b/api/core/mcp/session/base_session.py
@@ -333,9 +333,10 @@ class BaseSession[
                                     )
                                 )
                         else:
-                            self._handle_incoming(
-                                RuntimeError(f"Received response with an unknown request ID: {message}")
-                            )
+                            # No request is waiting on this transport error: the request it belonged to
+                            # has already returned or timed out. Dropping it must not tear down the
+                            # receiver, otherwise every later message on this session is lost too.
+                            logger.warning("Discarding HTTP error with no request awaiting a response: %s", message)
                     case Exception():
                         self._handle_incoming(message)
                     case SessionMessage(message=JSONRPCMessage(root=JSONRPCRequest())):
@@ -389,7 +390,13 @@ class BaseSession[
                         if response_queue is not None:
                             response_queue.put(response_root)
                         else:
-                            self._handle_incoming(RuntimeError(f"Server Error: {message}"))
+                            # Orphaned response: duplicate, late, or for a request that already timed
+                            # out. Log and keep going so one stray message cannot kill the session.
+                            logger.warning(
+                                "Discarding response with an unknown or expired request ID %s: %s",
+                                response_root.id,
+                                message,
+                            )
             except queue.Empty:
                 continue
             except Exception:
--- a/api/tests/unit_tests/core/mcp/session/test_base_session.py
+++ b/api/tests/unit_tests/core/mcp/session/test_base_session.py
@@ -476,39 +476,92 @@ def test_check_receiver_status_fail(streams):
     executor.shutdown()
 
 
-@pytest.mark.timeout(10)
-def test_receive_loop_unknown_request_id(streams):
-    read_stream, write_stream = streams
-    session = MockSession(read_stream, write_stream, ReceiveRequest, ReceiveNotification)
+def _keep_alive_notification():
+    return SessionMessage(
+        message=JSONRPCMessage(
+            JSONRPCNotification(jsonrpc="2.0", method="test/notification", params={"message": "still alive"})
+        )
+    )
 
-    with session:
-        resp = JSONRPCResponse(jsonrpc="2.0", id=999, result={"ok": True})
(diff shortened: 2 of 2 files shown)


crazywoola:
Closing this duplicate PR under the author-first policy: PR #41499, authored by the issue opener, is being retained for issue #41482.

PR #41494, opened 2026-08-30
fix(mcp): keep receiver alive after orphaned response in BaseSession

### LINKED ISSUE #41482: MCP client session receiver exits on orphaned responses

Harsh23Kashyap:
## Summary

Fixes #41482 — an orphaned MCP response (or a bare `HTTPStatusError` with no waiting response queue) used to terminate the entire MCP session receiver, dropping every subsequent valid message. Now the receiver logs a warning and continues processing the rest of the stream.

The orphaned response was routed through `BaseSession._handle_incoming(RuntimeError(...))`. Under the default `ClientSession` message handler, the exception is converted to a `ValueError` that escapes `_receive_loop`, terminates the receiver future, and the outer `try/except` then logs `Error in message processing loop` and stops. After that, any later notification, request, or response on the same stream is silently dropped.

## Root cause

`api/core/mcp/session/base_session.py`, two reachable sites plus one defensive site:

- **`case HTTPStatusError():`** (line 321) — when `_response_streams.get(self._request_id - 1)` returns `None`, the code constructed `RuntimeError("Received response with an unknown request ID: …")` and routed it through `_handle_incoming`. This is path 1 in the issue.
- **`case _:`** (line 382, the fallback for `SessionMessage` whose root is `JSONRPCResponse`/`JSONRPCError`) — when `response_queue` is `None` (i.e. the id is no longer in `_response_streams`), the code constructed `RuntimeError("Server Error: …")` and routed it through `_handle_incoming`. This is path 2 in the issue.
- **`case _:`** defensive check (line 385) — when the root is somehow not a `Response`/`Error` at all. Unreachable in the current MCP type union, but kept consistent so future root types do not regress this.

The `case Exception():` branch (line 349) is intentionally **not** changed — it catches real protocol/transport failures that the outer `try/except` (line 416-418) is meant to surface and re-raise.

## Reproduction

1. Create a `BaseSession` with a handler that has the same exception behavior as `ClientSession._default_message_handler`.
2. Queue a JSON-RPC response for an unknown request ID, followed by a valid notification.
3. Run `_receive_loop()`.

Pre-fix: the receiver raises and exits before processing the notification. The receiver logs `Error in message processing loop` and stops.
Post-fix: a warning is logged for the orphan, the notification is delivered, and the receiver keeps running.

## Changes

- **`api/core/mcp/session/base_session.py`** — three `_handle_incoming(RuntimeError(...))` calls replaced with `logger.warning(...)` + `continue`. Each site has a 4-6 line comment explaining the orphan condition and why the receiver must stay alive. The `case Exception():` branch is unchanged.
- **`api/tests/unit_tests/core/mcp/session/test_base_session.py`** —
  - `test_receive_loop_unknown_request_id` (updated, was asserting the orphan reached `_handle_incoming`): now asserts the warning is logged (`caplog.text` contains "Received response with an unknown request ID") and `_handle_incoming` was not called.
  - `test_receive_loop_http_error_unknown_id` (updated, same shape): now asserts the warning is logged and `_handle_incoming` was not called.
  - `test_receive_loop_orphaned_response_does_not_block_subsequent_messages` (new, the core regression): queues an orphaned response followed by a valid notification, asserts the notification was delivered. Pre-fix this would have failed because the receiver stopped after the orphan.
  - `test_receive_loop_non_jsonrpc_response_is_dropped` (new): covers the `case _:` fallback for a `JSONRPCResponse` with an unknown id, asserts the next message is still processed.

## Out of scope (deliberate)

- The `case Exception():` branch (line 349-350) is intentionally not changed. It catches real protocol/transport failures that the outer `try/except` (line 416-418) is meant to surface. If the handler raises, the receiver still re-raises and the parent `check_receiver_status` (line 172) propagates the error. That is the correct behavior for real failures; only the orphan-specific cases are too aggressive.
- Did not change the `case SessionMessage(...)` branches for valid requests and notifications — those still route through `_handle_incoming` as designed.
- Did not change `ClientSession._default_message_handler` itself — the orphan happens upstream of that handler, so the handler's conversion of exceptions to `ValueError` is no longer triggered for the orphan cases.

## Testing

- `uv run python -m pytest -q tests/unit_tests/core/mcp/session/test_base_session.py` — 28/28 pass (was 26/26 before; 2 existing tests updated + 2 new tests added).
- `uv run ruff check` — All checks passed.
- `uv run ruff format --check` — clean on both files.
- `uv run pyrefly check core/mcp/session/base_session.py` — 0 diagnostics.
- `/consult-kiro` (openai/gpt-5 --variant low) confirmed: "The fix is good and scoped correctly; it converts the three orphaned-response exits to warning+continue without masking real errors." Suggested the additional non-JSON-RPC fallback test — applied.

## Risk

Low. The change is to error-handling paths only — three sites where the receiver was being torn down on a recoverable condition. The `case Exception():` branch still routes real errors through `_handle_incoming` (so the outer `try/except` still surfaces and re-raises actual transport failures). The new tests pin the new behavior; the updated tests prove the old behavior is gone.

## Impact

- **Users**: MCP integrations that previously saw sessions die after a single late/duplicate/401 response now stay alive. Notifications, progress updates, and request/response pairs on the same stream after the orphan are delivered as expected.
- **Maintainers**: a 29-line production diff with 4 new tests covering both orphan paths plus the regression that proves the receiver stays alive. No new patterns introduced; the fix removes a class of `Error in message processing loop` log lines that previously meant "the MCP session is dead."

### DIFF
--- a/api/core/mcp/session/base_session.py
+++ b/api/core/mcp/session/base_session.py
@@ -333,9 +333,19 @@ class BaseSession[
                                     )
                                 )
                         else:
-                            self._handle_incoming(
-                                RuntimeError(f"Received response with an unknown request ID: {message}")
+                            # Orphaned response/error: the request ID is no longer
+                            # in `_response_streams` (timed out, cancelled, or
+                            # already fulfilled). The default `ClientSession`
+                            # handler converts this to a ValueError that escapes
+                            # `_receive_loop` and terminates the receiver — every
+                            # subsequent valid MCP message is then dropped. Log
+                            # and continue instead; transport errors that are not
+                            # orphaned responses keep the existing path.
+                            logger.warning(
+                                "Received response with an unknown request ID, dropping: %s",
+                                message,
                             )
+                            continue
                     case Exception():
                         self._handle_incoming(message)
                     case SessionMessage(message=JSONRPCMessage(root=JSONRPCRequest())):
@@ -382,14 +392,25 @@ class BaseSession[
                     case _:  # Response or error
                         response_root = message.message.root
                         if not isinstance(response_root, (JSONRPCResponse, JSONRPCError)):
-                            self._handle_incoming(RuntimeError(f"Server Error: {message}"))
+                            # Bare HTTPStatusError (or any non-JSON-RPC message)
+                            # that has no waiting response queue. Logging here
+                            # instead of routing through `_handle_incoming` keeps
+                            # the receiver alive for subsequent valid messages.
+                            logger.warning("Server error, dropping: %s", message)
                             continue
 
                         response_queue = self._response_streams.get(response_root.id)
                         if response_queue is not None:
                             response_queue.put(response_root)
                         else:
-                            self._handle_incoming(RuntimeError(f"Server Error: {message}"))
+                            # Orphaned JSON-RPC response/error (unknown request
+                            # ID). Log and continue so the receiver keeps
+                            # processing subsequent messages.
+                            logger.warning(
+                                "Server error with an unknown request ID, dropping: %s",
+                                message,
+                            )
+                            continue
             except queue.Empty:
                 continue
             except Exception:
--- a/api/tests/unit_tests/core/mcp/session/test_base_session.py
+++ b/api/tests/unit_tests/core/mcp/session/test_base_session.py
@@ -477,38 +477,119 @@ def test_check_receiver_status_fail(streams):
 
 
 @pytest.mark.timeout(10)
-def test_receive_loop_unknown_request_id(streams):
-    read_stream, write_stream = streams
-    session = MockSession(read_stream, write_stream, ReceiveRequest, ReceiveNotification)
+def test_receive_loop_unknown_request_id(streams, caplog: pytest.LogCaptureFixture):
+    """Regression for #41482: an orphaned JSON-RPC response (unknown request ID)
+    must be logged as a warning and the receiver must continue running, not
+    raise through `_handle_incoming` and tear down the session."""
+    with caplog.at_level(logging.WARNING, logger="core.mcp.session.base_session"):
+        read_stream, write_stream = streams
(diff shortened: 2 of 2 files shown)


crazywoola:
Closing this duplicate PR under the author-first policy: PR #41499, authored by the issue opener, is being retained for issue #41482.

PR #41531, opened 2026-08-31
refactor(models): pass session into CustomizedSnippet accessors

moneycat957:
> [!IMPORTANT]
>
> 1. Associated issue: #40372 (claimed the `snippet.py` slice in [this comment](https://github.com/langgenius/dify/issues/40372#issuecomment-5474759541)).

## Summary

Part of #40372.

Follow-up to #40370 and #41370: converts the `CustomizedSnippet` accessors in
`api/models/snippet.py` that reach for the global `db.session` internally into plain
methods taking `session: Session` explicitly, per the pattern established in those PRs:

- `CustomizedSnippet.graph_dict` — was `@property` doing `db.session.get(Workflow, ...)`;
  now `def graph_dict(self, session: Session) -> dict[str, Any]`.
- `CustomizedSnippet.tags` — was `@property` doing `db.session.scalars(...)`;
  now `def tags(self, session: Session) -> Sequence[Tag]`.
- `CustomizedSnippet.created_by_account` — same conversion;
  now `def created_by_account(self, session: Session) -> Account | None`.
- `CustomizedSnippet.updated_by_account` — same conversion;
  now `def updated_by_account(self, session: Session) -> Account | None`.
- `CustomizedSnippet.author_name` — delegates to `created_by_account`, so it threads the
  caller's session through; now `def author_name(self, session: Session) -> str | None`.

`input_fields_list` and `version_str` touch no database and remain `@property`.
The now-unused `from .engine import db` import is dropped — `snippet.py` no longer
references the global session at all.

No production callers needed updating: the only references to these accessor names
elsewhere in the codebase belong to other models.

## Tests

`api/tests/unit_tests/models/test_snippet.py` previously needed a `snippet_session`
fixture that monkeypatched `snippet_module.db.session` onto the SQLite session, purely
to work around the accessors reaching for the global session. That fixture is now
obsolete and has been removed; the existing tests take `sqlite_session` directly and
pass it into each accessor.

Adds `TestCustomizedSnippetSessionAccessors::test_accessors_return_none_when_related_rows_are_absent`,
covering the case where `created_by` / `updated_by` hold ids with no matching row — a
distinct path from the existing "no id at all" test, since it exercises `session.get()`
returning `None` rather than the falsy-id short circuit. It also asserts the empty-result
paths for `graph_dict` and `tags`.

## Checklist

From Codex

### DIFF
--- a/api/models/snippet.py
+++ b/api/models/snippet.py
@@ -1,17 +1,17 @@
 import json
+from collections.abc import Sequence
 from datetime import datetime
 from enum import StrEnum
 from typing import Any
 
 import sqlalchemy as sa
 from sqlalchemy import DateTime, String, func
-from sqlalchemy.orm import Mapped, mapped_column
+from sqlalchemy.orm import Mapped, Session, mapped_column
 
 from libs.uuid_utils import uuidv7
 
 from .account import Account
 from .base import Base
-from .engine import db
 from .model import Tag, TagBinding
 from .types import AdjustedJSON, LongText, StringUUID
 
@@ -65,13 +65,12 @@ class CustomizedSnippet(Base):
         DateTime, nullable=False, server_default=func.current_timestamp(), onupdate=func.current_timestamp()
     )
 
-    @property
-    def graph_dict(self) -> dict[str, Any]:
+    def graph_dict(self, session: Session) -> dict[str, Any]:
         """Get graph from associated workflow."""
         if self.workflow_id:
             from .workflow import Workflow
 
-            workflow = db.session.get(Workflow, self.workflow_id)
+            workflow = session.get(Workflow, self.workflow_id)
             if workflow:
                 return json.loads(workflow.graph) if workflow.graph else {}
         return {}
@@ -81,10 +80,9 @@ class CustomizedSnippet(Base):
         """Parse input_fields JSON to list."""
         return json.loads(self.input_fields) if self.input_fields else []
 
-    @property
-    def tags(self):
+    def tags(self, session: Session) -> Sequence[Tag]:
         """Get snippet tags."""
-        tags = db.session.scalars(
+        tags = session.scalars(
             sa.select(Tag)
             .join(TagBinding, Tag.id == TagBinding.tag_id)
             .where(
@@ -97,24 +95,21 @@ class CustomizedSnippet(Base):
 
         return tags or []
 
-    @property
-    def created_by_account(self) -> Account | None:
+    def created_by_account(self, session: Session) -> Account | None:
         """Get the account that created this snippet."""
         if self.created_by:
-            return db.session.get(Account, self.created_by)
+            return session.get(Account, self.created_by)
         return None
 
-    @property
-    def author_name(self) -> str | None:
+    def author_name(self, session: Session) -> str | None:
         """Get the creator account name."""
-        account = self.created_by_account
+        account = self.created_by_account(session)
         return account.name if account else None
 
-    @property
-    def updated_by_account(self) -> Account | None:
+    def updated_by_account(self, session: Session) -> Account | None:
         """Get the account that last updated this snippet."""
         if self.updated_by:
-            return db.session.get(Account, self.updated_by)
+            return session.get(Account, self.updated_by)
         return None
 
     @property
--- a/api/tests/unit_tests/models/test_snippet.py
+++ b/api/tests/unit_tests/models/test_snippet.py
@@ -5,7 +5,6 @@ import json
 import pytest
 from sqlalchemy.orm import Session
 
-from models import snippet as snippet_module
 from models.account import Account
 from models.enums import TagType
 from models.model import Tag, TagBinding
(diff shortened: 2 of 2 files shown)


asukaminato0721 (approved):

asukaminato0721 (changes requested):
ci

moneycat957:
Thanks for the review. The CI failure points at something bigger than a broken test, so
I'd rather flag it than push a green-CI patch.

## Root cause of the failing checks

`tests/unit_tests/fields/test_snippet_fields.py:34` monkeypatches `models.snippet.db`:

    monkeypatch.setattr(snippet_module.db, "session", sqlite_session)

This PR drops the now-unused `from .engine import db` import from `models/snippet.py`, so
that attribute no longer exists. That's what both `Run Type Checks`
(`No attribute 'db' in module 'models.snippet' [missing-attribute]`) and
`API Unit Tests (3.12)` are reporting. I missed this file — I only updated
`tests/unit_tests/models/test_snippet.py`.

## But fixing that test is not sufficient

That test exists because `fields/snippet_fields.py` consumes these accessors through
Pydantic's `from_attributes` (`ResponseModel` sets `from_attributes=True`):

    class SnippetListItemResponse(ResponseModel):
        tags: list[SnippetTagResponse]
        author_name: str | None

    class SnippetResponse(ResponseModel):
        graph:      dict[str, Any] = Field(validation_alias="graph_dict")
        tags:       list[SnippetTagResponse]
        created_by: SimpleAccountResponse | None = Field(validation_alias="created_by_account")
        updated_by: SimpleAccountResponse | None = Field(validation_alias="updated_by_account")

All five accessors this PR converts are read by attribute name during response validation.
Once they are methods rather than properties, Pydantic receives the bound method object.
`controllers/console/workspace/snippets.py` calls `dump_response(SnippetResponse, snippet)`
directly on the ORM instance at three sites (lines 180, 203, 252), with no session in scope.

Reproduced on this branch:

    snippet.author_name
    -> <bound method CustomizedSnippet.author_name of <models.snippet.CustomizedSnippet ...>>

    dump_response(SnippetListItemResponse, snippet)
    -> ValidationError: 2 validation errors for SnippetListItemResponse
       tags:        Input should be a valid list   [input_type=method]
       author_name: Input should be a valid string [input_type=method]

So as it stands this PR would 500 the customized-snippet console endpoints. I've converted
it to draft so it doesn't get merged while we decide.

## How would you like to proceed?

**(a) Keep the model change and move response construction off `from_attributes`.**
The controller (or `SnippetService`) would build the response explicitly with a session it
already owns, e.g. `author_name=snippet.author_name(session)`. This is the change that
actually completes the refactor, but it reaches into `fields/snippet_fields.py`,
`controllers/console/workspace/snippets.py` and possibly `services/snippet_service.py` —
well beyond the slice I claimed, so I don't want to do it unilaterally.

**(b) Drop `snippet.py` from #40372's scope for now.**
Unlike the `ChildChunk` / `WorkflowAppLog` slices, every accessor on `CustomizedSnippet` is
load-bearing for response serialization, so there's no subset that can be converted in
isolation. If you'd rather land the serialization refactor first and revisit the session
threading afterwards, I'll close this PR and note the finding on #40372 so nobody else
claims the slice and hits the same wall.

I'm happy to do (a) in this PR if you confirm the direction — including whether you'd
prefer the explicit construction to live in the controller or in `SnippetService`. Just
didn't want to expand the diff by 3 files without asking.

moneycat957:
Following up on my own question — I went looking for precedent in the tree and found the
answer, so I'd like to propose option (a) concretely rather than leave this open.

`api/fields/` already has an established pattern for exactly this problem:
`DatasetDetailResponseSource` in `dataset_fields.py`, and `ConversationResponseSource` /
`MessageResponseSource` / `_SessionResponseSource` / `_FeedbackResponseSource` /
`_AnnotationResponseSource` in `conversation_fields.py`. Each holds the ORM object plus a
`Session`, exposes the session-dependent fields as properties that delegate to
`get_*(session=...)` on the model, and falls through to the model for everything else via
`__getattr__`. Response validation then runs against the wrapper instead of the raw model,
so the model accessors are free to take an explicit session.

`snippet_fields.py` has no such wrapper, which is why converting the `CustomizedSnippet`
accessors broke serialization here — I skipped a step that the `Dataset` and `Conversation`
paths had already taken.

Proposed plan for this PR, mirroring `DatasetDetailResponseSource`:

1. Add `get_graph_dict`, `get_tags`, `get_created_by_account`, `get_updated_by_account` and
   `get_author_name`, each taking `session: Session`, to `CustomizedSnippet`, keeping the
   existing `@property` accessors in place for now.
2. Add `SnippetResponseSource` (plus a `snippet_response_source(snippet, *, session)`
   factory) to `fields/snippet_fields.py`, following the `DatasetDetailResponseSource`
   shape.
3. Update the three `dump_response(SnippetResponse | SnippetListItemResponse, snippet)`
   call sites in `controllers/console/workspace/snippets.py` to wrap the model with a
   session first.
4. Remove the legacy `@property` accessors, leaving the session-aware methods as the only
   path — the step this PR originally jumped straight to.

That is more than the `snippet.py` slice I claimed on #40372, so I'd rather confirm before
expanding the diff: is this the direction you want, and would you prefer it all in this PR,
or steps 1–3 in a separate PR with this one rebased on top to do step 4?

asukaminato0721 (approved):

PR #41538, opened 2026-08-31
feat(web): add marketplace creator profiles and homepage redesign

CodingOnStar:
> [!IMPORTANT]
>

## Summary

Shared marketplace UI used by Dify console and marketplace.dify.ai.

- Creator profiles: public `/marketplace/creator/[uniqueHandle]` pages with publisher sidebar, creations, and Creator Center avatars.
- Homepage redesign: hero, banner carousel (blog / event / ad / recommend), sticky search, catalog tabs, tag and language filters.
- Mobile: stacked banners, swipe, search that stays below the header, partner header, and collection layout.
- Banner polish: crop event/ad art from the right, keep blog artwork at 400px, ellipsize the green blog tag on one line, wrap the title, and hide mobile Read more because the whole blog slide is already the link.
- Creator Center entry in the sidebar help menu; hide bundles from plugin categories.
- Banner impression/click tracking and marketplace site events.

## Checklist

From Grok


CodingOnStar:
Pushed CI fixes for this draft (still draft, not ready-for-review).

- Merged current `origin/main` (`762019a160`) so `HEAD` is no longer based on stale `8b8031710b`.
- Restored the skill/agent/API stack from main (`AgentSkillBindingSnapshot`, `SkillManagementService(session=...)`, `copy_agent_bindings` / `publish_agent_bindings`, `console_account_admission` skills controller, migrations, skills UI/tests/i18n).
- Marketplace lint: no leaked `ReactNode && (` in `home-catalog-navigation.tsx`; oxfmt on the TS Common files; restored `.editorconfig` `[oxlint-suppressions.json]` (no glob); overlay `className` vs stale `popupClassName`; banner click property narrowing; `pnpm -w check` and web tsslint both exit 0 locally.

Head is `ce65f9ae4c`. CI has not re-run yet from this agent’s perspective.

CodingOnStar:
Pushed `c48f932985` (`fix(web): drop unused marketplace knip exports`).

- Un-exported the 7 file-private knip hits in creator-profile `model.ts` and `use-banner-viewability.ts`.
- Restored `model-provider-page/__tests__/hooks.spec.ts` from `origin/main` (Web Tests 1/4 was failing on stale `model-list` query keys).
- Marked standalone marketplace `hydration-server.tsx` / `server-budget.ts` as knip entries so `knip --production --include files` stays clean.
- Local `vp run knip` and `vp run knip:production` both exit 0. Draft unchanged.

CodingOnStar:
Pushed two CI follow-ups (still draft):

- `6864ba33cd` `fix(web): keep marketplace knip production exports live`
  - `shouldSearchMarketplacePlugins` is used by production catalog search mode
  - `HERO_GRID_PITCH_PX` drives hero icon/grid pitch
  - `getStandaloneCreationHref` stays exported from `creator-profile/model.ts` (marketplace consumer + knip entry)
  - dropped test-only `hasManualTemplateLanguages` and unused `MarketplaceCollection` import
  - `vp run knip`, `vp run knip:production`, and `knip:production-unused-check` all exit 0 locally
- `5dec7c76d4` `fix(web): finish banner loop reset without rAF`
  - wrapping carousel settles to idle with rAF plus a 50ms timeout so clones unmount when frames are delayed
  - local browser spec `home-trending-layout.browser.spec.tsx` 9/9 passed, including the loop-clone case

web/app/layout.tsx
```diff
@@ -69,6 +86,23 @@ export default async function RootLayout({ children }: { children: React.ReactNo
   return (
     <html lang={locale ?? 'en'} className="h-full" suppressHydrationWarning>
       <head>
+        <link rel="manifest" href="/manifest.json" />
```
CodingOnStar: **[bug]** Console root layout now injects marketplace-style PWA/chrome tags (`/manifest.json`, `/apple-touch-icon.png`, `/icon-192x192.png`, `/browserconfig.xml`) and a capture-phase ResizeObserver error suppressor. None of those public files exist in `web/public/` (only `favicon.ico` does), so cloud.dify.ai will 404 the icons/manifest, advertise itself as an installable “Dify” web app, and swallow ResizeObserver errors for every console route. knip comments show marketplace.dify.ai consumes these components from a separate app, so this does not belong in the shared console layout. OAuth/proxy/`client.ts` changes in this PR are related; this block is leftover drift.

**Suggestion:** Revert the PWA tags and ResizeObserver script from the console root layout (or gate them on `IS_MARKETPLACE` in the marketplace app). Keep favicon ownership in `generateMetadata`.

web/app/(commonLayout)/templates/[[...category]]/page.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const parseSortOrder = (value?: string) =>
+  value && TEMPLATE_SORT_ORDERS.has(value) ? value : undefined
+
+const parsePage = (value?: string) => {
+  const parsed = Number(value)
+  return Number.isInteger(parsed) && parsed >= 1 ? parsed : 1
+}
+
+export default async function TemplatesPage({ params, searchParams }: TemplatesPageProps) {
```
Comment (CodingOnStar) [inline web/app/(commonLayout)/templates/[[...category]]/page.tsx]: **[bug]** `/templates` is an async Server Component page under client `MainNavLayout`. The plugins homepage and creator-profile routes were deliberately made sync (with an async child) because an async page module under this shell Flight-double-resolves streamed children (`reason.enqueueModel` on cloud.dify.dev). The marketplace page test even asserts the route is not `async`. `/templates` is the same layout and does heavier server fetching (`EmbeddedTemplatesMarketplace`), so it is exposed to the failure they already hit on `/marketplace`.

**Suggestion:** Match the creator-profile pattern: keep the route module synchronous and move data loading into an async child, or fetch templates on the client like `EmbeddedMarketplace`.

web/app/components/plugins/marketplace/home/home-trending.tsx
```diff
@@ -0,0 +1,394 @@
...
+  isDragging: boolean
+  isMarketplacePlatform: boolean
+  page: MarketplaceBannerPage
+}) {
+  const slideRef = useRef<HTMLDivElement>(null)
+
+  useBannerViewability(
+    slideRef,
+    () => {
+      trackEvent('marketplace_banner_impression', {
```
CodingOnStar: **[bug]** Banner impressions only call Amplitude `trackEvent('marketplace_banner_impression', …)`. Clicks in `home-trending-slides.tsx` dual-write Amplitude plus `trackMarketplaceSiteEvent('marketplace_banner_click', …)`. `trackMarketplaceSiteEvent` is the marketplace.dify.ai pipeline (`data-is-marketplace` + `window.__marketplaceTracking__`). On the standalone site Amplitude is typically not the source of truth, so homepage impression tracking claimed in the PR will not land there. Amplitude on console is consent-gated, which is fine; the gap is the missing site event.

**Suggestion:** Dual-write impressions the same way clicks are dual-written: `trackEvent(...)` and `trackMarketplaceSiteEvent('marketplace_banner_impression', { banner_id, sort, page, language, style_type })`.

web/app/components/plugins/marketplace/templates/index.tsx
```diff
@@ -0,0 +1,314 @@
...
+        page: pageCount,
+        query: normalizedQuery,
+        sortBy,
+        sortOrder,
+        view,
+      }),
+    )
+  }
+
+  const templates =
```
CodingOnStar: **[bug]** When the user has not picked `languages`, each search/category page is fetched without a language constraint, then `filterTemplatesForLocale` runs on that page only. Pagination and `pageCount` still use the API total, while the visible count is `templates.length` after the client filter. A German (or other) user on `/templates/marketing` or `?q=` can see a short/wrong page, pagination that continues into English-fallback or unfiltered slices, or a “results” number that is only the filtered page size. The in-file comment already notes that the API total does not describe what is on screen.

**Suggestion:** Pass the UI locale (or its language prefix) into `searchMarketplaceTemplates` as `languages` when the filter is unset, and paginate only the API result. Keep the client fallback for collection previews if the collection endpoint has no language param.

web/app/components/plugins/marketplace/creator-profile/data.server.ts
```diff
@@ -0,0 +1,173 @@
...
+    sortOrder: CreatorSortOrder,
+  ): Promise<LoadedCreatorProfile | null> => {
+    const [creatorResult, pluginsResult, templatesResult] = await Promise.allSettled([
+      getPublisher(uniqueHandle, publisherType),
+      getPublisherPlugins(uniqueHandle, sortField, sortOrder),
+      getPublisherTemplates(uniqueHandle, sortField, sortOrder),
+    ])
+
+    const creator = creatorResult.status === 'fulfilled' ? creatorResult.value : undefined
+    if (!creator) return null
```
CodingOnStar: **[bug]** `Promise.allSettled` plus `if (!creator) return null` maps every publisher failure to `notFound()`. A 404/empty creator is correct; a 500, network error, or the new 15s `marketplaceClient` deadline is not. Plugin/template fetches are allowed to fail partial, but a rejected `creatorDetail`/`organizationDetail` still 404s the whole profile. Combined with the timeout in `web/service/client.ts`, a slow creators API becomes a missing creator.

**Suggestion:** If `creatorResult.status === 'rejected'`, rethrow (or throw a digestible error) so the route error boundary/retry can run. Return null only when the request succeeded and `creator` is missing.

web/app/components/plugins/marketplace/creator-profile/data.server.ts
```diff
@@ -0,0 +1,173 @@
...
+import {
+  adaptCreatorProfile,
+  parseCreatorSortField,
+  parseCreatorSortOrder,
+  sortCreatorCreations,
+  toPublisherSortQuery,
+} from './model'
+import 'server-only'
+
+const PAGE_SIZE = 40
```
CodingOnStar: **[suggestion]** Publisher plugins and templates are fetched with `page: 1, page_size: 40` and never paginated. A creator with more than 40 plugins or 40 templates silently drops the rest, including after sort changes (the extra items never leave the API).

**Suggestion:** Either raise the page size to a documented cap and surface “showing first N”, or add load-more using the publisher `total`.

web/app/components/plugins/marketplace/home/home-trending-slides.tsx
```diff
@@ -0,0 +1,496 @@
...
+}) {
+  const resolved = resolveEventAdBannerImageSrcs({
+    desktop: getMarketplaceAssetURL(banner.content.images.desktop),
+    tablet: getMarketplaceAssetURL(banner.content.images.tablet) || undefined,
+    mobile: getMarketplaceAssetURL(banner.content.images.mobile) || undefined,
+  })
+
+  return (
+    <Link
+      href={banner.content.link}
```
CodingOnStar: **[suggestion]** Blog/event/ad slides (and embedded recommend `card.link`) are rendered as `Link href={banner.content.link}` / `card.link` after zod only requires a non-empty string. Creator social links reject non-http(s) schemes; banner CMS/partner payloads do not. `javascript:` or other schemes become an XSS/open-redirect on the console homepage, where recommend cards prefer delivery `card.link` when `isMarketplacePlatform` is false.

**Suggestion:** Allow only `http:`, `https:`, and same-origin relative paths (the same rule as `normalizeCreatorSocialLink`). Drop the banner/card if the href is not safe.

web/service/client.ts
```diff
@@ -36,6 +36,31 @@ function getMarketplaceHeaders() {
   })
 }
 
+// The Marketplace API is public and unauthenticated, so a healthy search
```
CodingOnStar: **[suggestion]** Several new comments narrate incident history and implementation choices (“search spins and never settles”, AbortSignal.any baseline, Flight `enqueueModel`, why a timeout exists) rather than a short constraint. That leaks design history into production code and will go stale.

**Suggestion:** Keep a one-line why (“15s deadline so a stalled Marketplace fetch can error/retry”) and drop the ticket/incident prose. Same for the Flight comments once the async-page rule is encoded in the tests that already exist.

web/app/components/plugins/marketplace/detail-dialog/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  useEffect(
+    () => () => {
+      timeoutIdsRef.current.forEach((id) => window.clearTimeout(id))
+      timeoutIdsRef.current.clear()
+    },
+    [],
+  )
+
+  const handleMessage = useCallback(
```
lyzno1, line 58: [P1] Re-check the installation permission at this message-handling boundary. The card only uses `canInstallPlugin` to hide its hover install button; every user can still open this dialog, and this handler accepts an `install-plugin` message and calls `install(plugin)` without a permission gate or passing a disabled state to the iframe. At minimum, users without `plugin.install` get an unauthorized entry point and a failing request. Please enforce the permission in the install/message owner and reuse the standard installation flow.

lyzno1 (commented):
This PR is not ready to merge in its current form. The diff combines creator profiles, the Marketplace home/catalog rewrite, iframe installation, OAuth/CSP integration, standalone-repository adapters, global Main Nav changes, and a performance experiment across 257 files (+18,871/-836). The result includes user-visible regressions and a substantial amount of code that has no valid owner or execution path.

Blocking scope and ownership changes requested:

- Keep creator profiles, the Marketplace home/catalog, and iframe/OAuth/CSP integration as separate vertical slices.
- Revert the unrelated Main Nav order and visual refresh.
- Remove the entire Marketplace performance E2E/CI bundle. The workflow cannot run before merge, and the scenario is not a stable performance gate.
- Move standalone-only server/search adapters to the standalone repository, or expose a versioned public package boundary. Do not use Knip entries to bless internal path-alias imports.
- Remove implementation-coupled tests that inspect source text, function shape, CSS classes, private DOM, fixed delays, or reverted experiments. This is a P2 maintenance problem in aggregate, not harmless test coverage.
- Preserve the necessary `response_type` redirect allowlist and Marketplace `frame-src` CSP changes; those have valid production call chains.

The inline comments below identify the concrete P1/P2 failures and the cleanup boundaries.

web/service/client.ts
```diff
@@ -99,9 +116,11 @@ const marketplaceLink = new OpenAPILink(marketplaceRouterContract, {
   url: MARKETPLACE_API_PREFIX,
   headers: () => getMarketplaceHeaders(),
   fetch: (request, init) => {
+    const requestInit = init as RequestInit | undefined
     return globalThis.fetch(request, {
-      ...init,
+      ...requestInit,
       cache: 'no-store',
+      signal: withRequestDeadline(requestInit?.signal ?? request.signal),
```
lyzno1: [P1] Scope this deadline to the operations that need it. `marketplaceLink` also serves the existing `downloadPlugin` Blob operation, so any large package or slow connection that takes more than 15 seconds is now aborted. Exclude binary downloads or make the deadline an operation context owned by the small JSON queries instead of changing the shared transport.

web/app/(commonLayout)/marketplace/page.tsx
```diff
@@ -1,19 +1,16 @@
...
   return (
     <div
-      id="marketplace-container"
-      className="flex h-full min-h-0 flex-col overflow-y-auto bg-background-default-subtle pr-1"
+      id={MARKETPLACE_CONTAINER_ID}
+      className="flex h-full min-h-0 flex-col overflow-y-auto bg-background-default"
     >
       <MarketplaceInstallPermissionProvider>
-        <Marketplace searchParams={searchParams} isMarketplacePlatform showInstallButton />
+        <EmbeddedMarketplace showInstallButton variant="home" />
```
lyzno1, line 13: [P1] Please restore request-aware server prefetching for the Dify route. This change drops `searchParams` and the server entry, so direct catalog/search/category URLs now ship an empty client shell and wait for hydration plus another network request. The old server entry/hydration code is still maintained elsewhere in this PR but has no Dify production consumer. Either keep SSR dehydration for this route or move the standalone-only server adapter to its actual owner.

web/service/marketplace-template-discovery.ts
```diff
@@ -0,0 +1,148 @@
...
+      page: 1,
+      page_size: 100,
+    },
+  })
+  const collections = response.data?.collections ?? []
+  const entries: (readonly [string, MarketplaceTemplate[]])[] = []
+
+  // Bounded fan-out: fetch collection previews in small batches instead of
+  // firing one uncached request per collection all at once.
+  for (
```
lyzno1: [P1] The batch size bounds concurrency but not route latency. This loop can await 20 batches for 100 collections; with the new 15-second per-request deadline, `/templates` can wait roughly 300 seconds before rendering. Add a route-level budget/abort, pagination, or a bulk endpoint rather than sequentially exhausting every collection.

web/app/components/plugins/marketplace/utils.ts
```diff
@@ -70,72 +70,97 @@ export const getPluginDetailLinkInMarketplace = (
...
+  const marketplaceCollections = marketplaceCollectionsDataJson.data?.collections || []
+  const marketplaceCollectionPluginsMap: Record<string, Plugin[]> = {}
+
+  const pending = [...marketplaceCollections]
+  const fetchCollectionPlugins = async () => {
+    for (let collection = pending.shift(); collection; collection = pending.shift()) {
+      try {
+        marketplaceCollectionPluginsMap[collection.name] =
+          await getMarketplacePluginsByCollectionId(collection.name, query, options)
+      } catch {
```
lyzno1: [P2] Rethrow cancellation here. This catch also swallows the caller's `AbortError`; when a user changes category/search, the cancelled query resolves successfully with empty carousels and TanStack Query can cache that state for five minutes. Only degrade real per-collection service failures; propagate an aborted signal.

web/service/marketplace-template-discovery.ts
```diff
@@ -0,0 +1,148 @@
...
+      ...(await Promise.all(
+        batch.map(async (collection) => {
+          try {
+            const collectionResponse = await marketplaceClient.templateCollectionTemplates({
+              params: { collectionName: collection.name },
+              body: { limit: COLLECTION_PREVIEW_TEMPLATE_LIMIT },
+            })
+
+            return [collection.name, collectionResponse.data?.templates ?? []] as const
+          } catch {
```
lyzno1: [P2] Do not convert a partial upstream failure into a successful empty collection. The function later returns `ok: true` and caches this result for five minutes, so a transient failure silently removes the collection even after the API recovers. Preserve a degraded/error state and do not put partial failures into the success cache.

web/app/components/plugins/marketplace/creator-profile/data.server.ts
```diff
@@ -0,0 +1,201 @@
...
+const MAX_PAGES = 5
```
lyzno1: [P2] This silently truncates creator inventory. When `total` is absent, the fallback makes the first page look complete; when it is present, anything beyond 200 items is dropped. The UI has no pagination, `hasMore`, or truncation notice. Add real pagination/load-more, or return an explicit truncated result instead of presenting incomplete data as complete.

web/app/components/plugins/marketplace/detail-dialog/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+    (data: unknown, reply: (payload: unknown) => void) => {
+      if (!isInstallRequest(data, plugin.latest_package_identifier)) return
+
+      const uniqueIdentifier = plugin.latest_package_identifier
+      const timeoutId = window.setTimeout(() => {
```
lyzno1, line 79: [P2] The timeout does not settle the operation. It sends `timeout`, but a later `install(plugin)` resolution still reaches the second `reply` below and sends `success` or `failed` for the same request. Use a single-settle guard or `Promise.race` and ignore late results after timeout.

.github/workflows/marketplace-performance-e2e.yml
```diff
@@ -0,0 +1,70 @@
+name: Marketplace Performance E2E
+
+# Opt-in diagnostic: single-sample timing budgets are too noisy to gate every
+# PR, so this lane is only run on demand instead of from the main CI pipeline.
+on:
+  workflow_dispatch:
```
lyzno1: [P2] Please remove this workflow and its dedicated E2E bundle. GitHub only delivers `workflow_dispatch` when the workflow file exists on the default branch (https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow), while this PR also excludes `@marketplace-performance` from the default/full Cucumber runs. The workflow API currently returns 404 and the PR check rollup contains no Marketplace Performance run, so this test has no pre-merge execution path. Its single-sample timing/DOM thresholds are not a stable performance gate either.

e2e/scripts/run-cucumber.ts
```diff
@@ -15,8 +16,27 @@ import './env-register'
...
+  }
+
+  if (process.env.E2E_CUCUMBER_TAGS) expressions.push(process.env.E2E_CUCUMBER_TAGS)
+
+  return expressions
+}
+
+const selectsMarketplacePerformance = (forwardArgs: string[]) =>
+  collectTagExpressions(forwardArgs).some((expression) =>
+    /(?<!not\s)@marketplace-performance/.test(expression),
```
lyzno1: [P2] Do not parse Cucumber tag expressions with this regex. A valid expression such as `not (@marketplace-performance)` still matches, so a run that explicitly excludes the scenario starts the Marketplace stub and bakes its API prefix into the shared Web build. Remove this one-off runner path with the unusable workflow; if a profile remains, select it explicitly rather than guessing the tag AST.

web/app/components/main-nav/routes.ts
```diff
@@ -88,6 +79,15 @@ export const MAIN_NAV_ROUTES = [
     activeIcon: 'i-custom-vender-main-nav-skill-active',
     visibility: SKILL_ENABLED_FOR_WORKSPACE,
   },
+  {
+    key: 'datasets',
```
lyzno1, line 74: [P2] Revert this unrelated global navigation reorder. Moving Skills before Knowledge changes the visible order and keyboard Tab order for every workspace with Skills enabled, and this PR also deletes the existing assertion that protected the complete route order. A Marketplace creator/home change does not own this behavior; deliver any intentional Main Nav reorder separately with product/design evidence.

knip.config.ts
```diff
@@ -15,6 +15,14 @@ const config: KnipConfig = {
         'tsslint.config.ts',
         'dev-proxy.config.ts',
         'plugins/eslint/index.js',
+        // Consumed by the dify-marketplace repository, which mounts this
+        // repo as a submodule and imports these modules via path aliases.
+        'app/components/plugins/marketplace/index.tsx',
```
lyzno1: [P2] Do not use Knip entries to suppress dead code that another repository imports through private path aliases. These files are not Dify application roots or a versioned public API, and several have no Dify production consumer beyond tests/each other. Move standalone-only adapters to the standalone repository, or expose a deliberate package entry point with explicit exports and checks.

web/global.d.ts
```diff
@@ -19,5 +19,18 @@ declare global {
   interface Window {
     gtag?: Gtag
     dataLayer?: unknown[]
+    __marketplaceTracking__?: {
```
lyzno1, line 26: [P2] This bridge has no producer in Dify's production tree, and the body marker used by the consumer is not set either, so the new Marketplace tracking calls are no-ops in a Dify build. If an external host injects this API, document and own that public contract; otherwise remove the ambient global, bridge, calls, and tests.

web/app/components/plugins/marketplace/home/home-trending.tsx
```diff
@@ -0,0 +1,397 @@
...
+  banners: PluginBanner[]
+  isMarketplacePlatform: boolean
+  page: MarketplaceBannerPage
+}) {
+  const { t } = useTranslation('plugin')
+  const carouselRootRef = useRef<HTMLDivElement>(null)
+  const swipeGestureRef = useRef<SwipeGesture | null>(null)
+  const suppressClickRef = useRef(false)
+  const suppressClickTimerRef = useRef<number | null>(null)
+  const [selectedIndex, setSelectedIndex] = useState(0)
```
lyzno1, line 100: [P2] Reconcile the selection when `banners` changes. If the user is on slide 3 and a locale/refetch replaces the list with one banner, `selectedIndex` and `trackIndex` remain out of range: the only slide is marked inactive/inert and the track stays translated off-screen. Store a stable banner id or clamp/derive the index from the new list, and cover a rerender where the list shrinks.

web/app/__tests__/layout.spec.tsx
```diff
@@ -130,4 +133,16 @@ describe('Root layout System Features bootstrap', () => {
 
     expect(queryClient.getQueryData(['console', 'system-features'])).toBeUndefined()
   })
+
+  it('does not inject marketplace PWA chrome or a global ResizeObserver filter', () => {
```
lyzno1: [P2] Delete this test. The production root layout is unchanged; this reads source text only to assert that strings from a reverted PWA/ResizeObserver experiment are absent. It can fail on a comment and pass for equivalent unwanted behavior, so it protects no observable contract and leaves reverted branch residue in the suite.

web/app/components/plugins/marketplace/home/__tests__/home-trending-layout.browser.spec.tsx
```diff
@@ -0,0 +1,275 @@
...
+    expect(getComputedStyle(artwork!).borderBottomLeftRadius).toBe('16px')
+  })
+
+  it('moves forwards into the first slide clone before resetting the loop', async () => {
+    const screen = await render(
+      <HomeTrending banners={carouselBanners} isMarketplacePlatform page="plugins" />,
+    )
+
+    await screen.getByRole('button', { name: 'Third banner' }).click()
+    await new Promise((resolve) => setTimeout(resolve, 450))
```
lyzno1: [P2] Replace or remove this implementation-coupled Browser Mode case. Fixed 450 ms/50 ms sleeps plus private `data-carousel-*` attributes and inline transform assertions are explicitly outside the frontend testing contract and create a flake source. Keep only the public browser-owned behavior (visible slide, focus/pointer behavior, and user navigation) and wait for an observable state transition.

web/app/components/main-nav/components/nav-link.tsx
```diff
@@ -30,7 +32,10 @@ const MainNavLink = ({ item, pathname, children }: MainNavLinkProps) => {
       )}
     >
       <NavIcon icon={item.icon} className="group-aria-[current=page]:hidden" />
-      <NavIcon icon={item.activeIcon} className="hidden group-aria-[current=page]:block" />
+      <NavIcon
```
lyzno1: [P2] Revert this from the Marketplace PR. The global 20 px → 18 px icon change and hard-coded active drop shadow belong to a separate Main Nav visual refresh and have no owner or design/browser evidence in this change. Keeping it here makes Marketplace review and rollback responsible for unrelated global navigation styling.

web/app/components/plugins/marketplace/detail-dialog/index.tsx
CodingOnStar: Fixed. The install message handler now checks `canInstallPlugin` and replies `failed` without calling `install()`. The iframe URL includes `canInstall` so the embedded detail page can hide the install action.

web/service/client.ts
CodingOnStar: Fixed. Package downloads whose path ends with `/download` no longer get the 15s shared deadline; JSON catalog requests still do.

web/service/marketplace-template-discovery.ts
CodingOnStar: Fixed with a route-level abort using the existing 2.5s server prefetch budget, rather than a bulk endpoint.

web/app/components/plugins/marketplace/utils.ts
CodingOnStar: Fixed. If the caller's signal is aborted, the error is rethrown instead of resolving as empty carousels.

web/service/marketplace-template-discovery.ts
CodingOnStar: Fixed. A failed collection now returns `ok: false` and is not written to the 5-minute success cache.

web/app/components/plugins/marketplace/detail-dialog/index.tsx
CodingOnStar: Fixed with a single-settle guard so a late `install()` result cannot reply after timeout.

.github/workflows/marketplace-performance-e2e.yml
CodingOnStar: Removed the workflow, feature, steps, stub, and the `e2e:marketplace-performance` script.

e2e/scripts/run-cucumber.ts
CodingOnStar: Removed the tag-regex runner path together with the performance bundle.

web/app/components/main-nav/routes.ts
CodingOnStar: Reverted the Skills/Knowledge order and restored the route-order assertion. `/templates` still marks Marketplace active.

knip.config.ts
CodingOnStar: Replaced the private path-alias Knip entries with `standalone/server.ts` and `standalone/client.ts` as the host surface.

web/global.d.ts
CodingOnStar: Documented the host contract: standalone marketplace injects `data-is-marketplace` and `__marketplaceTracking__`; Dify builds remain no-ops.

web/app/components/plugins/marketplace/home/home-trending.tsx
CodingOnStar: Fixed. When the banner list shrinks, `selectedIndex`/`trackIndex` are clamped back into range.

web/app/__tests__/layout.spec.tsx
CodingOnStar: Deleted the source-text assertion.

web/app/components/plugins/marketplace/home/__tests__/home-trending-layout.browser.spec.tsx
CodingOnStar: Replaced the sleep/private-DOM loop case with a wait on the visible slide / `aria-current`.

web/app/components/main-nav/components/nav-link.tsx
CodingOnStar: Reverted the 18px icon size and the hard-coded active drop shadow.

web/app/(commonLayout)/marketplace/page.tsx
CodingOnStar: Leaving this for a follow-up. Embedded `/marketplace` stays a client shell under `MainNavLayout` because async pages Flight-double-resolve. Standalone still SSRs through `HydrateQueryClient`.

web/app/components/plugins/marketplace/creator-profile/data.server.ts
CodingOnStar: Leaving load-more/pagination for a follow-up. The 200-item cap stays in this PR.

web/app/(commonLayout)/marketplace/page.tsx
CodingOnStar: Not restoring full request-aware prefetch on this Dify route.

The empty client shell is intentional for the console. `/marketplace` sits under the client `MainNavLayout`; an async page (or `HydrateQueryClient` wrapping `EmbeddedMarketplace`) Flight-double-resolves and threw `reason.enqueueModel` on cloud.dify.dev. Standalone `marketplace.dify.ai` still SSRs through `Marketplace` + `HydrateQueryClient`. That adapter is consumed there via `standalone/server.ts` — it is not a Dify-route leftover.

Putting main's prefetch back would also undo a measured console regression. On main, Marketplace navigation was blocked in the SSR/RSC prefetch: collections, then seven plugin groups, then a ~672KB / ~6.56MB decoded RSC, then a full mount (~121 cards / ~6k DOM). Search still prefetched collections before the client search ran. The SaaS console perf report and [ECO-441](https://linear.app/dify/issue/ECO-441) treat that as aggregating first-screen-unfit catalog data, not as missing SSR. Console is behind auth, so this route does not need SEO HTML.

Follow-up, if any, is a small first-screen prefetch (one collection / first search page) above the client shell — not putting main's full dehydration back on this page.

CodingOnStar:
Rebased onto current `main` by merging `origin/main`.

- Resolved the remaining `oxlint-suppressions.json` conflict (keep marketplace barrel-file / list-with-collection updates, take main’s query-options entries, prune unused counts).
- Restored CI: `vp check` formatting, partner badge query after the aria-label removal, autocomplete popup width tolerance, unused suppression prune.

Local `pnpm -w check` and the previously failing web unit/browser specs passed. Waiting on GitHub Main CI.

e2e/scripts/common.ts
```diff
@@ -47,6 +47,12 @@ export const e2eWebEnvOverrides = {
   NEXT_PUBLIC_API_PREFIX: 'http://127.0.0.1:5001/console/api',
   NEXT_PUBLIC_ENABLE_AGENT_V2: 'true',
   NEXT_PUBLIC_PUBLIC_API_PREFIX: 'http://127.0.0.1:5001/api',
+  // NEXT_PUBLIC_* values are inlined into the client bundle at build time, so the
+  // marketplace prefix (e.g. the frozen fixture stub started by run-cucumber.ts)
+  // must reach both the build environment and the build stamp hash.
+  ...(process.env.NEXT_PUBLIC_MARKETPLACE_API_PREFIX
+    ? { NEXT_PUBLIC_MARKETPLACE_API_PREFIX: process.env.NEXT_PUBLIC_MARKETPLACE_API_PREFIX }
+    : {}),
```
lyzno1:
```suggestion
```

CodingOnStar:
Merged `origin/main` and resolved the marketplace collection carousel conflicts.

Kept this branch's pages-based carousel / deferred collections, and applied #41738's `aria-labelledby` so each collection heading still names its carousel.

lin-snow (approved):
LGTM

PR #41374, opened 2026-08-27
refactor(service_api): dep-inject payloads with @model_validate

ShousenZHANG:
## Summary

part of #36659

Moves the inline `Payload.model_validate(service_api_ns.payload or {})` calls in four `service_api` handlers onto the existing `@model_validate` decorator, so the payload is validated before the handler body runs and arrives as an argument. Claimed in [this comment](https://github.com/langgenius/dify/issues/36659#issuecomment-5435152938).

Seven call sites across four files:

- `service_api/app/annotation.py` — 3 sites
- `service_api/dataset/metadata.py` — 2 of its 3 sites
- `service_api/app/human_input_form.py` — 1 site
- `service_api/app/message.py` — 1 site

`@model_validate` goes innermost in each decorator stack, so the model is injected as the first argument after `self` and the outer decorators still supply `session` / `app_model` / path params — matching how `console/agent/composer.py` and the other 73 converted console handlers already stack it. `service_api/app/annotation.py` already imported from `controllers.console.wraps`, and the recently merged web-layer conversions (#40796, #40856) import `model_validate` from the same place.

## Why only seven of the twenty-four remaining sites

I went through every remaining inline call before picking the slice. The other seventeen are not mechanical, because hoisting the parse changes which error a bad request gets back:

- **Inside a `try`** — `service_api/app/audio.py:190`, `web/audio.py:137`. The `try` maps `AppModelConfigBrokenError` / `NoAudioUploadedServiceError` onto API errors; a decorator parse runs outside it.
- **After a `raise` guard** — `web/completion.py:119`, `web/workflow.py:66`. Both sit behind `raise NotCompletionAppError()` / `NotWorkflowAppError()`, so converting them makes a malformed payload return a validation error where the API currently returns the app-mode error.
- **Parse is deep in the handler** — `segment.py` (4 sites, 23-27 statements before the parse), `conversation.py` (2), `rag_pipeline_workflow.py` (2), `metadata.py:326`, `web/remote_files.py:111`, `web/conversation.py:163`. Each runs permission, embedding or DB work first.
- **Not a handler** — `console/remote_files.py:54` is a plain helper, so there is no route to decorate.

Happy to take any of those on separately if maintainers would rather they were converted with the behaviour change made explicit.

## Screenshots

N/A — backend refactor, no UI.

## How this was verified

[...]

Issue #36659, opened 2026-05-26
dep inject model validate


asukaminato0721:
@dosu since we can dep inject the current user id, why dont we also inject the validated parameter?

like

```
...
@with_user
@validate(Query)
def post(self, args: Query, user: Account):
    ...
```

asukaminato0721:
> Source detection — could infer args vs json from the HTTP method
> Swagger integration — the current pattern registers schema models and uses @ns.expect() / @ns.doc(params=...). A @validate decorator could auto-register these too.

@dosu show me how to do it

asukaminato0721:
@dosu let's keep manually register the _ns, but avoid manually "args", "json"

asukaminato0721:
@dosu don't consider Pass source="args" or source="json" explicitly to override.

asukaminato0721:
@dosu tell me current progress and what to do next.

ddddajie:
Hi, I'd like to work on this issue. Could you assign it to me?

a-yeyang:
I'd like to work on this. Opening a PR that extends the existing `@model_validate` decorator (already used in `controllers/web/human_input_form.py` and across several `controllers/console`/`controllers/console/explore` handlers) to the remaining `controllers/web` endpoints that still call `Schema.model_validate(...)` inline:

- `SavedMessageListApi.get` / `.post` (`controllers/web/saved_message.py`)
- `MessageListApi.get`, `MessageFeedbackApi.post`, `MessageMoreLikeThisApi.get` (`controllers/web/message.py`)

This is a scoped, mechanical follow-up (not a full rewrite of the decorator itself, which already exists) — no behavior change, just dependency-injecting the already-validated model instead of parsing it inside the handler body. Will link the PR here once it's up.

yun520-1:
Dependency injection is one of those things that looks like a style preference until you hit the exact case where the validator is initialized before the config is bound. Then every test either stubs the wrong object or silently passes with defaults.

The fix is small but structural:
- accept the validated config as an explicit constructor argument rather than reading it from a global `db.session` registry
- keep the validator as a pure function of `(config, model_spec)` so it can be unit-tested without a database
- fail loudly on missing fields instead of coercing to `None`

This is the same category of problem as HeartFlow's `gate.js` refactor: the logic was always correct, but the wiring made it impossible to verify at rest.

asukaminato0721:
@dosu tell me current progress and what to do next.

Terminator666666:
[...]

PR #41540, opened 2026-08-31
refactor(inner_api): dep-inject workspace payloads with @model_validate

ShousenZHANG:
## Summary

part of #36659

Moves the three inline parses in `inner_api/workspace/workspace.py` onto the `@model_validate` decorator — `EnterpriseWorkspace.post`, `EnterpriseWorkspaceNoOwnerEmail.post` and `EnterpriseWorkspaceMember.post`. Same change as #41374, #41490, #41491, #41500, #41501 and #41539.

Nothing else in `inner_api` this round: `agent/{llm,tools,files}.py` and `knowledge/retrieval.py` raise custom `*HttpError(400)` from a local `except ValidationError`, and `app/dsl.py:110` returns its own `{"code": "invalid_workflow_id"}, 400` — converting those would change documented error contracts rather than just move the parse.

## How did you test it?

- `pytest tests/unit_tests/controllers/inner_api/` — **155 passed**, against 152 on `main`; the three extra are the parametrized decorator test added here
- `ruff check` / `ruff format --check` — clean; `pyrefly check` — 0 diagnostics
- The three handler test classes patched `inner_api_ns` to supply the payload; they now put it in the request context and pass the validated model where the decorator injects it. Eight call sites, all in `test_workspace.py`, verified by handler identity
- Added a bound-method test for the decorators themselves, parametrized over all three handlers and sent as POST so the decorator reads the JSON body these endpoints actually use. Every other test in the file unwraps the view and builds the model by hand, so without it the decorators would ship unexercised. Verified by mutation: dropping any one of the three fails exactly one case
- Runtime signature check on all three wrapped handlers

This PR was fully generated with an AI assistant. I have reviewed the changes and run the relevant tests.

### DIFF
--- a/api/controllers/inner_api/workspace/workspace.py
+++ b/api/controllers/inner_api/workspace/workspace.py
@@ -56,7 +56,6 @@ class EnterpriseWorkspace(Resource):
     )
     @model_validate(WorkspaceCreatePayload)
     def post(self, args: WorkspaceCreatePayload):
-
         account = db.session.scalar(select(Account).where(Account.email == args.owner_email).limit(1))
         if account is None:
             return {"message": "owner account not found."}, 404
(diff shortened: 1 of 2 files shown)


asukaminato0721 (approved):

ShousenZHANG:
@asukaminato0721 this one has been approved and green since Aug 31 — I think it just got skipped when #41571 and #41575 went through this morning. Still `CLEAN`, nothing needed from my side.

PR #40681, opened 2026-08-13
fix(rag): prevent empty keyword results for parent-child chunks

pineapple880066:
## Summary

Fixes #40680

This fixes a silent correctness bug in parent-child keyword retrieval. Keyword hits contain `ChildChunk.index_node_id` values, but `Jieba.search()` previously queried only `DocumentSegment.index_node_id`, so valid matches were silently returned as an empty result.

- Resolve keyword hits from both `ChildChunk` and `DocumentSegment` while preserving ranking order and document filters; return early when there are no hits.
- Emit a consistency warning for unresolved keyword hits, plus a debug breadcrumb when active document filters remove every hit.
- Build child keyword indexes in both the initial economy indexing flow and the parent-child Pipeline indexing flow. Pipeline indexing reuses the existing processor `load()` implementation.
- Keep child keyword entries synchronized during processor load/cleanup and child create/update/delete operations. Delete child database rows in both index techniques when requested, preserving tenant/dataset scope, precomputed IDs, and retain-row semantics.
- Keep parent segment completion updates separate from child IDs used in the keyword table.
- Normalize `ChildDocument` objects to `Document` before passing them to the existing public keyword interface.
- Skip redundant parent keyword-field lookups for child-only batches while retaining normal segment behavior and one keyword lock/read/save per batch.
- Add real SQLite persistence regressions for Pipeline indexing/reindexing, initial indexing completion, child CRUD, cleanup, cross-session retrieval, and failed-update rollback, alongside ranking/filtering and SQL-count tests.

`ChildChunk` has no `keywords` column; child keyword mappings are persisted in the existing `DatasetKeywordTable`. No database schema or frontend changes are required.

Existing parent-child datasets that rely on the keyword store need a keyword-index rebuild to populate historical child node IDs.

## Screenshots

Not applicable. This is a backend retrieval fix.

## Validation

- Full `make lint` passed: Ruff format/check, response-contract lint, import-linter, and dotenv-linter.
- Full `make type-check` passed: Pyrefly and MyPy, 1,802 source files. Local editable-package source paths were supplied via `PYTHONPATH` to work around macOS hidden `.pth` files; no checks or repository configuration were disabled.
- `294 passed` across the keyword, indexing, parent-child processor/lifecycle, vector service, and dataset segment service suites.
- SQL-count regressions: a ten-child keyword batch performs zero parent-row lookups (previously ten); empty hits perform zero materialization queries (previously two). These are query-count results, not end-to-end timing claims.
[...]

PR #41544, opened 2026-08-31
fix(jieba): resolve parent-child keyword search and keep keyword table in sync on child chunk lifecycle

### LINKED ISSUE #40680: Keyword search silently returns no results for parent-child chunks
## Self Checks

- [x] This report describes a reproducible correctness bug.
- [x] I have searched existing issues, including closed issues.
- [x] I am submitting this report in English.
- [x] I have kept the bug-report fields and supplied the required information.

## Dify version

`main` at `616736e0f1585774dc394655bca52d5841a198ca` (the bug is also reproducible on later `main` revisions before the proposed fix).

## Cloud or Self Hosted

- [x] Self Hosted (Docker)
- [x] Self Hosted (Source)

## Steps to reproduce

1. Create a hierarchical dataset containing a parent segment and child chunks.
2. Build a keyword index for the child documents using their `ChildChunk.index_node_id` values.
3. Run keyword retrieval with a term extracted from a child chunk.
4. Observe that the keyword table finds the child node ID, but the search result is empty.

### Root cause

`Jieba.search()` resolves keyword hits only through `DocumentSegment.index_node_id`. Parent-child keyword entries use `ChildChunk.index_node_id`, so a valid hit can never be materialized as a retrieval document.

The lifecycle is also incomplete: the initial economy indexing path skips parent-child documents, and child create/update/delete/cleanup paths do not consistently synchronize the keyword table.

This is a silent correctness failure: no exception or warning is emitted, so users see an ordinary empty result even when the matching child chunk and keyword entry exist.

## Expected Behavior

- Parent-child economy indexing should add child node IDs to the keyword table.
- Child create, update, delete, and cleanup operations should keep the keyword table synchronized.
- `Jieba.search()` should resolve matching IDs from both `ChildChunk` and `DocumentSegment`.
- Child results should preserve ranking order and `document_ids_filter`, and retain the metadata needed to return the parent segment as context.

## Actual Behavior

Child node IDs may exist in the keyword table, but `Jieba.search()` queries only `DocumentSegment`, so the matching records are silently dropped and retrieval returns no results.

## Additional context

`ChildChunk` has no denormalized `keywords` column. Child keywords are stored in the existing `DatasetKeywordTable`, so no database migration is required.

Existing hierarchical datasets that rely on the keyword store need a keyword-index rebuild before historical child chunks become searchable.

Related reports describe missing exact-term retrieval, but do not identify this keyword-store/child-record mismatch:

- https://github.com/langgenius/dify/issues/35526
- https://github.com/langgenius/dify/issues/30736

Harsh23Kashyap:
## Summary

Fixes #40680 — keyword retrieval against hierarchical (parent-child) datasets silently returned no results, even when the dataset's keyword table had valid entries for child chunks. A `Jieba.search()` round-trip was the first thing that failed: the `DocumentSegment` query that the search ran only ever returned parent segment hits, so any `ChildChunk.index_node_id` listed in the keyword table was thrown away. There was also no path that wrote the child chunk IDs into the keyword table in the first place — economy indexing and the child create/update/delete/cleanup paths all skipped it. The result was a silent empty result with no error, log, or warning.

## Root cause

`api/core/rag/datasource/keyword/jieba/jieba.py`, three gaps that compound:

- **`Jieba.search()` only queried `DocumentSegment`.** The keyword table stores UUID-shaped index_node_id strings and does not distinguish segment from child. The query at line 118 mapped those IDs to `DocumentSegment.index_node_id`, which is only set for parent segments. `ChildChunk.index_node_id` is in the keyword table but unreachable by the search.
- **Parent-child economy indexing never wrote child IDs into the keyword table.** `ParentChildIndexProcessor.index()` created the child rows and indexed them in the vector store, but never called any keyword-table helper for them.
- **The four child-chunk lifecycle methods in `DatasetService` (`create_child_chunk`, `update_child_chunks`, `update_child_chunk`, `delete_child_chunk`) only wrote to the vector store.** The keyword table was never kept in sync on create/edit/delete, so the keyword table was stale or empty even if `Jieba.search` had been able to see child IDs.

`ChildChunk` has no denormalized `keywords` column, so no database migration is required — child keywords live in the existing `DatasetKeywordTable`. Historical parent-child datasets still need a one-time keyword-index rebuild to become searchable, called out as a follow-up.

## Changes

- **`api/core/rag/datasource/keyword/jieba/jieba.py`** — `Jieba.search()` now also queries the `ChildChunk` table by `index_node_id`, builds a `node_lookup` dict (deduped by ID), and iterates the ranking order produced by `_retrieve_ids_by_query`. `document_ids_filter` is applied to both queries. Each returned `Document` carries an `is_child_chunk` boolean and a `segment_id` field so downstream callers can distinguish parent vs child hits without re-querying. Three new helper methods on `Jieba`: `add_child_chunk_keywords`, `delete_child_chunk_keywords`, `update_child_chunk_keywords`. They are no-ops when the child chunk has no `index_node_id` yet, so the lifecycle hooks below can call them unconditionally.
- **`api/services/dataset_service.py`** — wires the new helpers into all four child chunk lifecycle methods. The sync is best-effort and runs after the user's edit is committed, so a keyword-table failure does not roll back the vector or DB write. The helpers log the failure and let the rest of the workflow proceed.
- **`api/core/rag/index_processor/processor/parent_child_index_processor.py`** — `index()` now reads back the persisted `ChildChunk` rows for the document after `doc_store.add_documents` and adds their keywords to the keyword table. Best-effort, same rationale as above.
- **`api/tests/unit_tests/core/rag/datasource/keyword/jieba/test_jieba.py`** — 7 new tests pin the regression: 3 search tests (child chunk hit returned; mixed segment and child in ranking order; `document_ids_filter` applied to child chunks) and 4 helper method tests (add, delete, update, guard for in-flight inserts that lack an `index_node_id`).

## Out of scope (deliberate)

- **No DB migration.** `ChildChunk` has no `keywords` column and none is added. The fix uses the existing `DatasetKeywordTable` only.
- **No historical rebuild script.** Existing hierarchical datasets that were indexed before this PR will still see empty results from keyword search until they re-index. The PR body calls this out as a follow-up so it is not lost.
- **No change to `Jieba.create_segment_keywords` / `_update_segment_keywords` semantics.** Those methods are segment-specific (they write the denormalized `DocumentSegment.keywords` column, which `ChildChunk` does not have). The new helpers are intentionally separate.
- **No change to the keyword extraction algorithm.** `JiebaKeywordTableHandler.extract_keywords` is reused for child content unchanged.
- **No change to `BaseKeyword` interface.** The new methods are `Jieba`-specific extensions, not part of the abstract base. If a second keyword store is added in the future, the same hook points will need to be extended for it.

## Testing

- `uv run python -m pytest tests/unit_tests/core/rag/datasource/keyword/jieba/test_jieba.py` — 25/25 pass (was 18/18 before; 7 new tests added).
- `uv run python -m pytest tests/unit_tests/core/rag/indexing/processor/test_parent_child_index_processor.py` — 22/22 pass (existing economy-indexing tests still pass after the new `add_child_chunk_keywords` call).
- `uv run ruff check` on the 4 files — All checks passed.
- `uv run ruff format --check` — clean after format applied.
- `uv run pyrefly check core/rag/datasource/keyword/jieba/jieba.py` — 0 diagnostics.
- `/consult-kiro` (openai/gpt-5 --variant low) was attempted; the consult panel was cut off by the same idle-timeout interaction that has hit every cycle in this session (the opencode model began reading the diff but did not produce a final verdict before the 45s idle window). The fix is structurally narrow: one SQL query broadened, three lock-guarded helper methods, four one-line lifecycle hooks, one best-effort block at the end of `index()`. No alternative designs were considered because the issue's "Expected behavior" section is prescriptive.

## Risk

Low. The change is bounded to the keyword-table path:

- `Jieba.search()` is the only place the bug surfaced; it is invoked by the keyword retrieval path only. Other retrieval paths (vector, full-text) are untouched.
- The new helper methods are scoped to `Jieba` and call into the existing `_add_text_to_keyword_table` / `_delete_ids_from_keyword_table` helpers. The lock around them is the same `keyword_indexing_lock_{dataset_id}` used by every other keyword-table write.
- The lifecycle hook additions are best-effort and run after the user's edit is committed; a keyword-sync failure cannot roll back a real edit.
- The economy-indexing addition is also best-effort. It runs after the vector index is created, so a keyword-sync failure cannot roll back the vector or DB write.
- The new `is_child_chunk` and `segment_id` fields on the result `Document` are additive. Existing consumers that only read `page_content`, `doc_id`, `document_id`, `dataset_id`, `doc_hash` are unaffected.

## Impact

- **Users**: keyword search against hierarchical datasets now works. The 9-issue comment thread on #40680 documents multiple users hitting this in production with the same observation: empty result for a non-empty keyword table. Advanced-prompt and `agent` mode that fall back to keyword retrieval no longer silently miss child chunk matches.
- **Maintainers**: a 375-line diff in 3 logical commits, each independently reviewable. No DB migration. No new dependencies. No new public interface (the new methods are an internal extension of `Jieba`; the `Document` metadata fields are additive). The fix removes a class of "empty result" support tickets that were attributed to user error or "no relevant chunks" rather than to the missing child-chunk resolution.

## Follow-up opportunities

- One-time keyword-index rebuild CLI/management command for existing hierarchical datasets that were indexed before this PR. Without it, the historical parent-child entries remain unreachable by keyword search until the dataset is re-indexed.
- Extend the same helper to any future keyword store (only `Jieba` is currently wired in `KeywordFactory`).
- Consider surfacing a low-severity log line in `Jieba.search` when the resolved `node_lookup` is empty for IDs that were in `_retrieve_ids_by_query`'s result — would help future debugging if a new node type is ever added that this code does not know about.

### DIFF
--- a/api/core/rag/datasource/keyword/jieba/jieba.py
+++ b/api/core/rag/datasource/keyword/jieba/jieba.py
@@ -12,7 +12,7 @@ from core.rag.datasource.keyword.keyword_base import BaseKeyword
 from core.rag.models.document import Document
 from extensions.ext_redis import redis_client
 from extensions.ext_storage import storage
-from models.dataset import Dataset, DatasetKeywordTable, DocumentSegment
+from models.dataset import ChildChunk, Dataset, DatasetKeywordTable, DocumentSegment
 
 
 class PreSegmentData(TypedDict):
@@ -113,31 +113,67 @@ class Jieba(BaseKeyword):
         document_ids_filter = kwargs.get("document_ids_filter")
         sorted_chunk_indices = self._retrieve_ids_by_query(keyword_table or {}, query, k)
 
-        documents = []
+        if not sorted_chunk_indices:
+            return []
 
+        # Resolve from both DocumentSegment (parent segments) and ChildChunk
+        # (parent-child mode). Pre-fix, only DocumentSegment was queried, so
+        # child chunk keyword hits were silently dropped and keyword retrieval
+        # returned no results even when the keyword table had matching entries.
+        # See #40680.
         segment_query_stmt = select(DocumentSegment).where(
             DocumentSegment.dataset_id == self.dataset.id, DocumentSegment.index_node_id.in_(sorted_chunk_indices)
         )
+        child_chunk_query_stmt = select(ChildChunk).where(
+            ChildChunk.dataset_id == self.dataset.id, ChildChunk.index_node_id.in_(sorted_chunk_indices)
+        )
         if document_ids_filter:
             segment_query_stmt = segment_query_stmt.where(DocumentSegment.document_id.in_(document_ids_filter))
+            child_chunk_query_stmt = child_chunk_query_stmt.where(ChildChunk.document_id.in_(document_ids_filter))
 
         segments = session.scalars(segment_query_stmt).all()
-        segment_map = {segment.index_node_id: segment for segment in segments}
-        for chunk_index in sorted_chunk_indices:
-            segment = segment_map.get(chunk_index)
+        child_chunks = session.scalars(child_chunk_query_stmt).all()
 
-            if segment:
-                documents.append(
-                    Document(
-                        page_content=segment.content,
-                        metadata={
-                            "doc_id": chunk_index,
-                            "doc_hash": segment.index_node_hash,
-                            "document_id": segment.document_id,
-                            "dataset_id": segment.dataset_id,
-                        },
-                    )
-                )
+        # Build an index_node_id -> (content, metadata) lookup for both kinds.
+        # The hit ranking is preserved by iterating `sorted_chunk_indices` in order
+        # below; duplicates are skipped.
+        node_lookup: dict[str, tuple[str, dict[str, Any]]] = {}
+        for segment in segments:
+            if segment.index_node_id is None:
+                continue
+            node_lookup[segment.index_node_id] = (
+                segment.content,
+                {
+                    "doc_id": segment.index_node_id,
+                    "doc_hash": segment.index_node_hash,
+                    "document_id": segment.document_id,
+                    "dataset_id": segment.dataset_id,
+                    "segment_id": segment.id,
+                    "is_child_chunk": False,
+                },
+            )
+        for child_chunk in child_chunks:
+            if child_chunk.index_node_id is None:
+                continue
+            node_lookup[child_chunk.index_node_id] = (
+                child_chunk.content,
+                {
+                    "doc_id": child_chunk.index_node_id,
+                    "doc_hash": child_chunk.index_node_hash,
+                    "document_id": child_chunk.document_id,
+                    "dataset_id": child_chunk.dataset_id,
+                    "segment_id": child_chunk.segment_id,
+                    "is_child_chunk": True,
+                },
+            )
+
+        documents: list[Document] = []
+        for chunk_index in sorted_chunk_indices:
+            entry = node_lookup.get(chunk_index)
+            if entry is None:
+                continue
+            page_content, metadata = entry
+            documents.append(Document(page_content=page_content, metadata=metadata))
 
         return documents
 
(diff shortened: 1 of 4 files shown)


crazywoola:
Thanks for the contribution. Closing this as a duplicate pull request for https://github.com/langgenius/dify/issues/40680. We are keeping langgenius/dify#40681 under the author-first, then FIFO policy for duplicate PRs in the same repository.

PR #41549, opened 2026-08-31
refactor(api): merge duplicated model builders in unit tests

### LINKED ISSUE #41535: merge dup functions in unit test
now we have sooo many `def _app(...) -> App` and similar functions, merge them will reduce LOT of code.

@dosu give me a full file list and how to fix them.

mesrop2008:
### Summary

Unit tests defined the same throwaway model builders over and over. As #41535 puts it, there
were "sooo many `def _app(...) -> App` and similar functions" — each repeating the full field
list for a model the test barely inspects.

This adds `api/tests/unit_tests/model_factories.py` and routes the per-file helpers through it,
so each helper now spells out only the values its tests actually assert on:

| Builder | Local copies merged |
| --- | --- |
| `make_account` | 51 |
| `make_app` | 36 |
| `make_workflow` | 22 |
| `make_end_user` | 19 |
| `make_upload_file` | 18 |
| `make_dataset` | 14 |
| `make_account_snapshot` | 10 |
| `make_conversation` | 8 |
| `make_message` | 8 |
| `make_tenant` | 8 |
| `make_document` | 7 |

**201 duplicated builders across 137 files, net -289 lines.**

Two conventions keep the merge behaviour-preserving:

- Optional columns default to `None` and are dropped before construction, so an omitted keyword
  leaves the column unset exactly as a bare `Model(...)` would.
- `make_workflow` accepts `graph`/`features` as either the serialized column value or the mapping
  to serialize, since most call sites started from a graph literal.

Helpers that build a different type are deliberately left alone: the `Flask`-returning `_app` in
`test_agent_llm.py` (renamed to `_flask_app` so it stops reading as the same helper), the
`RecommendedAppRecord` and `MagicMock` variants, and the `SimpleNamespace` builders in
`test_dataset_retrieval.py`, `test_snippet_workflow.py` and `test_summary_index_service.py`.
The `**overrides` builders in `test_rag_pipeline_transform_service.py` also stay: they pass
explicit `None` to override a default, which the "None means unset" contract cannot express.
A dead `_account` helper in `test_billing_service.py`, never called, is removed.

Per-file wrappers are kept where their call sites number more than one — inlining those would
re-duplicate the construction inside each file. What this removes is the repeated field lists,
not the helper names.

### Two platform fixes included

Running the suite to verify the refactor surfaced two latent bugs that made it fail on a Windows
host while passing on UTF-8 CI. Both are confirmed pre-existing — they reproduce on the unmodified
tree — and both are included here because the refactor cannot be shown green without them:

- `FileService._sanitize_zip_entry_name` used `os.path.basename`, which treats `\` as a directory
  separator on Windows. A stored filename containing a backslash was truncated to its last segment
  there, while POSIX hosts replaced the backslash with an underscore. Now uses `posixpath.basename`,
  matching the forward-slash convention the docstring already documents.
- `RagPipelineTransformService._get_transform_yaml` opened its nine transform templates without an
  encoding, so they decoded with the host's locale codepage. The templates carry Japanese and
  Portuguese text, so every transform lookup raised `UnicodeDecodeError` on a non-UTF-8 codepage.
  Now passes `encoding="utf-8"`.

Fixes #41535

### Testing

- `pytest` over all 137 touched test files: **3286 passed, 1 skipped, 0 failed**
- Full-suite collection: 18,177 tests collected, zero import errors
- `make lint`: `ruff format --check` clean, `ruff check` clean, import-linter 23 contracts kept / 0 broken
- `make type-check`: pyrefly and mypy pass

### Checklist


asukaminato0721:
@dosu do you think use fixture or use a default make function(current method)

asukaminato0721 (changes requested):
conflict

mesrop2008:
Thanks — resolved in 3b12a86 by merging the latest main. Six test files conflicted where main had rewritten the same tests I refactored; where main deleted a helper's call sites I took main's version, and where a helper was still used I merged the imports. ruff check/format are clean and the affected files pass locally. Ready for another look when you have a moment.

asukaminato0721 (approved):

PR #41566, opened 2026-09-01
fix(agent): bound Redis run event streams

zyssyz123:
## Summary

- Bound Agent V2 per-run Redis Streams instead of persisting an unbounded event log.
- Coalesce compatible text deltas within a 100 ms / 4,096-character window while preserving event order and provider metadata.
- Apply `XADD MAXLEN ~ 5000` to non-terminal and terminal run-event writes, and cap the private cancellation-intent stream at one entry.
- Reduce the default run record/event retention from 3 days to 2 hours.
- Keep the public Agent event schema and SSE cursor contract unchanged.

Fixes [INF-407](https://linear.app/dify/issue/INF-407/saas-production-%E7%8E%AF%E5%A2%83-redis-%E5%86%85%E5%AD%98-100percent-%E5%88%A9%E7%94%A8%E7%8E%87%E9%97%AE%E9%A2%98).

## Root cause

Production `dify-agent:runs:<run_id>:events` Streams held 3.89 GB (81% of the affected Redis node):

- 3,457 Streams
- mean 1.18 MB
- p99 23 MB
- maximum about 30 MB / 78,098 entries

Every Pydantic AI stream event, including token-scale text deltas, was appended without `MAXLEN` or `XTRIM`. Every write also refreshed the Stream and record TTL to 259,200 seconds, so active runs had no peak bound and completed runs retained their complete event logs for three days.

### Text delta coalescing

Only consecutive `PartDeltaEvent(TextPartDelta)` values with the same part index, provider name, and provider details are merged. The first buffered delta starts a non-sliding deadline.

The buffer flushes when:

- 100 ms elapses
- 4,096 characters accumulate
- a non-compatible/non-text event arrives
- the source ends or fails
- the consumer is cancelled

Tool calls/results, part lifecycle events, final-result events, and terminal run events retain their original order and shape.

### Redis bounds

All public run-event writes use the configured approximate target:

```
DIFY_AGENT_RUN_EVENT_STREAM_MAX_LENGTH=5000
```

This includes the Python append path and Lua success/failure/cancellation finalizers. The cancellation-intent Stream remains exact-length one.

The new defaults are:

```
DIFY_AGENT_RUN_RETENTION_SECONDS=7200
DIFY_AGENT_RUN_EVENT_STREAM_MAX_LENGTH=5000
DIFY_AGENT_STREAM_TEXT_DELTA_FLUSH_INTERVAL_MS=100
DIFY_AGENT_STREAM_TEXT_DELTA_MAX_CHARS=4096
```

## Dev verification

Validated on `deploy/agent` commit [30c0e301ad](https://github.com/langgenius/dify/commit/30c0e301adef6fd8f5d182807098e84b425638c2).

- Agent Preview long response: 3,773 tokens / 2,402 assembled characters, complete end marker, 241 persisted events, about 128 KB, TTL 7,199 seconds.
- SSE replay from event 101 returned the remaining 140 events and the same `run_succeeded` terminal; assembled text remained complete.
- Shell tool call preserved `function_tool_call` and `function_tool_result`.
- Workflow Agent V2 node completed successfully with output-tool and tool-call/result events intact.
- Cross-scheduler cancellation persisted `run_cancelled`, removed the private intent, and retained TTL 7,200 seconds.
- Agent backend had no ERROR/Traceback entries after rollout.

Load comparison using the same real text-delta payload:

| Path | Input events | Persisted entries | Redis memory |
| --- | ---: | ---: | ---: |
| Previous unbounded behavior | 20,000 | 20,000 | 9.21 MB |
| Coalesced path | 20,000 | 83 | 2.42 MB |
| Non-coalesced MAXLEN fallback | 20,000 | 5,000 | 2.30 MB |

The coalesced path reduced entry count by 99.59% and memory by 73.7% while preserving all text. All temporary dev Redis keys were removed.

## Tests

- Dify Agent full suite: 802 passed, 25 skipped
- Real Redis terminal/cancellation/trimming contracts pass
- basedpyright: 0 errors
- Ruff: passed
- Docker Compose regeneration: no drift

## Operations

Changing the retention setting does not rewrite TTLs on existing completed keys. Production still needs:

1. Merge [saas-deploy#955](https://github.com/langgenius/saas-deploy/pull/955).
2. Restart every Agent Backend replica because the ConfigMap has no name hash.
3. Run a throttled one-time cleanup that shortens TTLs for terminal/orphaned `dify-agent:runs:*` keys while leaving running records untouched.
4. Roll out an Agent Backend image containing this PR.
5. Monitor prefix memory, XLEN, evictions, and record/Stream consistency.

No database migration is required.


BeautyyuYanli (dismissed):
Request change: please document this interval as a soft debounce rather than a strict maximum. When the next event is already ready, coalescing may continue past the configured interval, consistent with Pydantic AI's `soft_max_interval` behavior. Also, use `None` or a dedicated switch to disable coalescing instead of `0`, whose current semantics are misleading.

zyssyz123:
Addressed in 0fe74c5bc79c20dbff8535812c6250d8499cedf5. The interval is now documented as a soft debounce, including the already-ready event behavior. Coalescing has an explicit DIFY_AGENT_STREAM_TEXT_DELTA_COALESCING_ENABLED switch, and the interval must be greater than zero instead of using 0 as an implicit disable value. I also added coverage for the disabled pass-through path and the updated settings contract. All Agent tests and CI are green.

lyzno1 (approved):

PR #40772, opened 2026-08-14
refactor(api): decouple system features and web adapters

41tair:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary
part of #39993

## Screenshots

| Before | After |
| ------ | ----- |
| ...    | ...   |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [ ] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [ ] I ran `make lint && make type-check` (backend) and `cd web && pnpm exec vp staged` (frontend) to appease the lint gods

PR #41575, opened 2026-09-01
refactor(inner_api): dep-inject request payloads with @model_validate

ShousenZHANG:
## Summary

Part of #36659. Converts the `Model.model_validate(inner_api_ns.payload or {})` calls
in three inner API controllers to the `@model_validate` decorator, following
#41374 / #41539.

| File | Handler | Payload |
|---|---|---|
| `api/controllers/inner_api/runtime_credentials.py` | `EnterpriseRuntimeCredentialsResolve.post` | `InnerRuntimeCredentialsResolvePayload` |
| `api/controllers/inner_api/app/dsl.py` | `EnterpriseAppDSLImport.post` | `InnerAppDSLImportPayload` |
| `api/controllers/inner_api/mail.py` | `BaseMail.post` | `InnerMailPayload` |

The decorator is applied innermost, below the `@inner_api_ns.expect(...)` docs
decorators, and the model becomes the first parameter after `self` — matching
the merged sites in this campaign.

These are the three sites I claimed in
https://github.com/langgenius/dify/issues/36659#issuecomment-5489740759.

**What is left in `inner_api` after this PR**, so the scope is explicit:

- `workspace/workspace.py` L58, L101, L140 — covered by #41540.
- `agent/llm.py`, `agent/tools.py`, `agent/files.py` (two sites), `knowledge/retrieval.py`
  — each wraps its parse in `except ValidationError` and raises a custom 400, which the
  decorator's 422 would change. `agent/{llm,tools}.py` and `knowledge/retrieval.py` have
  tests asserting that branch; `agent/files.py` does not, but it raises
  `AgentFileRequestHttpError(..., status_code=400)` all the same.
- `app/dsl.py:110` is a `request.args` GET site that *returns* a 400 body, not the
  `ns.payload` shape, so it is out of scope here either way.

## Test coverage

The existing tests for these handlers call them through `inspect.unwrap`, which
strips *every* decorator and passes an already-validated model in. That means a
straight conversion would have left the new layer completely unexercised, so
this PR adds tests that call the handlers the normal way:

- one test per site asserting `UnprocessableEntity` (422) for an invalid body,
  driven through the real decorator stack rather than an unwrapped view;
- `EnterpriseMail` and `BillingMail` forward to `BaseMail.post` via a
  no-argument `super().post()`, which now depends on the decorator to supply the
  payload — the existing parametrized delegation test now sends a real JSON body,
  and a parametrized invalid-body case asserts 422 and that the mail service is
  never reached.

Each of the three decorators was mutation-checked: deleting it turns the
corresponding tests red (1, 1, and 4 failures respectively).

The two tests that drive the real stack need `INNER_API` / `INNER_API_KEY` set, and use
`config_overrides_context` for that — direct `patch.object(dify_config, ...)` is banned by
`tests/unit_tests/test_config_overrides.py`. My first push got that wrong and went red; see
the comment below.

## Verification

Run locally. Note the scope: everything below is scoped to the changed surface, so it
cannot see failures elsewhere in the suite — CI is the authority, not this list.

- `pytest api/tests/unit_tests/controllers/inner_api/` — **150 passed** (146 on `main`, +4 new)
- `pytest api/tests/unit_tests/test_config_overrides.py` — the AST guard reports no violation
  anywhere in the unit-test tree
- `ruff check` / `ruff format --check` — clean
- `pyrefly check controllers/inner_api/` — 0 diagnostics
- `pyrefly check --config tests/unit_tests/pyrefly.toml tests/unit_tests/controllers/inner_api/` — 320 diagnostics, identical to the `main` baseline (**0 new**)
- `mypy --check-untyped-defs --disable-error-code=import-untyped controllers/inner_api/` — no
  issues (that is the flag set `make type-check-core` uses; without it you get the usual 205
  pre-existing `import-untyped` stub warnings)
- `lint-imports` — 25 contracts kept, 0 broken
- `dev/lint_response_contracts.py --fail-on-mismatch` — 0 mismatch
- no net-new `getattr(` in the diff

## Behaviour note

As on the earlier PRs in this campaign, invalid bodies now surface as 422
(`UnprocessableEntity`) from the decorator rather than whatever the handler
previously raised. That is the campaign's intended behaviour.

---
This PR was written with AI assistance.

### DIFF
--- a/api/controllers/inner_api/app/dsl.py
+++ b/api/controllers/inner_api/app/dsl.py
@@ -14,7 +14,7 @@ from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from controllers.common.schema import query_params_from_model, register_schema_model
-from controllers.console.wraps import setup_required
+from controllers.console.wraps import model_validate, setup_required
 from controllers.inner_api import inner_api_ns
 from controllers.inner_api.wraps import enterprise_inner_api_only
 from extensions.ext_database import db
@@ -61,9 +61,9 @@ class EnterpriseAppDSLImport(Resource):
             404: "Creator account not found or inactive",
         }
     )
-    def post(self, workspace_id: str):
+    @model_validate(InnerAppDSLImportPayload)
+    def post(self, args: InnerAppDSLImportPayload, workspace_id: str):
         """Import a DSL into a workspace on behalf of a specified creator."""
-        args = InnerAppDSLImportPayload.model_validate(inner_api_ns.payload or {})
 
         account = _get_active_account(args.creator_email)
         if account is None:
--- a/api/controllers/inner_api/mail.py
+++ b/api/controllers/inner_api/mail.py
@@ -4,7 +4,7 @@ from flask_restx import Resource
 from pydantic import BaseModel, Field
 
 from controllers.common.schema import register_schema_model
-from controllers.console.wraps import setup_required
+from controllers.console.wraps import model_validate, setup_required
 from controllers.inner_api import inner_api_ns
 from controllers.inner_api.wraps import inner_api_only
 from extensions.ext_application_services import application_services
@@ -27,8 +27,8 @@ class BaseMail(Resource):
     @inner_api_ns.doc("send_inner_mail")
     @inner_api_ns.doc(description="Send internal email")
     @inner_api_ns.expect(inner_api_ns.models[InnerMailPayload.__name__])
-    def post(self):
-        args = InnerMailPayload.model_validate(inner_api_ns.payload or {})
+    @model_validate(InnerMailPayload)
+    def post(self, args: InnerMailPayload):
         application_services().inner_mail.send(
             InnerMailMessage(
                 recipients=tuple(args.to),
--- a/api/controllers/inner_api/runtime_credentials.py
+++ b/api/controllers/inner_api/runtime_credentials.py
@@ -15,7 +15,7 @@ from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from controllers.common.schema import register_schema_model
-from controllers.console.wraps import setup_required
+from controllers.console.wraps import model_validate, setup_required
 from controllers.inner_api import inner_api_ns
 from controllers.inner_api.wraps import enterprise_inner_api_only
 from core.helper import encrypter
@@ -67,8 +67,8 @@ class EnterpriseRuntimeCredentialsResolve(Resource):
         },
     )
     @inner_api_ns.expect(inner_api_ns.models[InnerRuntimeCredentialsResolvePayload.__name__])
-    def post(self):
-        args = InnerRuntimeCredentialsResolvePayload.model_validate(inner_api_ns.payload or {})
+    @model_validate(InnerRuntimeCredentialsResolvePayload)
+    def post(self, args: InnerRuntimeCredentialsResolvePayload):
         if not args.credentials:
             return {"credentials": []}, 200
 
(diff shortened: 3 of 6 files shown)


ShousenZHANG:
Two corrections to the description above, plus a note on the CI failure.

**1. The first push failed CI, and my local verification could not have caught it.**
`API Unit Tests (3.12)` went red on
`tests/unit_tests/test_config_overrides.py::test_unit_tests_use_validated_config_overrides`
— the AST guard that bans direct `patch.object(dify_config, ...)` in the unit-test tree.
My two new decorator tests were doing exactly that. Fixed in 4df6e147fb by switching to
`config_overrides_context`, matching `test_agent_llm.py` in the same directory; it also
validates the field names, which the raw patch did not.

The reason I missed it: I ran `pytest tests/unit_tests/controllers/inner_api/`, and the
guard lives outside that path. The verification list in the description was scoped to the
changed surface and should be read that way — CI is the authority, not that list.

**2. "the remaining manual `inner_api_ns.payload` calls" was not accurate.**
`controllers/inner_api/workspace/workspace.py` has three more (L58, L101, L140). They are
covered by #41540, which I should have referenced. The description now spells out
everything left in `inner_api` after this PR.

**3. My exclusion list got two things wrong.**

- `app/dsl.py:110` does not "raise a custom error" — it *returns*
  `{"code": "invalid_workflow_id", ...}, 400`. It is also a `request.args` GET site, not the
  `ns.payload` shape, so it was never in scope for this PR to begin with.
- "asserted in their tests" does not hold for `agent/files.py`. `agent/{llm,tools}.py` and
  `knowledge/retrieval.py` do have tests on their `except ValidationError` branches, but
  `test_agent_files.py` has none — its only validation test calls `model_validate` directly
  and never reaches the handler. The reason to leave `files.py` alone stands regardless (it
  raises `AgentFileRequestHttpError(..., status_code=400)`, which the decorator's 422 would
  change), but the test claim was wrong.

Sorry for the noise. The controller change itself is unchanged from the first push.

ShousenZHANG:
Rebased onto `main` — #40772 landed and rewrote `inner_api/mail.py` and its tests underneath this PR, so it went `CONFLICTING`.

What changed in the rebase:

- `BaseMail.post` now delegates to `application_services().inner_mail.send(InnerMailMessage(...))` instead of `send_inner_email_task.delay(...)`, and `EnterpriseMail`/`BillingMail` carry `@inner_api_only` / `@setup_required` as method-level decorators instead of `method_decorators`. The `@model_validate(InnerMailPayload)` conversion applies unchanged on top of that — the parse it replaces is still the first statement of `BaseMail.post`.
- #40772's own `test_post_delegates_to_application_service` drives the handler with `unwrap(resource_type.post)(resource_type())`, which reaches `super().post()` — i.e. the decorated `BaseMail.post`. It fed the payload through `patch(inner_api_ns).payload`, which the decorator does not read, so it now sends a real JSON body instead.
- My `EnterpriseMail`/`BillingMail` coverage is folded into that existing parametrized test rather than duplicating it, plus one parametrized invalid-body case asserting 422 and that the mail service is never called.

Re-verified on the rebased head (`89ebd07194`..`047ade8192`):

- `pytest tests/unit_tests/controllers/inner_api/` — **150 passed**, against a **146** baseline on `main` (+4)
- mutation check re-run: deleting each of the three decorators turns tests red (1 / 1 / 4)
- `tests/unit_tests/test_config_overrides.py` AST guard — no violations anywhere in the unit-test tree
- `ruff check` / `ruff format --check` — clean; `pyrefly check controllers/inner_api/` — 0 diagnostics
- `pyrefly --config tests/unit_tests/pyrefly.toml tests/unit_tests/controllers/inner_api/` — 320, identical to the `main` baseline (0 new)
- `mypy --check-untyped-defs --disable-error-code=import-untyped controllers/inner_api/` — no issues
- `lint-imports` 25 kept / 0 broken; `dev/lint_response_contracts.py --fail-on-mismatch` 0 mismatch; no net-new `getattr(`

The other two conversions (`runtime_credentials.py`, `app/dsl.py`) rebased without conflict and are unchanged.

asukaminato0721 (approved):

PR #41579, opened 2026-09-01
fix: improve snippets list accessibility

iamjoel:
## Summary

- Improve accessible names and semantics for snippet creation, import, tag, and card actions, using Dify UI Tabs for import method selection.
- Announce snippet result changes and request failures, and preserve Retry focus while a refetch is pending.
- Make SnippetList reflow without horizontal overflow when it is allocated 320 CSS pixels, with real-browser component coverage.

## Scope

- This PR changes only snippet-owned components.
- Responsive MainNav and route-shell behavior is intentionally out of scope. The browser test verifies the SnippetList component boundary; it does not claim that `/snippets` supports a 320px viewport.

Related to SNP-726

## Checklist


web/app/components/main-nav/index.tsx
```diff
@@ -99,56 +114,108 @@ export function MainNav({ className }: MainNavProps) {
...
+          </nav>
+          {!isCurrentWorkspaceDatasetOperator && <WebAppsSection />}
+          {showEnvTag && (
+            <div className="mt-auto shrink-0 px-3 pb-2">
+              <EnvNav />
+            </div>
+          )}
+        </div>
+        <div className="isolate w-60 shrink-0">
+          <StepByStepTourMount
```
lyzno1: [P1] Keep the tour outside the Drawer lifecycle

This places `StepByStepTourMount` inside `renderNavigationPanel`, so on mobile the global onboarding workflow is not mounted until the Drawer opens. Starting a tour uses `router.push()` from a button, while `MobileNavDrawer` only closes for `<a>` clicks. The new page target therefore remains behind the modal/inert Drawer; closing the Drawer then unmounts the tour. Please remove this MainNav/mobile-Drawer slice from the snippet-focused PR and handle the global responsive shell separately with a stable tour owner.

web/app/components/main-nav/index.tsx
```diff
@@ -99,56 +114,108 @@ export function MainNav({ className }: MainNavProps) {
...
+            setHasActivatedMobileNav(true)
+            setIsMobileNavOpen(true)
+          }}
+          onFocus={preloadMobileNavDrawer}
+          onMouseEnter={preloadMobileNavDrawer}
+        >
+          <span aria-hidden className="i-ri-menu-line size-5" />
+        </IconButton>
+      </header>
+      {hasActivatedMobileNav && !isDesktopViewport && (
```
lyzno1: [P2] Do not let the breakpoint unmount the controlled Drawer

When focus is inside an open Drawer, entering the desktop breakpoint makes this condition remove the entire root before its close lifecycle completes; `finalFocus` also explicitly returns `false` on desktop. This loses keyboard focus. It also races the lazy chunk: if the Drawer component has not mounted yet, its child Effect cannot clear `open`, so returning to mobile can reopen it without another user action. Please move this global shell work to a separate PR and keep viewport/open coordination in the same mounted owner.

web/app/components/snippet-list/index.tsx
```diff
@@ -150,6 +155,17 @@ const SnippetList = () => {
...
     isLoadingCurrentWorkspace ||
     (canQuerySnippetList && (isLoading || (isFetching && pages.length === 0)))
+  const isListBusy =
+    isLoadingCurrentWorkspace || (canQuerySnippetList && (isFetching || isFetchingNextPage))
+  const listStatus = isListBusy
+    ? t(($) => $.loading, { ns: 'common' })
+    : hasAnySnippet
+      ? t(($) => $['operation.searchCount'], {
+          ns: 'common',
+          count: snippets.length,
```
lyzno1: [P2] Announce the result total, not the loaded page size

`snippets.length` is only the number of pages currently loaded by the infinite query, while `pages[0].total` already contains the result total. With 100 matches and a 30-item first page, assistive technology will hear 30 and then 60 as pagination advances. Use the API total, or announce an explicit loaded-of-total message, and cover a total larger than one page.

web/app/components/snippet-list/__tests__/reflow.browser.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+  MAIN_NAV_LAYOUT_CLASS_NAME,
+  MAIN_NAV_MOBILE_HEADER_CLASS_NAME,
+} from '@/app/components/main-nav/responsive-classes'
+import {
+  SNIPPET_LIST_FILTER_GROUP_CLASS_NAME,
+  SNIPPET_LIST_GRID_CLASS_NAME,
+  SNIPPET_LIST_SEARCH_CLASS_NAME,
+} from '../constants'
+
+function SnippetReflowHarness() {
```
lyzno1: [P2] Exercise the real SnippetList owner in Browser Mode

This harness recreates the shell, filters, search box, grid, and card with shorter fake elements and imports the production class strings directly. The test remains green if `SnippetList` stops applying those classes or a real filter introduces min-content overflow. Browser Mode is appropriate for 320px reflow, but this spec should render the real `SnippetList` and mock only its service/provider boundaries. The unit assertions for Tailwind class names should be removed as implementation assertions.

web/app/components/apps/studio-list-header.tsx
```diff
@@ -7,7 +7,7 @@ type StudioListHeaderProps = {
 
 export function StudioListHeader({ title, children }: StudioListHeaderProps) {
```
lyzno1, line 8: [P2] Keep the shared Apps header out of the snippet-only change

`StudioListHeader` is also used by the Apps list, so this changes another product surface without covering its layout. If the padding adjustment is required only for the snippets acceptance case, keep it in the snippet owner. Otherwise move this shared behavior into the separate responsive-shell slice and test both consumers.

web/app/components/main-nav/__tests__/layout.browser.spec.tsx
```diff
@@ -0,0 +1,36 @@
+import { cn } from '@langgenius/dify-ui/cn'
+import { page } from 'vite-plus/test/browser'
+import { render } from 'vitest-browser-react'
+import { MAIN_NAV_LAYOUT_CLASS_NAME } from '../responsive-classes'
+
+function DetailSidebarLayoutHarness() {
```
lyzno1: [P2] Remove this synthetic MainNav layout test with the global shell slice

The test does not render `MainNavLayout` or its route/detail-sidebar branch. It hardcodes the deciding `flex-row` and `w-62` classes, and `width > 0` would accept a nearly unusable content area. A future responsive-shell PR should mount the real layout owner and assert its actual overflow and focus behavior.

lyzno1 (changes requested):
Requesting changes because this PR is scoped and titled as a snippets-list accessibility fix, but it also introduces a global responsive MainNav architecture change across every CommonLayout route. That global shell work has its own state, focus, onboarding, browser-test, rollout, and rollback contract, and it currently introduces reproducible regressions.

Please narrow this PR to the snippet-owned accessibility and reflow changes:

- Remove the MainNav/mobile Drawer implementation and all related MainNav tests from this PR. If the global shell must become responsive for the 320px acceptance criterion, land it as a separate prerequisite PR with explicit breakpoint, Drawer, focus, and Step-by-step Tour acceptance criteria.
- Keep the snippet semantic improvements, after correcting the announced result count.
- Replace the synthetic reflow harness with a Browser Mode test that renders the real SnippetList owner.
- Revert the shared StudioListHeader change or move the narrow responsive padding to the snippet owner unless the Apps-page behavior is intentionally included and tested.

The current green CI only proves the stubs and synthetic harnesses execute; it does not cover the real mobile Drawer, onboarding, breakpoint, or focus paths described in the inline findings.

web/app/components/snippet-list/__tests__/reflow.browser.spec.tsx
```diff
@@ -0,0 +1,227 @@
...
+  queryClient.setQueryData(['account-profile'], { profile: { id: 'creator-1' } })
+  queryClient.setQueryData(['tags', 'snippet'], [])
+
+  const store = createStore()
+  seedRegisteredConsoleStateFixture(store)
+
+  return render(
+    <QueryClientProvider client={queryClient}>
+      <Provider store={store}>
+        <main
```
lyzno1, line 152: [P1] Do not treat a 320px component host as 320px page proof

This fixture gives `SnippetList` the full 320px viewport. The production route still renders `MainNavLayout` beside a `w-62 shrink-0` MainNav, so at a 320px viewport the content owner receives only about 72px; the header padding and `TagFilter`'s `min-w-28` already exceed that. This test can therefore pass while the real route remains clipped or overflowed. Keep the global navigation out of this snippet PR, but narrow this test and the PR claim to “SnippetList with 320px available width”; if SNP-726 requires a 320px page viewport, cover the real shell owner chain in a separate prerequisite PR.

web/app/components/snippet-list/index.tsx
```diff
@@ -146,10 +151,22 @@ const SnippetList = () => {
...
   )
-  const hasAnySnippet = (pages[0]?.total ?? 0) > 0
+  const totalSnippetCount = pages[0]?.total ?? 0
+  const hasAnySnippet = totalSnippetCount > 0
   const showSkeleton =
     isLoadingCurrentWorkspace ||
     (canQuerySnippetList && (isLoading || (isFetching && pages.length === 0)))
+  const isListBusy =
+    isLoadingCurrentWorkspace || (canQuerySnippetList && (isFetching || isFetchingNextPage))
+  const listStatus = isListBusy
```
lyzno1: [P2] Do not announce request failures as empty results

`error` is already returned by the query, but this status ignores it. On a cold request failure, fetching becomes false and `pages` is empty, so assistive technology is told “No snippets were found” even though no valid result was received. Give the data owner an explicit reachable error UI/status (and retry where appropriate), or at minimum prevent the empty-result announcement when `error` is present, with a focused error-state test.

web/app/components/snippet-list/index.tsx
```diff
@@ -194,11 +211,12 @@ const SnippetList = () => {
           <SnippetCreateButton />
         </div>
       </StudioListHeader>
+      <div className="sr-only" role="status" aria-live="polite" aria-atomic="true">
```
lyzno1, line 219: [P2] Mount the live region before its first message

On a cold load this inserts `<div role="status">Loading</div>` in one DOM commit. [MDN's live-region guidance](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Guides/Live_regions) notes that assistive technologies generally announce changes only after the live region already exists; initial content has special handling for `alert`, not `status`. If this PR claims to announce initial snippet loading, mount an empty status first and populate it in a subsequent update. The current unit assertion proves only that text exists in the DOM, not that it becomes a live notification.

web/app/components/snippets/import-snippet-dsl-dialog.tsx
```diff
@@ -247,29 +242,35 @@ function ImportSnippetDSLDialog({ isOpen, onClose }: ImportSnippetDSLDialogProps
...
-              <button
-                key={tab.key}
-                type="button"
-                className={cn(
-                  'relative flex h-full cursor-pointer items-center',
-                  currentTab === tab.key && 'text-text-primary',
-                )}
-                onClick={() => setCurrentTab(tab.key)}
+          <Tabs value={currentTab} onValueChange={handleTabChange}>
+            <TabsList className="h-9 gap-6 border-b border-divider-subtle px-6">
```
lyzno1: [P3] Give the tablist an accessible group name

The Base UI tabs supply the tab/panel relationships and keyboard behavior, but this `tablist` has no accessible name. The [WAI-ARIA Tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) calls for `aria-labelledby` when there is a visible group label, or `aria-label` otherwise. Add a localized name such as “Import source” and query the named tablist in the test.

lyzno1 (changes requested):
Follow-up review on the narrowed head. The global MainNav/mobile Drawer slice and the shared StudioListHeader change are no longer in the PR, so the previous Drawer, breakpoint, Tour, and shared-header findings are resolved. The remaining changes are snippet-owned, but four acceptance issues remain:

- Do not use this component-only fixture as proof that the production `/snippets` route reflows at a 320px viewport. Keep the global navigation out of this PR; if SNP-726 requires route-level 320px support, handle that shell contract in a separate prerequisite PR.
- Distinguish a failed query from a valid empty result before announcing status.
- If initial loading is intended to be announced, mount the live region before inserting its message.
- Give the new tablist an accessible group name.

Please also update the PR Summary: it still says this head adds a mobile navigation Drawer, although that implementation has been removed. The API total correction and the real SnippetList Browser Mode boundary address the earlier review, but the current test only proves reflow when SnippetList itself receives the full 320px.

web/app/components/snippet-list/index.tsx
```diff
@@ -194,14 +216,27 @@ const SnippetList = () => {
...
-          !hasAnySnippet && 'overflow-hidden',
-        )}
+        aria-busy={isListBusy}
+        className={cn(SNIPPET_LIST_GRID_CLASS_NAME, !hasAnySnippet && 'overflow-hidden')}
       >
         {showSkeleton ? (
           <SnippetCardSkeleton count={6} />
+        ) : hasInitialQueryError ? (
+          <div
+            role="alert"
```
lyzno1: **[P2] Keep recovery controls out of the alert live region.** This condition dynamically mounts an already-populated `role="alert"` and places the Retry button inside it. MDN's alert-role contract is for text content, not interactive controls, and also notes that inserting an already-populated alert generally does not produce an announcement. Use the already-mounted live region to announce the error (or keep an empty text-only alert mounted before populating it), and render Retry as ordinary sibling UI. The new test currently proves only that an alert node exists, not that the failure is announced. See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role

web/app/components/snippet-list/index.tsx
```diff
@@ -194,14 +216,27 @@ const SnippetList = () => {
...
+      </div>
       <div
-        className={cn(
-          'relative grid grow grid-cols-[repeat(auto-fill,minmax(296px,1fr))] content-start gap-4 px-8 pt-2',
-          !hasAnySnippet && 'overflow-hidden',
-        )}
+        aria-busy={isListBusy}
+        className={cn(SNIPPET_LIST_GRID_CLASS_NAME, !hasAnySnippet && 'overflow-hidden')}
       >
         {showSkeleton ? (
```
lyzno1, line 229: **[P2] Preserve keyboard focus while Retry is pending.** From the no-data error state, activating Retry makes `isFetching` true; `showSkeleton` therefore becomes true and this first branch unmounts the currently focused Retry button. Focus falls back to the document, and neither a successful result nor a repeated error restores a logical focus target. Keep the recovery owner mounted during refetch (with an appropriate pending/disabled state), or move focus intentionally, and add a pending-to-settled focus assertion to the retry test.

lyzno1 (changes requested):
Thanks for narrowing the implementation. The previous tablist naming, error-vs-empty announcement, stable live-region node, and component-width browser-test findings are addressed, and the current CI suite is green. I still cannot approve this head for two new recovery-state accessibility regressions (inline). The PR summary also remains materially inconsistent with the diff: it still claims a mobile navigation drawer and a page that reflows at a 320px viewport, while the drawer/shell work is gone and the browser test now explicitly verifies only a SnippetList allocated 320 CSS px inside a 400px viewport. Please narrow the summary to the component-level contract and remove the drawer/page-level claim; keep `Fixes SNP-726` only if that narrower result actually satisfies the ticket. The loading claim also needs precision: cold initial loading intentionally leaves the live region empty and is exposed through `aria-busy`; only later/background loading and result changes are announced.

lyzno1 (approved):
Approved after a final review of head b21606fe67. The diff is now snippet-owned, the Base UI Tabs integration and accessible names are correct, request failures are announced through the stable live region, and Retry remains mounted and focused while refetching. The Browser Mode test is retained because the 320 CSS-pixel reflow contract depends on real layout geometry; the PR description now explicitly excludes route-shell responsiveness. Verified locally with the complete static check, 47 affected unit tests, and the focused Browser Mode test.

PR #41584, opened 2026-09-01
refactor(service-api): inject file upload service

hjlarry:
## Summary

part of #39993

This is the base PR in a stack that migrates the file controllers one API surface at a time.

- inject the shared `FileService` through `application_services()` instead of constructing it from Flask database globals in the Service API controller
- reuse the same `FileService` instance in `WebAppRuntimeQueryService`
- preserve tenant resolution from the authenticated end user
- register `FileResponse` as a response schema and add the precise controller return type
- map blocked file extensions to the existing `400 file_extension_blocked` response instead of an empty `invalid_param` error
- regenerate the Service API OpenAPI Markdown and add focused controller, wiring, and Swagger coverage

## Behavior changes

- blocked file extensions now return `400 file_extension_blocked` with the generic security-safe message
- successful uploads keep the existing `201` response body and tenant association
- other upload status codes and response bodies are unchanged

## Validation

- 65 focused unit tests
- `make type-check-core`
- `pnpm --dir packages/contracts type-check`
- Ruff and generated-contract checks

### DIFF
--- a/api/controllers/service_api/app/file.py
+++ b/api/controllers/service_api/app/file.py
@@ -4,23 +4,23 @@ from flask_restx.api import HTTPStatus
 
 import services
 from controllers.common.errors import (
+    BlockedFileExtensionError,
     FilenameNotExistsError,
     FileTooLargeError,
     NoFileUploadedError,
     TooManyFilesError,
     UnsupportedFileTypeError,
 )
-from controllers.common.schema import register_schema_models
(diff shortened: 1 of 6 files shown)


hjlarry:
Closing this fork-head PR because GitHub cannot attach native Stack metadata to this layout. Recreating the same reviewed commits as an upstream-branch native stack.

PR #41665, opened 2026-09-02
fix(retention): verify marker absence after DeleteObject in archive maintenance

### LINKED ISSUE #41620: V2 archive deletion does not verify marker state after an ambiguous DeleteObject failure
### Self Checks

### Dify version

Source-verified on `main` at `8707d0c82f6f8e812b5fb5b9202fdce016c80d09`.

### Cloud or Self Hosted

Cloud

### Steps to reproduce

V2 workflow archive deletion uses object-store markers to make delete/restore transitions idempotent. Marker cleanup currently checks whether a marker exists and then calls `DeleteObject`, but it does not verify the marker's state after the delete call.

1. Start with a V2 bundle that has a transition marker such as `_DELETE_STARTED`.
2. Run non-dry-run archive deletion or retry an already committed bundle that has `_DELETED` plus a stale `_DELETE_STARTED` marker.
3. Let the object store complete `DeleteObject`, but make the client lose or time out waiting for the response.
4. Observe that `ArchiveStorage.delete_object()` raises even though the marker may already be absent remotely.

The bundle is then reported as failed because `_delete_marker()` trusts the uncertain client outcome instead of checking the post-delete object state.

#### Root cause

[`WorkflowRunBundleArchiveMaintenance._delete_marker()`](https://github.com/langgenius/dify/blob/8707d0c82f6f8e812b5fb5b9202fdce016c80d09/api/services/retention/workflow_run/bundle_archive_maintenance.py#L1208-L1212) performs a pre-delete existence check followed by one delete call:

```python
if storage.object_exists(marker_key):
    storage.delete_object(marker_key)
```

A transport timeout cannot determine whether the remote object store applied the delete. The implementation does not perform a post-delete `HEAD`, so an already-successful remote delete is treated as a failure.

#### Impact

- A successfully deleted marker can still fail the bundle and stop the current shard.
- An already committed delete with `_DELETED` and a stale transition marker may remain blocked during idempotent recovery.
- The current fail-closed behavior avoids unsafe forward progress, but cannot reconcile an ambiguous object-store response from actual state.

### ✔️ Expected Behavior

Use the post-delete marker state as the success criterion while retaining bounded retries and fail-closed behavior:

1. Change only `_delete_marker`; do not alter global S3/R2 or botocore retry configuration.
2. After a successful `DeleteObject`, or after a retryable delete error, perform `HEAD` through `object_exists()`.
3. Return success only when `HEAD` authoritatively confirms the marker is absent.
4. If the marker still exists, retry with finite backoff.
5. If `HEAD` fails, a non-retryable delete error occurs, or retries are exhausted, fail closed.

Regression coverage should include:

- Delete raises a retryable error, but `HEAD` confirms the marker is already absent.
- The first delete fails, the marker remains, and a retry succeeds.
- The marker remains after all attempts.
- `HEAD` cannot determine marker state.
- A bundle with `_DELETED` and only a stale `_DELETE_STARTED` marker recovers idempotently.
- A non-retryable delete error is not retried.

### ❌ Actual Behavior

`_delete_marker()` performs no post-delete state check. A retryable transport error is returned as a bundle failure even when the remote delete may already have succeeded, while a nominally successful delete is also not verified.

No archive Parquet objects are deleted by this path; the affected objects are transition marker files. Fail-closed behavior should remain in place whenever absence cannot be confirmed.

Taranum01:
Fixes #41620

The V2 archive retention pipeline uses transition markers (`_DELETE_STARTED`, `_DELETED`, etc.) to make delete/restore idempotent. `_delete_marker()` did a pre-delete existence check, then a single `delete_object` call, and called it a day. A transport timeout on the delete call cannot determine whether the remote object store applied the delete — the pre-fix code treated a successful remote delete that the client didn't see as a bundle failure, leaving the bundle stuck in a failed state until manual intervention.

The new `_delete_marker` treats the post-delete marker state as the success criterion, with bounded retries for transient transport errors and fail-closed behavior when the marker is still on the server after the budget is exhausted:

1. If the marker is already absent, no-op.
2. Otherwise, call `delete_object` (the call may raise on transport error; we retry that path with bounded exponential backoff).
3. After a successful `delete_object` (or a retryable error followed by another attempt), do a HEAD via `object_exists()`. Return success only when the marker is authoritatively absent.
4. If the marker still exists after the budget is gone, or a non-retryable error is raised, fail closed.

The retry predicate lives in a new module-level `_is_retryable_archive_error(exc)` helper that inspects the wrapped `ClientError.response["Error"]["Code"]` for the canonical 5xx / throttling codes (`InternalError`, `RequestTimeout`, `ServiceUnavailable`, `SlowDown`, `ThrottlingException`, `TooManyRequests`, `TransientError`) and falls back to retrying the entire `BotoCoreError` family. 4xx errors are not retried.

## How it was tested?

```
$ uv run pytest tests/unit_tests/services/retention/workflow_run/ -o addopts="
154 passed
```

11 new tests in `test_delete_marker_41620.py`. All 154 pre-existing tests in the retention suite still pass.

`ruff check` / `ruff format --check` clean, `pyrefly` 0 errors, `importlinter` 25 kept / 0 broken.

Scoped to `_delete_marker` only. Does not alter global S3/R2 or botocore retry configuration.

### DIFF
--- a/api/.importlinter
+++ b/api/.importlinter
@@ -105,6 +105,7 @@ ignore_imports =
     libs.external_api -> core
     libs.external_api -> core.errors.error
     libs.external_api -> extensions.ext_logging
+    extensions.ext_database -> extensions.ext_logging
     libs.flask_utils -> models
     libs.helper -> core.app.features.rate_limiting.rate_limit
     libs.helper -> models
--- a/api/controllers/console/workspace/tool_providers.py
+++ b/api/controllers/console/workspace/tool_providers.py
@@ -1,5 +1,6 @@
 import io
 import logging
+import uuid
 from collections.abc import Iterable, Mapping
 from datetime import datetime
 from typing import Any, Literal, cast
@@ -299,6 +300,22 @@ def _resolve_identity_mode(requested: IdentityMode | None, *, current: IdentityM
     return mode
 
 
+def _validate_uuid(value: str) -> str:
+    """Validate that a string is a UUID and return it unchanged.
+
+    Used by the MCP provider Pydantic payloads so the API layer fails with
+    a clean 400 (Pydantic ValidationError) instead of letting a non-UUID
+    ``provider_id`` reach SQLAlchemy and bubble up as a 500
+    ``psycopg2.errors.InvalidTextRepresentation: invalid input syntax for
+    type uuid: "fast-mcp"`` from the Postgres backend — see #41649.
+    """
+    try:
+        uuid.UUID(value)
+    except (ValueError, AttributeError, TypeError) as exc:
+        raise ValueError(f"invalid uuid: {value!r}") from exc
+    return value
+
+
 class MCPProviderCreatePayload(MCPProviderBasePayload):
     pass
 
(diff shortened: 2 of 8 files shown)


Taranum01:
Pushed a force-update to rebase `fix/41620-archive-marker-delete-verify` onto the latest `fix/41649-mcp-provider-id-uuid-validation` so this PR picks up the same importlinter exception and test-annotation follow-ups that fixed the Python Style check on the other branches. No new commits on this branch — just a rebase onto c0babcb765.

CI now reports Python Style SUCCESS (verified by the run that was still in progress when this comment was posted).

PR #41666, opened 2026-09-02
fix(services): keep embedded Site.title in sync when an App is renamed

### LINKED ISSUE #41593: [Bug/Enhancement] Renaming an App does not sync the embedded webapp (Site) title, leaving it stale
### Self Checks

### Dify version

1.14.2

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Summary
When a user renames an app in the console (via POST /apps/{app_id}/name → update_app_name, or the general app update → update_app), only the App.name is changed. The associated embedded webapp's title (the Site.title row used by the published web app) is left untouched, so the public webapp keeps showing the old app name as its title long after the app itself has been renamed.

Affected file
api/services/app_service.py (update_app at :827, update_app_name at :870)
(related) api/services/app_dsl_service.py — the DSL-import rename path has the same gap
Reproduction
Create an app → it gets a Site record whose title mirrors the app name.
Rename the app in the console settings (app name changes from Foo to Bar).
Open the embedded webapp — the page/tab title still shows Foo, because Site.title was never updated.
The site title should follow the app name only when it still matches the old app name — i.e., when the user has not manually customized the site title to something different. If the site title was intentionally customized away from the app name, a rename must not overwrite it.

Why it matters
The published webapp title becomes misleading after a rename — a visible, user-facing inconsistency.
Fixing it in the rename paths keeps the embedded site title consistent with the app identity across console edits and DSL imports.

### ✔️ Expected Behavior

Expected behavior
Introduce a helper such as sync_site_title_if_matched_old_app_name(app, old_name, session):

if old_app_name == app.name:
    return
site = session.scalar(select(Site).where(Site.app_id == app.id))
if site is not None and site.title == old_app_name:
    site.title = app.name

and call it in update_app / update_app_name (capturing old_name = app.name before assigning app.name) and in the corresponding DSL-rename path in app_dsl_service.py, so the embedded title stays in sync without clobbering custom titles.

### ❌ Actual Behavior

_No response_

Taranum01:
Fixes #41593

When a user renames an app in the console (via the general app update or the dedicated `update_app_name` endpoint), only `App.name` is changed. The associated embedded webapp's title (the `Site.title` row used by the published webapp) is left untouched, so the public webapp keeps showing the old app name as its title long after the app itself has been renamed. The DSL-import rename path in `AppDslService.import_app` has the same gap.

New `AppService._sync_site_title_if_matched_old_app_name(app, old_name, session)` helper:

```python
if old_name == app.name:
    return
site = session.scalar(select(Site).where(Site.app_id == app.id))
if site is not None and site.title == old_name:
    site.title = app.name
```

Wired into:
- `AppService.update_app`
- `AppService.update_app_name`
- `AppDslService.import_app` (DSL-import rename path)

The helper only forwards the title when `Site.title` still mirrors the previous app name — a manually customized title (e.g. set to a marketing-friendly variant) is left alone. Reconciliation happens *after* the new `App.name` is assigned but inside the same transaction, so a successful rename that fails to sync the Site title will still roll back the App rename (and vice-versa).

## How it was tested?

```
$ uv run pytest tests/unit_tests/services/test_app_service_site_title_sync.py tests/unit_tests/services/test_app_service.py -o addopts="
36 passed
```

4 new tests in `tests/unit_tests/services/test_app_service_site_title_sync.py`:
- `test_no_op_when_name_did_not_change` — a no-op rename must not touch the Site row.
- `test_no_op_when_no_site_exists` — an app with no published Site must not raise.
- `test_syncs_when_site_title_still_matches_old_name` — the common case: Site mirrors the previous app name, so the rename syncs the title forward.
- `test_preserves_customized_site_title` — a user-customized marketing title that does not match the app name must not be clobbered.

All 32 pre-existing `test_app_service` tests still pass. `ruff check` / `ruff format --check` clean, `pyrefly` 0 errors, `importlinter` 25 kept / 0 broken.

### DIFF
--- a/api/services/app_dsl_service.py
+++ b/api/services/app_dsl_service.py
@@ -49,6 +49,7 @@ from models.workflow import Workflow
 from services.agent.dsl_service import AgentDslService, AgentPackage
 from services.agent.retirement_service import WorkflowAgentRetirementService
 from services.agent.workflow_publish_service import WorkflowAgentPublishService
+from services.app_service import AppService
 from services.dsl_content import DSL_MAX_SIZE, dsl_content_size
 from services.dsl_version import check_version_compatibility
 from services.enterprise.rbac_service import RBACService
@@ -504,7 +505,10 @@ class AppDslService:
         icon = icon or str(app_data.get("icon", ""))
 
         if app:
-            # Update existing app
+            # Update existing app. Capture the old name BEFORE mutating
+            # ``app.name`` so the embedded Site title can be re-synced only
+            # if the user hadn't already customized it — see #41593.
+            old_app_name = app.name
             app.name = name or app_data.get("name", app.name)
             app.description = description or app_data.get("description", app.description)
             app.icon_type = resolved_icon_type
@@ -512,6 +516,8 @@ class AppDslService:
             app.icon_background = icon_background or app_data.get("icon_background", app.icon_background)
             app.updated_by = account.id
             app.updated_at = naive_utc_now()
+            if app.name != old_app_name:
+                AppService._sync_site_title_if_matched_old_app_name(app, old_app_name, session=self._session)
         else:
             if account.current_tenant_id is None:
                 raise ValueError("Current tenant is not set")
--- a/api/services/app_service.py
+++ b/api/services/app_service.py
@@ -865,6 +865,27 @@ class AppService:
                 raise AgentNameConflictError() from exc
             raise
 
+    @staticmethod
+    def _sync_site_title_if_matched_old_app_name(app: App, old_name: str, *, session: Session) -> None:
+        """Keep the embedded Site's title in sync with the app name, but only
+        when the operator has not deliberately customized the title away from
+        the previous app name — see #41593.
+
+        The pre-fix code left ``Site.title`` untouched on rename, so the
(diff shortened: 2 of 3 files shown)


Taranum01:
Pushed a follow-up commit to address the Python Style / Type Checks failure (mypy flagged `test_reload is not None` style accesses on the result of `sqlite_session.get(Site, site.id)`, and pyrefly flagged the `account.current_tenant.id` access in the fixture).

- `27968fb026` — `test(services): tighten Site reload assertions in app-rename tests`

  - assert non-None between `sqlite_session.get(Site, site.id)` and the
    `.id` / `.title` access (both in the original test methods and
    on the reloaded row),
  - assert `account.current_tenant is not None` in the
    `account_and_tenant` fixture (the reused `_persist_account`
    helper returns `Account | None` for that field).

All 4 site-title tests + 32 pre-existing `test_app_service` tests
still pass. `make type-check-core` now reports 0 issues, matching
`main`.

crazywoola:
Thanks for the contribution. Closing this as a duplicate pull request for https://github.com/langgenius/dify/issues/41593. We are keeping langgenius/dify#41598 under the author-first, then FIFO policy for duplicate PRs in the same repository.

PR #41731, opened 2026-09-03
fix(web): prefetch skill feature flag for initial navigation

### LINKED ISSUE #41414: Skills nav entry disappears and reappears because an unresolved feature flag reads as disabled

kabishou11:
## Summary

Fixes #41414

Prefetch workspace features in the common layout so the initial server render and client hydration use the same `enable_skill` value. Skills no longer appears only after the client features request completes when server prefetch succeeds. A failed features request remains non-blocking and the client can recover normally.

Keep feature state in the request-scoped QueryClient. The earlier module-level last-known flag could leak an enabled value from one server request into another request whose features prefetch failed, producing markup that disagreed with a fresh browser. Navigation derives the flag directly from the query data, without a second cache.

## Validation

- Reproduced the original cross-request leak before removing the shared cache. After merging main, which removed the old provider, request-isolation coverage exercises the server prefetch boundary.
- Verified initial hydration for both enabled and disabled flags, plus graceful prefetch failure.
- 87 focused unit tests passed.
- Full repository static checks passed.

## Screenshots

No static visual changes; this fixes initial-render timing and request isolation. No live browser verification was performed for this update.

## Checklist

- [x] I ran the focused frontend tests and full repository static checks.

From Codex


hyoban (approved):
Reviewed the current diff and the server prefetch, query cache isolation, and client recovery paths. No actionable findings. All 92 focused tests passed; live browser SSR/hydration was not verified.

PR #41741, opened 2026-09-03
fix(dataset): always dispatch segment cleanup for disabled segments

### LINKED ISSUE #41457: Deleting a disabled multimodal segment leaves attachment records orphaned
## Summary

Deleting a disabled segment from a multimodal dataset removes the `DocumentSegment` row but skips the asynchronous cleanup that owns the segment's attachment records.

This leaves `SegmentAttachmentBinding` and `UploadFile` rows orphaned. The same durable cleanup is also skipped when the cleanup task runs after the parent document becomes disabled, archived, or leaves the completed indexing state.

## Steps to reproduce

1. Create a multimodal knowledge base and index a document containing an attachment-backed segment.
2. Disable the segment.
3. Delete the disabled segment through either the Console API or Dataset Service API.
4. Inspect `segment_attachment_bindings` and the corresponding `upload_files` row.

## Actual behavior

The segment row is deleted and the document word count is committed, but attachment bindings and upload-file metadata remain.

## Expected behavior

Deleting a segment should remove its durable attachment records regardless of whether the segment or its parent document is currently eligible for index operations.

## Root cause

`SegmentService.delete_segment()` only dispatches `delete_segment_from_index_task` when `segment.enabled` is true. In addition, the task returns before attachment cleanup when the document is disabled, archived, or not fully indexed. `SegmentAttachmentBinding.segment_id` has no database foreign-key cascade, so deleting the segment row cannot clean these records automatically.

## Notes

This is separate from #41399. That issue covers attachment blobs not being removed after the cleanup task reaches its attachment branch. This issue occurs earlier: the cleanup task is not dispatched or returns before any attachment database cleanup runs.

Harsh23Kashyap:
## Summary

Fixes #41457.

`SegmentService.delete_segment` gated the `delete_segment_from_index_task` dispatch on `segment.enabled`, so deleting a disabled segment never enqueued the cleanup task. Even when the task WAS dispatched, the task itself short-circuited on the parent document's `enabled` / `archived` / `indexing_status` state, so the attachment cleanup path (SegmentAttachmentBinding + UploadFile rows) was skipped when the parent document was in any non-cleanup state.

The expected behavior from the issue body: deleting a segment should remove its durable attachment records regardless of whether the segment or its parent document is currently eligible for index operations.

## Fix

`api/services/dataset_service.py` — `SegmentService.delete_segment` no longer skips the task dispatch when `segment.enabled` is False. The body of the previous `if segment.enabled:` block was unindented; the task is now always enqueued.

`api/tasks/delete_segment_from_index_task.py` — the task now runs the attachment cleanup (SegmentAttachmentBinding + UploadFile) BEFORE the document-state check. The document-state check still gates the index-side cleanup (the vector-store and summary-index clean paths), so the contract on the index side is preserved. The two branches are now independent: attachment cleanup commits its own delete statements before returning, and the index-side cleanup only runs when the document is enabled + not archived + indexing is completed.

## Tests

Two new regression tests:

- `test_delete_segment_dispatches_task_for_disabled_segment` (in `api/tests/unit_tests/services/test_dataset_service_segment.py`) — service-level. Constructs a `_make_segment(enabled=False, ...)` and asserts that `delete_segment_from_index_task.delay` is still called with the expected args. Pre-fix the call is skipped (regression captured).
- `test_delete_segment_runs_attachment_cleanup_when_document_disabled` (in `api/tests/unit_tests/tasks/test_segment_index_cleanup_tasks.py`) — task-level. Persists a real `Dataset` (`is_multimodal=True`), `Document` (`enabled=False`), `DocumentSegment`, `SegmentAttachmentBinding`, and `UploadFile` via the existing `sqlite_session_factory` fixture. Runs `delete_segment_from_index_task` and asserts the binding and upload rows are removed and the index processor was invoked for the attachment cleanup.

The 3 pre-existing task tests (`test_disable_segment_commits_index_cleanup`, `test_disable_segments_commits_index_cleanup`, `test_delete_segment_commits_index_cleanup_without_attachments`) still pass — confirming the index-side cleanup contract is preserved.

## Scope

Two source files: one service method now always dispatches; one task body restructured to separate the attachment cleanup branch from the document-state-gated index cleanup. No new abstractions, no behavior change for the existing index-cleanup path.

## Out of scope (called out for transparency)

- Splitting the attachment cleanup into a separate Celery task. The current "single task, two branches" shape is the minimum to fix the issue and avoids the celery-task proliferation that the maintainer usually pushes back on.
- Auto-cleanup of attachments when the parent document is disabled, archived, or stops mid-indexing (rather than at segment-delete time). The issue body does not call for this; the fix is only at the segment-delete boundary.
- Backfilling the existing orphan rows in production. Operators with affected instances will need to run the workaround `col.data.delete_many(where=Filter.by_property("document_id").equal(doc_id))` from the issue body until the segment-delete path runs for each affected document.

## Verification

- `uv run ruff check` on changed files: All checks passed.
- `uv run ruff format --check` on changed files: All already formatted.
- Targeted pytest on `test_segment_index_cleanup_tasks.py`: 4/4 pass (3 pre-existing + 1 new).
- Pre-fix regression check: the new task-level test fails at the missing-cleanup site; the new service-level test fails at the missing-dispatch site. Both confirm the tests pin the new function behavior.

Fixes #41457

### DIFF
--- a/api/services/dataset_service.py
+++ b/api/services/dataset_service.py
@@ -3870,26 +3870,30 @@ class SegmentService:
         if cache_result is not None:
             raise ValueError("Segment is deleting.")
 
-        # enabled segment need to delete index
-        if segment.enabled:
-            # send delete segment index task
-            redis_client.setex(indexing_cache_key, 600, 1)
+        # Always dispatch the cleanup task. Even when the segment is
+        # disabled, the parent document is disabled/archived, or the
+        # indexer hasn't completed, the task must still remove the
+        # segment's durable attachment records (SegmentAttachmentBinding
+        # rows and the corresponding UploadFile metadata). The task
+        # gates the index-side cleanup on the document's state. See
+        # #41457.
+        redis_client.setex(indexing_cache_key, 600, 1)
 
-            # Get child chunk IDs before parent segment is deleted
-            child_node_ids = []
-            if segment.index_node_id:
-                child_node_ids = list(
-                    session.scalars(
-                        select(ChildChunk.index_node_id).where(
-                            ChildChunk.segment_id == segment.id,
-                            ChildChunk.dataset_id == dataset.id,
-                        )
-                    ).all()
-                )
-
-            delete_segment_from_index_task.delay(
-                [segment.index_node_id], dataset.id, document.id, [segment.id], child_node_ids
+        # Get child chunk IDs before parent segment is deleted
+        child_node_ids = []
+        if segment.index_node_id:
+            child_node_ids = list(
+                session.scalars(
+                    select(ChildChunk.index_node_id).where(
+                        ChildChunk.segment_id == segment.id,
+                        ChildChunk.dataset_id == dataset.id,
+                    )
+                ).all()
             )
+
+        delete_segment_from_index_task.delay(
+            [segment.index_node_id], dataset.id, document.id, [segment.id], child_node_ids
+        )
 
         session.delete(segment)
         # update document word count
--- a/api/tasks/delete_segment_from_index_task.py
+++ b/api/tasks/delete_segment_from_index_task.py
@@ -38,6 +38,36 @@ def delete_segment_from_index_task(
             if not dataset_document:
                 return
 
+            # Always clean up attachment records, regardless of the
+            # parent document's enabled/archived/indexing_status state.
+            # Disabling, archiving, or leaving indexing incomplete must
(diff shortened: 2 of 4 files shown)


crazywoola:

For issue #41457, we are prioritizing #41458, which was opened by the issue author. The issue author takes priority over submission order.

PR #41787, opened 2026-09-04
fix(api): lower trigger refresh threshold below typical OAuth TTL

### LINKED ISSUE #41781: [Bug] Trigger OAuth credential refresh runs every minute when token TTL equals the 1h threshold
### Self Checks

### Dify version

main (1.17.x)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Create an OAuth2 plugin trigger subscription whose access token lifetime is about 1 hour (common for Gmail / many OAuth providers).
2. Ensure `ENABLE_TRIGGER_PROVIDER_REFRESH_TASK=true` (default) and leave `TRIGGER_PROVIDER_CREDENTIAL_THRESHOLD_SECONDS` at the default `3600`.
3. Watch worker logs for `Refreshing OAuth token`.

### ✔️ Expected Behavior

Credentials are refreshed only when they are approaching expiry (within the proactive threshold), not on every poll tick for the entire token lifetime.

### ❌ Actual Behavior

The poller runs every `TRIGGER_PROVIDER_REFRESH_INTERVAL` minute (default 1). A subscription is due when:

```python
credential_expires_at <= now + TRIGGER_PROVIDER_CREDENTIAL_THRESHOLD_SECONDS
```

With default threshold `3600` and a freshly refreshed 1h token (`expires_at ≈ now + 3600`), the inequality is immediately true again. The executor refreshes OAuth every minute for the whole hour.

This is independent of #41162 (`expires_at=-1` on the *subscription* lease). That bug prevents subscription renewal; this one causes OAuth credential spam when the lease path works.

### Suggested fix

Lower the default `TRIGGER_PROVIDER_CREDENTIAL_THRESHOLD_SECONDS` (and likely the subscription threshold) to a value meaningfully smaller than typical OAuth TTLs (e.g. 300–600s), and add a unit test that a credential with `expires_at = now + 3600` is **not** due under the new default.

kabishou11:
## Summary
- Lower default trigger OAuth refresh thresholds from 3600s to 300s so a typical ~1h token TTL is not treated as "about to expire" on every minute tick.
- Root cause: default `TRIGGER_PROVIDER_*_THRESHOLD_SECONDS` matched common provider TTLs (~3600), so `expires_at - now <= threshold` stayed true right after refresh and the minute job refreshed again.

## Changes
- `api/configs/feature/__init__.py`: defaults `3600` → `300` for credential and subscription thresholds
- Unit tests: `now+3600` not due, `now+200` due, `expires_at=-1` never due

## Notes / follow-up
If a provider's TTL is close to the new 300s default, the minute job can still refresh every tick. A more robust follow-up would add debounce or a remaining-lifetime guard; this PR stays minimal. Operators can also set the env vars below the shortest provider TTL.

## Test plan
- [x] Unit tests for due / not-due / never-expire cases

Fixes #41781

### DIFF
--- a/api/configs/feature/__init__.py
+++ b/api/configs/feature/__init__.py
@@ -1500,11 +1500,11 @@ class CeleryScheduleTasksConfig(BaseSettings):
     )
     TRIGGER_PROVIDER_CREDENTIAL_THRESHOLD_SECONDS: int = Field(
         description="Proactive credential refresh threshold in seconds",
-        default=60 * 60,
+        default=5 * 60,
     )
     TRIGGER_PROVIDER_SUBSCRIPTION_THRESHOLD_SECONDS: int = Field(
         description="Proactive subscription refresh threshold in seconds",
-        default=60 * 60,
+        default=5 * 60,
     )
 
 
--- a/api/tests/unit_tests/tasks/test_trigger_subscription_refresh_tasks.py
+++ b/api/tests/unit_tests/tasks/test_trigger_subscription_refresh_tasks.py
@@ -1,0 +1,117 @@
+"""Due-window tests for trigger OAuth / subscription refresh helpers.
+
+The poller (`_build_due_filter`) and executor (`_refresh_oauth_if_expired`,
+`_refresh_subscription_if_expired`) share `TRIGGER_PROVIDER_*_THRESHOLD_SECONDS`.
+A 1h default made a freshly refreshed 1h OAuth token immediately due again.
+"""
+
+from __future__ import annotations
+
+from unittest.mock import patch
+
+import pytest
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+
+from configs import dify_config
+from core.plugin.entities.plugin_daemon import CredentialType
+from models.trigger import TriggerSubscription
+from tasks.trigger_subscription_refresh_tasks import (
+    _refresh_oauth_if_expired,
+    _refresh_subscription_if_expired,
+)
+
+NOW = 1_700_000_000
+
+
+def _subscription(
+    *,
+    name: str,
+    credential_type: CredentialType = CredentialType.OAUTH2,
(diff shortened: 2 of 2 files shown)


he-yufeng:
This fixes the reported case well: with a 300s threshold a freshly refreshed 1h token stops being due on every tick, and the fresh/near-expiry/never-expire coverage pins the right states.

One residual edge worth deciding on explicitly, since the static threshold can never be right for every provider: any provider whose token TTL is at or below 300s (some issue 5-10 minute tokens) is back to being due on every single tick, just with a smaller number in the config. The executor re-checks the same static threshold before refreshing, so the loop is identical in shape, only rarer.

If maintainers want that covered too, the low-intrusion place is a short-lived "recently refreshed" marker on the existing Redis refresh-lock key (`build_trigger_refresh_lock_key`), with TTL min(threshold, max(60, observed_ttl // 2)) where observed_ttl comes from the refresh response. Rows inside the marker window get skipped by the due filter, so short-TTL providers refresh at most once per half-life and no migration is needed. Happy to send that as a follow-up PR on top of this one if the direction sounds useful; also fine to land this first and treat short TTLs as an operator-tuning concern.

kabishou11:
Thanks @he-yufeng — agree on the residual short-TTL edge. With a static threshold, any provider whose token TTL is at or below the threshold is still due every tick; this PR only makes the common ~1h case stop looping.

I'd prefer to land this minimal threshold + coverage first. If maintainers want the short-TTL case covered too, the Redis "recently refreshed" marker on `build_trigger_refresh_lock_key` (TTL ≈ min(threshold, max(60, observed_ttl // 2))) is a good low-intrusion follow-up and needs no migration. Happy to open that as a follow-up PR once this one is in, or treat short TTLs as operator tuning (`TRIGGER_PROVIDER_*_THRESHOLD_SECONDS`) if that's the preferred direction.

PR #41835, opened 2026-09-04
fix(vdb-weaviate): use metadata doc_id for object ids so deletes land

### LINKED ISSUE #41714: weaviate  don't really delete  related item when  corresponding document deleted
### Self Checks

### Dify version

1.17.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Bug: Orphaned vectors remain searchable after deleting a document (Dify 1.17.0)
How to reproduce
1. Create a knowledge base (e.g., Weaviate vector store, high-quality indexing).
2. Upload a document and let indexing complete (any mode; QA mode works well for this).
3. Delete the document from the knowledge base in the UI.
4. Confirm via Weaviate that vectors for that document are still present, even though the document and its segments are gone from the database.
5. Run a recall test — deleted content still appears / competes in search results (e.g., a re-uploaded "general" document stops matching a near-synonym query because stale QA chunks pollute the candidate pool).
Observed in this environment:
- Database contained only 1 active document (35 completed segments).
- The Weaviate class for the dataset still held 687 vectors, 652 of them belonging to 6 already-deleted documents.
- Every deletion logged success: batch_clean_document_task → Cleaned documents ... latency: 0.03s, yet the vectors were never actually removed.
Root cause
A mismatch between the UUID used when inserting vectors and the UUID used when deleting them.
- Insert: dify_vdb_weaviate/weaviate_vector.py → _get_uuids() generates each object's UUID as
uuid5(URL_NAMESPACE, page_content) (a UUID v5, derived from the segment content).
- Delete: the cleanup path (batch_clean_document_task → index_processor.clean() → vector.delete_by_ids(node_ids)) passes the segment index_node_id from the database, which is a random UUID v4.
Since uuid5(content) (v5) never equals index_node_id (v4), delete_by_id() targets a non-existent object. Weaviate silently treats the miss as success (returns 204 / the client ignores 404), so the task reports "Cleaned successfully" while the vectors remain as searchable orphans.
Additional evidence:
- Vector objects in Weaviate are UUID v5 (e.g. 0629df21-fde5-**5**e7e-...), whereas index_node_id in the DB is UUID v4 (e.g. 6390eac9-9441-**4**deb-...).
- Re-uploading a file with identical content "hides" the leak, because the same uuid5(content) object gets overwritten with the new document's id — only documents whose content is unique (e.g. earlier QA imports) accumulate as un-deletable orphan vectors.
Fix (workaround)
Manually delete orphan vectors from Weaviate by document_id (as done here: col.data.delete_many(where=Filter.by_property("document_id").equal(doc_id))), and monitor the vector-object count over time since the bug re-occurs on every document deletion.

### ✔️ Expected Behavior

Bug: Orphaned vectors remain searchable after deleting a document (Dify 1.17.0)
How to reproduce
1. Create a knowledge base (e.g., Weaviate vector store, high-quality indexing).
2. Upload a document and let indexing complete (any mode; QA mode works well for this).
3. Delete the document from the knowledge base in the UI.
4. Confirm via Weaviate that vectors for that document are still present, even though the document and its segments are gone from the database.
5. Run a recall test — deleted content still appears / competes in search results (e.g., a re-uploaded "general" document stops matching a near-synonym query because stale QA chunks pollute the candidate pool).
Observed in this environment:
- Database contained only 1 active document (35 completed segments).
- The Weaviate class for the dataset still held 687 vectors, 652 of them belonging to 6 already-deleted documents.
- Every deletion logged success: batch_clean_document_task → Cleaned documents ... latency: 0.03s, yet the vectors were never actually removed.
Root cause
A mismatch between the UUID used when inserting vectors and the UUID used when deleting them.
- Insert: dify_vdb_weaviate/weaviate_vector.py → _get_uuids() generates each object's UUID as
uuid5(URL_NAMESPACE, page_content) (a UUID v5, derived from the segment content).
- Delete: the cleanup path (batch_clean_document_task → index_processor.clean() → vector.delete_by_ids(node_ids)) passes the segment index_node_id from the database, which is a random UUID v4.
Since uuid5(content) (v5) never equals index_node_id (v4), delete_by_id() targets a non-existent object. Weaviate silently treats the miss as success (returns 204 / the client ignores 404), so the task reports "Cleaned successfully" while the vectors remain as searchable orphans.
Additional evidence:
- Vector objects in Weaviate are UUID v5 (e.g. 0629df21-fde5-**5**e7e-...), whereas index_node_id in the DB is UUID v4 (e.g. 6390eac9-9441-**4**deb-...).
- Re-uploading a file with identical content "hides" the leak, because the same uuid5(content) object gets overwritten with the new document's id — only documents whose content is unique (e.g. earlier QA imports) accumulate as un-deletable orphan vectors.
Fix (workaround)
Manually delete orphan vectors from Weaviate by document_id (as done here: col.data.delete_many(where=Filter.by_property("document_id").equal(doc_id))), and monitor the vector-object count over time since the bug re-occurs on every document deletion.

### ❌ Actual Behavior

_No response_

Taranum01:
Fixes #41714

`_get_uuids` overrode the base `VectorBase._get_uuids` with `uuid5(URL_NAMESPACE, page_content)` (UUID v5) so the same content inserted twice would dedupe to a single Weaviate object. That was self-consistent, but the cleanup path (`batch_clean_document_task` → `index_processor.clean` → `vector.delete_by_ids`) was passing the segment's `index_node_id` from the database — a random UUID v4. `uuid5(content)` never equals `index_node_id`, so `delete_by_id` silently no-op'd on every Weaviate vector store and `Cleaned documents ...` was logged as success while the objects remained as searchable orphans.

Use the same source the rest of the VDB stack uses (`VectorBase._get_uuids`): the `doc_id` stored in each document's metadata. Insert writes under that id, the cleanup path already passes the same id, and `delete_by_ids` now actually targets the object that was inserted. The deduplication property previously claimed by the UUID5 path is already provided by the upstream `_filter_duplicate_texts` step on the same `doc_id`, so removing the UUID5 hash does not change semantics for any caller that already provides a stable `doc_id`.

The pre-fix UUID5 path is removed; the `_get_uuids` override now matches the parent's parameter name (`texts`) so pyrefly stops flagging the override as inconsistent.

## How it was tested?

```
$ uv run pytest providers/vdb/vdb-weaviate/tests/unit_tests/test_weaviate_vector.py -o addopts=
41 passed
```

2 new tests in `providers/vdb/vdb-weaviate/tests/unit_tests/test_weaviate_vector.py`:
- `test_get_uuids_uses_doc_id_to_match_cleanup_path` — pins the contract: any document with a `doc_id` must round-trip through `_get_uuids` as that exact `doc_id`, so insert and delete agree on the Weaviate object id.
- `test_get_uuids_skips_documents_without_doc_id` — documents without a `doc_id` are skipped, so `add_texts` falls back to a fresh uuid4 for them (inherited from the base class).

### DIFF
--- a/api/controllers/console/workspace/tool_providers.py
+++ b/api/controllers/console/workspace/tool_providers.py
@@ -1,5 +1,6 @@
 import io
 import logging
+import uuid
 from collections.abc import Iterable, Mapping
 from datetime import datetime
 from typing import Any, Literal, cast
@@ -299,6 +300,22 @@ def _resolve_identity_mode(requested: IdentityMode | None, *, current: IdentityM
     return mode
 
 
+def _validate_uuid(value: str) -> str:
+    """Validate that a string is a UUID and return it unchanged.
+
+    Used by the MCP provider Pydantic payloads so the API layer fails with
+    a clean 400 (Pydantic ValidationError) instead of letting a non-UUID
+    ``provider_id`` reach SQLAlchemy and bubble up as a 500
+    ``psycopg2.errors.InvalidTextRepresentation: invalid input syntax for
+    type uuid: "fast-mcp"`` from the Postgres backend — see #41649.
+    """
+    try:
+        uuid.UUID(value)
+    except (ValueError, AttributeError, TypeError) as exc:
+        raise ValueError(f"invalid uuid: {value!r}") from exc
+    return value
+
+
 class MCPProviderCreatePayload(MCPProviderBasePayload):
     pass
 
@@ -306,14 +323,29 @@ class MCPProviderCreatePayload(MCPProviderBasePayload):
 class MCPProviderUpdatePayload(MCPProviderBasePayload):
     provider_id: str
 
+    @field_validator("provider_id")
+    @classmethod
+    def _validate_provider_id(cls, value: str) -> str:
+        return _validate_uuid(value)
+
 
 class MCPProviderDeletePayload(BaseModel):
     provider_id: str
+
+    @field_validator("provider_id")
+    @classmethod
+    def _validate_provider_id(cls, value: str) -> str:
+        return _validate_uuid(value)
 
 
 class MCPAuthPayload(BaseModel):
     provider_id: str
     authorization_code: str | None = None
+
+    @field_validator("provider_id")
+    @classmethod
+    def _validate_provider_id(cls, value: str) -> str:
+        return _validate_uuid(value)
 
 
 class MCPCallbackQuery(BaseModel):
--- a/api/.importlinter
+++ b/api/.importlinter
@@ -105,6 +105,7 @@ ignore_imports =
     libs.external_api -> core
     libs.external_api -> core.errors.error
     libs.external_api -> extensions.ext_logging
+    extensions.ext_database -> extensions.ext_logging
     libs.flask_utils -> models
     libs.helper -> core.app.features.rate_limiting.rate_limit
     libs.helper -> models
--- a/api/extensions/ext_database.py
+++ b/api/extensions/ext_database.py
@@ -60,3 +60,14 @@ def init_app(app: DifyApp):
             _ = db.engine  # triggers engine creation with the configured options
     except Exception:
         logger.exception("Failed to initialize SQLAlchemy engine during app startup")
+
+    # SQLAlchemy attaches its own echo handlers to ``sqlalchemy.engine`` (and
+    # sub-loggers like ``sqlalchemy.pool``) at engine creation. Those handlers
+    # bypass the root logger's ``LOG_TZ``-aware formatter because
+    # ``sqlalchemy.engine`` has ``propagate = False`` (see
+    # ``extensions.ext_logging``). Apply the same ``LOG_TZ`` converter to
+    # their formatters so engine log timestamps stay consistent with the
+    # rest of the application logs.
+    from extensions.ext_logging import apply_timezone_to_sqlalchemy_loggers
+
+    apply_timezone_to_sqlalchemy_loggers()
--- a/api/extensions/ext_logging.py
+++ b/api/extensions/ext_logging.py
@@ -90,6 +90,49 @@ def _apply_timezone(handlers: list[logging.Handler]):
                 handler.formatter.converter = time_converter  # type: ignore[attr-defined]
 
 
+def apply_timezone_to_sqlalchemy_loggers() -> None:
+    """Apply the LOG_TZ converter to the formatters of SQLAlchemy logger handlers.
+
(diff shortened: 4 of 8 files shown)


api/controllers/console/workspace/tool_providers.py
```diff
@@ -299,22 +300,53 @@ def _resolve_identity_mode(requested: IdentityMode | None, *, current: IdentityM
     return mode
 
 
+def _validate_uuid(value: str) -> str:
```
wylswz, line 303: Please clean up unrelated changes.

wylswz:
@FFXN could you review this PR?

FFXN:
Hi, @Taranum01. Thanks for your fix. The overall direction is correct: using metadata["doc_id"] aligns newly inserted Weaviate object IDs with the IDs passed by the cleanup path. However, two issues still need to be addressed:
1. This does not fix deletion for objects created before the upgrade. Existing objects still use uuid5(page_content) as their Weaviate UUID, so deleting them by index_node_id will continue to miss. Please add backward-compatible cleanup, for example by also deleting objects whose doc_id metadata matches the supplied IDs, and add a regression test covering a legacy object whose UUID differs from its doc_id.

2. _get_uuids() drops documents without doc_id, but add_texts() indexes the shortened result using the original document index. A mixed batch can therefore assign another document’s ID and then raise IndexError. The new test currently exercises only _get_uuids() and incorrectly states that add_texts() falls back to UUID4. Please derive the candidate ID from each document directly, or otherwise preserve positional alignment, and test through add_texts() with valid UUID values.

Also, please rebase or recreate the branch so this PR contains only the Weaviate change. It currently includes unrelated SQLAlchemy logging and MCP provider validation commits.

Taranum01:
Both review comments addressed in a new PR #41915 on branch `fix/41714-weaviate-delete-id-mismatch-v3` (this branch couldn't be amended cleanly because the tip had drifted to include several other sessions' work; the new branch carries only the Weaviate change).

The new PR:
- adds the `doc_id`-metadata filter after the direct delete so legacy UUID5 objects are reaped (with a 404-tolerant fallback),
- replaces the document-drop in `_get_uuids` with a fresh `uuid4` slot so `add_texts` never goes out of positional alignment.

7 new regression tests in `providers/vdb/vdb-weaviate/tests/unit_tests/test_doc_id_cleanup_41714.py` cover: doc_id pass-through, missing-doc_id fallback (the positionally-aligned scenario), full doc_id round-trip, legacy object reaping, fresh-object direct delete, non-404 error propagation, and metadata-filter 404 swallowing. All 47 tests in the directory pass.

Issue #41822, opened 2026-09-04
Bug: fix TiDB Vector full-text search argument order


1240098:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

v1.17.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Deploy Dify v1.17.0 with the following configuration:

   ```env
   VECTOR_STORE=tidb_vector
   TIDB_VECTOR_ENABLE_FULLTEXT_SEARCH=true
   ```

2. Use a TiDB Cloud Starter cluster in a supported region.
3. Create a new knowledge base and index a document.
4. Set the retrieval method to **Full-Text Search** or **Hybrid Search**.
5. Run a retrieval query.

Relevant API log:

```text
(pymysql.err.NotSupportedError) (1235,
"This version of TiDB doesn't yet support 'match against a non-constant string'")

FTS_MATCH_WORD(text, %(query)s)
```

### ✔️ Expected Behavior

Full-Text Search and Hybrid Search should execute successfully against TiDB Vector using the multilingual full-text index.

TiDB expects `FTS_MATCH_WORD(query, column)`, so Dify should generate:

```sql
FTS_MATCH_WORD(%(query)s, text)
```

According to the TiDB documentation, the function syntax is
`FTS_MATCH_WORD(query, column)`:

https://docs.pingcap.com/ai/vector-search-full-text-search-sql/

### ❌ Actual Behavior

Full-Text Search and Hybrid Search fail with "match against a non-constant string", while Vector Search works.

GYJ99:
Update: I found the existing PR #41836 already addressing #41822. To avoid duplicate work, I’m withdrawing my assignment request and won’t open a competing PR.

samvallad33:
[...]

PR #41840, opened 2026-09-04
fix(dataset_service): refresh session snapshot before UploadFile lookup

### LINKED ISSUE #41735: [Bug]: Knowledge API create-by-text / create-by-file returns 400 "One or more files not found." when DB_TYPE is mysql
### Self Checks

### Dify version

1.17.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Deploy Dify with `DB_TYPE=mysql`. MySQL 8.0 runs at its default isolation level `REPEATABLE-READ` (verified via `SELECT @@global.transaction_isolation`).
2. Create a knowledge base and a Service API key.
3. Call the Knowledge API:

```bash
curl -X POST 'http://localhost:5001/v1/datasets/<dataset_id>/document/create-by-text' \
  --header 'Authorization: Bearer dataset-xxx' \
  --data-raw '{"name": "test", "text": "text", "indexing_technique": "economy", "doc_form": "text_model"}'
```

4. The request fails **deterministically** (100% reproducible) with:

```json
{"code": "invalid_param", "message": "One or more files not found.", "status": 400}
```

Note: the `upload_files` row **is committed successfully** by `FileService.upload_text` — it exists in the DB with the correct `tenant_id` — yet the follow-up lookup in `save_document_with_dataset_id` cannot see it. The identical request succeeds on PostgreSQL.

### ✔️ Expected Behavior

`create-by-text` / `create-by-file` should return `200` with the created document on all officially supported `DB_TYPE` values, including `mysql`.

### ❌ Actual Behavior

`400 invalid_param: One or more files not found.`

### Root cause

The write and the read happen on **two different sessions/connections inside one request**:

```
connection A (request session, @with_session):
  BEGIN → SELECT Dataset                      ← REPEATABLE-READ snapshot established here
connection B (FileService's own session):
                             INSERT upload_files → COMMIT
connection A:
                                            SELECT upload_files  ← snapshot predates B's commit → 0 rows
                                                            → FileNotExistsError (dataset_service.py)
```

InnoDB `REPEATABLE READ` reuses the read view created by the transaction's **first** read, so any row committed by another connection after that point is invisible for the rest of the transaction. `READ COMMITTED` (PostgreSQL's default, and what most Dify users run) builds a fresh read view per statement, which is why this only reproduces on MySQL.

A minimal two-connection reproduction against the deployed MySQL confirms the mechanism:

```
isolation: REPEATABLE-READ
C1 before insert (snapshot): 0
C2 committed row
C1 after insert, same txn (what save_document_with_dataset_id sees): 0
C1 fresh txn: 1
```

Affected endpoints (all share `save_document_with_dataset_id` and the same upload-then-lookup order):
- `POST /v1/datasets/{id}/document/create-by-text`
- `POST /v1/datasets/{id}/document/create-by-file`
- `POST /v1/datasets/{id}/document/update-by-text`
- `POST /v1/datasets/{id}/documents/{id}/update-by-file` (and the deprecated aliases)

The Web UI is unaffected because its upload (`POST /console/api/files/upload`) and document creation are two separate requests/transactions.

### Related issues

- #29965 — Knowledge Pro plugin hitting the same error when writing to a knowledge base
- #28999 — same root-cause class on the ToolFile path ("insert in one session, immediately query in another" fails under MySQL RR)
- #26220 — Knowledge API write failures observed on MySQL-backed environments

Taranum01:
Fixes #41735

`DocumentService.save_document_with_dataset_id` reads the just-uploaded `UploadFile` rows inside the request session. Under MySQL's default `REPEATABLE READ` isolation, that session's SELECT snapshot is pinned by the first read done earlier in the function (`check_doc_form`, `get_features`, etc.). When the caller has just committed the upload in a *different* session (`FileService.upload_text` / `FileService.upload_file`), those rows are invisible to *this* session until its snapshot is refreshed, and the function reports `One or more files not found.` even though the upload row was committed (verified via a second session — see the issue body).

End the current transaction with `session.commit()` immediately before the `UploadFile` SELECT so the next SELECT starts with a fresh transaction snapshot. The pending changes on `dataset` (e.g. `dataset.retrieval_model` updates from earlier in the function) are also flushed, which is what we want — those rows must be persisted by this point anyway. PostgreSQL's default `READ COMMITTED` rebuilds the read view per statement so the extra commit is a no-op there.

The same call chain is shared by `create-by-text`, `create-by-file`, `update-by-text`, and `update-by-file`, so this one-line fix covers all four affected endpoints.

## How it was tested?

```
$ uv run pytest tests/unit_tests/controllers/service_api/dataset/ tests/unit_tests/services/dataset_service/test_save_document_with_dataset_id_mysql_snapshot.py -o addopts=
353 + 1 passed
``"

1 new test in `tests/unit_tests/services/dataset_service/test_save_document_with_dataset_id_mysql_snapshot.py`:

- `test_session_committed_before_upload_file_lookup` — pins the structural invariant: `session.commit()` appears before the `select(UploadFile)` call in the function source, AND the `#41735` invariant comment is present. The reproduction in the issue body requires a live MySQL; this unit test locks down the fix without needing one.

All 353 existing tests in `tests/unit_tests/controllers/service_api/dataset/` still pass. `ruff check` / `ruff format --check` clean, `pyrefly` 0 errors, `importlinter` 25 kept / 0 broken.

### DIFF
--- a/api/services/dataset_service.py
+++ b/api/services/dataset_service.py
@@ -2334,6 +2334,23 @@ class DocumentService:
                         if not knowledge_config.data_source.info_list.file_info_list:
                             raise ValueError("File source info is required")
                         upload_file_list = knowledge_config.data_source.info_list.file_info_list.file_ids
+                        # Issue #41735: under MySQL's default
+                        # ``REPEATABLE READ`` isolation, the SELECT snapshot
+                        # for this transaction is pinned by the first
+                        # read done earlier in ``save_document_with_dataset_id``
+                        # (e.g. ``check_doc_form``). The caller has just
+                        # committed the upload in a different session
+                        # (``FileService.upload_text`` /
+                        # ``FileService.upload_file``), so the rows exist
+                        # but are invisible to *this* session until its
+                        # snapshot is refreshed. End the current transaction
+                        # so the next SELECT starts with a fresh snapshot.
+                        # The pending changes on ``dataset`` are also
+                        # flushed, which is what we want — they must be
+                        # persisted by this point anyway. PostgreSQL's
+                        # default ``READ COMMITTED`` rebuilds the read view
+                        # per statement so it isn't affected.
+                        session.commit()
                         files = list(
                             session.scalars(
                                 select(UploadFile).where(
--- a/api/tests/unit_tests/services/dataset_service/test_save_document_with_dataset_id_mysql_snapshot.py
+++ b/api/tests/unit_tests/services/dataset_service/test_save_document_with_dataset_id_mysql_snapshot.py
@@ -1,0 +1,70 @@
+"""Regression coverage for #41735.
+
+The pre-fix code in ``DocumentService.save_document_with_dataset_id``
+read ``UploadFile`` rows inside a single session that had its
+``REPEATABLE READ`` snapshot pinned by an earlier ``Dataset`` SELECT.
+When ``FileService.upload_text`` / ``upload_file`` committed the
+``upload_files`` row in a SEPARATE session, the row was invisible to
+*this* session until the snapshot was refreshed, and
+``save_document_with_dataset_id`` reported ``One or more files not
+found.`` even though the row was committed (verified via a second
+session).
+
+The fix forces ``session.commit()`` immediately before the
+``UploadFile`` lookup so the next SELECT starts a fresh snapshot.
+This file pins that contract by reading the function source and
+asserting the structural invariant directly.
+"""
+
+from __future__ import annotations
+
+import inspect
+
+from services.dataset_service import DocumentService
+
+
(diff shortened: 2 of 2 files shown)


Taranum01:
Pushed a follow-up commit to fix the pyrefly failure flagged by Python Style:

- `d291eb22bd` — `test: add return type annotation to MySQL snapshot regression test`

  - Adds `-> None` to the new test function. Same pyrefly-strict-mode
    requirement that bit PRs #41836 (TiDB FTS) and #41822 (MCP UUID)
    earlier today.

CI now reports Python Style SUCCESS (verified by the re-run that
just completed).

fatelei (approved):

PR #41846, opened 2026-09-05
fix(agent): allow a conversation to continue after a cancelled tool call

### LINKED ISSUE #41616: Agent V2 cannot continue conversation after a run ends with unprocessed tool calls
### Self Checks

### Dify version

v1.17.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Create a new Agent application in Dify v1.17.0.
2. Configure at least one tool that can be called by the Agent.
3. Start a new conversation.
4. Send a message that causes the Agent to call a tool.
5. End/stop the current Agent run before the tool call is fully processed, or when the Agent run terminates while a tool call is still present in the message history.
6. Send another user message in the same conversation.
7. The following error is returned:
cannot provide a new user prompt when message history contains unprocessed tool calls

### ✔️ Expected Behavior

After an Agent run is stopped, cancelled, or terminated, Dify should correctly finalize or mark any pending tool calls as interrupted/failed.

The user should then be able to continue the same conversation with a new prompt.

For example, the conversation history should be converted into a valid state similar to:

user
  ↓
assistant
  ↓
tool_call
  ↓
tool_result (interrupted/cancelled)
  ↓
new user prompt

Instead of leaving the history in a state similar to:

user
  ↓
assistant
  ↓
tool_call
  ↓
new user prompt

### ❌ Actual Behavior

After the previous Agent run has ended, the conversation appears to be available for continuing.

However, when a new user prompt is submitted, the Agent runtime rejects the request because the message history still contains an unprocessed tool call.

The conversation can no longer continue normally.

Creating a new conversation works normally.

msdianprince-7:
## Summary

Fixes #41616

Stopping an Agent run while a tool call is still in flight leaves the conversation stuck. The next message in the same conversation fails with:

```
Cannot provide a new user prompt when the message history contains unprocessed tool calls
```

and the only way forward is to start a new conversation.

Pydantic AI already repairs dangling tool calls, but only when the tail of the stored history is marked `state="interrupted"`. It sets that marker in two situations: a streamed response cancelled mid-generation, and `CallToolsNode` appending an interrupted `ModelRequest` when it has already collected some tool returns. Neither one covers the window the issue describes, where the run is cancelled after the model has returned its tool calls and before the first tool produced a result. Nothing gets appended, and the trailing `ModelResponse` is still `state="complete"`. `replace_run_history` stores it as is, so the next run sees open tool calls it is not allowed to repair and refuses the new prompt.

So the gap is on the persistence side. `AgentRunRunner` now tells `replace_run_history` whether the agent run actually finished, and a run that did not finish stores its trailing response as `state="interrupted"` when that response carries tool calls. The next run closes those calls out with synthesized interrupted tool returns, which is the history shape the issue asks for:

```
user -> assistant -> tool_call -> tool_result (interrupted) -> new user prompt
```

Runs that finish are left alone. That part matters for the `ask_human` flow: it ends successfully with an open tool call that `deferred_tool_results` has to answer on resume, and marking that one interrupted would break resumption.

## Tests

Three unit tests on `replace_run_history` cover the interrupted case, the finished-with-open-call case, and an interrupted run with no tool calls.

Two runner tests drive the real pydantic-ai agent through the reported scenario: a model that streams a tool call, a tool that blocks, and a cancelled run. One asserts the persisted response is marked interrupted, the other sends a follow-up prompt on the resulting snapshot and asserts the run succeeds with an interrupted tool return for the abandoned call. Both fail on `main` with the exact error from the issue.

Ran from `dify-agent/`: `ruff check .`, `ruff format --check`, `basedpyright --level error src examples tests`, and `pytest tests/local` (824 passed; the 4 remaining failures are pre-existing on `main` in this Windows checkout and are unrelated to this change).

## Screenshots

Not applicable, this is a backend runtime change.

## Checklist

- [x] This change requires a documentation update, included: the history-layer page under `dify-agent/docs` is updated in this PR. No change needed in [Dify Document](https://github.com/langgenius/dify-docs).
- [x] I ran the `dify-agent` equivalents of the lint gates (`make check`, `make typecheck`, `make test`). No frontend files are touched, so `vp staged` does not apply.

### DIFF
--- a/dify-agent/docs/dify-agent/user-manual/history-layer/index.md
+++ b/dify-agent/docs/dify-agent/user-manual/history-layer/index.md
@@ -54,7 +54,9 @@ cancelled runs. A failure or cancellation before the capture contains any
 messages preserves the previously restored history. An interrupted capture can
 include a partial response or tool-return request marked `state="interrupted"`;
 pydantic-ai repairs that state when the snapshot is used by a later independent
-run. Without this layer, compaction and interrupted messages affect only the
+run. A run cancelled after the model returned tool calls but before any of them
+ran is also stored with that marker, so the calls it never executed do not block
+the next user prompt. Without this layer, compaction and interrupted messages affect only the
 current run.
 
 ## Resume a conversation
@@ -101,7 +103,10 @@ Dify Agent handles memory conservatively:
    the previously restored history remains unchanged.
 6. Run-level system instructions are removed before history is persisted.
 7. Interrupted partial messages retain pydantic-ai's `state="interrupted"` marker
-   so a later independent run can repair and continue from the checkpoint.
+   so a later independent run can repair and continue from the checkpoint. A
+   trailing response whose tool calls the interrupted run never executed is
+   stored with the same marker; a run that ends in a deferred tool request keeps
+   its open call so the next run can answer it.
 8. Failed and cancelled runs keep their terminal status; their snapshot is a
    checkpoint, not a successful continuation of the interrupted run.
 
--- a/dify-agent/src/dify_agent/runtime/history.py
+++ b/dify-agent/src/dify_agent/runtime/history.py
@@ -5,7 +5,10 @@ named ``history``. Current system instructions belong to each run and are never
 stored. Once Pydantic AI binds and builds messages in the run capture, its
 complete captured history replaces the layer for every terminal outcome,
 including interrupted runs. A failure or cancellation before the capture
-contains messages preserves the previously restored history.
+contains messages preserves the previously restored history. When a run is
+interrupted with tool calls the model returned but the run never executed, the
+stored history records that the response was cut short so the next run can close
+those calls out instead of refusing a new user prompt.
 """
 
 from __future__ import annotations
@@ -14,7 +17,7 @@ from collections.abc import Sequence
 from dataclasses import replace
 from typing import Protocol
 
-from pydantic_ai.messages import ModelMessage, ModelRequest
+from pydantic_ai.messages import ModelMessage, ModelRequest, ModelResponse
 
 from agenton_collections.layers.pydantic_ai import PYDANTIC_AI_HISTORY_LAYER_TYPE_ID, PydanticAIHistoryLayer
 from dify_agent.protocol import DIFY_AGENT_HISTORY_LAYER_ID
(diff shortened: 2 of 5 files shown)


CryoThrust (approved):
Reviewed the persistence boundary and cancellation scenarios. The new `interrupted` flag is passed only when the agent run does not reach a terminal result, and `_mark_trailing_tool_calls_interrupted` changes only a trailing complete `ModelResponse` that actually contains tool calls. This correctly targets the gap between tool-call emission and the first tool return, while leaving successful deferred requests such as `ask_human` unchanged.\n\nThe history tests verify that the source response is not mutated, the interrupted state is persisted, no-tool responses remain unchanged, and the follow-up run receives exactly one synthesized interrupted return and completes. The runner tests exercise cancellation against a real blocked tool path rather than directly fabricating the snapshot. The implementation also preserves the existing instruction stripping and failure/cancellation handling. I did not find a correctness issue in the current diff; the remaining check is the repository’s broader CI coverage.

wylswz:
I didn't manage to reproduce the original issue by stopping the stream on UI while the tool executes. How did you reproduce that?

msdianprince-7:
From the UI the window is narrow, which is probably why it didn't show up. The api only checks `queue_manager.is_stopped()` when the next SSE chunk arrives, and while a tool is running the only chunks are the 15s heartbeats. So the tool has to still be running when that heartbeat lands. A quick tool like web search returns first, pydantic-ai already holds a tool return by the time the cancel hits, and the next run repairs the history fine.

To hit it from the UI you need a tool that runs well past 15s (an HTTP request tool against a slow endpoint works), press stop while it shows the tool running, wait for the run to end as cancelled, then send another message in the same conversation.

Below the UI it's deterministic. `POST /runs/{id}/cancel` while a tool is blocked makes `RunScheduler` call `runner_task.cancel()`, and `finalize_cancellation` stores a snapshot whose last message is a `ModelResponse` with `state="complete"` and an open call. I just ran that through `RunScheduler.cancel_run` with the blocking tool from the PR tests: on main the stored tail is `complete` with `call-1` open, and on this branch it's `interrupted`. The two runner tests in the PR also fail on main with the error from the issue.

wylswz (approved):
LGTM

PR #41861, opened 2026-09-05
fix(messages): stop dropping doc_metadata and friends from GET /messages citations

### LINKED ISSUE #41860: GET /messages silently drops retriever_from/page/doc_metadata/title/files from citations
### Self Checks

- [x] I have read the Contributing Guide.
- [x] This is a bug report, not a question.
- [x] Submitted in English.

### Dify version

1.17.0 (also reproduced by reading `main`'s current source — not fixed there either)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Create a chat app whose knowledge retrieval returns citations carrying extra metadata beyond
   `RetrieverResource`'s declared fields — e.g. an external knowledge API (or a custom retrieval
   node) that returns `retriever_from`, `page`, `doc_metadata`, `title`, or `files` in a citation's
   `RetrievalSourceMetadata` (`api/core/rag/entities/citation_metadata.py`, which declares all
   five of these fields).
2. Send a message through the streaming endpoint (`POST /chat-messages`). The response's citation
   correctly includes all fields, `doc_metadata` included — confirmed by inspecting the raw SSE
   response.
3. Reload the conversation (or call `GET /messages` / `GET /installed-apps/.../messages` / the
   equivalent service-api endpoint directly) for that same message.
4. Inspect the `retriever_resources` entry for the same citation in the response.

### ✔️ Expected Behavior

`GET /messages` returns the same citation data the streaming response did — `retriever_from`,
`page`, `doc_metadata`, `title`, and `files` all present when they were populated at
message-completion time.

### ❌ Actual Behavior

`GET /messages` silently drops `retriever_from`, `page`, `doc_metadata`, `title`, and `files` from
every citation. Root cause, confirmed by reading source directly (not inferred):

- The full citation, including these five fields, **is correctly stored** in
  `messages.message_metadata` (a JSON column) at write time — nothing is lost in storage. Verified
  directly against a Postgres row for a real citation with a populated `doc_metadata` dict.
- `RetrieverResource` (`api/fields/message_fields.py`), the Pydantic response model used by
  `GET /messages` and every other `MessageListItem`-derived history endpoint, declares none of
  these five fields — while `RetrievalSourceMetadata`, the model that actually produces each
  citation, declares all five.
- `ResponseModel`'s `model_config` (`api/fields/base.py`) sets `extra="ignore"`. When
  `controllers/web/message.py`'s `MessageListApi.get()` (and the console/service-api equivalents)
  re-validates the stored `Message.retriever_resources` property (`api/models/model.py`, itself
  just `json.loads`-ing `message_metadata`) against `RetrieverResource`, Pydantic silently drops
  every key the response model doesn't declare.
- The streaming path doesn't go through `RetrieverResource` and isn't affected, which is why the
  gap is easy to miss — a client that branches on `doc_metadata` (e.g. rendering a video player
  keyed off `doc_metadata.start_seconds` for a video citation, our case) works during a live
  answer and silently degrades to a plain citation the moment the same conversation reloads
  through history.

I have a fix and a PR ready — a straightforward field-parity addition to `RetrieverResource`, no
migration or backfill needed since the underlying data was never actually lost. Will link it here.

From Claude Code

tim-gregory-pt:
Fixes #41860

## Summary

`RetrieverResource` (`api/fields/message_fields.py`), the response model behind `GET /messages` (and every other `MessageListItem`-derived history endpoint), declares fewer fields than `RetrievalSourceMetadata` (`api/core/rag/entities/citation_metadata.py`) — the model that actually produces each citation at message-completion time. `ResponseModel`'s `model_config` sets `extra="ignore"`, so when a stored `Message.retriever_resources` dict is re-validated against `RetrieverResource`, Pydantic silently drops every key the response model doesn't declare: `retriever_from`, `page`, `doc_metadata`, `title`, `files`.

The streaming `POST /chat-messages` response doesn't go through this model and isn't affected — which is what makes the gap easy to miss: a client that branches on `doc_metadata` (e.g. rendering a video player keyed off `doc_metadata.start_seconds`) works during a live answer and silently degrades to a plain citation the moment the same conversation is reloaded through history.

The data was never actually lost — `message_metadata` already stores the full dict at write time — so this is a response-model-only fix. No migration, no backfill: every already-stored citation starts returning full metadata as soon as this ships.

**Verified against a live, patched 1.17.0 instance**, not just unit tests: applied this same five-field addition to a running self-hosted deployment, restarted the `api` service, and confirmed `GET /messages` for a conversation with a video citation now returns `doc_metadata`. A client rendering a video player from that field went from a plain-text citation (broken) to a working player (fixed) on conversation reload, with no other change on either side.

**Why typed fields instead of `extra="allow"`:** ran a read-only type check across every stored `retriever_resources` entry in a production database (~10.6k rows) for these five fields before choosing this shape — `page`/`files`/`title` are always absent, `doc_metadata` is always an object or absent, `retriever_from` is always a string. No legacy data would fail validation against the types this PR adds.

## Screenshots

N/A — backend response-model fix; the fix is easiest to see in the field-diff below, not a UI screenshot. (The reporter's downstream symptom was a video player disappearing on conversation reload; not a Dify-UI-visible change.)

## Test plan

- [x] New tests in `api/tests/unit_tests/fields/test_message_fields.py`, confirmed **red** against the pre-fix model and **green** against the fix (temporarily reverted just the model change, re-ran, restored)
- [x] `pytest api/tests/unit_tests/fields/` (23 tests) passes
- [x] `ruff check` / `ruff format --check` clean
- [x] Regenerated API docs and frontend contracts and included them in this PR (rather than relying on autofix.ci): `uv run dev/generate_swagger_markdown_docs.py --swagger-dir ../packages/contracts/openapi --markdown-dir openapi/markdown --keep-swagger-json` and `pnpm --dir packages/contracts gen-api-contract`
- [x] Verified live on a patched self-hosted 1.17.0 instance (see Summary)

## Checklist

From Claude Code


CryoThrust (commented):
I compared the response model with `RetrievalSourceMetadata`: all five previously dropped fields are present with matching nullability and JSON-compatible types, and the generated console/service/web contracts are updated consistently. The tests cover both direct `RetrieverResource` validation and propagation through `MessageListItem`, which is the serialization boundary described in the issue. No blocking issue from my side.

PR #41863, opened 2026-09-05
fix(core): don't classify a Squid-forwarded 401 as an SSRF block

### LINKED ISSUE #41434: Upstream 401 response is incorrectly classified as ToolSSRFError when passing through Squid
### Self Checks

### Dify version

1.16.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

The target service itself returns:

HTTP/1.1 401 Unauthorized
Server: nginx

When the same request passes through Dify's Squid SSRF proxy, the response becomes:

HTTP/1.1 401 Unauthorized
Server: nginx
Via: 1.1 xxxx (squid/6.13)

The Squid access log shows:

TCP_MISS/401

which indicates that Squid successfully forwarded the request to the upstream server and received a 401 response, rather than blocking it with an ACL rule.

However, Dify still raises ToolSSRFError.

### ✔️ Expected Behavior

If the upstream server returns 401 or 403 and Squid only forwards that response, Dify should preserve the original upstream status instead of raising ToolSSRFError.

For example:

Server: nginx
Via: ... squid

should not by itself be considered sufficient evidence that the response was generated by the SSRF proxy.

Only responses actually generated by Squid due to SSRF ACL rejection should be converted into ToolSSRFError.

### ❌ Actual Behavior

HTTP node failed and output "Access to '{url}' was blocked by SSRF protection. The URL may point to a pricate or local network address."

Taranum01:
Fixes #41434

The pre-fix code raised `ToolSSRFError` for every response that went through Squid with status 401 / 403, including the case where the status came from the upstream target server and Squid only added the `Via` header while forwarding. That's a Squid forwarding signature, not a Squid ACL-deny signature — Squid's own ACL denials set the `Server: squid/x.x.x` header because Squid writes the response itself. Trusting `Via` alone turned every upstream 401/403 into a false-positive SSRF block and sent users down the wrong remediation path (they edited the SSRF allowlist when the real problem was on the target).

Tighten the Squid-detection condition to `squid in Server` header — the Squid-generated error page is the only signal we can rely on without false positives. `Via` alone is left out of the classification on purpose.

## How it was tested?

```
$ uv run pytest tests/unit_tests/core/helper/test_ssrf_proxy.py -o addopts=""
36 passed
``"

2 updated + 1 new regression test:

- `test_squid_401_via_header_only_does_not_trigger_error` — renamed and repurposed: a 401 with `Server: ""` and `Via: 1.1 squid (squid/4.10)` (Squid forwarded an upstream response) now passes through unchanged, not raised as `ToolSSRFError`.
- `test_upstream_401_with_squid_via_header_does_not_raise_ssrf_error` — exact reproduction of issue #41434: `Server: nginx`, `Via: 1.1 xxxx (squid/6.13)`, status 401. The upstream application returned 401; Squid only forwarded. Must not raise.
- `test_squid_block_raises_actionable_tool_ssrf_error` — keeps the original "Squid is the source of the response" path: with `Server: squid/4.10` and `Via: 1.1 squid (squid/4.10)` (Squid wrote the response), the actionable ToolSSRFError still fires.

`ruff check` / `ruff format --check` clean, `pyrefly` 0 errors, `importlinter` 25 kept / 0 broken.

### DIFF
--- a/api/controllers/console/workspace/models.py
+++ b/api/controllers/console/workspace/models.py
@@ -320,7 +320,6 @@ class ModelProviderModelApi(Resource):
 
         return SimpleResultResponse(result="success").model_dump(mode="json"), 200
 
-    @console_ns.expect(console_ns.models[ParserDeleteModels.__name__])
     @console_ns.response(204, "Model deleted successfully")
     @setup_required
     @login_required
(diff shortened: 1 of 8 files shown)


crazywoola:
Hi @Taranum01, thank you for contributing to Dify!

For issue #41434, we are keeping the earlier PR #41441 under the FIFO rule, after checking issue-author priority.

PR #41871, opened 2026-09-06
fix(agent): persist Chatflow Agent V2 memory by conversation_id

### LINKED ISSUE #41734: Agent App memory is not working when invoked from Chatflow in Dify 1.16
### Self Checks

### Dify version

1.16.1

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Test the Agent application directly

I created an Agent application with conversation memory enabled.

Conversation:

User: Please query the first page of results.

Agent: [returns the first page]

User: second page

Agent: [correctly understands that "second" refers to the previous query]

The Agent application can correctly use the previous conversation context.

2. Put the same Agent application into a Chatflow

I then use the same Agent application inside a Chatflow.

Conversation:

User: Please query the first page of results.

Chatflow
  -> Agent Application

Agent: [returns the first page]

User: second

Chatflow
  -> Agent Application

Agent: "Hello! You mentioned "page 2", but I don't have the specific context of the previous query..."

The Agent does not understand that "second" refers to the previous request.

Important observation

I checked the streaming output of the second Chatflow execution.

The Agent node contains a conversation_id, for example:

{
  "conversation_id": "ec2df33b-2d06-4ddc-956d-5719d7e4dc20"
}

Therefore, it appears that the Chatflow conversation ID is being passed to the Agent.

However, the runtime information contains:

{
  "memory": "reserved_not_executed"
}

There is also an agent_session_history related runtime record, but the previous conversation does not appear to be actually loaded into the Agent's context.

### ✔️ Expected Behavior

When the same Agent application is invoked multiple times within the same Chatflow conversation, I expect:

Chatflow
   |
   | sys.conversation_id = A
   v
Agent Application
   |
   | conversation_id = A
   v
Agent Memory
   |
   +-- Previous conversation
   |
   v
Agent response

The Agent should be able to use the previous conversation context, just as it does when the Agent application is used directly.

### ❌ Actual Behavior

The conversation_id appears to be passed correctly, but the memory runtime reports:

memory: reserved_not_executed

As a result, the Agent behaves as if there is no previous conversation context.

kabishou11:
Key Chatflow Agent V2 sessions by conversation_id so memory persists across turns. Fixes #41734


wylswz:
Hi @kabishou11, as far as I know, the Agent V2 node is currently unavailable in Chatflow apps, primarily because its context is scoped to individual workflow runs. Since this PR addresses that limitation, could you also enable the Agent V2 node for Chatflow apps?

kabishou11:
Thanks @wylswz — will enable the Agent V2 node for Chatflow apps on top of the conversation_id memory fix and push an update.

kabishou11:
Follow-up for @wylswz: enabled the Agent V2 node for Chatflow (advanced-chat) in the node picker when the feature flag is on (3568642), matching workflow enablement. Backend memory-by-conversation_id behavior from the earlier commit is unchanged.

Happy to tweak the gate/tests if you want a different enablement path.

wylswz:
I just noticed that Both the E2B and Enterprise runtime backends reject `create_binding()` when `existing_workspace_ref` is provided. Since each physical sandbox represents both the binding and workspace, a later Chatflow turn cannot create a new binding and attach it to the conversation’s existing workspace.

For Chatflow, I believe we should instead create one binding per stable conversation/node scope—such as conversation_id + node_id + workflow_agent_binding_id—persist its ID, and `acquire` that same binding on subsequent turns. This would also make the snapshot-inheritance workaround unnecessary, since the session snapshot would remain on the reused binding.

We should also define the binding lifecycle:

- Release the operation-scoped lease after each turn without retiring the binding.
- Keep the binding alive for the lifetime of the conversation.
- Garbage collect the binding after an inactive period.
- Retire and collect it when the conversation is deleted or reset.
- Handle expired/lost backend bindings explicitly.

kabishou11:
Thanks @wylswz — agreed. E2B/Enterprise rejecting `existing_workspace_ref` means the snapshot-inheritance path cannot reattach a later Chatflow turn to the conversation workspace.

I'll switch Chatflow to: one stable binding per conversation/node scope (e.g. conversation_id + node_id + workflow_agent_binding_id), persist that binding id, and `acquire` the same binding on subsequent turns; drop the snapshot-inheritance workaround for that path. Lifecycle: release the turn lease without retiring the binding; keep it for the conversation lifetime; GC after inactivity; retire on conversation delete/reset; handle expired/lost backend bindings explicitly.

Working on this on the same PR branch.

kabishou11:
Follow-up for @wylswz on `12b8027`:

- Chatflow Agent V2 now resolves an ACTIVE binding for the stable conversation/node scope (`conversation_id` + node scope key + agent) via `resolve_active_binding_for_scope()` **before** `create_binding`, so later turns `acquire` the same participant instead of creating with `existing_workspace_ref` (which E2B/Enterprise reject).
- Removed the cross-binding snapshot-inheritance workaround; session snapshot stays on the reused binding.
- Per-turn: conversation-scoped workspaces are not retired by the workflow-run retirement layer (lease ends with the turn; binding stays ACTIVE).
- On conversation delete: `retire_all_for_conversation()` + enqueue physical collection.
- Unit tests cover reuse (no second `create_execution_binding_sync`), no `existing_workspace_ref` on create, and conversation-delete retirement.

Still open for a possible follow-up if you want them in-tree: independent inactive-TTL sweeper, and explicit recreate-on-acquire-failure when the backend binding is gone but DB still says ACTIVE.

wylswz:
> Follow-up for @wylswz on `12b8027`:
> 
> * Chatflow Agent V2 now resolves an ACTIVE binding for the stable conversation/node scope (`conversation_id` + node scope key + agent) via `resolve_active_binding_for_scope()` **before** `create_binding`, so later turns `acquire` the same participant instead of creating with `existing_workspace_ref` (which E2B/Enterprise reject).
> * Removed the cross-binding snapshot-inheritance workaround; session snapshot stays on the reused binding.
> * Per-turn: conversation-scoped workspaces are not retired by the workflow-run retirement layer (lease ends with the turn; binding stays ACTIVE).
> * On conversation delete: `retire_all_for_conversation()` + enqueue physical collection.
> * Unit tests cover reuse (no second `create_execution_binding_sync`), no `existing_workspace_ref` on create, and conversation-delete retirement.
> 
> Still open for a possible follow-up if you want them in-tree: independent inactive-TTL sweeper, and explicit recreate-on-acquire-failure when the backend binding is gone but DB still says ACTIVE.

Tested locally and confirmed that acquire and release are invoked as expected.

I also added the `NEXT_PUBLIC_ENABLE_AGENT_V2_IN_CHATFLOW` feature flag to keep Agent V2 disabled in Chatflow for now. We’ll enable it once we’re confident it’s ready. If you’re self-hosting and want to use Agent V2 in Chatflow, remember to set this environment variable accordingly.

wylswz (approved):

Issue #14650, opened 2025-03-02
Get empty retriever_resources parameter in API with /messages method


lyc280705:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

v1.0.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

In webpage there are retrieved resources:

![Image](https://github.com/user-attachments/assets/35e3274e-590e-4ca4-b986-17d1332c056c)
In API and only with /messages method the list is empty, it works well with /chat-message method:
```
{
    "limit": 20,
    "has_more": false,
    "data": [
        {
            "id": "a87431be-ab19-4bb2-bc14-8e4d07bb489c",
            "conversation_id": "65e69008-a4d2-48f8-8d37-ee02c7cf46ec",
            "parent_message_id": "00000000-0000-0000-0000-000000000000",
            "inputs": {},
            "query": "什么是勾股定理",
            "answer": "勾股定理是一个基本的几何定理，它描述了直角三角形中三边之间的关系。具体来说，勾股定理指出：**在直角三角形中，两条直角边的平方和等于斜边的平方**。\n\n用数学公式表示就是：  \n$$\na^2 + b^2 = c^2\n$$\n其中，$a$ 和 $b$ 是直角三角形的两条直角边，$c$ 是斜边（即直角三角形中与直角相对的最长边）。\n\n---\n\n### **勾股定理的来源**\n在中国古代，直角三角形被称为“勾股形”，其中较短的直角边称为“勾”，较长的直角边称为“股”，斜边称为“弦”。因此，这个定理在中国也被称为“勾股定理”或“商高定理”。\n\n---\n\n### **勾股定理的几何意义**\n勾股定理揭示了直角三角形中三边之间的数量关系。它不仅是一个数学公式",
            "message_files": [],
            "feedback": null,
            "retriever_resources": [],         //null
            "created_at": 1740830044,
            "agent_thoughts": [],
[...]

PR #41689, opened 2026-09-03
fix(api): default missing retriever_resources to empty list

leilei3167:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 1. Ensure there is an associated issue and you have been assigned to it
> 1. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary

Embedded web apps fail to load conversation history when any message has `message_metadata` but no `retriever_resources` key (common for Chatflow/Agent apps without Knowledge Retrieval).

`GET /messages` builds `WebMessageListItem` from `Message.retriever_resources`. That property did:

```python
return self.message_metadata_dict.get("retriever_resources") if self.message_metadata else []
```

If metadata exists but the key is absent, `.get(...)` returns `None`, and Pydantic rejects the required `list[RetrieverResource]` field:

```
1 validation error for WebMessageListItem
retriever_resources
  Input should be a valid list
  [type=list_type, input_value=None, input_type=NoneType]
```

The service API path was already guarded in #17304; the web path still used the unguarded property. This PR defaults the missing key to `[]`, matching the no-metadata branch and the expected schema.

Fixes #41688

## Screenshots

| Before | After |
| ------ | ----- |
| N/A (API validation) | N/A (API validation) |

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [x] I ran `make lint && make type-check` (backend) and `vp staged` (frontend) to appease the lint gods

PR #41880, opened 2026-09-06
fix(api): default Message.retriever_resources to [] when metadata lacks the key

### LINKED ISSUE #41688: Bug: Web app /messages returns `retriever_resources Input should be a valid list` when a message has no retriever_resources
### Self Checks

### Dify version

1.17.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

1. Create a Chatflow / Agent app that does **not** use a Knowledge Retrieval node. For example, a flow with an SQL query node (or HTTP request node, or Agent node with tool calling).
2. Publish the app and embed it into an external page using the "Embed" snippet (the web app, not the service API).
3. Send a message so that the conversation history contains at least one message without any `retriever_resources`.
4. Reload the embedded page so it loads the message history via `GET /messages`.

### ✔️ Expected Behavior

Message history loads normally, with `retriever_resources` as an empty list (`[]`) for messages that had no retrieval.

### ❌ Actual Behavior

The embedded page throws:

```
1 validation error for WebMessageListItem
retriever_resources
  Input should be a valid list
  [type=list_type, input_value=None, input_type=NoneType]
```

And the history fails to render.

### Root cause

`api/models/model.py` (around line 1823):

```python
@property
def retriever_resources(self) -> Any:
    return self.message_metadata_dict.get("retriever_resources") if self.message_metadata else []
```

When `message_metadata` exists but does **not** contain the `retriever_resources` key, `.get("retriever_resources")` returns `None` instead of `[]`. This `None` then reaches the required field `WebMessageListItem.retriever_resources: list[RetrieverResource]` in `api/fields/message_fields.py` and fails Pydantic validation.

Note: the service_api path (`controllers/service_api/app/message.py`) was already fixed in PR #17304 (issue #14650), but the **web path** (`controllers/web/message.py` → `WebMessageListItem`) still uses the un-guarded property above.

### Suggested fix

Give the `.get()` a default value so a missing key resolves to an empty list:

```python
return self.message_metadata_dict.get("retriever_resources", []) if self.message_metadata else []
```

I have verified this one-line change fixes the issue on a self-hosted 1.17.0 Docker deployment.

Harsh23Kashyap:
## Summary

Closes a 1.17.0 regression in the web-app embedded conversation history. When a chatflow or agent message has `message_metadata` set but does not include the `retriever_resources` key, the embedded page throws:

```
1 validation error for WebMessageListItem
retriever_resources
  Input should be a valid list
  [type=list_type, input_value=None, input_type=NoneType]
```

and the history fails to render.

Root cause: `Message.retriever_resources` does `self.message_metadata_dict.get("retriever_resources")` when `message_metadata` is set. `dict.get` without a default returns `None` for a missing key, and `None` then fails Pydantic validation on the required `list[RetrieverResource]` field declared by `MessageListItem.retriever_resources`.

Fix: add the missing default to `dict.get` so a metadata blob that does not carry the `retriever_resources` key resolves to `[]`.

Fixes #41688

## Why this is a regression, not a feature

- The `MessageListItem` Pydantic field is a required `list[RetrieverResource]` with no default; the contract is "always a list, possibly empty".
- The property already had the right contract for the `message_metadata is None` branch (`else []`). The `message_metadata is not None but key missing` branch was an oversight.
- The sibling `service_api` path was already fixed in PR #17304 (issue #14650) on a different serializer; this brings the web path in line with the same default-empty contract.

## Changes

`api/models/model.py::Message.retriever_resources`

```python
@property
def retriever_resources(self) -> Any:
-    return self.message_metadata_dict.get("retriever_resources") if self.message_metadata else []
+    return self.message_metadata_dict.get("retriever_resources", []) if self.message_metadata else []
```

No other call sites need to change; the property is the single source of truth for `retriever_resources` on the model.

## Tests

Three new unit tests in `api/tests/unit_tests/models/test_app_models.py` (`TestMessageModel`):

- `test_retriever_resources_default_when_key_missing` — a Message whose `message_metadata` JSON does not contain `retriever_resources` returns `[]` from the property. **This test fails on the unpatched property with `assert None == []`**; verified by stashing only the source file and re-running, then restoring.
- `test_retriever_resources_none_when_message_metadata_none` — a Message with `message_metadata=None` still returns `[]` (the existing `else` branch).
- `test_retriever_resources_returns_existing_value` — a Message whose metadata does carry `retriever_resources` returns the stored value unchanged.

3/3 pass on the new tests. 57/57 pass on the full `test_app_models.py` file. ruff check + format clean on both touched files.

## Verification

- `pytest tests/unit_tests/models/test_app_models.py` — 57/57 pass.
- `ruff check` — clean.

## Risk

Very low. The change is a 1-character default on a `dict.get` call. The two other branches (key present, metadata None) are unchanged. No schema, API, or protocol change. The web `MessageListItem.retriever_resources` field already required a list, so the contract is preserved.

A small audit of other `message_metadata_dict.get(...)` call sites in the model layer to confirm no other property has the same missing-key shape mismatch with its serializer. Grep for `message_metadata_dict.get` in `api/models/` would surface them. None were reported as broken in #41688, but a sweep would harden the contract.

### DIFF
--- a/api/models/model.py
+++ b/api/models/model.py
@@ -1817,7 +1817,7 @@ class Message(Base):
 
     @property
     def retriever_resources(self) -> Any:
-        return self.message_metadata_dict.get("retriever_resources") if self.message_metadata else []
+        return self.message_metadata_dict.get("retriever_resources", []) if self.message_metadata else []
 
     @property
     def message_files(self) -> list[MessageFileInfo]:
--- a/api/tests/unit_tests/models/test_app_models.py
+++ b/api/tests/unit_tests/models/test_app_models.py
@@ -857,6 +857,77 @@ class TestMessageModel:
         # Assert
         assert result == {}
 
+    def test_retriever_resources_default_when_key_missing(self):
+        """Regression for langgenius/dify#41688: when message_metadata is set
+        but does not contain the 'retriever_resources' key, the property must
+        return [] (not None) so MessageListItem.retriever_resources
+        (a required list[...] field) does not fail Pydantic validation.
+        """
+        # Arrange — message_metadata exists but has no retriever_resources key
+        message = Message(
+            app_id=str(uuid4()),
+            conversation_id=str(uuid4()),
+            query="Test query",
+            message={"role": "user", "content": "Test"},
+            answer="Test answer",
+            message_unit_price=Decimal("0.0001"),
+            answer_unit_price=Decimal("0.0002"),
+            currency="USD",
+            from_source=ConversationFromSource.API,
+            message_metadata=json.dumps({"usage": {"tokens": 100}}),
+        )
+
+        # Act
+        result = message.retriever_resources
+
+        # Assert — must default to empty list, not None
+        assert result == []
+
+    def test_retriever_resources_none_when_message_metadata_none(self):
+        """When message_metadata is None, the property returns [] (the else branch)."""
+        # Arrange
+        message = Message(
+            app_id=str(uuid4()),
+            conversation_id=str(uuid4()),
+            query="Test query",
+            message={"role": "user", "content": "Test"},
+            answer="Test answer",
+            message_unit_price=Decimal("0.0001"),
+            answer_unit_price=Decimal("0.0002"),
+            currency="USD",
+            from_source=ConversationFromSource.API,
+            message_metadata=None,
+        )
+
+        # Act
+        result = message.retriever_resources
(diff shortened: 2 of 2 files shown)


crazywoola:

For issue #41688, we are keeping the earlier PR #41689 under the FIFO rule, after checking issue-author priority.

PR #41881, opened 2026-09-06
fix: suggested questions empty for reasoning models

### LINKED ISSUE #41854: suggested-question API request always return empty content
### Self Checks

### Dify version

1.17.0

### Cloud or Self Hosted

Cloud

### Steps to reproduce

suggested-question API request always return empty content, its setting is keeping in default

Image

Image

### ✔️ Expected Behavior

return three questions

### ❌ Actual Behavior

the content AI answers is pretty rich, but sugestion-question always return empty content

however-yir:
## Summary

Fixes #41854

When the suggested-questions-after-answer generator falls back to the workspace default LLM and that model is a reasoning model, the completion can contain a `<think>...</think>` block before the JSON payload. `SuggestedQuestionsAfterAnswerOutputParser` searched the raw text for the first `[...]` pair, so a bracket pair inside the reasoning block (e.g. a probability vector `[0.2, 0.8]` or a token like `[MASK]`) was matched first: it either failed to decode (logged warning + empty list) or decoded into non-string values that were filtered away — both leaving the API returning `[]` even though the model answered fine.

This change strips `<think>...</think>` blocks (DOTALL) before searching for the payload array. Output from non-reasoning models is unaffected, as is the behavior when no reasoning block is present.

## Screenshots

N/A — behavior is covered by unit tests.

## Checklist

From ZCode

### DIFF
--- a/api/core/llm_generator/output_parser/suggested_questions_after_answer.py
+++ b/api/core/llm_generator/output_parser/suggested_questions_after_answer.py
@@ -24,10 +24,13 @@ class SuggestedQuestionsAfterAnswerOutputParser:
 
     def parse(self, text: str) -> Sequence[str]:
         stripped_text = text.strip()
-        # Reasoning models may prepend a <think>...</think> block before the payload. Any
-        # bracket pair inside it would otherwise be matched first (and either fail to parse
-        # or parse as non-strings), leaving the suggested questions empty.
-        stripped_text = re.sub(r"<think>.*?</think>", "", stripped_text, flags=re.DOTALL).strip()
+        # Reasoning models may prepend a <think>...</think> or <thought>...</thought> block
+        # before the payload. Unclosed blocks (e.g. truncated responses) are discarded through
+        # end-of-input. Any bracket pair inside reasoning would otherwise be matched first
+        # (and either fail to parse or parse as non-strings), leaving the suggested questions empty.
+        stripped_text = re.sub(
+            r"<(think|thought)>.*?(?:</\1>|$)", "", stripped_text, flags=re.DOTALL | re.IGNORECASE
+        ).strip()
         action_match = re.search(r"\[.*?\]", stripped_text, re.DOTALL)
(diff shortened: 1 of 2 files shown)


CryoThrust:
The parser-level fix matches the reported failure and the regression cases are useful. One edge case worth covering before merge: if the model emits an unclosed <think> block (for example because the provider truncates the completion), this regex leaves the reasoning text in place and the first bracket pair can still be selected. Could the parser explicitly handle an unmatched opening tag (discard through end-of-input), and add a regression test? It may also be worth using a case-insensitive tag match if providers vary capitalization.

however-yir:
@CryoThrust Thanks for the review! Good catch on the truncated completion edge case.

I have updated the regex in `SuggestedQuestionsAfterAnswerOutputParser` to `r"<(think|thought)>.*?(?:</\1>|$)"` with `flags=re.DOTALL | re.IGNORECASE`:
- **Unclosed reasoning tags**: If the completion is truncated leaving an unclosed `<think>` or `<thought>` opening tag, all content through the end of the input is discarded, preventing any brackets inside the truncated reasoning from being parsed as suggested questions.
- **Case insensitivity**: Handled via `re.IGNORECASE` so tags like `<Think>...</Think>` or `<THINK>...` are matched.
- **Tags**: Also covers `<thought>` in addition to `<think>`.

Unit tests have been added covering:
- Truncated responses with unclosed `<think>` and `<thought>` tags containing bracket tokens.
- Mixed-case closed (`<Think>...</THINK>`) and unclosed (`<THINK>...`) blocks.
- Closed `<thought>` tags.

All local pytest unit tests, mypy, and ruff checks pass.

however-yir:
@CryoThrust Thanks again for the feedback. Just a quick follow-up: the requested edge cases are now covered in `f24f694` (unclosed `<think>/<thought>` blocks and case-insensitive tag handling), with regression tests added. The latest CI is green as well. Appreciate the review!

PR #41898, opened 2026-09-07
fix(console): stop advertising the education plan outside Dify Cloud

lin-snow:
## Summary

`features.education.enabled` followed `EDUCATION_ENABLED` on every deployment edition, but the four `/account/education*` endpoints are admitted for `CLOUD` only (`console_account_admission(editions=...)`). A COMMUNITY or ENTERPRISE deployment with the flag on therefore advertised a capability its own gate rejects: the frontend fired six requests, the gate answered `abort(404)`, and flask-restx appended url-map suggestions to that body — so the toast read `The requested URL was not found on the server ... did you mean /console/api/account/education`, naming the path just requested.

Two independent defects, both fixed here.

**The flag promised what the gate rejects.** `EDUCATION_EDITIONS` now lives with the education application service (`services/account_education_service.py`) and is read by both the Console admission gate and `_fulfill_params_from_env`, so the two cannot drift apart again. No frontend gating changes were needed: all six probe call sites already gate on `education.enabled`, and `/education/verify` and `/education/apply` already redirect home when it is false.

**Every 404 body leaked url-map suggestions.** The suppression used to live in `ExternalApi._help_on_404`, and only applied to the single surface that installs a canonical error formatter (`/openapi/v1`) — console, web, service_api, files, mcp and inner_api all leaked. It is now `RESTX_ERROR_404_HELP = False` in `create_flask_app_with_configs()`, flask-restx's own documented setting, which retires the private-hook override and its startup-guard entry. Those suggestions enumerate the url map to anonymous callers and, on a policy 404 for a route that exists, suggest that very route.

The education status probe is also `silent` by default. It is passive — every call site already degrades to "not a student", so a failure has nothing to tell the user.

Operator-visible: `EDUCATION_ENABLED=true` is now a no-op outside CLOUD. Nothing usable is lost — those endpoints have been CLOUD-only since `@only_edition_cloud` — but the config docstring now says so.

## Verification

- COMMUNITY/ENTERPRISE report `education.enabled=false` and issue no request; CLOUD is unchanged.
- `test_get_features_community_edition` asserted `education.enabled is True`, i.e. the bug was pinned by a test. That assertion is flipped, with the reason in a comment.
- The startup guard on the one remaining patched flask-restx hook now has a test of its own: deleting
  `Api._should_use_fr_error_handler` must break construction, not the first 404.
- The new 404 test runs through the real app factory on the real bug path; setting `RESTX_ERROR_404_HELP` back to `True` reproduces `... but did you mean /console/api/account/education or ...`.
- `pytest api/tests/unit_tests`: 18676 passed, 4 skipped. `ruff`, `lint-imports` (27 contracts), `make type-check-core`, the response-contract linter and both ast-grep guards are clean.

## Screenshots

N/A — no visual change. The fix removes a request and the raw server toast it produced.

## Checklist

From Claude Code

### DIFF
--- a/api/app_factory.py
+++ b/api/app_factory.py
@@ -105,6 +105,10 @@ def create_flask_app_with_configs() -> DifyApp:
     dify_app = DifyApp(__name__)
     dify_app.config.from_mapping(dify_config.model_dump())
     dify_app.config["RESTX_INCLUDE_ALL_MODELS"] = True
+    # flask-restx appends url-map suggestions to 404 bodies: they enumerate routes to
+    # anonymous callers, and a policy 404 on an existing route (edition admission)
+    # ends up suggesting the very path that was just requested.
+    dify_app.config["RESTX_ERROR_404_HELP"] = False
 
     # add before request hook
     @dify_app.before_request
--- a/api/configs/feature/__init__.py
+++ b/api/configs/feature/__init__.py
@@ -1614,7 +1614,7 @@ class AccountConfig(BaseSettings):
     )
 
     EDUCATION_ENABLED: bool = Field(
-        description="whether to enable education identity",
+        description="whether to enable education identity (CLOUD deployments only)",
         default=False,
     )
 
--- a/api/controllers/console/workspace/account.py
+++ b/api/controllers/console/workspace/account.py
@@ -47,13 +47,13 @@ from controllers.console.workspace.error import (
     RepeatPasswordNotMatchError,
 )
 from controllers.console.wraps import model_validate, setup_required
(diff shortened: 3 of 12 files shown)


zyssyz123 (dismissed):

wylswz:
@lin-snow please resolve conflicts so we can move forward.

lin-snow:
> @lin-snow please resolve conflicts so we can move forward.

Conflicts have been resolved. We should be good to move forward now!

wylswz (approved):

PR #41088, opened 2026-08-22
fix(web): do not use builtin credential routes for providers that have none

sainikhiljuluri:
# Summary

Fixes #39169 — opening an agent's configure page returns a 500 (`PluginNotFoundError`) when the agent uses a custom API tool.

**Root cause.** `useGetApi` returned the *builtin* tool-provider credential routes for every provider in the tool category, ignoring `providerType`:

```ts
// web/app/components/plugins/plugin-auth/hooks/use-get-api.ts
export const useGetApi = ({ category = AuthCategory.tool, provider }: PluginPayload) => {
  if (category === AuthCategory.tool) {
    return {
      getCredentialSchema: (t) => `/workspaces/current/tool-provider/builtin/${provider}/credential/schema/${t}`,
```

A custom API tool provider is keyed by a `tool_api_providers` UUID, not a plugin id, so the client requests `.../builtin/<uuid>/credential/schema/api-key`. `ToolBuiltinProviderCredentialsSchemaApi` (`api/controllers/console/workspace/tool_providers.py:843-861`) passes `provider` straight to `BuiltinToolManageService.list_builtin_provider_credentials_schema`, the plugin daemon fails the lookup, and it surfaces as a 500 — matching the traceback in the issue.

This fires on mount from `features/agent-v2/.../provider-tool/item.tsx:62-76`, where the payload is fed into `usePluginAuth(pluginPayload, true)` unconditionally.

**Fix.** Credential endpoints are registered only under `/tool-provider/builtin/...`; there is no equivalent for `api`, `mcp` or `workflow`. So exclude exactly those three and let them fall through to the existing no-op route set.

Important detail: this is a deny-list, not an allow-list of `builtin`. `providerType` is a `ToolProviderType`, which also includes `'plugin'` — and plugin-backed providers **do** serve credentials from the builtin routes. `tool-provider-catalog.ts:48-49` registers `'plugin'` alongside `CollectionType.builtIn` for the same provider list, and `:92` treats the two as one class. Gating on `=== builtin` alone would have silently broken credential loading for plugin-backed agent tools, which is worse than the bug being fixed. There's a regression test pinning that.

An absent `providerType` also keeps the builtin routes, so existing callers are unaffected.

# Screenshots / Test

```
pnpm test app/components/plugins/plugin-auth/hooks/__tests__/use-get-api.spec.ts   # 16 passed
pnpm type-check                                                                    # clean
pnpm lint:tss                                                                      # 5961 passed
```

Five tests added:
[...]

PR #40686, opened 2026-08-13
fix: route MCP provider credential lookup away from builtin auth (fix #40612)

Sskift:
## Summary

Fixes #40612.

`getProviderCredentialType` (in `web/features/agent-v2/agent-detail/configure/tool-provider-catalog.ts`) classified any tool provider with stored `team_credentials` as credential type `api-key`, checking `team_credentials` before `provider.type`. An MCP provider that had stored team credentials was therefore misclassified as `api-key`, which routed it into the plugin-auth UI path and set `credentialVariant: 'unauthorized'`. `UnauthorizedCredentialStatus` then mounts `usePluginAuth`, whose `useGetApi` always builds URLs like `/workspaces/current/tool-provider/builtin/<provider>/credential/info`. For an MCP server, `<provider>` is the MCP server id (not a builtin/plugin provider id), so the backend returns 500.

## Root cause

The `team_credentials` check in `getProviderCredentialType` was not scoped to builtin providers, so any provider type (including `mcp`) with stored team credentials resolved to `'api-key'`.

## Fix

Scope the `team_credentials` → `'api-key'` classification to `CollectionType.builtIn` only, mirroring the existing `oauth2` branch. MCP (and `custom`/`workflow`/`model`) providers now resolve to no credential type, so they no longer enter the plugin-auth path.

## Testing

- Added `web/features/agent-v2/agent-detail/configure/__tests__/tool-provider-catalog.spec.ts` covering `getProviderCredentialType` for builtin (api-key / oauth2) and MCP (with/without team_credentials) providers.
- RED: before the fix, the MCP-with-team_credentials case returned `'api-key'` and the test failed (`expected 'api-key' to be undefined`).
- GREEN: after the fix, the focused spec passes (5/5).
- Regression: `pnpm --config.verifyDepsBeforeRun=false test configure` → 26 files, 379 tests passed.

Note: this addresses the primary root cause (credential-type misclassification). The related hardening suggested in the issue — adding an `isBuiltinProvider` guard on the `UnauthorizedCredentialStatus` render path in `provider-tool/item.tsx` so non-builtin providers short-circuit before `usePluginAuth` mounts — was intentionally left out of this PR to keep the change atomic and scoped to the actual 500 trigger; it can be a follow-up if maintainers want defense-in-depth.

From Claude

PR #41837, opened 2026-09-04
fix(tools): translate PluginNotFoundError to ToolProviderNotFoundError

Taranum01:
Fixes #41805

## What problem does this PR solve?

`ToolManager.get_builtin_provider` falls back to `PluginToolManager.fetch_tool_provider` when the requested identifier isn't in the builtin cache. The plugin daemon raises `PluginNotFoundError` (and the lower-level `PluginDaemonNotFoundError`) when the provider is unknown to it, and neither was caught at the boundary. The error bubbled up to the console API generic exception handler as a 500.

## What is changed and how it works?

Translate both to `ToolProviderNotFoundError` at the `get_plugin_provider` boundary so the API returns a controlled 4xx (response matches the existing "missing provider" path that already returns `ToolProviderNotFoundError`). This is defense in depth behind the frontend fixes in #40686 and #41088 — the Agent V2 MCP-server-id path referenced in the report is no longer exercised, but any other caller hitting the builtin credential endpoints with a non-builtin identifier now gets a clean 4xx instead of a 500.

## How it was tested?

```
$ uv run pytest tests/unit_tests/core/tools/test_tool_manager.py -o addopts=
46 passed
```

1 new test (parameterized over both daemon error types):
- `test_get_plugin_provider_translates_plugin_not_found_to_domain_error[daemon_error0]` — `PluginNotFoundError`
- `test_get_plugin_provider_translates_plugin_not_found_to_domain_error[daemon_error1]` — `PluginDaemonNotFoundError`

Both confirm the exception is wrapped to `ToolProviderNotFoundError` matching the "plugin provider <name> not found" message. The existing `test_get_plugin_provider_raises_when_provider_missing` test unchanged (covers the `None`-from-fetch path).

PR #41924, opened 2026-09-07
fix(api): translate PluginNotFoundError to ToolProviderNotFoundError in get_plugin_provider

### LINKED ISSUE #41805: Builtin tool credential endpoints return 500 for unknown provider identifiers
### Self Checks

### Dify version

main (6208b33e), also reproduced on 1.16.0

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Related to #39169 and #40612, but this report covers the remaining backend error handling after the frontend stops sending the invalid request.

1. Sign in to a self-hosted Dify workspace.
2. Request a builtin credential endpoint with an identifier that is not a builtin or plugin provider, for example:
   `GET /console/api/workspaces/current/tool-provider/builtin/<unknown-provider>/credential/info`
3. The same path occurs when the current Agent V2 bug sends an MCP server identifier to the builtin endpoint.

`ToolManager.get_builtin_provider()` falls back to `PluginToolManager.fetch_tool_provider()`. The plugin daemon raises `PluginNotFoundError`, which is not a `ValueError`, so the console API generic exception handler returns HTTP 500. The same risk applies to the builtin credential schema and credential-list endpoints.

### ✔️ Expected Behavior

An unknown or non-builtin provider identifier should produce a controlled 4xx response (consistent with the existing `ToolProviderNotFoundError` domain error), without surfacing an internal server error.

The frontend fixes in #40686 and #41088 should still prevent invalid requests from Agent V2; this backend guard is defense in depth for other callers.

### ❌ Actual Behavior

The request returns HTTP 500 with `PluginNotFoundError: plugin not found`.

A minimal fix is to translate `PluginNotFoundError` to `ToolProviderNotFoundError` at the `ToolManager.get_plugin_provider()` boundary, with a focused unit test. I can prepare the PR if a maintainer agrees and assigns this issue to me.

AI-assisted investigation: Codex.

Harsh23Kashyap:
## Summary

Closes a 1.16.0 / 1.17.0 regression in the console builtin tool credential endpoints. When a request lands on a builtin-tool URL with an identifier that is not a builtin and not a known plugin, the fallback path `ToolManager.get_builtin_provider()` → `get_plugin_provider()` calls `PluginToolManager.fetch_tool_provider()`. The plugin daemon raises `PluginNotFoundError`, which is a `PluginDaemonClientSideError` (not a `ValueError` subclass) and falls through the `external_api` error handlers to the generic `Exception` handler — surfacing as HTTP 500 with body `PluginNotFoundError: plugin not found`.

Translate the error at the `get_plugin_provider` boundary so the existing `ToolProviderNotFoundError` path is taken. The console API generic handler already maps that to a controlled 4xx response (`invalid_param`).

Fixes #41805

## Why this is a regression, not a feature

- `ToolManager.get_builtin_provider()` already raises `ToolProviderNotFoundError` for the missing-hardcoded case (and many other call sites in the same file do the same). The `PluginNotFoundError` leak was an oversight on the fallback path.
- The 500 response leaks internal plugin-daemon state to the client. The 4xx response is the same one the rest of `ToolManager` already produces.
- The front-end fixes in #40686 and #41088 already prevent invalid requests from Agent V2; this is defense in depth for other callers (the issue body reports the bug surfaces from a self-hosted credential lookup with an unknown identifier).

## Changes

`api/core/tools/tool_manager.py::ToolManager.get_plugin_provider`

- Import `PluginNotFoundError` from `core.plugin.impl.exc`.
- Wrap the `manager.fetch_tool_provider(tenant_id, provider)` call in a `try/except PluginNotFoundError` and re-raise as `ToolProviderNotFoundError(f"plugin provider {provider} not found")` (matching the message used by the existing `if not provider_entity` branch). Use `from None` to suppress the original `PluginNotFoundError` chain since the boundary has translated the failure.

No other call sites change. The fix is additive: the existing `if not provider_entity` branch is unchanged, so the `manager.fetch_tool_provider` returning `None` path keeps the same behavior.

## Tests

One new unit test in `api/tests/unit_tests/core/tools/test_tool_manager.py`:

- `test_get_plugin_provider_translates_plugin_not_found_error` — when the plugin daemon's `fetch_tool_provider` raises `PluginNotFoundError`, `ToolManager.get_plugin_provider` must translate it to `ToolProviderNotFoundError`. **On the unpatched path this test fails with the raw `PluginNotFoundError` leaking out**; verified by stashing only `api/core/tools/tool_manager.py` and re-running, then restoring.

48/48 pass on the full `test_tool_manager.py`. ruff check + format clean on both touched files.

## Verification

- `pytest tests/unit_tests/core/tools/test_tool_manager.py` — 48/48 pass.
- `ruff check` — clean.

## Risk

Very low. The fix is a single `try/except` that translates one exception type to another at one site. The new path is gated on the exact `PluginNotFoundError` from the plugin daemon; other errors (timeouts, transport failures, etc.) still flow through their existing handlers. The existing `if not provider_entity: raise ToolProviderNotFoundError(...)` branch is unchanged, so the daemon-returns-None path keeps the same behavior.

A small audit of other plugin-daemon-error call sites in `ToolManager` to confirm no other exception type is leaking through the same way. The grep for `PluginNotFoundError` in `api/core/tools/` is now down to zero (after this change), but sibling errors (`PluginPermissionDeniedError`, `PluginUniqueIdentifierError`, etc.) may have the same shape mismatch. A separate sweep would harden the contract.

### DIFF
--- a/api/core/tools/tool_manager.py
+++ b/api/core/tools/tool_manager.py
@@ -22,6 +22,7 @@ from core.entities import PluginCredentialType
 from core.helper.module_import_helper import load_single_subclass_from_source
 from core.helper.position_helper import is_filtered
 from core.helper.provider_cache import ToolProviderCredentialsCache
+from core.plugin.impl.exc import PluginNotFoundError
 from core.plugin.impl.tool import PluginToolManager
 from core.tools.__base.tool import Tool
 from core.tools.__base.tool_provider import ToolProviderController
@@ -161,7 +162,10 @@ class ToolManager:
                 return plugin_tool_providers[provider]
 
             manager = PluginToolManager()
-            provider_entity = manager.fetch_tool_provider(tenant_id, provider)
+            try:
+                provider_entity = manager.fetch_tool_provider(tenant_id, provider)
+            except PluginNotFoundError:
+                raise ToolProviderNotFoundError(f"plugin provider {provider} not found") from None
             if not provider_entity:
                 raise ToolProviderNotFoundError(f"plugin provider {provider} not found")
 
--- a/api/tests/unit_tests/core/tools/test_tool_manager.py
+++ b/api/tests/unit_tests/core/tools/test_tool_manager.py
@@ -300,6 +300,28 @@ def test_get_plugin_provider_raises_when_provider_missing():
                     ToolManager.get_plugin_provider("provider-a", "tenant-1")
 
 
+def test_get_plugin_provider_translates_plugin_not_found_error():
+    """Regression for langgenius/dify#41805: when the plugin daemon raises
+    PluginNotFoundError, translate it to ToolProviderNotFoundError so the
+    console API generic exception handler returns a controlled 4xx response
+    instead of HTTP 500.
+    """
+    from core.plugin.impl.exc import PluginNotFoundError
+
+    provider_context = _SimpleContextVar()
+    lock_context = _SimpleContextVar()
+    lock_context.set(threading.Lock())
+
+    with patch("core.tools.tool_manager.contexts.plugin_tool_providers", provider_context):
+        with patch("core.tools.tool_manager.contexts.plugin_tool_providers_lock", lock_context):
+            with patch("core.tools.tool_manager.PluginToolManager") as mock_manager_cls:
+                mock_manager_cls.return_value.fetch_tool_provider.side_effect = PluginNotFoundError(
+                    "plugin not found: my-mcp-tools"
+                )
+                with pytest.raises(ToolProviderNotFoundError, match="plugin provider my-mcp-tools not found"):
+                    ToolManager.get_plugin_provider("my-mcp-tools", "tenant-1")
+
+
 def test_get_tool_runtime_builtin_without_credentials():
     tool = Mock()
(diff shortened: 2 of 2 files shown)


CryoThrust:
I traced the changed boundary and the exception mapping looks correctly scoped: only `PluginNotFoundError` is translated, so daemon transport/readiness failures continue through their existing error path. The regression test also exercises the real failure shape rather than the previous `None`-return-only branch.

One small suggestion for the final review: please keep the provider identifier in the translated message (as this PR does) and consider an assertion that an unrelated plugin-daemon exception is not translated, so the 4xx/503 distinction remains explicit.

Harsh23Kashyap:
Good call - added a companion test that pins the boundary: `PluginDaemonInternalServerError` from `fetch_tool_provider` now propagates unchanged, so only `PluginNotFoundError` maps to `ToolProviderNotFoundError` and the existing 4xx/503 split stays explicit. The translated message keeps the provider identifier as suggested.

crazywoola:

For issue #41805, we are keeping the earlier PR #41837 under the FIFO rule, after checking issue-author priority.

PR #41971, opened 2026-09-08
refactor(api): isolate installed app conversations

hjlarry:
## Summary

Part of #39993. Stacked on #41957.

Explore conversation list, delete, rename, pin, and unpin handlers now use Console account admission and an injected conversation application service instead of request-scoped ORM objects. The repository owns bounded transactions and account/app ownership checks; automatic naming runs between the read and write transactions, and deletion cleanup is dispatched after commit. Successful response schemas and pagination behavior are preserved, while missing conversations, cursors, and first messages receive specific error codes. Invalid list queries use the shared 422 validation response.

Capture `app_mode` during installed-app admission and reuse it in conversation, generation, saved-message, and stop handlers, removing their extra mode query. Mode decisions use the admission snapshot: a stop request can still stop its task if the app is subsequently deleted; generation revalidates the app in the runtime after recording usage. Share conversation retirement in a persistence helper so binding retirement and soft deletion stay in one transaction for both legacy and migrated callers. Legacy input/file restoration and the existing Agent Workspace persistence methods remain in place.

Replace the obsolete container-based mock controller tests with HTTP and SQLite coverage, and add an import rule preventing conversation persistence from directly calling the legacy `ConversationService`.


api/repositories/installed_app_conversation_repository.py
```diff
@@ -0,0 +1,219 @@
...
+        with self._session_factory.begin() as session:
+            app = self._get_app(session=session, installed_app=installed_app)
+            conversation = self._get_conversation(
+                session=session, app_id=app.id, account_id=account_id, conversation_id=conversation_id
+            )
+            conversation.name = name
+            if updated_at is not None:
+                conversation.updated_at = updated_at
+            session.flush()
+            return self._to_record(conversation=conversation, session=session)
```
41tair: We handle the row lock until _to_record finished.

api/repositories/conversation_lifecycle.py
```diff
@@ -0,0 +1,55 @@
...
+    # Reuse the existing Workspace lifecycle owner. These two methods only
+    # read and mutate the supplied session; they do not perform external I/O.
+    if binding_id is not None:
+        owner_scope = WorkspaceOwnerScope(
+            tenant_id=app_model.tenant_id,
+            app_id=app_model.id,
+            owner_type=AgentWorkspaceOwnerType.CONVERSATION,
+            owner_id=conversation.id,
+        )
+        binding = AgentWorkspaceService.get_active_binding(
```
41tair, line 32: Why here still use AgentWorkSpaceService ?

hjlarry: cause create a AgentWorkSpaceRepo in this PR seems strange, the methods in `AgentWorkspaceService` basiclly all repo layer,  create a unique PR to resolve this maybe better?
the `conversation_lifecycle.py` also plan to remove in the later refactor

linw1995 (approved):

PR #41973, opened 2026-09-08
fix(api): validate web app site icon ownership

linw1995:
## Summary

- return an actionable configuration error when a Web App site icon is unavailable
- reject missing or cross-tenant site icon references when updating site settings or publishing workflows
- log invalid runtime asset references with internal app, tenant, and file identifiers


linw1995:
Superseded by #41977, which targets main.

PR #41980, opened 2026-09-08
fix(web): keep recommend banner logos intact and show template authors

CodingOnStar:
## Summary

Recommend banner cards were cover-cropping non-square plugin icons (for example Alsa Go-to-Market's 83×98 SVG). Fit logos with `object-contain`. Template cards now render the served `creator` the same way plugins do (`by {author}`).

Marketplace companion: https://github.com/langgenius/dify-marketplace/pull/168

## Tests

- `home-trending.spec.tsx` (23)
- `vp check` on the two touched files


lin-snow (approved):
LGTM

PR #41993, opened 2026-09-08
fix: keep Notion table cell rich text segments in one Markdown column

### LINKED ISSUE #41992: Notion extractor corrupts Markdown tables for cells with mixed formatting or empty data cells
## Description

`NotionExtractor._read_table_rows` builds one Markdown column per rich text segment instead of per cell. In the Notion API, each table cell is an array of rich text segments, and a cell containing mixed formatting (for example plain text followed by bold text) arrives as multiple segments. Each segment currently becomes its own column, so the row has more columns than the header and the table is corrupted in the content indexed into the knowledge base.

A related case: an empty data cell (`[]`) is skipped entirely, so the row loses a column and all following cells in that row shift left.

## Steps to reproduce

```python
from unittest import mock
from core.rag.extractor import notion_extractor

def _mock_response(data):
    r = mock.Mock()
    r.status_code = 200
    r.json.return_value = data
    return r

ex = notion_extractor.NotionExtractor(
    notion_workspace_id="ws", notion_obj_id="obj", notion_page_type="page",
    tenant_id="tenant", notion_access_token="token",
)

page = {
    "results": [
        {"table_row": {"cells": [
            [{"text": {"content": "Name"}}],
            [{"text": {"content": "Desc"}}],
        ]}},
        {"table_row": {"cells": [
            [{"text": {"content": "item1"}}],
            # one cell, two rich text segments (e.g. plain + bold)
            [{"text": {"content": "plain "}}, {"text": {"content": "bold"}}],
        ]}},
    ],
    "next_cursor": None,
}

with mock.patch("httpx.request", side_effect=[_mock_response(page)]):
    print(ex._read_table_rows("tbl-1"))
```

Output on current `main`:

```
| Name | Desc |
| --- | --- |
| item1 | plain  | bold |
```

The data row has three columns against a two-column header. With an empty cell (`[]`) in a data row, the row instead comes out one column short and subsequent cells shift into the wrong columns.

## Expected behavior

Each Notion cell maps to exactly one Markdown column: its rich text segments are concatenated, and empty cells render as empty columns.

```
| Name | Desc |
| --- | --- |
| item1 | plain bold |
```

## Root cause

In `api/core/rag/extractor/notion_extractor.py`, `_read_table_rows` iterates the rich text segments inside each cell and appends every segment as a separate column entry, and skips empty data cells instead of emitting an empty column.

Harsh23Kashyap:
Fixes #41992

## What

`NotionExtractor._read_table_rows` treated every rich text segment inside a table cell as its own Markdown column. In the Notion API a cell is an array of rich text segments, so any cell with mixed formatting (plain + bold, links, mentions, etc.) produced extra columns and corrupted the table. Empty data cells (`[]`) were skipped entirely, which dropped a column and shifted the remaining cells in the row left.

## Fix

Build each cell as the concatenation of its rich text segments (skipping non-text segments such as mentions without a `text` payload, matching the existing convention in `_read_block`), and always emit one column per cell, including empty ones. Applied to both header and data rows.

## Verification

- Reproduced on current `main` with a mocked Notion API response: a two-segment cell produced the row `| item1 | plain  | bold |` (3 columns) against a 2-column header; an empty middle cell collapsed the row to fewer columns.
- Added two regression tests in `tests/unit_tests/core/rag/extractor/test_notion_extractor.py`:
  - `test_read_table_rows_joins_rich_text_segments_within_cell`
  - `test_read_table_rows_preserves_empty_data_cells_as_columns`
  Both fail before the fix and pass after.
- Full extractor suite: `pytest tests/unit_tests/core/rag/extractor/` - 229 passed.
- `ruff check` clean on both changed files.

Note: #37836 (guarding empty `results` pages) is complementary - it handles zero rows returned by the API, while this PR handles per-cell segment/column structure.

## Environment note

Tests were run locally with Python 3.12 against the unit test suite only; the Notion API is mocked in the new tests, so no external credentials are required.

### DIFF
--- a/api/core/rag/extractor/notion_extractor.py
+++ b/api/core/rag/extractor/notion_extractor.py
@@ -316,12 +316,14 @@ class NotionExtractor(BaseExtractor):
             table_header_cell_texts = []
             table_header_cells = data["results"][0]["table_row"]["cells"]
             for table_header_cell in table_header_cells:
-                if table_header_cell:
-                    for table_header_cell_text in table_header_cell:
-                        text = table_header_cell_text["text"]["content"]
-                        table_header_cell_texts.append(text)
-                else:
-                    table_header_cell_texts.append("")
+                # A cell is an array of rich text segments; join them so one
+                # cell always maps to one Markdown column.
+                text = "".join(
+                    table_header_cell_text["text"]["content"]
+                    for table_header_cell_text in table_header_cell
+                    if "text" in table_header_cell_text
+                )
+                table_header_cell_texts.append(text)
             # Initialize Markdown table with headers
             markdown_table = "| " + " | ".join(table_header_cell_texts) + " |\n"
             markdown_table += "| " + " | ".join(["---"] * len(table_header_cell_texts)) + " |\n"
@@ -331,11 +333,15 @@ class NotionExtractor(BaseExtractor):
             for i in range(len(results) - 1):
                 column_texts = []
                 table_column_cells = data["results"][i + 1]["table_row"]["cells"]
-                for j in range(len(table_column_cells)):
-                    if table_column_cells[j]:
-                        for table_column_cell_text in table_column_cells[j]:
-                            column_text = table_column_cell_text["text"]["content"]
-                            column_texts.append(column_text)
+                for table_column_cell in table_column_cells:
+                    # Join the rich text segments of each cell and keep empty
+                    # cells as empty columns so the column count stays stable.
+                    column_text = "".join(
+                        table_column_cell_text["text"]["content"]
+                        for table_column_cell_text in table_column_cell
+                        if "text" in table_column_cell_text
+                    )
+                    column_texts.append(column_text)
                 # Add row to Markdown table
                 markdown_table += "| " + " | ".join(column_texts) + " |\n"
             result_lines_arr.append(markdown_table)
--- a/api/core/rag/extractor/notion_extractor.py
+++ b/api/core/rag/extractor/notion_extractor.py
@@ -290,6 +290,13 @@ class NotionExtractor(BaseExtractor):
         result_lines = "\n".join(result_lines_arr)
         return result_lines
 
+    @staticmethod
+    def _get_cell_text(cell: list[dict[str, Any]]) -> str:
+        # A cell is an array of rich text segments (text, mention, equation);
+        # join them so one cell always maps to one Markdown column, keeping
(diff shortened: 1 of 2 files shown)


api/core/rag/extractor/notion_extractor.py
```diff
@@ -316,12 +316,14 @@ def _read_table_rows(self, block_id: str) -> str:
...
-                if table_header_cell:
-                    for table_header_cell_text in table_header_cell:
-                        text = table_header_cell_text["text"]["content"]
-                        table_header_cell_texts.append(text)
-                else:
-                    table_header_cell_texts.append("")
+                # A cell is an array of rich text segments; join them so one
+                # cell always maps to one Markdown column.
+                text = "".join(
+                    table_header_cell_text["text"]["content"]
```
fatelei: i think use plain text more better:https://developers.notion.com/reference/rich-text

because in notion document text | mention | equation

api/core/rag/extractor/notion_extractor.py
```diff
@@ -331,11 +333,15 @@ def _read_table_rows(self, block_id: str) -> str:
...
                 table_column_cells = data["results"][i + 1]["table_row"]["cells"]
-                for j in range(len(table_column_cells)):
-                    if table_column_cells[j]:
-                        for table_column_cell_text in table_column_cells[j]:
-                            column_text = table_column_cell_text["text"]["content"]
-                            column_texts.append(column_text)
+                for table_column_cell in table_column_cells:
+                    # Join the rich text segments of each cell and keep empty
+                    # cells as empty columns so the column count stays stable.
+                    column_text = "".join(
```
fatelei: you can add a helper method like 

```
def _get_cell_text(cell: list[dict[str, Any]]) -> str:
      return "".join(
          segment.get("plain_text")
          or segment.get("text", {}).get("content", "")
          for segment in cell
      )
```

make the code less duplicate

Harsh23Kashyap:
Thanks for the review. Both points are addressed in the latest push:

1. Cell text now prefers each segment's `plain_text`, so mention and equation segments (which have no `text` object) are included, falling back to `text.content` when `plain_text` is absent.
2. The shared `_get_cell_text` helper now backs both the header and body loops, removing the duplication.

Added a unit test covering mixed text/mention/equation segments and the empty-cell case. Extractor test suite passes (230 tests).

fatelei (approved):

PR #41983, opened 2026-09-08
refactor(models): remove the unread App.author_name db.session wrapper

ShousenZHANG:
## Summary

Part of #40372. Deletes `App.author_name` in `api/models/model.py`, a legacy `@property` whose entire body was `return self.author_name_with_session(session=db.session())`. The twin `author_name_with_session` is kept unchanged.

Same shape as my merged #41942 and #41964.

Claimed in https://github.com/langgenius/dify/issues/40372#issuecomment-5584206374.

## Why this is a pure deletion

Nothing reads the raw property. Every `.author_name` occurrence in `api/` has a different receiver:

| Site | Receiver |
|---|---|
| `tests/.../console/app/test_app_response_models.py:468` | `AppResponseView` |
| `tests/unit_tests/fields/test_dataset_fields.py:238` | `DatasetDetailResponse` |
| `tests/.../repositories/test_app_definition_query_repository.py:296` | the repository's result object |
| `tests/unit_tests/services/test_app_service.py:500` | `RecentAppListItem` |

`RecentAppListItem` (`services/app_service.py:140`) is a `@dataclass(frozen=True)` that declares `author_name: str | None` as its own field. That test also asserts the recent-apps query issues exactly **one** SELECT — a per-row lazy property would have broken that assertion, so the raw property demonstrably was not in play.

No `getattr`/`hasattr` on the name, and no `validation_alias` referencing it.

`Dataset.author_name` is a different accessor on a different model and is untouched.

## Verification

- `pytest tests/unit_tests/models/` — 423 passed, **identical to the `main` baseline** (2 pre-existing failures, 2 errors, unrelated)
- `pytest tests/unit_tests/services/test_app_service.py` — 33 passed, same on `main`
- `pytest tests/unit_tests/fields/test_dataset_fields.py` — 5 passed, same on `main`
- `pytest tests/unit_tests/repositories/test_app_definition_query_repository.py` — 22 passed, same on `main`
- `ruff check` / `ruff format --check` — clean
- `pyrefly check models/model.py` — 0 diagnostics
- `mypy --check-untyped-defs --disable-error-code=import-untyped models/model.py` — no issues
- `lint-imports` — 31 contracts kept, 0 broken
- `dev/lint_response_contracts.py --fail-on-mismatch` — 0 mismatch
- no net-new `getattr(` in the diff

`test_app_response_models.py` segfaults on my Windows box on clean `main`, so I could not run that one locally — relying on CI, as on my previous two PRs in this campaign.

---
[...]

PR #41998, opened 2026-09-08
refactor(models): remove the App.bound_agent_id and App.is_agent wrappers

ShousenZHANG:
## Summary

Part of #40372. Deletes two legacy `@property` wrappers on `App` in `api/models/model.py`:

| Removed | Kept (unchanged) |
|---|---|
| `App.bound_agent_id` | `bound_agent_id_with_session(self, *, session: Session)` |
| `App.is_agent` | `is_agent_with_session(self, *, session: Session)` |

Both bodies were `return self.<name>_with_session(session=db.session())`. Same shape as my merged #41942 and #41964.

Claimed in https://github.com/langgenius/dify/issues/40372#issuecomment-5585872421.

## Readers

No production code reads either property. The apparent readers:

| Site | What it actually is |
|---|---|
| `test_app_response_models.py:459,464` | `AppResponseView`, not the model — untouched |
| `test_hitl_service_api.py:403` | sets `is_agent` on a `MagicMock` — untouched |
| `test_app_service.py:648` | real read — updated |
| `test_app_models.py:187` | real read — updated |

## The two test changes

**`test_bound_agent_id_is_none_for_non_agent_app`** asserts a non-agent app short-circuits without a DB round trip. `agent_app_binding_with_session` returns on `self.mode != AppMode.AGENT` before the session is used, so the test now calls the twin with `session=None` and still proves exactly that — arguably more directly than before, since a session that cannot be used is stronger evidence than one that merely isn't.

**`test_app_is_agent_property_false`** had a `monkeypatch.setattr(model_module.db, "session", scoped_session(...))` whose only purpose was to make the wrapper resolve. Calling the twin drops that scaffolding, along with the now-unused `scoped_session` and `model_module` imports. Renamed to `test_app_is_agent_false_when_not_configured_as_agent` to match what it asserts.

## Verification

- `pytest tests/unit_tests/models/` — 428 passed, **identical to the `main` baseline** (2 pre-existing failures, 2 errors, unrelated)
- `pytest tests/unit_tests/services/test_app_service.py` — 33 passed, same on `main`
- `ruff check` / `ruff format --check` — clean
- `pyrefly check models/model.py` — 0 diagnostics
- `mypy --check-untyped-defs --disable-error-code=import-untyped models/model.py` — no issues
- `lint-imports` — 31 contracts kept, 0 broken
- `dev/lint_response_contracts.py --fail-on-mismatch` — 0 mismatch
- no net-new `getattr(` in the diff

`test_hitl_service_api.py` and `test_app_response_models.py` both segfault on my Windows box **on clean `main`**, so I could not run them locally. Neither touches the raw properties — one uses `MagicMock`, the other `AppResponseView` — but I'm relying on CI to confirm.

---
This PR was written with AI assistance.

### DIFF
--- a/api/models/model.py
+++ b/api/models/model.py
@@ -492,10 +492,6 @@ class App(Base):
 
         return None
 
-    @property
-    def bound_agent_id(self) -> str | None:
-        return self.bound_agent_id_with_session(session=db.session())
-
     def bound_agent_id_with_session(self, *, session: Session) -> str | None:
         """For an Agent App (mode=agent), the roster Agent it is backed by.
 
@@ -549,10 +545,6 @@ class App(Base):
     @property
     def tenant(self) -> Tenant | None:
         return db.session.scalar(select(Tenant).where(Tenant.id == self.tenant_id))
-
-    @property
-    def is_agent(self) -> bool:
-        return self.is_agent_with_session(session=db.session())
 
     def is_agent_with_session(self, *, session: Session) -> bool:
         """Detect legacy agent mode, committing the compatible app mode through the supplied session."""
--- a/api/tests/unit_tests/controllers/console/app/test_model_config_api.py
+++ b/api/tests/unit_tests/controllers/console/app/test_model_config_api.py
@@ -25,8 +25,9 @@ def _poison_implicit_app_config_properties(monkeypatch: pytest.MonkeyPatch) -> N
     def fail(_app: App) -> Never:
         raise AssertionError("implicit App model-config property was accessed")
 
+    # `App.is_agent` used to be poisoned here too. It no longer exists as an
+    # implicit property, so its absence enforces the same thing outright.
     monkeypatch.setattr(App, "app_model_config", property(fail))
-    monkeypatch.setattr(App, "is_agent", property(fail))
 
 
 @pytest.mark.parametrize("app_mode", [AppMode.CHAT, AppMode.COMPLETION])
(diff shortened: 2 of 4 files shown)


ShousenZHANG:
Correction — the first run of `API Unit Tests` here failed on me, not on infrastructure, and I want to be explicit about what I missed.

Three cases in `tests/unit_tests/controllers/console/app/test_model_config_api.py` failed with `AttributeError: <class 'models.model.App'> has no attribute 'is_agent'`. The cause is `_poison_implicit_app_config_properties` (`:24-29`), which installs a raising `property` over `App.is_agent` to assert the controller never touches it:

```python
monkeypatch.setattr(App, "is_agent", property(fail))
```

It names the attribute as a **string**, so my `git grep '\.is_agent\b'` sweep for readers never saw it. Fixed in `88cecfdd3f`: the poison line is dropped, since a deleted property enforces "never accessed" outright and more strongly than a poisoned one. The `app_model_config` poison on the line above stays — that property still exists and is untouched by this PR.

I then re-swept for the string form across every accessor I have removed in this campaign. `is_agent` had exactly this one; `bound_agent_id`'s string hits are all `payload.pop("bound_agent_id")`-style operations on already-serialised dicts, which don't touch the model class. I also went back over the accessors in my merged #41942 / #41964 / #41983 — their string hits are dict keys, SQL labels, module names and the `validation_alias` entries that go through `AppResponseView`, with no `setattr` on the class anywhere. So nothing needs revisiting there.

The earlier `Step canceled by GitHub` failures on this PR were a separate, genuine runner issue; that is what the empty commit re-triggered.

laipz8200 (approved):

asukaminato0721 (approved):

PR #42015, opened 2026-09-08
fix: parse first complete unfenced JSON value

### LINKED ISSUE #42006: Question Classifier can fail with JSONDecodeError "Extra data" when model returns multiple JSON objects
### Self Checks

- [x] I have read the Contributing Guide and searched for existing issues, including closed ones.
- [x] This is a bug report.
- [x] I am submitting this report in English.

### Dify version

Observed on self-hosted Dify app version 1.13.3 (Enterprise chart 3.9.10 and 3.9.11). The same parser implementation is still present on current `main` in `api/libs/json_in_md_parser.py`.

### Cloud or Self Hosted

Self Hosted

### Steps to reproduce

The failure occurs when a Question Classifier model returns more than one JSON object without a code fence.

A minimal parser-level reproduction is:

```python
from libs.json_in_md_parser import parse_json_markdown

text = '{"a": 1}\n{"a": 2}'
parse_json_markdown(text)
```

`parse_json_markdown()` finds the first opening `{` but uses `rfind("}")` / `rfind("]")` to determine the end of the JSON block. Therefore the extracted slice contains both JSON objects:

```text
{"a": 1}
{"a": 2}
```

Passing that slice to `json.loads()` raises:

```text
json.decoder.JSONDecodeError: Extra data
```

In a real Question Classifier workload we reproduced the same class of failure at low load, so it is not a saturation/concurrency issue. The exact `Extra data` position varies because the model output shape is nondeterministic.

### Expected Behavior

Question Classifier should parse the first valid JSON object returned by the model, or otherwise reject malformed output in a way that does not incorrectly concatenate multiple JSON objects into one parse attempt.

### Actual Behavior

`parse_json_markdown()` currently anchors from the first `{` / `[` to the last `}` / `]` in the entire model response. If the response contains two JSON objects (or JSON plus another JSON-like block), `json.loads()` receives both and raises `Extra data`.

This can cause a Question Classifier workflow execution to fail even when the model produced a valid first JSON object.

### Relevant implementation

Current `api/libs/json_in_md_parser.py` does roughly:

```python
start_candidates = [i for i in (json_string.find("{"), json_string.find("[")) if i != -1]
start_index = min(start_candidates)
end_index = max(json_string.rfind("}"), json_string.rfind("]"))
extracted_content = json_string[start_index:end_index + 1].strip()
return json.loads(extracted_content)
```

The `rfind()` end selection is the problematic part when more than one JSON value is present.

### Suggested fix

Use `json.JSONDecoder().raw_decode()` starting at the first JSON token and consume only the first complete JSON value, instead of slicing through the last closing bracket in the whole response.

For example, conceptually:

```python
decoder = json.JSONDecoder()
obj, end = decoder.raw_decode(json_string[start_index:])
return obj
```

The exact implementation should preserve the existing fenced-content fallback behavior and expected-key validation.

### Additional context

- Reproduced under low concurrency, so this is independent of load.
- The failure position varies across runs, consistent with varying model output shape rather than a fixed input-specific problem.
- Structured output is not available for this Question Classifier path in the observed version, so the parser should be robust to ordinary model output variation.
- No customer-specific data is required to reproduce this issue; the two-object example above reproduces the parser behavior directly.

fritz-fujii:
## Summary

Fix `parse_json_markdown()` so multiple unfenced JSON values do not get concatenated into a single `json.loads()` call.

The parser now consumes the first complete unfenced JSON value while preserving the existing fenced-content and scalar behavior.

## Problem

Given:

```text
{"a": 1}
{"a": 2}
```

the current implementation selects the first opening brace and the last closing brace, causing both values to be passed to `json.loads()` and raising `JSONDecodeError: Extra data`.

This surfaces in the Question Classifier node when a model emits more than one JSON value in a row.

## Fix

Use `json.JSONDecoder().raw_decode()` for the unfenced JSON path so parsing stops after the first complete JSON value.

Existing fenced JSON behavior is intentionally preserved — in particular, `test_parse_and_check_json_markdown_multiple_blocks_fails` (multiple fenced blocks still fail) is left unchanged.

## Tests

Added regression coverage for multiple unfenced JSON values and verified the existing JSON-in-Markdown parser tests continue to pass.

New cases: first value is an object / an array / contains nested objects / contains brackets inside string values / is followed by prose, a malformed first value still raises, and the reported `parse_and_check_json_markdown` failure mode now resolves.

`tests/unit_tests/libs/test_json_in_md_parser.py`: **20 passed** (13 existing + 7 new).

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs) — not needed, internal parser behavior only
- [ ] I've updated the documentation accordingly. — N/A

> Note on the last item: the full `make` targets need `uv`, which was not available in my environment. I ran the equivalent checks scoped to the two changed files instead: `ruff check` (all checks passed), `ruff format --check` (already formatted), and `mypy --check-untyped-defs` (no issues). Happy to run anything else the full pipeline flags.

Fixes #42006

From Claude Code

### DIFF
--- a/api/libs/json_in_md_parser.py
+++ b/api/libs/json_in_md_parser.py
@@ -21,7 +21,13 @@ def parse_json_markdown(json_string: str):
         # than one JSON value in a row, and slicing to the last bracket hands
         # all of them to json.loads() at once, which fails with "Extra data".
         # Decode only the first complete value instead.
-        if "`" not in json_string:
+        #
+        # A lone backtick is not a reliable "this is fenced" signal: it can
+        # legitimately occur inside a JSON string value (e.g. "use `print`")
+        # or in surrounding prose. Only a real markdown code fence (```) means
+        # the block below - which anchors on the last bracket and therefore
+        # still fails on multiple fenced blocks - should run instead.
+        if "```" not in json_string:
             first_value, _ = json.JSONDecoder().raw_decode(json_string, start_index)
             return first_value
 
--- a/api/libs/json_in_md_parser.py
+++ b/api/libs/json_in_md_parser.py
@@ -15,6 +15,16 @@ def parse_json_markdown(json_string: str):
     start_candidates = [i for i in (json_string.find("{"), json_string.find("[")) if i != -1]
     if start_candidates:
         start_index = min(start_candidates)
+
+        # Without code fences there is no end marker to anchor on, so the last
+        # "}" / "]" may belong to a *later* value. Models sometimes emit more
+        # than one JSON value in a row, and slicing to the last bracket hands
+        # all of them to json.loads() at once, which fails with "Extra data".
+        # Decode only the first complete value instead.
+        if "`" not in json_string:
+            first_value, _ = json.JSONDecoder().raw_decode(json_string, start_index)
+            return first_value
+
         end_index = max(json_string.rfind("}"), json_string.rfind("]"))
         if end_index != -1 and start_index < end_index:
             end_index += 1
--- a/api/tests/unit_tests/libs/test_json_in_md_parser.py
+++ b/api/tests/unit_tests/libs/test_json_in_md_parser.py
@@ -165,3 +165,28 @@ def test_parse_and_check_json_markdown_multiple_unfenced_objects():
     src = '{"keywords": ["a"], "category_id": "1", "category_name": "x"}\n{"category_id": "2"}'
     obj = parse_and_check_json_markdown(src, ["keywords", "category_id", "category_name"])
     assert obj == {"keywords": ["a"], "category_id": "1", "category_name": "x"}
+
+
+def test_parse_json_markdown_backtick_in_string_value_followed_by_second_unfenced_value():
+    """A backtick inside a string value must not disable the first-value-only fix.
+
+    Regression case from PR review: a single stray backtick anywhere in the
+    input previously fell back to anchoring on the *last* bracket, which
+    reintroduces the original "Extra data" bug from #42006 once a second
+    unfenced value follows.
+    """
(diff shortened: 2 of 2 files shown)


api/libs/json_in_md_parser.py
```diff
@@ -15,6 +15,16 @@ def parse_json_markdown(json_string: str):
     start_candidates = [i for i in (json_string.find("{"), json_string.find("[")) if i != -1]
     if start_candidates:
         start_index = min(start_candidates)
+
+        # Without code fences there is no end marker to anchor on, so the last
+        # "}" / "]" may belong to a *later* value. Models sometimes emit more
+        # than one JSON value in a row, and slicing to the last bracket hands
+        # all of them to json.loads() at once, which fails with "Extra data".
+        # Decode only the first complete value instead.
+        if "`" not in json_string:
```
41tair: if a string like {"code":"use` `print` function","n":1} the raw_decode can't called.

41tair (changes requested):

api/libs/json_in_md_parser.py
fritz-fujii: You're right — checking for the presence of any single backtick anywhere in the string was too broad, since a legitimate JSON string value can contain one (e.g. `"use `print` function"`). Fixed to only treat a real markdown fence (triple backtick, ``````) as "fenced"; anything else now always goes through `raw_decode()` for the first complete value.

Added a regression test for exactly this case: a backtick inside a JSON string value followed by a second unfenced value (this used to fall back to the old last-bracket anchor and reproduce the original "Extra data" bug from #42006).

Verified 156 related tests, ruff, mypy, and pyrefly all pass with the fix.

PR #42021, opened 2026-09-09
fix: validate declared blob size before allocating buffer in chunk merger

### LINKED ISSUE #42020: Plugin chunk merger pre-allocates buffer from unvalidated plugin-declared size
## Description

`merge_blob_chunks` in `api/core/plugin/utils/chunk_merger.py` merges streaming `BLOB_CHUNK` messages from plugins into complete blobs. For each new file it immediately pre-allocates a `bytearray` of the plugin-declared `total_length` - before any validation against `max_file_size`.

Because `total_length` comes from the plugin:

- A plugin declaring a multi-GB `total_length` causes the worker to allocate that much memory on the very first chunk, even though the documented limit is 30MB (`PLUGIN_MAX_FILE_SIZE`) - memory exhaustion from untrusted input.
- A plugin declaring an absurd `total_length` (for example `2**62`) raises an unhandled `MemoryError`/`OverflowError` instead of the documented `ValueError` size-limit error.

The existing size checks only run afterwards, against `bytes_written + len(blob_data)`, so they never get a chance to fire before the allocation.

## Steps to reproduce

```python
from core.plugin.utils.chunk_merger import merge_blob_chunks
from core.tools.entities.tool_entities import ToolInvokeMessage

def gen():
    yield ToolInvokeMessage(
        type=ToolInvokeMessage.MessageType.BLOB_CHUNK,
        message=ToolInvokeMessage.BlobChunkMessage(
            id="f1", sequence=0, total_length=2**62, blob=b"x", end=False
        ),
    )

list(merge_blob_chunks(gen()))
```

On current `main` this raises `MemoryError` (or, for claims like 8GB, actually allocates 8GB before any limit is enforced).

## Expected behavior

The declared `total_length` is validated against `max_file_size` before the buffer is allocated, so an oversized claim raises the same `ValueError("File is too large which reached the limit of ...MB")` as an oversized actual payload.

## Root cause

`api/core/plugin/utils/chunk_merger.py`: `files[chunk_id] = FileChunk(total_length)` runs before any bound check on `total_length`.

Harsh23Kashyap:
Fixes #42020

## What

`merge_blob_chunks` pre-allocates a `bytearray` of the plugin-declared `total_length` for every new blob-chunk stream, before validating it against `max_file_size`. Since `total_length` is plugin-controlled, a plugin claiming a huge size either exhausts worker memory on the first chunk (the 30MB `PLUGIN_MAX_FILE_SIZE` limit is never consulted first) or crashes with an unhandled `MemoryError`/`OverflowError` instead of the documented `ValueError` size-limit error.

## Fix

Validate `total_length > max_file_size` before creating the `FileChunk` buffer, raising the same `ValueError` message used for oversized actual payloads. Streaming behavior for in-limit files is unchanged; the existing per-chunk and cumulative checks still apply.

## Verification

- Reproduced on current `main`: a chunk declaring `total_length=2**62` raises `MemoryError` before the fix.
- Added two regression tests in `tests/unit_tests/core/plugin/utils/test_chunk_merger.py`:
  - `test_merge_blob_chunks_rejects_declared_size_above_limit` (declared 1025 with limit 1024 and a tiny payload - previously accepted)
  - `test_merge_blob_chunks_rejects_absurd_declared_size` (declared `2**62` - previously `MemoryError`)
  Both fail before the fix and pass after.
- `pytest tests/unit_tests/core/plugin/utils/` - 63 passed. Broader `tests/unit_tests/core/plugin/` - 465 passed; `test_plugin_runtime.py` has 2 failures that also fail on a clean checkout of `main` in my environment (they assert `127.0.0.1` where local resolution yields `localhost`), unrelated to this change.
- `ruff check` clean on both changed files.

## Environment note

Tests were run locally with Python 3.12 against the unit test suite only; no plugin daemon is required.

### DIFF
--- a/api/core/plugin/utils/chunk_merger.py
+++ b/api/core/plugin/utils/chunk_merger.py
@@ -61,9 +61,7 @@ def merge_blob_chunks[T: ToolInvokeMessage | AgentInvokeMessage](
                 # (or raise an unhandled MemoryError/OverflowError) before the
                 # size limit below is ever checked.
                 if total_length > max_file_size:
-                    raise ValueError(
-                        f"File is too large which reached the limit of {max_file_size / 1024 / 1024}MB"
-                    )
+                    raise ValueError(f"File is too large which reached the limit of {max_file_size / 1024 / 1024}MB")
                 files[chunk_id] = FileChunk(total_length)
 
             # Check if file is too large (before appending)
--- a/api/core/plugin/utils/chunk_merger.py
+++ b/api/core/plugin/utils/chunk_merger.py
@@ -55,6 +55,15 @@ def merge_blob_chunks[T: ToolInvokeMessage | AgentInvokeMessage](
 
             # Initialize buffer for this file if it doesn't exist
             if chunk_id not in files:
+                # Validate the declared size before allocating the buffer.
+                # total_length comes from the plugin and the whole buffer is
+                # pre-allocated, so an oversized claim would exhaust memory
+                # (or raise an unhandled MemoryError/OverflowError) before the
+                # size limit below is ever checked.
+                if total_length > max_file_size:
+                    raise ValueError(
+                        f"File is too large which reached the limit of {max_file_size / 1024 / 1024}MB"
+                    )
                 files[chunk_id] = FileChunk(total_length)
 
             # Check if file is too large (before appending)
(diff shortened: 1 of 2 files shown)


fatelei (approved):

CryoThrust:
The pre-allocation guard and regression coverage look good. One small non-blocking boundary question: if a malformed plugin sends a negative total_length, the comparison passes and bytearray(total_length) raises its own negative-count exception rather than the chunk merger's documented size-limit ValueError. If the message schema guarantees a non-negative integer this is already enforced upstream; otherwise it may be worth folding total_length < 0 into the same validation branch (or adding a dedicated invalid-length error) so all untrusted length claims fail before allocation with a stable API error.

PR #42034, opened 2026-09-09
refactor(models): pass session into WorkflowRun accessors and drop dead ones

keith991001:
> [!IMPORTANT]
>
> 1. Associated issue: #40372 (claimed in [this comment](https://github.com/langgenius/dify/issues/40372#issuecomment-5580534358) alongside the merged #41968; this covers the remaining `workflow.py` accessors).

## Summary

Part of #40372, follow-up to #41968 — this finishes `models/workflow.py`: no `db.session`-backed `@property` remains in the file.

- `WorkflowRun.created_by_account` / `created_by_end_user` (both marked `@deprecated`) → converted to methods taking `session`, matching the `WorkflowAppLog` / `WorkflowNodeExecutionModel` pairs in the same file. Their only remaining readers were implicit `from_attributes=True` reads by the run list / detail response models, now routed through a session-carrying `WorkflowRunResponseSource` (plus a pagination helper that wraps each run in the page) at the seven console validation sites (app workflow-runs list ×2 / detail, snippets list / detail, rag-pipeline list / detail). Each site reuses `db.session()` at the boundary. The advanced-chat list keeps working through `WorkflowWithMessage`'s attribute passthrough — the source resolves the accessor on the underlying run while `message_id` / `conversation_id` keep proxying.
- `WorkflowRun.message` / `WorkflowRun.workflow` (also `@deprecated`) → removed: an untruncated sweep across production, serializers, helper-indirected `dump_response` sites, events, and tests found zero readers.

## Tests

- New `test_workflow_run_accessors.py`: both accessors against the real `sqlite_session` fixture (role dispatch + None paths), the detail source resolving through its wrapped session while proxying other attributes, and the pagination helper wrapping every run in a page.
- `test_workflow_run_api.py` contract tests adapted: the `db.session` monkeypatch is now a callable (matching the scoped-session contract the controllers rely on), and the two mocked paginations use object-shaped results matching what the repository layer actually returns.
- Full `tests/unit_tests` run: 18661 passed; the only failures are the same pre-existing `test_check_no_new_getattr` environment failures that fail identically on a clean `main` checkout on this machine.

## Screenshots

N/A (backend-only refactor).

## Checklist

From Claude Code


keith991001:
Hi @asukaminato0721 — a friendly ping on this one whenever you have a moment. CI is all green and the branch is up to date. Happy to rebase or adjust anything if needed. Thanks for your time! 🙏

api/fields/workflow_run_fields.py
```diff
@@ -185,6 +185,38 @@ def _normalize_timestamp(cls, value: datetime | int | None) -> int | None:
...
+
+    @property
+    def created_by_account(self) -> Any:
+        return self.workflow_run.created_by_account(self.session)
+
+    @property
+    def created_by_end_user(self) -> Any:
+        return self.workflow_run.created_by_end_user(self.session)
+
+    def __getattr__(self, name: str) -> Any:
```
asukaminato0721: can we avoid getattr here?

api/controllers/console/datasets/rag_pipeline/rag_pipeline_workflow.py
asukaminato0721, line 918: can we inject session here?

asukaminato0721 (changes requested):

api/fields/workflow_run_fields.py
keith991001: @asukaminato0721 Good catch, thank you! You're right that this was adding yet another `no-new-getattr` exemption. Fixed in 26c3537: I promoted the existing `_SessionResponseSource` base from `conversation_fields.py` to `fields/base.py` and made both workflow-run wrappers subclass it, so the delegation now lives in the one place that already carries the exemption and this PR no longer introduces any new `getattr`. Please let me know if you'd prefer a different structure!

api/controllers/console/datasets/rag_pipeline/rag_pipeline_workflow.py
keith991001: Thanks for pointing this out! Fixed in 26c3537 — the endpoint now obtains the session once at the top and passes it to both the service and the response source. I applied the same change to the workflow-run detail endpoint in this file, which had the same double `db.session()` call. Happy to adjust further if you had a different injection pattern in mind!

keith991001:
Hi @asukaminato0721 — just a quick note that both review comments have been addressed in 26c3537 (shared `_SessionResponseSource` base instead of a new `__getattr__`, and single-session injection in the rag-pipeline endpoints), with replies in the respective threads. CI is green. Whenever you get a chance to take another look — and of course, happy to adjust further! 🙏

api/controllers/console/app/workflow_run.py
```diff
@@ -158,7 +161,10 @@ def get(self, req_data: WorkflowRunListQuery, request_context: RequestContext, a
             triggered_from=_triggered_from(req_data.triggered_from),
         )
 
-        return dump_response(AdvancedChatWorkflowRunPaginationResponse, result)
+        return dump_response(
+            AdvancedChatWorkflowRunPaginationResponse,
+            workflow_run_pagination_response_source(result, session=db.session()),
```
asukaminato0721: with_session

asukaminato0721 (changes requested):

api/controllers/console/app/workflow_run.py
keith991001: Done in 9247d1d — the eight read handlers this PR touched (workflow_run / rag_pipeline_workflow / snippet_workflow) now obtain their session via `@with_session(write=False)` instead of inline `db.session()`, and the handler unit tests pass the session explicitly (real sqlite fixture). Also rebased onto latest main.

asukaminato0721 (changes requested):
ci

keith991001:
> ci

thx! I fixed CI. plz check again

asukaminato0721 (approved):

PR #42068, opened 2026-09-09
fix(web): restore focus after closing overlays

iamjoel:
## Summary

Fixes [SNP-776](https://linear.app/dify/issue/SNP-776/修复关闭弹窗后焦点没有恢复).

Closing several overlays could leave focus on the page body or an unrelated control. For example, opening feedback or template details and moving the pointer away hid the button that should receive focus on close.

- Connect template details and chat feedback to the existing Dialog through DialogTrigger. Keep their actions keyboard-accessible and visible when focused, and restore feedback focus to the replacement rating button after submission.
- Connect both workflow and conversation log details to DrawerTrigger buttons while preserving whole-row clicks, URL state, and refresh behavior.
- Restore the Lexical editor's focus and saved selection after Escape cancels note-link editing, while preserving focus on outside controls when they dismiss the editor.

## Validation

- 165 focused unit tests passed.
- 7 Chromium tests passed, covering rendered visibility, keyboard access, cancellation, feedback submission, and typing at the restored note selection.
- Changed-file formatting, lint, and type checks passed.
- Repository-wide `vp check --no-fmt` and `vp run lint:eslint` passed with existing warnings.
- Full `vp run -w check` stops on an existing formatting issue in the unchanged `web/app/device/page.tsx`.

## Checklist

- [x] Linked the issue describing the change.
- [x] Added regression coverage for the changed behavior.
- [x] Frontend staged checks passed.
- [ ] Documentation update required (not applicable: no API or setup changes).

From Codex

### DIFF
--- a/web/app/components/base/chat/chat/answer/operation.tsx
+++ b/web/app/components/base/chat/chat/answer/operation.tsx
@@ -3,7 +3,6 @@ import type { ChatItem, Feedback } from '../../types'
 import { Button } from '@langgenius/dify-ui/button'
 import { cn } from '@langgenius/dify-ui/cn'
 import {
-  createDialogHandle,
   Dialog,
   DialogClose,
   DialogContent,
@@ -103,7 +102,6 @@ function Operation({
     readonly,
   } = useChatContext()
   const [isShowReplyModal, setIsShowReplyModal] = useState(false)
-  const [feedbackDialogHandle] = useState(() => createDialogHandle())
   // Submitting replaces the dialog trigger with the current rating button.
   const userFeedbackRef = useRef<HTMLButtonElement>(null)
   const adminFeedbackRef = useRef<HTMLButtonElement>(null)
@@ -189,12 +187,10 @@ function Operation({
     const succeeded = await handleFeedback('dislike', feedbackContent, feedbackTarget)
     if (!succeeded) return
 
-    setFeedbackContent('')
     setIsShowFeedbackModal(false)
   }
 
   const handleFeedbackCancel = () => {
-    setFeedbackContent('')
     setIsShowFeedbackModal(false)
   }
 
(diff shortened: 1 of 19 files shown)


web/app/components/base/chat/chat/answer/operation.tsx
```diff
@@ -459,67 +464,71 @@ function Operation({
...
-          open
-          onOpenChange={(open) => {
-            if (!open) handleFeedbackCancel()
-          }}
+      <Dialog
+        handle={feedbackDialogHandle}
+        open={isShowFeedbackModal}
+        onOpenChange={(open) => {
+          if (open) setIsShowFeedbackModal(true)
+          else handleFeedbackCancel()
```
lyzno1: [P2] Defer feedback draft cleanup until the close transition completes

Keeping the Root mounted correctly restores the exit lifecycle, but `handleFeedbackCancel()` still clears `feedbackContent` at the same time it sets `open` to false. Submission does the same. With animations enabled, I reproduced an empty textarea while the popup still had `data-ending-style` and opacity greater than zero: the text disappears before the dialog finishes closing.

Close immediately, but move draft cleanup to `onOpenChangeComplete(false)`:

```jsx
<Dialog
  open={isShowFeedbackModal}
  onOpenChange={setIsShowFeedbackModal}
  onOpenChangeComplete={(open) => {
    if (!open) setFeedbackContent('')
  }}
>
  {/* Existing feedback triggers and content */}
</Dialog>
```

The explicit cancel handler and successful submit path should only request closure; remove their immediate `setFeedbackContent('')` calls. Preserve the current behavior that a failed submission keeps the dialog and draft open. Add an animation-enabled regression case, since the shared browser setup disables Base UI animations by default. [Base UI lifecycle API](https://base-ui.com/react/components/dialog#api-reference)

web/app/components/workflow/note-node/note-editor/plugins/link-editor-plugin/hooks.ts
```diff
@@ -22,9 +22,16 @@ export const useOpenLink = () => {
...
+          const {
+            selectedLinkUrl,
+            selectedIsLink,
+            linkEditorDismissed,
+            setLinkAnchorElement,
+            setLinkOperatorShow,
+          } = noteEditorStore.getState()
           if (selectedIsLink) {
+            // Restoring the editor selection must not reopen a dismissed surface.
+            if (linkEditorDismissed) return
```
lyzno1: [P2] Scope dismissal to the link that was closed

`linkEditorDismissed` suppresses every subsequent link selection, rather than just reopening the dismissed link. I reproduced this in Chromium with two adjacent links, `hello` and `world`, pointing to different URLs: place the caret inside the first link, press Escape, then use End followed by ArrowLeft to enter the second link. The native selection is inside `world`, but its open/edit/unlink toolbar never appears because this branch still returns early.

Track the dismissed link's identity and clear the suppression when selection moves to a different link. Keep suppression when restoring the original selection or typing within the same dismissed link, so the original Escape regression stays fixed. Please cover keyboard movement between different links in addition to the existing mouse-click reopening case.

lyzno1 (changes requested):
Requesting changes for the two reproduced regressions below and a simpler trigger/root structure.

### Remove the unnecessary detached handles

All four new handles connect parts already owned by the same feature: template details, chat feedback, workflow logs, and conversation logs. Base UI supports regular triggers anywhere under the same Root, including inside child components. Root renders no DOM, and Portal placement does not break React context. Neither a separate component file nor a table requires a detached trigger.

Please keep the real Trigger elements, focus visibility changes, mounted Root lifecycle, and feedback `finalFocus`, but move each Root to the smallest common ancestor of its triggers and content. This removes the four handle instances and their wiring without changing the feature's state ownership.

For template details, the minimal structure is:

```jsx
<>
  <Dialog open={showDetailModal} onOpenChange={setShowDetailModal}>
    <Actions />
    <DialogContent>
      <Details onClose={closeDetailsModal} />
    </DialogContent>
  </Dialog>

  {/* Keep the independent edit and delete dialogs as siblings. */}
  <EditDialog />
  <DeleteDialog />
</>

// Inside Actions; keep its existing business props and other actions.
<DialogTrigger render={<Button>Details</Button>} />
```

These are structural sketches, not requests to introduce new wrapper components. Existing layout and business props can stay where they are. In `Operation`, use the same arrangement around the feedback actions and feedback content, keeping unrelated dialogs outside that Root. Preserve the feedback refs and `finalFocus`: submission replaces the original trigger with the rating button, so that explicit return target remains necessary.

For both log lists, the existing Drawer Root can enclose the table and its Portal:

```jsx
<Drawer open={showDrawer} onOpenChange={handleDrawerOpenChange}>
  <table>
    <tbody>
      {logs.map(log => (
        <tr key={log.id}>
          <td>
            <DrawerTrigger onClick={() => selectLogAndOpen(log)}>
              {formatTime(log.created_at)}
            </DrawerTrigger>
          </td>
        </tr>
      ))}
    </tbody>
  </table>
  <DrawerPortal>{/* Existing drawer anatomy and content */}</DrawerPortal>
</Drawer>
```

Keep the existing whole-row click forwarding, URL synchronization, selection, and refresh behavior; removing the handle does not require redesigning those paths. Multiple triggers can share a regular Root.

References: [Dialog detached/multiple triggers](https://base-ui.com/react/components/dialog#detached-triggers), [Drawer detached triggers](https://base-ui.com/react/components/drawer#detached-triggers).

### Verification

Reviewed commit `98fa612fc3af0daea30639770e122748f30539f2`. The 100 focused unit tests and 13 existing Chromium cases I reran passed. Two additional temporary Chromium probes reproduced the inline findings: feedback text disappearing during exit, and a dismissed link preventing the next adjacent link's toolbar from opening. The existing browser setup sets `BASE_UI_ANIMATIONS_DISABLED = true`; the feedback regression is observable when animations are enabled.

lyzno1 (approved):

PR #42096, opened 2026-09-10
fix(web): correct form labels, validation feedback, and keyboard navigation

iamjoel:
## Summary

- Associate form fields, switches, descriptions, and validation errors with their visible labels; focus the first visible invalid field within the submitting form.
- Make upload-method selection and option creation keyboard accessible, preserve recipient Tab navigation and variable-form submission, and keep credential actions and read-only textarea focus visible.
- Add regression coverage and remove the obsolete accessibility lint suppressions for option creation.

Fixes SNP-751

## Validation

- Initial changes: 142 related tests passed; affected flows were manually checked in Chrome.
- Base form follow-up: 115 Web tests and 7 Textarea Chromium tests passed.
- Full repository static checks and commit hooks passed (existing lint warnings remain).

### DIFF
--- a/web/app/components/base/form/components/base/base-field.tsx
+++ b/web/app/components/base/form/components/base/base-field.tsx
@@ -14,7 +14,7 @@ import {
   SelectValue,
 } from '@langgenius/dify-ui/select'
 import { useStore } from '@tanstack/react-form'
-import { isValidElement, memo, useCallback, useMemo } from 'react'
+import { isValidElement, memo, useCallback, useId, useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 import { CheckboxList } from '@/app/components/base/checkbox-list'
 import { FormItemValidateStatusEnum, FormTypeEnum } from '@/app/components/base/form/types'
@@ -143,6 +143,35 @@ const BaseField = ({
     help,
   } = formSchema
   const disabled = propsDisabled || formSchemaDisabled
+  const controlId = useId()
+  const labelId = `${controlId}-label`
+  const descriptionId = `${controlId}-description`
+  const messageId = `${controlId}-message`
+  const meta = useStore(field.form.store, (state) => state.fieldMeta[field.name])
+  const errors = fieldState?.errors ?? (meta?.isTouched ? meta.errors : [])
+  const validateStatus =
+    fieldState?.validateStatus ?? (errors?.length ? FormItemValidateStatusEnum.Error : undefined)
+  const messages =
+    validateStatus === FormItemValidateStatusEnum.Error
+      ? errors
+      : validateStatus === FormItemValidateStatusEnum.Warning
+        ? fieldState?.warnings
+        : undefined
+  const hasMessage = !!messages?.length
+  const controlProps = {
+    'aria-describedby':
+      [description && descriptionId, hasMessage && messageId].filter(Boolean).join(' ') ||
+      undefined,
+    'aria-invalid': validateStatus === FormItemValidateStatusEnum.Error || undefined,
+    'aria-required': required || undefined,
+  }
+  const isSingleControl = [
+    FormTypeEnum.textInput,
+    FormTypeEnum.secretInput,
+    FormTypeEnum.textNumber,
+    FormTypeEnum.select,
+    FormTypeEnum.dynamicSelect,
+  ].includes(formItemType)
 
   const [
     translatedLabel,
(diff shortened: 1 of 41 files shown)


lyzno1:
The native form boundary, explicit submit/cancel button types, label associations, and keyboard-access improvements address useful parts of the form guidance. However, I would not describe the affected forms as fully compliant yet.

For this PR, please replace the manually wired Select labels with `SelectLabel` from `@langgenius/dify-ui/select`, composed inside the corresponding `Select`. This applies to the field-type selector and both default-value selectors in `config-modal/form-fields.tsx`. The current `titleId` + `aria-labelledby` relationship supplies an accessible name, but the title is a plain `div`: clicking it does not focus the trigger. `SelectLabel` owns both the association and label activation. Please cover clicking the visible label and checking that the trigger receives focus, in addition to the existing accessible-name assertions.

To clarify the Field question: the `Field` imported from `./field` is a local layout component, not Base UI Field. Native `label htmlFor` + input `id` is valid; those attributes are not inherently a problem. For ordinary form rows, Dify UI `Field`, `FieldLabel`, and a participating input primitive would avoid maintaining these relationships manually. Simply replacing the outer wrapper is insufficient while the legacy input remains.

There are also existing gaps in the affected configuration form:
- Field validation still reports errors through `toast.error()`, without associating an error with the invalid control or moving focus to the first invalid field.
- `config-select/index.tsx` still renders option inputs without accessible labels and an “Add option” action as a click-only `div`.

These are pre-existing gaps, not regressions introduced by this PR. They can be tracked separately if outside the intended scope, but the PR should be described as targeted accessibility fixes rather than complete form-guidance compliance.

References: [Base UI form labeling guidance](https://base-ui.com/react/handbook/forms#naming-form-controls), [Dify UI forms contract](https://github.com/langgenius/dify/blob/3e14a0fc92c97f05a71359fd694eec768577769d/packages/dify-ui/docs/forms.md), and [accessible names and descriptions](https://github.com/langgenius/dify/blob/3e14a0fc92c97f05a71359fd694eec768577769d/packages/dify-ui/docs/accessible-names-and-descriptions.md).

web/app/components/base/form/components/base/base-field.tsx
```diff
@@ -143,6 +143,35 @@ const BaseField = ({
     help,
   } = formSchema
   const disabled = propsDisabled || formSchemaDisabled
+  const controlId = useId()
+  const labelId = `${controlId}-label`
+  const descriptionId = `${controlId}-description`
+  const messageId = `${controlId}-message`
+  const meta = useStore(field.form.store, (state) => state.fieldMeta[field.name])
+  const errors = fieldState?.errors ?? (meta?.isTouched ? meta.errors : [])
```
lyzno1: [P2] Preserve client validation after clearing server errors

An empty fieldState.errors array takes precedence over meta.errors here. This is reachable through the existing subscription-creation flow: handleApiKeyCredentialsChange and handleVerify call setFields([{ name, errors: [] }]) to clear server errors. Subsequent client validation can therefore block submission and focus an empty required input while leaving it without aria-invalid or an associated inline error. I reproduced this through the real BaseForm ref API: setFields with errors: [], then getFormValues({}) on an empty required token field; the focus assertion passes but toBeInvalid() fails. Clearing server errors should allow current client errors to become visible again; please define the precedence between the two error sources and cover this transition.

web/app/components/base/form/components/base/base-field.tsx
```diff
@@ -242,26 +271,35 @@ const BaseField = ({
     <>
       <div className={cn(fieldClassName)}>
         <div className={cn(labelClassName, formLabelClassName)}>
-          {translatedLabel}
+          {isSingleControl ? (
+            <label id={labelId} htmlFor={controlId}>
```
lyzno1, line 286: [P2] Use SelectLabel for the shared Select fields

isSingleControl includes both select and dynamicSelect, so this change labels their triggers through a native label/htmlFor relationship. The current tests demonstrate an accessible name and label activation, so this is a documented composition-contract issue rather than a missing-name defect: packages/dify-ui/docs/forms.md explicitly requires trigger-based Select fields to use SelectLabel. Native label activation also opens the popup in the added test, whereas SelectLabel owns the intended label-to-trigger focus behavior. Please compose SelectLabel inside each Select owner, as the updated configuration dialog already does, and retain the description/error relationships.

lyzno1 (changes requested):
Reviewed the full PR at fa512043858914613aa71cbde6d2583d91cb419c. Requesting changes for the two inline findings: clearing server errors suppresses subsequent client validation feedback, and the shared BaseField Select composition still does not follow the documented SelectLabel contract.

The configuration dialog now uses SelectLabel, option creation is a native button, upload-method selection uses RadioGroup, and Enter in the extension input no longer submits the configuration form.

Two previously identified gaps remain. These are existing issues, not regressions introduced by this PR, but they mean the affected forms are not yet fully compliant with the form guidance:
- ConfigModal.handleConfirm still reports validation errors only through toast.error. The new inline-error and first-invalid-field focus behavior belongs to BaseForm, which this dialog does not use.
- ConfigSelect's editable option inputs still have no associated labels or other accessible names, although the Add option action is now keyboard accessible.

Validation: all 131 tests in the 14 changed Web test files passed locally. An additional temporary regression test using the real BaseForm public ref API reproduced the first inline finding; the temporary test was then removed. CI had no failed or pending checks when reviewed. I did not independently run a real-browser or screen-reader check for this revision.

lyzno1 (approved):

PR #42125, opened 2026-09-10
chore: restore grid and table accessibility

iamjoel:
## Summary

- Restore native table and column-header semantics in document, log, annotation, CSV preview, member, and archive views without adding unnecessary `scope="col"` attributes.
- Expose sorting and unread states, label document switches, and make document navigation, retrieval-history reuse, and Webhook row deletion accessible by keyboard.
- Give calendar buttons localized Gregorian date names, selected/today states, native disabled behavior, and visible focus.

Fixes SNP-788

[Linear issue](https://linear.app/dify/issue/SNP-788)

## Validation

- Scoped format, lint, and type checks passed with 0 errors (25 warnings).
- Full repository check failed in unchanged main-navigation files: missing local `bowser` dependency (TS2307) and unsupported React `onBrowserBailout` option (TS2353).

- 3,608 unit tests passed across 219 files in the expanded regression run.
- 2 Chromium Browser Mode tests passed for the date picker and Webhook deletion focus.
- E2E locator audit found no affected steps; full E2E execution was not performed because the local runtime services were unavailable.
- Actual screen-reader testing and full-page visual verification remain outstanding.

From Codex

### DIFF
--- a/web/app/components/base/date-and-time-picker/calendar/item.tsx
+++ b/web/app/components/base/date-and-time-picker/calendar/item.tsx
@@ -2,11 +2,14 @@ import type { FC } from 'react'
 import type { CalendarItemProps } from '../types'
 import { cn } from '@langgenius/dify-ui/cn'
 import * as React from 'react'
+import { useTranslation } from 'react-i18next'
 import { useLocale } from '@/context/i18n'
 import dayjs from '../utils/dayjs'
 
 const Item: FC<CalendarItemProps> = ({ day, selectedDate, onClick, isDisabled }) => {
   const locale = useLocale()
+  const { t } = useTranslation('common')
+  const selectedDescriptionId = React.useId()
   const { date, isCurrentMonth } = day
   const isSelected = selectedDate?.isSame(date, 'date')
   const isToday = date.isSame(dayjs(), 'date')
@@ -20,7 +23,7 @@ const Item: FC<CalendarItemProps> = ({ day, selectedDate, onClick, isDisabled })
     <button
       type="button"
       aria-label={dateLabel}
-      aria-pressed={!!isSelected}
+      aria-describedby={isSelected ? selectedDescriptionId : undefined}
       aria-current={isToday ? 'date' : undefined}
       disabled={isDisabled}
       onClick={() => !isDisabled && onClick(date)}
@@ -34,6 +37,11 @@ const Item: FC<CalendarItemProps> = ({ day, selectedDate, onClick, isDisabled })
       )}
     >
       {date.date()}
+      {isSelected && (
+        <span id={selectedDescriptionId} className="sr-only">
+          {t(($) => $['calendar.selectedDate'])}
+        </span>
+      )}
       {isToday && (
         <div
           aria-hidden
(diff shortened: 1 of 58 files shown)


web/app/components/datasets/hit-testing/components/records.tsx
```diff
@@ -86,17 +97,23 @@ const Records = ({ records, onClickRecord }: RecordsProps) => {
...
                 </td>
                 <td className="w-48 p-3 pr-2">
-                  {formatTime(
-                    created_at,
-                    t(($) => $.dateTimeFormat, { ns: 'datasetHitTesting' }) as string,
-                  )}
+                  <button
+                    type="button"
+                    className="w-full cursor-pointer rounded-sm text-left focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-state-accent-solid"
+                    onClick={() => onClickRecord(record)}
```
lyzno1, line 110: [P3] Non-blocking: make the history reuse action explicit

The new button is named only by its timestamp, but `handleClickRecord` replaces the current query input with this record's queries. The English date format also stops at minutes, so different records created in the same minute have identical button names. Table navigation provides surrounding context, but Tab/button-list navigation does not identify which query will be reused or explain the action. Consider a localized reuse label with query/file context, preserving the visible timestamp in the accessible name if it remains the button text. This follows the descriptive-label intent of [WCAG 2.4.6](https://www.w3.org/WAI/WCAG22/Understanding/headings-and-labels.html).

web/app/components/base/date-and-time-picker/calendar/item.tsx
```diff
@@ -2,19 +2,33 @@ import type { FC } from 'react'
...
+    timeZone: 'UTC',
+  }).format(new Date(Date.UTC(date.year(), date.month(), date.date())))
 
   return (
     <button
       type="button"
+      aria-label={dateLabel}
+      aria-describedby={isSelected ? selectedDescriptionId : undefined}
+      aria-current={isToday ? 'date' : undefined}
+      disabled={isDisabled}
```
lyzno1, line 28: [P3] Non-blocking: reuse the native disabled state for presentation

Now that the button has `disabled={isDisabled}`, its disabled styling can use `disabled:cursor-not-allowed`, `disabled:text-text-quaternary`, and `disabled:hover:bg-transparent` instead of the `isDisabled && ...` class branch below. Native disabled behavior also makes the `!isDisabled` click guard redundant for user activation. This keeps the DOM state as the styling source and removes duplicate handling; it is a cleanup suggestion, not a WCAG failure. See [MDN :disabled](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:disabled).

lyzno1 (approved):
Reviewed the full diff at a76b08c1d2, including the affected consumers, table/header semantics, sorting state, accessible names, keyboard entry points, focus styling, and CSS state handling. No blocking regression found. The calendar aria-pressed issue has already been addressed in this head.

Two non-blocking follow-ups are inline: make the retrieval-history reuse action clearer, and let native disabled state own the calendar's disabled styling/activation.

Validation: 69 focused unit tests across 6 files and 2 Chromium Browser Mode tests passed locally. Web unit/browser, dify-ui/Storybook, and style CI checks have passed; Chromium full E2E and WebKit smoke are still running. This review does not establish full WCAG conformance: actual screen-reader and full-page visual/zoom checks were not performed.

PR #42149, opened 2026-09-10
fix(api): mark async trigger log failed when celery dispatch raises

### LINKED ISSUE #42148: Async workflow trigger log stuck in PENDING when Celery dispatch fails
When `AsyncWorkflowService.trigger_workflow_async` fails to enqueue the Celery task (e.g. broker down, `execute_workflow_*_team.delay()` raises), the trigger log row created just before stays in `PENDING` forever.

No worker will ever pick it up since the task was never queued, the quota reservation is refunded, and `get_failed_logs_for_retry` only surfaces `FAILED`/`RATE_LIMITED` logs — so the stuck `PENDING` log is invisible to the retry flow (`reinvoke_trigger`).

The quota-exceeded path in the same method already handles its failure correctly (marks the log `RATE_LIMITED` with the error and commits before re-raising). The dispatch-failure path should do the same: mark the log `FAILED` with the dispatch error before re-raising.

soroush5:
Fixes #42148

Summary:
- If the Celery `delay()` call in `trigger_workflow_async` raises, the trigger log kept its initial `PENDING` status even though no task was ever queued. It stayed that way permanently: no worker picks it up, and the retry query only returns `FAILED`/`RATE_LIMITED` logs.
- The dispatch-failure handler now marks the log `FAILED` with the dispatch error and commits before re-raising, mirroring the existing quota-exceeded path in the same method. This also makes failed dispatches retryable via `reinvoke_trigger`.

Checklist:
- [x] Added regression test (`test_should_mark_log_failed_when_celery_dispatch_raises`)
- [x] New test fails on base (`pending` instead of `failed`), passes with the fix
- [x] Full `test_async_workflow_service.py` suite green (18 passed); neighboring trigger/task suites green (132 passed, 1 skipped)
- [x] `ruff format --check` and `ruff check` clean; no new mypy errors in the changed file

### DIFF
--- a/api/services/async_workflow_service.py
+++ b/api/services/async_workflow_service.py
@@ -175,8 +175,22 @@ class AsyncWorkflowService:
             else:  # SANDBOX
                 task = execute_workflow_sandbox.delay(task_data_dict)
             quota_charge.commit()
-        except Exception:
+        except Exception as e:
             quota_charge.refund()
+            # The Celery task was never enqueued, so no worker will pick it up.
+            # Mark the log FAILED (with the dispatch error) instead of leaving it
+            # PENDING forever, invisible to the failed-log retry query.
+            trigger_log.status = WorkflowTriggerStatus.FAILED
+            trigger_log.error = f"Failed to dispatch workflow task: {e}"
+            trigger_log_repo.update(trigger_log)
+            session.commit()
+            logger.exception(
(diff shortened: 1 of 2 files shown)


soroush5:
Closing as stale, no activity for a while. Happy to rebase and reopen if there is interest. Thanks for your time.

PR #42171, opened 2026-09-11
fix(rag): preserve Notion mention and equation text

### LINKED ISSUE #42170: Notion page extraction drops mention and equation rich-text segments
## Description

`NotionExtractor` preserves all Notion rich-text segments in database properties and table cells, but the normal page/block extraction paths still keep only segments that contain a `text` object.

Notion rich text can also contain `mention` and `equation` objects. Those segments expose their rendered value through `plain_text`, so they are currently skipped when a Notion page is imported into a knowledge base.

This affects both:

- `_get_notion_block_data()` for top-level page blocks;
- `_read_block()` for recursively-read child blocks.

## Reproduction

On current `main`, a paragraph payload containing mixed rich text such as:

```python
[
    {"type": "text", "plain_text": "See ", "text": {"content": "See "}},
    {"type": "mention", "plain_text": "Project Alpha", "mention": {"type": "page", "page": {"id": "page-id"}}},
    {"type": "equation", "plain_text": "E = mc^2", "equation": {"expression": "E = mc^2"}},
]
```

is processed by the current loop as only:

```text
See 
```

The mention and equation text are omitted because the implementation checks `if "text" in rich_text` and then reads only `rich_text["text"]["content"]`.

## Expected behavior

All rendered rich-text segments should be preserved in extracted page content, using each segment's `plain_text` value (with the existing `text.content` shape retained only as a fallback if needed).

For the payload above, the extracted text should contain:

```text
See Project AlphaE = mc^2
```

## Upstream precedent

This is consistent with the recently merged table-cell fix in #41993. During review, `plain_text` was explicitly preferred because Notion rich text can contain `text`, `mention`, and `equation` segments, and the merged `_get_cell_text()` helper now follows that rule.

The recently merged database-property fix in #42050 also joins all `plain_text` segments instead of taking only a subset.

The two ordinary page/block paths appear to be the remaining sibling omission in the same extractor.

## Proposed scope

Keep the fix narrow:

1. preserve `plain_text` for every rich-text segment in `_get_notion_block_data()`;
2. do the same in `_read_block()`;
3. add focused regression coverage for top-level and nested blocks containing text + mention + equation segments;
4. leave database-property and table-cell behavior unchanged.

I have the scoped change ready and will link the PR here.

Bruce-Yii:
## Summary

Fixes #42170.

Notion page imports currently keep only rich-text segments that contain a `text` object in the ordinary page/block paths. Valid `mention` and `equation` segments therefore disappear from indexed knowledge content even though Notion provides their rendered value through `plain_text`.

This change keeps the existing extraction structure and only changes how each rich-text segment is read:

- prefer the segment's `plain_text`, which covers text, mentions, and equations;
- fall back to the existing `text.content` shape when `plain_text` is absent;
- continue skipping genuinely empty segments;
- preserve the current block joining, indentation, recursion, table, and database-property behavior.

The same `plain_text` rule is already used by the merged table-cell path after #41993 and by database rich-text/title extraction after #42050.

## Validation

Added focused regression coverage for both affected paths:

- `_get_notion_block_data()` preserves a mixed text + page mention + equation payload;
- `_read_block()` preserves the same segments while retaining child-block indentation.

The tests exercise the real extractor methods with only the Notion HTTP response mocked. They are intended to fail on the pre-fix implementation because it drops the mention and equation segments.

Repository test/lint suites were not run locally in this connector-backed environment; CI is left to validate the submitted branch.

## Screenshots

Not applicable; backend extraction only.

## Checklist

- [ ] This change requires a documentation update — not expected; no API or configuration change.
- [x] Added focused regression coverage for the changed behavior.
- [ ] Ran the repository-wide backend lint/type-check locally — CI pending.

From ChatGPT

### DIFF
--- a/api/core/rag/extractor/notion_extractor.py
+++ b/api/core/rag/extractor/notion_extractor.py
@@ -205,9 +205,8 @@ class NotionExtractor(BaseExtractor):
                 else:
                     if "rich_text" in result_obj:
                         for rich_text in result_obj["rich_text"]:
-                            # skip if doesn't have text object
-                            if "text" in rich_text:
-                                text = rich_text["text"]["content"]
+                            text = rich_text.get("plain_text") or rich_text.get("text", {}).get("content", "")
+                            if text:
                                 cur_result_text_arr.append(text)
 
                     result_block_id = result["id"]
@@ -263,9 +262,8 @@ class NotionExtractor(BaseExtractor):
                 else:
                     if "rich_text" in result_obj:
                         for rich_text in result_obj["rich_text"]:
-                            # skip if doesn't have text object
-                            if "text" in rich_text:
-                                text = rich_text["text"]["content"]
+                            text = rich_text.get("plain_text") or rich_text.get("text", {}).get("content", "")
+                            if text:
                                 prefix = "\t" * num_tabs
                                 cur_result_text_arr.append(prefix + text)
                     result_block_id = result["id"]
--- a/api/tests/unit_tests/core/rag/extractor/test_notion_page_rich_text.py
+++ b/api/tests/unit_tests/core/rag/extractor/test_notion_page_rich_text.py
@@ -3,7 +3,7 @@ from unittest import mock
 from core.rag.extractor import notion_extractor
 
 
-def _mock_response(data):
+def _mock_response(data: dict[str, object]) -> mock.Mock:
     response = mock.Mock()
     response.status_code = 200
     response.text = ""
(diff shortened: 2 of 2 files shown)


41tair (dismissed):

Bruce-Yii:
@41tair Thanks for the review. I pushed one follow-up commit after the Pyrefly report showed two newly untyped symbols in the regression-test module. The commit only adds return type annotations in the test file; production code is unchanged. This push dismissed the previous approval/auto-merge state. Could you please re-review the latest head when convenient?

Bruce-Yii:
@41tair Quick follow-up on this one: your previous approval was dismissed only because I pushed the test-only return type annotations after the Pyrefly report; production code is unchanged, and there are currently no unresolved review threads.

Could you please re-review the latest head when convenient? Thanks.

41tair (approved):

PR #42186, opened 2026-09-11
fix(web): use tool-call capability for customizable Agent V2 models

### LINKED ISSUE #42092: Agent V2: clarify compatibility policy for customizable models with declared tool-call capability
## Problem

Agent V2 currently applies the same name-based compatibility baseline to both predefined and customizable models.

This seems reasonable for predefined models, where Dify may intentionally maintain a curated quality/reliability baseline beyond raw API capabilities.

For customizable models, however, this can produce a confusing result:

a model may explicitly declare function/tool-calling support through Dify's existing model capability metadata, but still be classified as `Incompatible` solely because its configured label matches one of the predefined model-family patterns.

For example, this class of behavior was reported in #39623 with a customizable OpenAI-compatible Qwen model.

## Current behavior

Agent V2 currently determines compatibility in:

`web/features/agent-v2/agent-detail/configure/model-compatibility.ts`

The compatibility predicate checks `modelItem.label.en_US` against a curated list of model-name patterns.

It does not currently distinguish between:

- `predefined-model`
- `customizable-model`

or use the model's declared function-calling capability for this decision.

This affects the product surface rather than directly blocking the runtime:

- incompatible models are hidden from the default selector view;
- users can explicitly reveal them through **Show incompatible models**;
- an active model may still be selectable after being revealed;
- once selected, it is still shown as **Incompatible**, with the tooltip saying that the model is not available in the current version.

## Existing capability metadata

Dify already exposes both `fetch_from` and `features` in the model contract.

The selector also already has function-call capability handling for model feature filtering.

So the question is not whether Dify should introduce a new capability-detection system.

The question is whether the **predefined-model compatibility baseline is intentionally meant to apply unchanged to customizable models even when those models explicitly declare the technical capability Agent V2 requires.**

## Proposed behavior

If this is not intentional, I suggest separating two concepts.

### Predefined models

Keep the existing curated Agent compatibility baseline unchanged.

Dify can continue deciding which predefined models meet the expected Agent quality/reliability baseline.

### Customizable models

When a customizable model explicitly declares supported function/tool-calling capability, it should not be classified as `Incompatible` solely because its user-configured label matches a predefined-model blacklist pattern.

This would only establish technical eligibility.

It would **not** make the model a Dify-recommended model.

## Suggested models

The existing `Suggested` policy should remain fully curated by Dify.

A customizable model becoming technically eligible should never automatically make it `Suggested`.

In other words:

**technical eligibility != Dify recommendation**

## Scope

The smallest implementation appears to be limited to the model-selection layer:

1. expose the existing `fetch_from` field on the local `ModelSelectorModel` type;
2. preserve the current compatibility behavior for predefined models;
3. use the existing declared function-calling capability when evaluating customizable models;
4. keep `isAgentSuggestedModel()` unchanged;
5. add focused regression tests for predefined and customizable behavior.

The model-list API already carries `fetch_from`, so this should not require a new backend contract or runtime capability-probing mechanism.

## Non-goals

This proposal does not attempt to:

- guarantee that arbitrary self-hosted models work correctly with Agent V2;
- make every tool-calling model Agent-compatible;
- remove Dify's curated predefined-model baseline;
- automatically recommend customizable models;
- change Agent V2 runtime behavior;
- probe model endpoints automatically;
- solve provider-specific vLLM/OpenAI-compatible runtime issues.

Those provider/runtime compatibility problems should remain separate.

## Acceptance cases

Existing predefined-model compatibility and Suggested behavior should remain unchanged.

For customizable models:

- a customizable model with explicitly declared supported function-calling capability should not become `Incompatible` solely because its label matches a predefined-model pattern;
- a customizable model without the required declared capability should remain unsupported;
- a technically eligible customizable model should not automatically become `Suggested`.

## Maintainer question

Is applying the predefined Agent compatibility baseline unchanged to `customizable-model` entries an intentional product policy?

If not, I would like to implement the scoped distinction above and add focused regression tests.

If this direction is acceptable, please let me know or assign the issue to me. I am also happy to adjust the scope before implementation based on the intended Agent V2 model policy.

Souravrajvi0:
## Summary

Fixes #42092

Customizable models were classified as `Incompatible` in Agent V2 when their configured label matched a predefined-model blacklist pattern, even when the model explicitly declared function/tool-calling support.

This PR separates the two policies:

- **Predefined models** — keep the existing curated Agent compatibility baseline (name-pattern blacklist unchanged)
- **Customizable models** — evaluate technical eligibility via declared function-calling capability (`tool-call`, `multi-tool-call`, or `stream-tool-call`)

`isAgentSuggestedModel()` is unchanged; technically eligible customizable models are not auto-suggested.

## Changes

- Expose `fetch_from` on `ModelSelectorModel` so compatibility logic can distinguish model origin
- Route customizable models through `supportFunctionCall()` instead of the predefined blacklist
- Add focused regression tests for customizable vs predefined behavior

## Test plan

- [x] `cd web && node_modules/.bin/vitest run features/agent-v2/agent-detail/configure/__tests__/model-compatibility.spec.ts` (11 tests pass)

### DIFF
--- a/web/app/components/header/account-setting/model-provider-page/model-selector/types.ts
+++ b/web/app/components/header/account-setting/model-provider-page/model-selector/types.ts
@@ -12,6 +12,7 @@ export type ModelSelectorValue = {
 export type ModelSelectorModel = Pick<
   ProviderModelWithStatusEntity,
   | 'deprecated'
+  | 'fetch_from'
   | 'has_invalid_load_balancing_configs'
   | 'label'
   | 'load_balancing_enabled'
--- a/web/features/agent-v2/agent-detail/configure/__tests__/model-compatibility.spec.ts
+++ b/web/features/agent-v2/agent-detail/configure/__tests__/model-compatibility.spec.ts
@@ -4,6 +4,7 @@ import type {
 } from '@dify/contracts/api/console/workspaces/types.gen'
 import {
   ConfigurationMethodEnum,
+  ModelFeatureEnum,
   ModelStatusEnum,
   ModelTypeEnum,
 } from '@/app/components/header/account-setting/model-provider-page/declarations'
@@ -172,9 +173,78 @@ describe('isAgentCompatibleModel', () => {
     expect(isAgentCompatibleModel(provider, createModelItem('grok-4'))).toBe(true)
     expect(isAgentCompatibleModel(provider, createModelItem('qwen3.7-max'))).toBe(true)
   })
+
+  it('should use declared function-calling capability for customizable models', () => {
+    const provider = createModel('openai-api-compatible')
+
+    expect(
+      isAgentCompatibleModel(
+        provider,
+        createModelItem('qwen-max', {
+          fetch_from: ConfigurationMethodEnum.customizableModel,
+          features: [ModelFeatureEnum.toolCall],
+        }),
+      ),
+    ).toBe(true)
+    expect(
+      isAgentCompatibleModel(
+        provider,
+        createModelItem('custom-model', {
+          fetch_from: ConfigurationMethodEnum.customizableModel,
+          features: [ModelFeatureEnum.streamToolCall],
+        }),
+      ),
+    ).toBe(true)
+    expect(
+      isAgentCompatibleModel(
+        provider,
+        createModelItem('custom-model', {
+          fetch_from: ConfigurationMethodEnum.customizableModel,
+          features: undefined,
+        }),
+      ),
+    ).toBe(false)
+    expect(
+      isAgentCompatibleModel(
+        provider,
+        createModelItem('custom-model', {
+          fetch_from: ConfigurationMethodEnum.customizableModel,
+          features: [ModelFeatureEnum.vision],
+        }),
+      ),
+    ).toBe(false)
+  })
+
+  it('should keep predefined-model blacklist behavior unchanged', () => {
+    const provider = createModel('qwen')
+
+    expect(isAgentCompatibleModel(provider, createModelItem('qwen-max'))).toBe(false)
+    expect(
+      isAgentCompatibleModel(
+        provider,
+        createModelItem('qwen-max', {
+          fetch_from: ConfigurationMethodEnum.predefinedModel,
+          features: [ModelFeatureEnum.toolCall],
+        }),
+      ),
+    ).toBe(false)
+  })
 })
 
 describe('isAgentSuggestedModel', () => {
+  it('should not auto-suggest customizable models that become technically eligible', () => {
+    const provider = createModel('openai-api-compatible')
+
+    expect(
+      isAgentSuggestedModel(
+        provider,
+        createModelItem('qwen-max', {
+          fetch_from: ConfigurationMethodEnum.customizableModel,
+          features: [ModelFeatureEnum.toolCall],
+        }),
+      ),
+    ).toBe(false)
+  })
   it('should suggest configured Agent baseline models by English label', () => {
     const provider = createModel('any-provider')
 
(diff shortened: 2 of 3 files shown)


Bruce-Yii:
Hi @Souravrajvi0 — I opened #42092 and explicitly noted there that I wanted to implement the scoped change and asked to be assigned if the direction was acceptable. I noticed this PR now implements essentially that proposal and closes the issue.

Could you clarify why you chose to take over the implementation without coordinating on the issue first? I’m mainly trying to understand the contribution workflow here so I can avoid duplicate work in the future.

Souravrajvi0:
> Hi @Souravrajvi0 — I opened #42092 and explicitly noted there that I wanted to implement the scoped change and asked to be assigned if the direction was acceptable. I noticed this PR now implements essentially that proposal and closes the issue.
> 
> Could you clarify why you chose to take over the implementation without coordinating on the issue first? I’m mainly trying to understand the contribution workflow here so I can avoid duplicate work in the future.

Sorry I missed your comment on #42092 asking to be assigned  I should have coordinated with you before implementing. Your write-up was very clear, and this PR follows that design closely.

I'm happy for you to review, co-author, or take the PR over. My goal was just to unblock the bug. Thanks for defining the scoped fix! If you want me to close this pr I'll be happy to do it as well

Bruce-Yii (commented):
Thanks for the clarification — no need to close or transfer this PR. I reviewed the current diff against the policy boundary I proposed in #42092, and the main implementation is aligned: customizable models are gated by declared function/tool-calling capability, predefined models keep the curated blacklist behavior, and the compatibility tests cover tool-call / multi-tool-call / stream-tool-call plus unsupported cases.

One product invariant still looks under-tested / potentially incomplete: the PR description says technically eligible customizable models should not become auto-suggested, but `isAgentSuggestedModel()` itself is unchanged. The new test uses a customizable `qwen-max`, which only proves the invariant if that label actually matches one of `agentSuggestedModelPatterns`. Please add a regression case using a label that does match the suggested-pattern set (or explicitly gate customizable models in `isAgentSuggestedModel()` if the intended policy is always `false` for customizable models).

Once that invariant is nailed down, I’m happy with the overall direction. I’d prefer keeping this collaborative PR rather than reopening a competing implementation. Since this implementation follows the problem framing / scoped design in #42092, attribution in the PR (and co-author attribution if convenient, as you offered) would also be appreciated.

Bruce-Yii (commented):
Thanks for adding the follow-up test. I checked it against the actual `agentSuggestedModelPatterns`, and the invariant is still not exercised: `qwen-max` does **not** match the current suggested patterns. A concrete matching case is `qwen3.7-max` (matched by `/^qwen[ .-]?3\.7[ .-]max$/i`).

So with the current unchanged `isAgentSuggestedModel()`, a customizable `qwen3.7-max` would still be auto-suggested even though #42092's policy boundary was: customizable models may become technically compatible from declared tool-calling capability, but should not automatically enter the curated suggested set.

Please either:

1. add an explicit customizable-model guard in `isAgentSuggestedModel()` and cover it with a matching label such as `qwen3.7-max`, or
2. if maintainers intentionally want customizable models to participate in the curated suggestion patterns, update the PR/issue policy wording accordingly.

The compatibility implementation itself still looks aligned.

Souravrajvi0:
@Bruce-Yii Thanks for the detailed review — addressed in the latest push:

1. **isAgentSuggestedModel()** now returns false for customizable-model entries so labels that match agentSuggestedModelPatterns (e.g. qwen3.7-max) are not auto-suggested.
2. **Regression test** updated to cover customizable qwen3.7-max with declared tool-call, plus the existing qwen-max case.

Rebased onto current main. Co-authored-by on the commit per your note; happy to adjust attribution however you prefer.

PR #42190, opened 2026-09-11
refactor(models): pass session into App.app_model_config and App.workflow

JessYanCoding:
> [!IMPORTANT]
>
> 1. Associated issue: #40372 (claimed in [this comment](https://github.com/langgenius/dify/issues/40372#issuecomment-5630755563)).

## Summary

Part of #40372, follow-up to #41942 / #41964 / #41983 / #41998. Removes two of the three remaining `@property` wrappers on `App` whose whole body was `return self.<name>_with_session(session=db.session())`:

| Removed | Kept (unchanged) |
| --- | --- |
| `App.app_model_config` | `app_model_config_with_session(self, *, session: Session)` |
| `App.workflow` | `workflow_with_session(self, *, session: Session)` |

`App.desc_or_prompt` (`api/models/model.py:453`) is the third and is left for #40814, whose hunk it sits in. Separately `App.tenant` (`api/models/model.py:538`) still reaches `db.session` directly and has no twin yet, so #40372 stays open for `App`.

The earlier slices were pure deletions because nothing read them. These two are read in production, so this threads a caller session through the readers instead.

## Readers

A receiver-level sweep of `api/` — AST over every `.workflow` / `.app_model_config` attribute node, classifying each receiver — gives the first four rows. That kind of sweep structurally cannot see a name passed as a string, and a literal-name sweep over `"app_model_config"` / `"workflow"` in `getattr`/`hasattr`/`setattr`/`patch.object` surfaces exactly one more site, the fifth row. Every reader already had a session in hand:

| Reader | Change |
| --- | --- |
| `controllers/mcp/mcp.py` `_get_user_input_form` | takes `session: Session`; its only caller already runs inside the `with sessionmaker(...).begin() as session:` block at `mcp.py:85` and passes the same session to `_process_mcp_message` on the next line |
| `services/tools/workflow_tools_manage_service.py` ×3 | workflow lookup folded into the existing `with ... as _session:` block |
| `services/agent/dsl_service.py` | `self.session` |
| `services/app_service.py` `AppService.get_app` | the `ModifiedApp` override moves onto the session-aware accessor — see below |
| `tests/unit_tests/controllers/console/app/test_model_config_api.py` | the string-literal site: its poison helper becomes an absence assertion — see below |
| `tests/.../services/test_agent_service.py`, `tests/.../services/workflow/test_workflow_converter.py` | swap the accessor for its twin, using `db_session_with_containers` |
| `tests/.../services/tools/test_workflow_tools_manage_service.py` | needs `db_session_with_containers.commit()` — see below |
| `core/plugin/backwards_invocation/app.py` | docstring only: it named `App.workflow`'s request-scoped session property |

No response model can reach the raw names: `AppResponseView` (`app_service.py:199,206`) defines its own `app_model_config` / `workflow` properties forwarding to the twins, and `AppResponseModel._use_request_session` converts every `App` — `ModifiedApp` included — into that view before validation.

## Two latent problems this removes

**1. A cross-transaction read in `workflow_tools_manage_service.py`.** All three sites loaded the `App` inside a `with sessionmaker(..., expire_on_commit=False).begin() as _session:` block and then read `app.workflow` **after that block had closed**, so the workflow row was fetched on the global `db.session` in a different transaction from the `App` it was derived from. The lookup now sits inside the block. `expire_on_commit=False` keeps the returned `Workflow` usable afterwards, and the only attributes read after the block are `graph_dict` (a plain `json.loads` of the loaded `graph` column) and `version` — `Workflow` declares no relationships and no deferred columns, so nothing lazy-loads on the detached instance.

The behavioural consequence worth stating: the service now requires the published workflow to be **committed**. Both production callers already are — the console controller commits its own session before calling, and `import_service._ensure_workflow_app_is_published` commits inside its own `sessionmaker(db.engine).begin()` block first.

**2. An inert override in `AppService.get_app`.** It built a local `class ModifiedApp(App)` overriding `app_model_config` **as a `@property`** to return the tool-parameter-masked config. Nothing read that property — `AppResponseView.app_model_config` resolves through `app_model_config_with_session`, so the masking survived only because the same session's identity map returned the same mutated row. The override now sits on the session-aware accessor, which is the path responses actually take, and `@override` stays meaningful (`mypy` reports it as overriding nothing once the base property is gone). It deliberately ignores its `session` argument: re-reading the row there would hand back the **unmasked** row, so this is also stricter than before for any caller passing a different session.

## Three tests that depended on the old implicit behaviour

- `tests/unit_tests/controllers/console/app/test_model_config_api.py:30` poisoned the accessor with `monkeypatch.setattr(App, "app_model_config", property(fail))`. With the property gone, `monkeypatch.setattr` raises `AttributeError` and 3 tests error. The helper's own comment already records the resolution for the previous symbol — `App.is_agent` "no longer exists as an implicit property, so its absence enforces the same thing outright" — so the poison becomes `assert not hasattr(App, ...)` for both names, which is a stronger guard than a raising property.
- `tests/test_containers_integration_tests/services/tools/test_workflow_tools_manage_service.py:530` mutates `workflow.graph` **without committing** and relied on the service seeing it through the shared global session. With the lookup inside the service's own transaction that mutation is invisible and `WorkflowToolHumanInputNotSupportedError` never fires. Added the commit, matching `test_update_workflow_tool_human_input_node_error` in the same file, which already commits after the identical mutation.
- Two more cases in that file (`..._with_file_parameter_default`, `..._with_files_parameter_default`) mutate `workflow.graph` without committing too. They stay green either way — `create_workflow_tool` only feeds the graph to `ensure_no_human_input_nodes`, and `WorkflowToolProviderController.from_db` is mocked — but without a commit the FILE / Array[File] graph each case is named for would no longer reach the service at all, so they get the same commit.

## Tests

Neither twin had unit coverage. Five new cases:

`tests/unit_tests/models/test_app_models.py` (+4)
- `app_model_config_with_session` / `workflow_with_session` resolve the linked row **through a session the model never owns** (a second `Session` over the same bind), and the result is attached to that session.
- Both return `None` when the foreign key is unset, asserted against an **unbound** `Session`, so dropping the `if self.<fk>:` guard raises rather than silently passing.

`tests/unit_tests/services/test_app_service.py` (+1)
- `get_app` serves the masked config through `app_model_config_with_session`. The base accessor is stubbed with `side_effect=[masked, refetched]` so the assertion distinguishes the override from a fall-through.

Mutation-checked against the otherwise-fixed tree, each run as
`pytest tests/unit_tests/models/test_app_models.py tests/unit_tests/services/test_app_service.py` (93 passed unmutated, 88 with the five new cases reverted):

| Mutation | Result |
| --- | --- |
| drop the `if self.<fk>:` guard on both twins | 2 failed, 91 passed (both "returns None" cases) |
| point both lookups at the wrong column (`.app_id` instead of `.id`) | 3 failed, 90 passed (both "caller session" cases, plus an existing `deleted_tools` case that shares the predicate) |
| leave the `ModifiedApp` override on the raw property | 1 failed, 92 passed (`DID NOT` serve the masked config) |

## Verification

On `a3b2786`, from the repo root:

- `pytest api/tests/unit_tests` — **18782 passed, 6 skipped, 138 subtests passed**, 0 failed
- `pytest api/tests/unit_tests/models/ api/tests/unit_tests/services/test_app_service.py api/tests/unit_tests/controllers/mcp/ api/tests/unit_tests/controllers/console/app/` — **935 passed**
- `make lint` — `ruff format`, `ruff check --fix`, `lint-imports` (31 contracts kept, 0 broken), `dotenv-linter`, `dev/lint_response_contracts.py` (528 valid, 0 mismatch): clean
- `mypy --exclude-gitignore --check-untyped-defs --disable-error-code=import-untyped .` — no issues in 1796 source files
- `pyrefly check` — 0 diagnostics
- `api/uv.lock` unchanged

Container-integration suites: all three touched files run locally against Docker — `pytest tests/test_containers_integration_tests/services/tools/test_workflow_tools_manage_service.py tests/test_containers_integration_tests/services/test_agent_service.py tests/test_containers_integration_tests/services/workflow/test_workflow_converter.py` → **51 passed**. Dropping the added `commit()` from `test_create_workflow_tool_human_input_node_error` reproduces the failure this PR fixes: `Failed: DID NOT RAISE WorkflowToolHumanInputNotSupportedError`.

## Screenshots

N/A (backend-only refactor).

## Checklist

From Claude Code

### DIFF
--- a/api/controllers/mcp/mcp.py
+++ b/api/controllers/mcp/mcp.py
@@ -88,7 +88,7 @@ class MCPAppApi(Resource):
             self._validate_server_status(mcp_server)
 
             # Get user input form
-            user_input_form = self._get_user_input_form(app)
+            user_input_form = self._get_user_input_form(app, session)
 
             # Handle notification vs request differently
             return self._process_mcp_message(
@@ -180,17 +180,19 @@ class MCPAppApi(Resource):
 
         return helper.compact_generate_response(result.model_dump(by_alias=True, mode="json", exclude_none=True))
 
-    def _get_user_input_form(self, app: App) -> list[VariableEntity]:
+    def _get_user_input_form(self, app: App, session: Session) -> list[VariableEntity]:
         """Get and convert user input form"""
         # Get raw user input form based on app mode
         if app.mode in {AppMode.ADVANCED_CHAT, AppMode.WORKFLOW}:
-            if not app.workflow:
+            workflow = app.workflow_with_session(session=session)
+            if not workflow:
                 raise MCPRequestError(mcp_types.INVALID_REQUEST, "App is unavailable")
-            raw_user_input_form = app.workflow.user_input_form(to_old_structure=True)
+            raw_user_input_form = workflow.user_input_form(to_old_structure=True)
         else:
-            if not app.app_model_config:
+            app_model_config = app.app_model_config_with_session(session=session)
+            if not app_model_config:
                 raise MCPRequestError(mcp_types.INVALID_REQUEST, "App is unavailable")
-            features_dict = app.app_model_config.to_dict()
+            features_dict = app_model_config.to_dict()
             raw_user_input_form = features_dict.get("user_input_form", [])
 
         # Convert to VariableEntity objects
--- a/api/core/plugin/backwards_invocation/app.py
+++ b/api/core/plugin/backwards_invocation/app.py
@@ -281,7 +281,10 @@ class PluginAppBackwardsInvocation(BaseBackwardsInvocation):
     @classmethod
     def _get_workflow(cls, app: App) -> Workflow | None:
         """
-        get workflow without relying on App.workflow's request-scoped session property
+        get workflow with an owned session, additionally scoped by tenant and app
+
+        `App.workflow_with_session` resolves by `workflow_id` alone; a plugin
+        invocation must also confirm the workflow belongs to this tenant and app.
         """
         if not app.workflow_id:
             return None
--- a/api/models/model.py
+++ b/api/models/model.py
@@ -470,19 +470,11 @@ class App(Base):
     def site_with_session(self, *, session: Session) -> Site | None:
         return session.scalar(select(Site).where(Site.app_id == self.id))
 
-    @property
-    def app_model_config(self) -> AppModelConfig | None:
-        return self.app_model_config_with_session(session=db.session())
-
     def app_model_config_with_session(self, *, session: Session) -> AppModelConfig | None:
         if self.app_model_config_id:
             return session.scalar(select(AppModelConfig).where(AppModelConfig.id == self.app_model_config_id))
 
         return None
-
-    @property
-    def workflow(self) -> Workflow | None:
-        return self.workflow_with_session(session=db.session())
 
     def workflow_with_session(self, *, session: Session) -> Workflow | None:
         if self.workflow_id:
--- a/api/services/agent/dsl_service.py
+++ b/api/services/agent/dsl_service.py
@@ -203,7 +203,7 @@ class AgentDslService:
             package=package,
             package_path="agent",
         )
-        if app.app_model_config is None:
+        if app.app_model_config_with_session(session=self.session) is None:
             model_config = AppModelConfig(app_id=app.id, created_by=account.id, updated_by=account.id)
             self.session.add(model_config)
             self.session.flush()
--- a/api/services/app_service.py
+++ b/api/services/app_service.py
@@ -809,9 +809,12 @@ class AppService:
                 def __init__(self, app):
                     self.__dict__.update(app.__dict__)
 
-                @property
                 @override
-                def app_model_config(self):
+                def app_model_config_with_session(self, *, session: Session) -> AppModelConfig | None:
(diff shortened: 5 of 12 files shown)


JessYanCoding:
The workflows on this PR are still sitting at `action_required`, so no CI has run since it was opened on Sep 11 — there is nothing for a reviewer to look at yet. Could someone approve the workflow run?

For context, this is the `App.app_model_config` / `App.workflow` slice of #40372, claimed [here](https://github.com/langgenius/dify/issues/40372#issuecomment-5630755563). Still `MERGEABLE` against current `main` (31 commits later), and it adds no new `getattr` and no new `db.session()` call — it removes two.

Everything was run locally against `a3b2786`:

- `pytest api/tests/unit_tests` — 18782 passed, 6 skipped
- the three touched container-integration files — 51 passed, including `test_create_workflow_tool_human_input_node_error`, which is the case the added `commit()` fixes
- `make lint` and `make type-check` — clean

Happy to rebase, or to split the `app_service.py` / `test_model_config_api.py` parts out if a smaller diff is easier to take.

asukaminato0721 (approved):

PR #42192, opened 2026-09-11
fix: strip reasoning blocks before suggested-questions parse

### LINKED ISSUE #41854: suggested-question API request always return empty content

nanjingwuyanzu:
> [!IMPORTANT]
>

## Summary

Fixes #41854

I previously asked on the issue whether @however-yir still plans to finish the abandoned #41881, and requested assignment. Happy to adjust if a maintainer prefers a different owner.

### Root cause

`SuggestedQuestionsAfterAnswerOutputParser.parse` searches the raw completion for the first `[...]` pair. `_default_suggested_questions_model_parameters` already tries to disable thinking when the model schema allows it, but reasoning models can still emit `<think>` / `<thought>` blocks. Brackets inside those blocks (for example `[0.2, 0.8]`, `[MASK]`, or a string list) are matched first, so the parser either fails to decode or keeps a non-string list and the suggested-question API returns `[]`.

Unclosed tags (truncated completions) have the same failure mode because the first bracket pair is still taken from the reasoning text.

### Fix

Strip closed and unclosed `<think>` / `<thought>` blocks (case-insensitive) in the parser **before** the JSON-array extract. The helper in `api/core/workflow/nodes/agent/think_tags.py` is stream-oriented (it repairs tags for UI streaming), so this uses a small dedicated strip local to the parser.

Non-reasoning output is unchanged: a clean JSON array still extracts as before.

### Test plan

Unit tests in `api/tests/unit_tests/core/llm_generator/output_parser/test_suggested_questions_after_answer.py`:

- Clean JSON array still extracts the questions
- Closed `<think>` with brackets inside does not steal the match; the payload after the block is used
- Unclosed `<think>` with brackets inside is discarded so inner arrays are not returned

```bash
make test TARGET_TESTS=./api/tests/unit_tests/core/llm_generator/output_parser/test_suggested_questions_after_answer.py
```

Ran the targeted unit tests (3 passed), plus `ruff check` / `ruff format --check` and mypy on the touched files.

## Screenshots

N/A — parser behavior is covered by unit tests.

## Checklist

### DIFF
--- a/api/core/llm_generator/output_parser/suggested_questions_after_answer.py
+++ b/api/core/llm_generator/output_parser/suggested_questions_after_answer.py
@@ -7,6 +7,22 @@ from core.llm_generator.prompts import DEFAULT_SUGGESTED_QUESTIONS_AFTER_ANSWER_
 
 logger = logging.getLogger(__name__)
 
+_CLOSED_REASONING_BLOCK_RE = re.compile(
+    r"<(think|thought)>.*?</\1>",
+    re.DOTALL | re.IGNORECASE,
+)
+_UNCLOSED_REASONING_BLOCK_RE = re.compile(
+    r"<(think|thought)>.*",
+    re.DOTALL | re.IGNORECASE,
+)
+
+
+def _strip_reasoning_blocks(text: str) -> str:
+    """Remove closed and unclosed ``<think>`` / ``<thought>`` blocks."""
+    stripped = _CLOSED_REASONING_BLOCK_RE.sub("", text)
+    stripped = _UNCLOSED_REASONING_BLOCK_RE.sub("", stripped)
+    return stripped.strip()
+
 
 class SuggestedQuestionsAfterAnswerOutputParser:
     def __init__(self, instruction_prompt: str | None = None) -> None:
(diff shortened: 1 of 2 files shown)


vex-7-agent:
Fixes the reported shape. One more failure class the strip won't catch: any bracket pair that is not inside think tags, before the payload. E.g. a preamble with a citation `[1]` (json.loads gives a list of ints, filtered to empty), or a bare score vector. Same `[]` outcome, different source. Since extraction still takes the first `[...]` match, consider making it validate as well: iterate `re.finditer(r"\[.*?\]", stripped_text, re.DOTALL)` and accept the first candidate that json-decodes to a list of strings. Keep the strip too: with validation alone, a quoted list inside an unstripped think block would be wrongly accepted (your `["bad question from reasoning"]` test is exactly that case). A test with a citation before the payload would lock the newer path in.

PR #42162, opened 2026-09-10
fix(web): send model delete params as query string on DELETE

Souravrajvi0:
## Summary

Fixes #42133 (and relates to #41570 / #40773).

Custom model providers such as **OpenAI-API-compatible** could not be deleted. The UI sent `model`, `model_type`, and (for credential removal) `credential_id` in the **JSON body** of `DELETE` requests. Many reverse proxies and gateways strip DELETE bodies, so the API received `{}` and Pydantic validation failed with missing-field errors.

This change sends those fields as **query parameters** instead. The backend already supports query params on DELETE via `validate_request()`; this aligns the web client with that behavior.

## Changes

- `web/service/use-models.ts`: use `params` instead of `body` for `useDeleteModel` and `useDeleteModelCredential`
- `api/controllers/console/workspace/models.py`: document both DELETE routes as query-param endpoints in OpenAPI
- `api/tests/.../test_models.py`: add regression tests for query-string DELETE on model and credential routes

## Test plan

- [x] `uv run pytest tests/unit_tests/controllers/console/workspace/test_models.py` — 23 passed

## Notes

This complements the approach in #41862 but fixes both the model and credential DELETE paths from the frontend without requiring a breaking API signature change.

PR #42219, opened 2026-09-11
fix(web): send MCP provider delete params in query

### LINKED ISSUE #42218: MCP provider deletion still sends provider_id in a DELETE request body
### Self Checks

- [x] I have searched for existing issues and pull requests, including closed ones.

### Dify version

main

### Cloud or Self Hosted

Both

### Problem

The Web MCP provider delete hook still sends `provider_id` in a DELETE request body:

```ts
del('/workspaces/current/tool-provider/mcp', {
  body: { provider_id: id },
})
```

This leaves the browser path vulnerable to intermediaries that drop DELETE request bodies. Adjacent model credential DELETE paths were moved to query parameters in #42162 for the same transport reason.

### Runtime contract

Dify's shared `@model_validate` handling already reads DELETE query parameters first and falls back to a JSON body. The backend therefore already accepts:

```text
DELETE /console/api/workspaces/current/tool-provider/mcp?provider_id=<id>
```

without removing compatibility for older body-based callers.

### Expected behavior

`useDeleteMCP()` should send `provider_id` through request query params rather than the DELETE body.

### Proposed bounded scope

- change the Web MCP delete hook from `body` to `params`;
- add a focused service-hook regression test that asserts the request shape;
- leave backend/service behavior unchanged.

The endpoint's OpenAPI annotation still describes the legacy DELETE body shape. That contract cleanup can be handled separately with generated-contract regeneration; it is not required for the browser runtime fix because DELETE query parsing is already supported upstream.

I have the scoped change ready.

Bruce-Yii:
## Summary

Fixes #42218.

`useDeleteMCP()` still sent `provider_id` in a DELETE request body. Dify's shared DELETE request validation already accepts query parameters first and retains JSON-body fallback compatibility, so the Web client can use the safer query form without changing backend behavior.

### Changes

- send MCP provider `provider_id` through `params` instead of `body`;
- add focused `useDeleteMCP` service-hook coverage that asserts the DELETE request shape using a valid UUID provider ID.

This is the same transport fix class as the adjacent model credential DELETE changes in #42162, but deliberately keeps this PR scoped to the browser runtime path. The endpoint's OpenAPI body annotation can be cleaned up separately together with generated-contract regeneration.

## Validation

The regression test exercises the mutation and asserts:

```ts
del('/workspaces/current/tool-provider/mcp', {
  params: { provider_id: '123e4567-e89b-12d3-a456-426614174000' },
})
```

No backend/service semantics are changed, and older body-based callers remain supported by the existing DELETE body fallback.

From ChatGPT

### DIFF
--- a/web/service/__tests__/use-tools.spec.tsx
+++ b/web/service/__tests__/use-tools.spec.tsx
@@ -3,15 +3,16 @@ import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import { act, waitFor } from '@testing-library/react'
 import { beforeEach, describe, expect, it, vi } from 'vite-plus/test'
 import { renderHook } from '@/test/console/render'
-import { useAllToolProviders, useRefreshMCPServerCode } from '../use-tools'
+import { useAllToolProviders, useDeleteMCP, useRefreshMCPServerCode } from '../use-tools'
 
-const { mockGet, mockPost } = vi.hoisted(() => ({
+const { mockDel, mockGet, mockPost } = vi.hoisted(() => ({
+  mockDel: vi.fn(),
   mockGet: vi.fn(),
   mockPost: vi.fn(),
 }))
 
 vi.mock('@/service/base', () => ({
-  del: vi.fn(),
+  del: mockDel,
   get: mockGet,
   post: mockPost,
   put: vi.fn(),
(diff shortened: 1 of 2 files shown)


Bruce-Yii:
Closing this to avoid duplicate review work: #42333 now covers the same MCP provider DELETE query-param fix and is already routed for maintainer review.

Issue #42230, opened 2026-09-12
i met an error when i try to delete a model.


18903653765:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.17

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

i try to delete a model from openai-api-compatible and met an error

### ✔️ Expected Behavior

delete that model

### ❌ Actual Behavior

[...]

PR #42244, opened 2026-09-13
fix(web): send provider credential DELETE params as query string

Souravrajvi0:
## Summary

Completes the same DELETE/query-param pattern as #42162 for **provider** credentials.

**#42230** (cannot delete a custom model) is already fixed on `main` via #42162 (`useDeleteModel` sends `model` / `model_type` as query params). This PR does not re-land that change; it fixes the remaining sibling path that still sent `credential_id` in the DELETE body.

## Changes

- `useDeleteProviderCredential`: `body` → `params`
- OpenAPI: document query params on provider credential DELETE
- Unit test: delete via query string

## Test plan

- [x] `pytest api/tests/unit_tests/controllers/console/workspace/test_model_providers.py::TestModelProviderCredentialApi::test_delete_via_query_params`


api/controllers/console/workspace/model_providers.py
```diff
@@ -270,7 +270,7 @@ def put(self, args: ParserCredentialUpdate, current_tenant_id: str, provider: st
 
         return SimpleResultResponse(result="success").model_dump(mode="json")
 
-    @console_ns.expect(console_ns.models[ParserCredentialDelete.__name__])
+    @console_ns.doc(params=query_params_from_model(ParserCredentialDelete))
```
laipz8200, line 273: I don't think the semantics of except and doc are the same here.

laipz8200 (changes requested):

api/controllers/console/workspace/model_providers.py
Souravrajvi0: @laipz8200 Good catch — agreed that `expect` and `doc` are not interchangeable here.

`model_validate` for DELETE reads **query first** and falls back to JSON body when the query string is empty (`validate_request` in `wraps.py`), so the handler supports both paths. Swapping `expect` → `doc` only documented the query case and dropped the body schema.

Updated on `feat/fix-model-delete-42230-main-38da`:

- **Keep** `@console_ns.expect(ParserCredentialDelete)` for the JSON fallback
- **Add** `@console_ns.doc(params=query_params_from_model(ParserCredentialDelete))` for the primary query-string path (what the web client now sends, same as #42162 for model / model-credential delete)
- OpenAPI markdown lists `credential_id` as a query param and documents the optional JSON body fallback

`test_delete_via_query_params` still passes; branch is rebased on current `main`. Let me know if you would rather document query-only and drop the body fallback from the spec instead.

PR #42251, opened 2026-09-13
chore(api): remove redundant str() conversions in legacy model type migration

HUAHAODIA:
## Summary

This PR removes the 50 `unnecessary-type-conversion` diagnostics that pyrefly reports in `api/services/legacy_model_type_migration.py`. All of them are `str()` calls applied to values that are already typed as `str` (SQLAlchemy `Mapped[str]` primary key columns), so the calls are runtime no-ops and safe to drop.

This is a first chunk of #39947. The remaining ~300 instances are spread across other modules (`services/`, `controllers/`, `core/`); I'm happy to follow up with additional chunked PRs if this approach looks good.

Verification:
- `pyrefly check --min-severity info`: 0 `unnecessary-type-conversion` remaining in the touched file (only a pre-existing, environment-dependent `missing-import` noise remains)
- `ruff check` and `ruff format` pass on the touched file
- `python -m py_compile` passes

Ref #39947

## Screenshots

N/A (no UI change)

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. (No behavior change — mechanical removal of no-op `str()` calls; no new tests needed)
- [x] I've updated the documentation accordingly. (N/A)
- [x] I ran `make lint && make type-check` (backend) and `vp staged` (frontend) to appease the lint gods (ran `ruff check` / `ruff format` and `pyrefly check` on the touched file)

From ZCode

PR #42253, opened 2026-09-13
chore(api): remove redundant str() conversions across services

HUAHAODIA:
## Summary

Follow-up to #42251. This chunk clears the 69 `unnecessary-type-conversion` diagnostics that pyrefly reports under `api/services/**` (26 files): `str()` calls applied to values already typed as `str` (SQLAlchemy `Mapped[str]` columns and `str`-annotated fields). All are runtime no-ops, so removal is behavior-preserving.

Where removing `str()` left a trivial pass-through comprehension (e.g. `{x for x in xs}`), it is rewritten as `set(xs)` / `list(xs)` per ruff C416 (`unnecessary-comprehension`), which would otherwise fail CI after the str() removal.

Remaining instances after this and #42251: 235 across `controllers/`, `core/`, `factories/`, `repositories/`, etc. — happy to continue in follow-up chunks.

Verification:
- `pyrefly check --min-severity info`: 0 `unnecessary-type-conversion` remaining under `api/services/`
- `ruff check` / `ruff format`: pass on all touched files
- `python -m py_compile`: pass on all touched files

Ref #39947

## Screenshots

N/A (no UI change)

## Checklist

- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. (No behavior change — mechanical removal of no-op `str()` calls plus equivalent comprehension rewrites)
- [x] I've updated the documentation accordingly. (N/A)
- [x] I ran `make lint && make type-check` (backend) and `vp staged` (frontend) to appease the lint gods (ran `ruff check` / `ruff format` and `pyrefly check` on all touched files)

From ZCode


HUAHAODIA:
Addressed the review feedback in 35bcbd7d:

- **dataset_service.py**: restored `str(upload_file_id)` — `data_source_info` is persisted JSON (`dict[str, Any]`) and the integration test stores `{"upload_file_id": 99}`, so the string normalization at this boundary is load-bearing.
- **workflow_run_service.py**: restored `int(args.get("limit", 20))` — `args` comes from the request layer where limit may still be a string.
- **restore_archived_workflow_run.py**: removed the leftover no-op self-assignment.

For the other removals, values are guaranteed `str` at runtime: SQLAlchemy rows loaded through `StringUUID.process_result_value` always return `str(value)` (see `api/models/types.py:40-43`), and `dify_config` fields are pydantic-validated.

asukaminato0721 (changes requested):
ci

crazywoola:
Hi @HUAHAODIA, thanks for contributing to Dify.

### Why this is being closed

For issue #39947, no competing PR was opened by the issue author, so we are keeping the earliest submission, #39950, under the FIFO rule. This PR overlaps with that retained implementation, so we are closing it as a duplicate to keep review focused.

### Next steps

Please add any unique test cases, edge cases, or implementation notes to #39950. If part of this change is independent, please open a focused PR for that scope and link its own tracking issue in the description.

PR #42257, opened 2026-09-13
fix(api): run draft variable storage cleanup outside the DB transaction

### LINKED ISSUE #42256: Object storage cleanup runs inside the DB transaction in delete_draft_variables_batch
### Self Checks

### Dify version

1.17.0 (self-hosted from source, main branch @ dde1d500b5)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

This is a code-level issue in the app-deletion path. The default OPENDAL_SCHEME=fs
(local filesystem) is not affected — it needs a remote object storage backend.

1. Configure remote object storage instead of the default local filesystem
   (S3 / OSS / GCS, etc.).
2. Create an app, then run a workflow in the editor a few times with large node
   outputs so draft variables get offloaded — string > 100000 chars, array > 1000
   items (see WORKFLOW_VARIABLE_TRUNCATION_* in api/configs/feature/__init__.py:870-883).
   Each offloaded variable leaves one object in storage.
3. Delete the app and watch the app_deletion worker.

Code path:
- api/tasks/remove_app_and_related_data_task.py:516 opens the per-batch transaction
- :534 calls _delete_draft_variable_offload_data inside it
- :598-600 loops over the batch (up to batch_size=1000) calling storage.delete()

### ✔️ Expected Behavior

Object storage cleanup runs outside the database transaction. The transaction should
only cover the fast DB reads and writes for one batch.

### ❌ Actual Behavior

delete_draft_variables_batch opens a transaction per batch (line 516) and performs up
to 1000 storage.delete() calls inside it (lines 598-600). With a remote backend each
call is a network round-trip, so a batch can hold the transaction open for seconds
while holding row locks it doesn't need.

api/AGENTS.md says: "Do not perform external I/O inside an open transaction unless a
documented consistency contract requires it."

I don't think the current grouping is load-bearing: storage deletion failures are
already swallowed and the DB rows are deleted anyway (:603-606), and object storage
doesn't roll back with the DB transaction regardless.

If the single-transaction grouping is intentional, let me know and I'll drop this.

mesrop2008:
Fixes #42256

## Summary

`delete_draft_variables_batch` opens one transaction per batch and, inside it, called
`storage.delete()` once per offloaded draft variable — up to `batch_size` (1000) times.
With a remote object storage backend (S3/OSS/GCS) every call is a network round-trip, so
a single batch could hold the transaction open for seconds while holding row locks on
`workflow_draft_variables`, `workflow_draft_variable_files`, and `upload_files` that the
storage calls don't need. `api/AGENTS.md`: "Do not perform external I/O inside an open
transaction unless a documented consistency contract requires it."

The grouping wasn't load-bearing — storage failures were already swallowed and the DB rows
deleted anyway, and object storage doesn't roll back with the transaction regardless.

Changes in `api/tasks/remove_app_and_related_data_task.py`:

- `_delete_draft_variable_offload_data` is now DB-only. It deletes the `upload_files` and
  `workflow_draft_variable_files` rows and **returns the storage keys** instead of deleting
  the objects. On a DB error it returns `[]`, so we never delete objects whose rows may
  still exist.
- New `_delete_storage_objects(storage_keys) -> int` performs the deletions, swallowing
  per-object failures with the same log message as before.
- `delete_draft_variables_batch` collects the keys inside the transaction and calls
  `_delete_storage_objects` after the `with session.begin()` block exits — once per batch,
  with the row locks already released.

Semantics are preserved: `total_files_deleted` still counts successfully deleted objects,
storage failures remain non-fatal, and DB rows are deleted regardless. The only behavior
change is timing — a crash between commit and storage deletion now leaks an object rather
than leaving an orphan row, which is the better failure direction, since the rows are what
must go.

## Screenshots

N/A — backend-only change to a Celery task's transaction boundary, no user-facing output.

## Testing

Added a regression test that pins the actual fix, plus updates to the tests whose
expectations encoded the old in-transaction behavior.

`api/tests/unit_tests/tasks/test_remove_app_and_related_data_task.py`

- `test_delete_draft_variables_batch_deletes_storage_after_commit` (new) — mocks the session
  factory and storage, and asserts the recorded event order is `commit` →
  `storage-delete:<key>`. This fails on the old code, where the delete happened before commit.
- `test_delete_draft_variable_offload_data_does_not_touch_storage` (new) — the helper returns
  the keys and never calls `storage.delete()`.
- Updated the empty-input and DB-failure tests for the new `list[str]` return type.

`api/tests/test_containers_integration_tests/tasks/test_remove_app_and_related_data_task.py`

- `..._offload_data_success` now asserts the returned keys match the upload file keys and
  that storage was not touched, while still checking both row sets are gone.
- `..._offload_data_storage_failure` drives `_delete_storage_objects` after the transaction:
  one key fails, one succeeds, the failure is logged and the count is 1.
- `..._batch_logs_progress` stubs both helpers and asserts the keys returned by the offload
  cleanup are handed to `_delete_storage_objects`.

Run:

    cd api && uv run pytest tests/unit_tests/tasks/test_remove_app_and_related_data_task.py -q

13 passed. `ruff check`, `ruff format --check`, `mypy --check-untyped-defs`, and
`pyrefly check` are clean on the three changed files. The testcontainers suite was not run
locally (needs Docker + Postgres); it runs in CI.

## Checklist


crazywoola:
Hi @mesrop2008, thank you for contributing to Dify!

For issue #42256, we are prioritizing #42258, which was opened by the issue author. The issue author takes priority over submission order.

PR #42276, opened 2026-09-14
fix(web): cover IPv6 and the full IPv4 loopback range in isPrivateOrLocalAddress

### LINKED ISSUE #41865: isPrivateOrLocalAddress() never matches IPv6 addresses — URL.hostname returns "[::1]", not "::1"
### Self Checks

### Dify version

1.17.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

`isPrivateOrLocalAddress()` in `web/utils/urlValidation.ts` never detects IPv6 addresses. The
check on line 33 compares `hostname` against the bare string `'::1'`:

```ts
// web/utils/urlValidation.ts:33
if (hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '::1') return true
```

That comparison is unreachable. The WHATWG URL parser **keeps the square brackets** in
`URL.hostname` for IPv6 literals. This can be confirmed in any browser console, no Dify
checkout required:

```js
new URL('http://[::1]:8080/x').hostname
// → "[::1]"     — never equal to "::1", so the check can never fire
```

Behaviour of the current function, for each input:

| URL | `URL.hostname` | Returns | Expected |
|---|---|---|---|
| `http://127.0.0.1` | `127.0.0.1` | `true` | `true` |
| `http://192.168.1.1` | `192.168.1.1` | `true` | `true` |
| `http://[::1]:8080/x` | `[::1]` | **`false`** | `true` |
| `http://[fd00::1]/x` | `[fd00::1]` | **`false`** | `true` |
| `http://127.0.0.2` | `127.0.0.2` | **`false`** | `true` |
| `http://0.0.0.0` | `0.0.0.0` | **`false`** | `true` |

For maintainers, this can be reproduced as a unit test by appending to
`web/utils/urlValidation.spec.ts` — all four cases fail:

```ts
import { isPrivateOrLocalAddress } from './urlValidation'

describe('isPrivateOrLocalAddress', () => {
  it.each([
    ['http://[::1]:8080/x', true],   // IPv6 loopback
    ['http://[fd00::1]/x', true],    // IPv6 unique local
    ['http://127.0.0.2', true],      // 127.0.0.0/8 beyond .1
    ['http://0.0.0.0', true],        // unspecified address
  ])('flags %s as private/local', (url, expected) => {
    expect(isPrivateOrLocalAddress(url)).toBe(expected)
  })
})
```

The two call sites are:

- `web/app/components/workflow/nodes/trigger-webhook/panel.tsx:199` —
  the `debugUrlPrivateAddressWarning` hint on the Webhook Trigger node
- `web/app/components/plugins/plugin-detail-panel/subscription-list/create/hooks/use-common-modal-state.helpers.ts:171` —
  the `callbackUrl.privateAddressWarning` on plugin subscription callback URLs

Both receive a server-generated URL derived from the deployment's configured base URL, so this
surfaces on self-hosted instances reachable at an IPv6 literal (e.g. `http://[::1]:5001`) or at a
`127.0.0.0/8` address other than `127.0.0.1`: the warning that should appear is silently absent.

> Note: this is **not** an SSRF report. `isPrivateOrLocalAddress` only drives a client-side
> advisory label; the server-side SSRF protection is separate and unaffected.

### ✔️ Expected Behavior

`isPrivateOrLocalAddress()` returns `true` for IPv6 loopback (`[::1]`) and IPv6 unique-local
(`fd00::/8`) hosts, for the whole `127.0.0.0/8` range, and for `0.0.0.0` — so the "this looks like
a private or local address" warning is shown for those URLs, as it already is for `127.0.0.1`,
`10.0.0.0/8`, `172.16.0.0/12` and `192.168.0.0/16`.

### ❌ Actual Behavior

It returns `false` for all of them, and no warning is rendered. The `'::1'` comparison on line 33
is unreachable, because `URL.hostname` yields `'[::1]'` for IPv6 literals.

A minimal fix is to strip the surrounding brackets before the comparison and extend the loopback
check to the full `127.0.0.0/8` range plus `0.0.0.0`. I'm happy to open a PR for this.

Harsh23Kashyap:
## Summary

Closes a 1.17.0 client-side advisory regression in the webhook trigger panel and the plugin subscription callback modal. `isPrivateOrLocalAddress()` in `web/utils/urlValidation.ts` only checked the literal string `'::1'` for IPv6 loopback, but the WHATWG URL parser keeps the surrounding brackets in `URL.hostname` for IPv6 literals — so the IPv6 branch was unreachable. The IPv4 loopback check only matched `127.0.0.1`, not the full `127.0.0.0/8` range, and `0.0.0.0` (unspecified) was not handled.

Fix the hostname handling in three steps:
- Use `URL.hostname` (already separated from the port) and strip the surrounding brackets from IPv6 literals so the comparisons work against the bare address.
- Extend the IPv4 loopback check to the full `127.0.0.0/8` range and add `0.0.0.0`.
- Add private IPv6 ranges: `fc00::/7` (unique-local) and `fe80::/10` (link-local).

The two call sites (`web/app/components/workflow/nodes/trigger-webhook/panel.tsx:199` and `web/app/components/plugins/plugin-detail-panel/subscription-list/create/hooks/use-common-modal-state.helpers.ts:171`) now correctly render the private-address warning for self-hosted deployments at an IPv6 literal or at any `127.x.x.x` / `0.0.0.0` address.

This is a client-side advisory label only; the server-side SSRF protection is separate and unaffected, as the issue body explicitly notes.

Fixes #41865

## Why this is the right scope

The issue body is unambiguous about both the root cause and the desired behavior. The fix is contained to one helper function (`isPrivateOrLocalAddress`) plus its tests. The two callers receive the corrected boolean and now render the warning at the right time. There is no need to touch the callers or the related server-side SSRF check.

## Changes

`web/utils/urlValidation.ts::isPrivateOrLocalAddress`

- Read `urlObj.hostname` (not `urlObj.host`) so the IPv6 literal is not suffixed by `:port` when a port is present.
- Strip the surrounding brackets from IPv6 literals into a `bare` string.
- Add `0.0.0.0` and the full `127.0.0.0/8` IPv4 loopback check (`a === 127`).
- Add `fc00::/7` (matches both `fc` and `fd` prefixes) and `fe80::/10` (matches `fe8`, `fe9`, `fea`, `feb` prefixes) for private IPv6.
- Keep the existing 10/8, 172.16/12, 192.168/16, 169.254/16 IPv4 checks and the `.local` domain check unchanged.

## Tests

`web/utils/urlValidation.spec.ts`

A new `isPrivateOrLocalAddress` test block (within the existing `URL Validation` describe) exercises 18 URL cases via `it.each`:

- IPv4 loopback: `127.0.0.1` (existing), `127.0.0.2`, `127.255.255.254` (full `127.0.0.0/8` range).
- IPv4 unspecified: `0.0.0.0`.
- IPv4 private ranges: `10.0.0.1`, `172.16.0.1`, `172.31.255.254`, `192.168.0.1`, `169.254.0.1`.
- IPv6 literals: `http://[::1]:8080/x`, `http://[fd00::1]/x`, `http://[fc00::1]/x`, `http://[fe80::1]/x` (loopback, unique-local `fd00::/8` and `fc00::/7`, link-local `fe80::/10`).
- DNS: `localhost`, `myhost.local`, `example.com`.
- Public IPv4/IPv6: `8.8.8.8`, `[2001:db8::1]`.
- Malformed URLs do not throw and return `false`.

7 of these cases fail on the unpatched helper — exactly the cases listed in the issue body. Verified by stashing only `web/utils/urlValidation.ts` and re-running the new tests, then restoring. The web test runner (`vp test --project unit`) is the source of truth for the CI run; the regression was captured with a Node.js reimplementation of both the pre-fix and post-fix functions to verify the contract without requiring the full `web/` install.

## Verification

- Node.js reimplementation of the post-fix function: 18/18 cases match expected.
- Node.js reimplementation of the pre-fix function: 7/18 cases fail (the exact regressions described in the issue body).
- `web/utils/urlValidation.spec.ts` updated with the 18-case test block; CI `vp test --project unit` will run it.

## Risk

Very low. The change is internal to one helper function. The two callers (`debugUrlPrivateAddressWarning` and `callbackUrl.privateAddressWarning`) only render an advisory label; they do not gate behavior or security. The server-side SSRF protection is separate and unaffected.

A small audit of the server-side equivalent in `api/core/helper/ssrf_proxy.py` and the broader SSRF-check helper would harden the corresponding server-side surface (this issue is explicitly client-side only). Not changed in this PR because the issue body notes the scope is the client-side warning and the server side is already separate.


crazywoola:

### Why this is being closed

For issue #41865, we are prioritizing #41866 because it was opened by the issue author. The author-first rule takes precedence over submission time. This PR overlaps with that retained implementation, so we are closing it as a duplicate to keep review focused.

### Next steps

Please add any unique test cases, edge cases, or implementation notes to #41866. If part of this change is independent, please open a focused PR for that scope and link its own tracking issue in the description.

PR #42299, opened 2026-09-15
fix(agent): validate workspace skill references on publish

### LINKED ISSUE #42298: Workflow publish rejects valid Agent V2 workspace Skill references
## Summary

Publishing a workflow fails when an Agent V2 node's system prompt references a workspace Skill that is validly bound to the Agent's published snapshot.

## Steps to reproduce

1. Create and publish a workspace Skill, for example `pai-diagnosis`.
2. Bind it to an Agent V2 roster agent.
3. Reference it in the Agent system prompt, for example `[§skill:pai-diagnosis:Pai Diagnosis§]`.
4. Publish the Agent so the Skill binding is captured in `agent_skill_binding_snapshots`.
5. Add the Agent to a workflow and publish the workflow.

## Expected behavior

The workflow publishes successfully because the referenced workspace Skill is bound to the selected Agent snapshot and is available at runtime.

## Actual behavior

The publish request returns HTTP 400:

```json
{
  "code": "invalid_param",
  "message": "Workflow Agent node <node-id> has invalid Agent Soul config refs: skill_ref_dangling: skill 'Pai Diagnosis' is not configured.",
  "status": 400
}
```

## Root cause

`WorkflowAgentPublishService._require_config_asset_refs_resolved_for_publish` validates Skill mentions only against embedded `agent_soul.config_skills`. It does not include workspace Skills from `AgentSkillBindingSnapshot`, although runtime Skill loading supports those snapshot bindings.

This makes publish-time validation inconsistent with runtime behavior.

## Environment

Observed on Dify 1.17.1.

ZhXZhao:
## Summary

- include workspace Skills bound to the selected Agent snapshot when validating Agent V2 prompt references
- reuse runtime Skill resolution so publish validation matches execution behavior
- avoid loading workspace Skills when every Skill reference is already satisfied by embedded configuration
- add regression coverage for publishing a workflow with a snapshot-bound workspace Skill reference

Closes #42298

## Test plan

- [x] Run `TestWorkflowAgentDraftBindingSync` (25 tests)
- [x] Run the runtime Skill published-identity unit test
- [x] Run Ruff lint and formatting checks for changed files
- [x] Run Pyrefly and Mypy for changed service files


api/services/agent/workflow_publish_service.py
```diff
@@ -184,21 +184,42 @@ def _validate_binding_composer_config_for_publish(
...
         from services.agent.prompt_mentions import MentionKind, parse_prompt_mentions
+        from services.skill_management_service import SkillManagementService
 
+        mentions = parse_prompt_mentions(agent_soul.prompt.system_prompt)
         configured_skill_names = {item.name for item in agent_soul.config_skills if not item.is_missing}
+        has_unresolved_skill_ref = any(
+            mention.kind == MentionKind.SKILL and mention.ref_id not in configured_skill_names for mention in mentions
+        )
+        if has_unresolved_skill_ref and binding.agent_id is not None:
+            configured_skill_names.update(
```
fatelei, line 212: this configured_skill_names is overwrite in configured_file_names = {item.name for item in agent_soul.config_files if not item.is_missing}

ZhXZhao: Thanks for pointing this out. `configured_skill_names` and `configured_file_names` are separate local variables, so assigning `configured_file_names` does not overwrite `configured_skill_names`. The validation below also reads them independently: Skill references are checked against `configured_skill_names`, while file references are checked against `configured_file_names`.

To make this behavior explicit, I've updated the regression test to include both a snapshot-bound workspace Skill reference and a config file reference. The test passes with both references present.

fatelei: sorry， i misread the variable name

fatelei: configured_skill_names seem not used

fatelei (approved):

PR #42308, opened 2026-09-15
fix(vdb): narrow text_exists except clauses to NotFoundError

### LINKED ISSUE #40423: bare except in VDB text_exists() swallows KeyboardInterrupt/SystemExit

Harsh23Kashyap:
## Summary

Narrow the `except Exception:` clause in `text_exists()` for the OpenSearch and Lindorm VDB backends to `except NotFoundError:`, so non-404 errors (transport failures, auth errors, etc.) actually surface instead of being silently swallowed and reported as 'document not found'.

Fixes #40423

## Background

Both `OpenSearchVector.text_exists()` and `LindormVectorStore.text_exists()` previously wrapped every `Exception` from the OpenSearch client's `get()` and returned `False`:

```python
@override
def text_exists(self, id: str) -> bool:
    try:
        self._client.get(index=self._collection_name.lower(), id=id)
        return True
    except Exception:
        return False
```

This conflates 'document truly does not exist' with 'something else went wrong'. The only client call where `False` is the correct semantic answer is when `opensearchpy` raises `NotFoundError` for an HTTP 404. Anything else (network timeout, broken DNS, missing credentials, malformed request, …) should propagate so the caller sees the real problem.

## Change

Import `NotFoundError` from `opensearchpy.exceptions` and replace `except Exception:` with `except NotFoundError:` at:

- `api/providers/vdb/vdb-opensearch/src/dify_vdb_opensearch/opensearch_vector.py` (`text_exists`)
- `api/providers/vdb/vdb-lindorm/src/dify_vdb_lindorm/lindorm_vector.py` (`text_exists`)

`NotFoundError` is the documented exception type for missing documents in `opensearch-py==3.1.0` (`opensearchpy.exceptions.NotFoundError`), matching the call the issue suggested.

## Testing

- `pytest api/providers/vdb/vdb-opensearch/tests/unit_tests/test_opensearch_vector.py` → **16/16 passed** (1 new test)
- `pytest api/providers/vdb/vdb-lindorm/tests/unit_tests/test_lindorm_vector.py` → **14/14 passed** (1 new test)
- `ruff check` and `ruff format --check` clean on all four changed files.

The new regression tests use a fake `opensearchpy` module (matching the existing pattern in those test files) and assert:

1. `NotFoundError` → `text_exists()` returns `False` (the intended 'doc not found' branch).
2. Any other `Exception` → propagates to the caller (regression for the previously-swallowed case).

## Compatibility and risk

The only behavioral change is for callers that were relying on transport/auth failures to be silently treated as 'missing document'. That interpretation is unsound — those callers were already misreporting the situation to their own callers as soon as the OpenSearch cluster was briefly unreachable. Existing test coverage of the happy paths (`return True`) and the intended `False` branch (`NotFoundError`) is preserved.

No schema, API, or configuration changes. Pure exception-handling narrow.


crazywoola:

### Why this is being closed

For issue #40423, we are prioritizing #40424 because it was opened by the issue author. The author-first rule takes precedence over submission time. This PR overlaps with that retained implementation, so we are closing it as a duplicate to keep review focused.

### Next steps

Please add any unique test cases, edge cases, or implementation notes to #40424. If part of this change is independent, please open a focused PR for that scope and link its own tracking issue in the description.

PR #42314, opened 2026-09-15
refactor(files): dep-inject query params with @model_validate

ColinHouse:
## Summary

Part of #36659. Converts the last `Model.model_validate(request.args.to_dict(flat=True))` statement in `api/controllers/files/` to the `@model_validate` decorator, following the merged query-param slices #41562, #41646 and #42318.

| File | Handler | Model |
|---|---|---|
| `api/controllers/files/tool_files.py` | `ToolFileApi.get` | `ToolFileQuery` |

It is a GET handler whose parse is the first statement of the method, with no guard clause and no `try`/`except` above it, and it already reads `request.args.to_dict(flat=True)` — exactly the decorator's GET branch.

Not converted, on purpose:

- `files/plugin_file_upload.py` — `PluginUploadFileApi.post` reads the signed query string on a **POST**; the decorator's non-GET branch reads the JSON body, so it does not fit as-is.
- `files/appdeploy_files.py` — each of its three parses raises a tested error of its own from `except ValidationError` (`InvalidFileRequestError` / `FileGrantInvalidError`).

Claimed in https://github.com/langgenius/dify/issues/36659#issuecomment-5682144391.

### Scope reduced after #42318

This PR originally also converted `ImagePreviewApi.get` and `FilePreviewApi.get` in `files/upload_file_delivery.py`. #42318 landed the identical conversion for both while this was open, so merging `main` in takes that file and its test from `main` unchanged; only the `tool_files.py` site — which #42318 did not touch — remains here.

## Behaviour change

`timestamp`, `nonce` and `sign` are required on `ToolFileQuery`. A request missing one used to raise `pydantic.ValidationError` inside the handler, which the blueprint's `ValueError` handler turned into a 400 `invalid_param`; it now returns 422 from the decorator, as at every previously converted site. Nothing asserted the old status.

## Tests

The existing tests drove the handler through a fake `request` object; they now build the query model and pass it in, like the converted `service_api` tests do. One rejection test pins the 422 path and asserts the service is never reached.

## Verification

Re-run after merging `main` (`bfd7ac0`):

- `pytest tests/unit_tests/controllers/files` — **89 passed** vs **88** on `main` (+1, the rejection test)
- Mutation-checked: deleting the decorator fails 11 tests in that file
- `ruff check` / `ruff format --check` — clean (the now-unused `from flask import request` is dropped)
- `pyrefly check controllers/files/tool_files.py` — 0 diagnostics; `pyrefly --config tests/unit_tests/pyrefly.toml` on the test file — 0 diagnostics, same as `main`
- `mypy --check-untyped-defs --disable-error-code=import-untyped` — no issues
- `lint-imports` — 31 kept / 0 broken; `dev/lint_response_contracts.py --fail-on-mismatch` — 0 mismatch


asukaminato0721 (approved):

asukaminato0721 (changes requested):
conflict

ColinHouse:
Conflict resolved by merging `main` in (no rebase, no force-push).

#42318 landed the identical conversion for `ImagePreviewApi.get` and `FilePreviewApi.get` while this was open, so I took `main`'s version of `upload_file_delivery.py` and its test unchanged. What remains here is only the `tool_files.py` / `ToolFileQuery` site, which #42318 did not touch — 2 files, +32/−106.

Re-verified on the merge: `pytest tests/unit_tests/controllers/files` 89 passed vs 88 on `main`; deleting the decorator fails 11 tests; ruff, pyrefly (0 diagnostics), mypy, lint-imports and the response-contract lint are all clean.

asukaminato0721 (approved):

PR #41876, opened 2026-09-06
fix(api): compute annotation list has_more with the effective page size

weike-zhang:
Fixes #41875

## Summary

The annotation list and hit-history endpoints computed `has_more` as `len(items) == limit`, while the query is served through `paginate_query(..., max_per_page=100)` which caps the page size at 100. Two wrong behaviours, both user-visible:

- **Last page exactly full** → `has_more=true` (phantom extra fetch / "loading more" spinner that returns nothing).
- **`limit` above the 100 cap with remaining rows** → `has_more=false` (pagination silently stops, remaining rows never fetched), and the response echoes the uncapped `limit` instead of the 100 actually used.

This is the same bug class already fixed for the dataset / document / segment Service API lists in #41784 / #41776; the annotation endpoints were missed.

### Changes

- `api/controllers/service_api/app/annotation.py` — Service API `GET /apps/annotations`
- `api/controllers/console/app/annotation.py` — Console `GET /apps/{app_id}/annotations` and `GET /apps/{app_id}/annotations/{annotation_id}/hit-histories`

Each now caps the page size (`effective_limit = min(limit, 100)`) and computes `has_more = page * effective_limit < total`, mirroring the accepted #41784/#41776 fix.

### Tests

6 regression tests added (2 per endpoint family), covering both the exact-full-last-page case and the `limit > 100` case. Before the fix all 6 fail; after, they pass.

- `tests/unit_tests/controllers/service_api/app/test_annotation.py` (2 new)
- `tests/unit_tests/controllers/console/app/test_annotation_api.py` (4 new)

Verification (local, dify `api` on `main`):
- Red: `test_has_more_*` → 6 failed
- Green after fix: `pytest tests/unit_tests/controllers/service_api/app/test_annotation.py` → 40 passed; `pytest .../console/app/test_annotation_api.py::TestAnnotationListHasMore TestConsoleAnnotationRefBoundaries` → 8 passed
- `ruff check` and `ruff format --check` clean on all changed files

### Follow-up (not in this PR)

The same `len(...) == limit` pattern remains on the console **datasets** controllers (`controllers/console/datasets/datasets.py`, `datasets_document.py`, `external.py`) and `controllers/console/explore/trial.py`; those are a separate service path and can be handled as a follow-up.

## Screenshots

N/A

## Checklist

- [ ] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs)
[...]

PR #42323, opened 2026-09-15
fix(api): report has_more from the page the query served

### LINKED ISSUE #42322: Annotation hit-history pager never ends for limit=0: has_more is computed from a page size the query does not use
### Self Checks

- [x] I have searched for existing issues, and this has not been reported before
- [x] I am using the latest `main`

### Dify version

`main` at `a4d74f7c`

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

`GET /console/api/apps/{app_id}/annotations/{annotation_id}/hit-histories?limit=0`

That route reads both pagination parameters straight off the query string with no bounds:

```python
page = request.args.get("page", default=1, type=int)
limit = request.args.get("limit", default=20, type=int)
effective_limit = min(limit, 100)
```

`min(limit, 100)` has no lower bound, but `libs/pagination.paginate_query` floors both at 1 before it runs the query:

```python
page = max(1, page)
per_page = max(1, per_page)
```

So the query is served with a page size of 1 while the response reports `limit: 0`, and `has_more` is computed from the requested value:

```python
has_more=page * effective_limit < total
```

`page * 0 < total` is true for every page, so a client that walks pages until `has_more` is false never stops. Past the end it keeps receiving empty pages that still claim there is more. `limit=-1` behaves the same way.

Driving the real `paginate_query` against a seven-row table, with the handler's own arithmetic:

```
--- client asks limit=0, walks pages until has_more is false ---
  page=1  rows=1 ids=[1] has_more=True   (paginate_query.has_next=True)
  ...
  page=7  rows=1 ids=[7] has_more=True   (paginate_query.has_next=False)
  page=8  rows=0 ids=[]  has_more=True   (paginate_query.has_next=False)
  page=9  rows=0 ids=[]  has_more=True
  page=10 rows=0 ids=[]  has_more=True
   -> client never stopped after 10 pages

--- the same walk with limit=3 ---
  page=3  rows=1 ids=[7] has_more=False -> client stops here
```

`PaginatedResult.has_next` is right in every one of those rows; it is the handler's separate recomputation from unclamped inputs that is not.

`page` has the same gap on this route: `?page=0` is served as page 1 while the response echoes `page: 0`.

The annotation list routes (`AnnotationApi.get`, service API `AnnotationListApi.get`) carry the same recomputation, introduced together with it in #41876. Those are currently protected because their query models declare `limit: int = Field(default=20, ge=1)`, so only the hit-history route is reachable today.

### ✔️ Expected Behavior

`has_more` is false once the last page has been served, and `limit`/`page` in the response are the values the query actually used.

### ❌ Actual Behavior

`has_more` stays true forever for `limit=0` or `limit=-1`, and the response reports a page size that was not used.

L4XB:
## Summary

Fixes #42322

`libs/pagination.paginate_query` floors both `page` and `per_page` at 1 before it runs the query. The annotation handlers computed `has_more` from the requested values instead, and the hit-history route reads both straight off the query string with no bounds, so `?limit=0` is served one row while the response reports a page size of 0. `page * 0 < total` is then true for every page, including the pages past the end that return nothing, and a client walking until `has_more` is false never stops. `limit=-1` behaves the same way.

Driving the real `paginate_query` over a seven-row table with the handler's arithmetic, before this change:

```
limit=0:  page 1..7 return one row each, page 8, 9, 10 ... return none,
          has_more true on every one of them
limit=3:  page 3 returns the last row, has_more false, the walk ends
```

after:

```
limit=0:  page 7 is the last one and has_more goes false
limit=-1: same
limit=3:  unchanged
limit=200: capped at 100, unchanged
```

and on every page `has_more` now equals `PaginatedResult.has_next`, which was already correct throughout.

The clamp moved into `clamp_pagination`, right next to the query that applies it, so the two cannot drift apart again. `paginate_query` calls it, and so do the three handlers. They now report the effective page and size rather than the requested ones, which is also why `?page=0` no longer echoes `page: 0` next to page 1's rows.

The two annotation list routes carry the same recomputation, introduced alongside it in #41876. They are not reachable today because their query models declare `limit: int = Field(default=20, ge=1)`, but they are routed through the same helper so the arithmetic cannot diverge if that validation ever moves.

## Second pass: the same guess in five more endpoints

While checking whether the annotation routes were the only ones, five more list endpoints turned up computing `has_more` as `len(items) == limit`, which is the form #41876 replaced for annotations. Their query models carry no `ge=1` either, so both failure directions are reachable. Measured over a forty-row table through the real `paginate_query`:

```
limit=20, page 2 is exactly full and also last
  before: has_more=True    -> one round trip that comes back empty
  after : has_more=False

limit=0 (accepted by these models)
  before: served 1 row, has_more=False -> the caller stops at 1 of 40
  after : has_more=True, and the walk ends on the real last page
```

The second is the worse one: a full page reading as "more" costs a request, a clamped page reading as "done" silently truncates the result.

- `datasets_document.py` already holds the `PaginatedResult` it built, so it now reports that object's `has_next`, `per_page` and `page` instead of recomputing anything.
- `datasets.py`'s dataset list had merged two branches under one expression. The `ids` branch calls `get_datasets_by_ids`, which pages with `per_page=len(ids)` and therefore has no next page at all; the paged branch uses the served numbers.
- `trial.py` is the same `ids` shape and was comparing against a `limit` its query never applied.
- `datasets.py`'s query list and `external.py` go through `clamp_pagination` like the annotation endpoints.

## Screenshots

| Before | After |
| ------ | ----- |
| `GET .../hit-histories?limit=0` → `{"limit": 0, "has_more": true}` on every page, forever | `GET .../hit-histories?limit=0` → `{"limit": 1, "has_more": false}` once the last row has been served |

## Checklist

- [x] This change requires a documentation update, included: not needed, no documented behaviour changes

On the last box: I could not run `make lint && make type-check` here, because the backend test environment needs dependencies I do not have installed (`tests/unit_tests/conftest.py` imports the app, which fails on `No module named 'graphon'`). What I could run: `ruff check` and `ruff format --check` are clean on all five touched files, and `tests/unit_tests/libs/test_pagination.py` passes standalone at 23 tests, since that module imports nothing from the app.

The two new cells in `tests/unit_tests/controllers/console/app/test_annotation_api.py` follow the shape of `test_hit_history_uses_annotation_ref` directly above them and are for CI to confirm.

Four mutations over `clamp_pagination`, all killed: the `per_page` floor dropped, the `page` floor dropped, the cap ignored, and the floor applied before the cap instead of after.


fatelei:
style check failed

L4XB:
Fixed and pushed. The Pyrefly diff bot pinned it exactly:

```
+ERROR `<` is not supported between `int` and `None` [unsupported-operation]
+   --> controllers/console/datasets/external.py:178:22
```

`ExternalDatasetService.get_external_knowledge_apis` is annotated
`tuple[list[ExternalKnowledgeApis], int | None]`, but it returns
`PaginatedResult.total`, which is declared `int` and built as
`session.scalar(count_stmt) or 0`. `None` was never one of the values; the
annotation was simply wrong, and this PR is the first code to compare that total
against anything.

So I corrected the annotation rather than guarding a value the function cannot
produce. `ExternalKnowledgeApiListResponse.total` stays `int | None` — `int` is
assignable to it, and no other caller exists.

Reproduced locally with the repo's own pinned checker, on the two files, before
and after:

```
$ pyrefly 1.2.0 check controllers/console/datasets/external.py \
      services/external_knowledge_service.py

before: ERROR `<` is not supported between `int` and `None` [unsupported-operation]
after:  (gone)
```

The other errors those two files report locally are pre-existing and identical
on both sides; they are the enterprise modules that are not installed here.

fatelei (approved):

PR #42335, opened 2026-09-16
refactor(models): drop implicit-session getters on Conversation/Message inputs and model_config

keith991001:
## Summary

Part of #40372 (claim: https://github.com/langgenius/dify/issues/40372#issuecomment-5694038614). Removes the last three implicit-session getters on `Conversation` / `Message` in `api/models/model.py`:

- `Conversation.inputs` (getter)
- `Conversation.model_config`
- `Message.inputs` (getter)

**Why this is safe** — audit on the base commit:

- Every serializer path already goes through `ConversationResponseSource` / `MessageResponseSource`, which override `inputs` / `model_config` with the `*_with_session` twins (13 validation call sites checked, including the nested `first_message` chain; the retention export service builds records from raw rows and never touches these properties).
- The only imperative production readers were `core/memory/token_buffer_memory.py` (`conversation.model_config`) and `core/ops/ops_trace_manager.py` (`message_data.inputs`), plus the models' own `to_dict()`. All three now call `*_with_session(session=db.session())` at the call boundary — same behavior, explicit dependency.

**Design note** — the `inputs` setters stay: they perform File-payload normalization and are used via assignment throughout the codebase (and by model constructors). `inputs` is therefore now a **write-only property** (`property(fset=...)`), so any missed reader fails loudly with `AttributeError: property ... has no getter` instead of silently reaching for the global session. Added unit tests pinning the write-only contract for both models.

Test migration: 3 unit tests and 3 container-integration tests that read the getters directly now call `inputs_with_session(...)`; the setter tests are unchanged.

## Screenshots

| Before | After |
| ------ | ----- |
| N/A    | N/A   |

## Checklist

- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This does not apply to typos!)

From Claude Code


api/tests/unit_tests/models/test_model.py
```diff
@@ -119,7 +119,7 @@ def test_inputs_restore_external_remote_url_file_mappings(owner_cls: type[Conver
         }
     }
 
-    restored_file = owner.inputs["file"]
+    restored_file = owner.inputs_with_session(session=Mock())["file"]
```
asukaminato0721: sqlite

asukaminato0721 (changes requested):

api/tests/unit_tests/models/test_model.py
keith991001: Done in 0b6569d — switched to the shared `sqlite_session` fixture here, and applied the same change to the two analogous round-trip tests in `test_app_models.py`.

asukaminato0721 (approved):

PR #42349, opened 2026-09-16
fix(rag): keep the block boundaries when extracting an HTML file

### LINKED ISSUE #42348: HTML knowledge-base files lose every block boundary: get_text() welds the words together
### Self Checks

### Dify version

main (`api/core/rag/extractor/html_extractor.py`)

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Upload an ordinary `.html` file to a knowledge base — or call the extractor
directly:

```python
from core.rag.extractor.html_extractor import HtmlExtractor

# <h1>Quarterly Report</h1><p>Revenue rose.</p><p>Costs fell.</p>
# <ul><li>Item one</li><li>Item two</li></ul>
print(HtmlExtractor("report.html")._load_as_text())
```

### ✔️ Expected Behavior

One block per line, the way the page reads:

```
Quarterly Report
Revenue rose.
Costs fell.
Item one
Item two
```

### ❌ Actual Behavior

```
Quarterly ReportRevenue rose.Costs fell.Item oneItem two
```

`_load_as_text` returns `soup.get_text()`, which concatenates every text node
with nothing in between. A page's block structure lives in the markup, not in the
text, so every boundary disappears and the words on either side are welded into
one: `ReportRevenue`, `fell.Item`, `oneItem`.

That string is what gets chunked and embedded. `ReportRevenue` is not a word in
any vocabulary, so the chunk that holds the heading is not retrievable by the
heading, and a chunk boundary can now fall in the middle of what used to be two
separate blocks.

The existing test pins only that the characters survive —
`"".join(page_content.split()) == "TitleHello"` — which is true either way, so
nothing catches it today.

Note for whoever picks this up: the obvious fix is wrong. `get_text(" ")` or
`get_text("\n")` also inserts the separator **between inline elements**, so
`<p>Hello <b>world</b>! See <a href="#">this</a>.</p>` turns into
`Hello world ! See this .` The end of each block-level element has to be marked
instead.

I have a patch ready with tests for both halves and will open the PR against this
issue.

L4XB:
## Summary

Fixes #42348

`HtmlExtractor._load_as_text` returns `soup.get_text()`, which concatenates every
text node with nothing in between. A page's block structure lives in the markup,
not in the text, so uploading an ordinary `.html` file to a knowledge base welds
the blocks together:

| | text stored |
|---|---|
| now | `Quarterly ReportRevenue rose.Costs fell.Item oneItem two` |
| after | `Quarterly Report\nRevenue rose.\nCosts fell.\nItem one\nItem two` |

`ReportRevenue`, `fell.Item` and `oneItem` are in no vocabulary, so the chunk
holding the heading is not retrievable by the heading, and a chunk boundary can
land in the middle of what were two separate blocks.

**Passing a separator is the obvious fix and is wrong.** `get_text(" ")` or
`get_text("\n")` also puts one between inline elements, so

```html
<p>Hello <b>world</b>! See <a href="#">this</a>.</p>
```

comes out as `Hello world ! See this .` — a different kind of damage in the same
text. The end of each block-level element is marked instead, `<br>` becomes a
newline, and runs of three or more newlines collapse to one blank line.

`<script>` and `<style>` contents were already excluded (BeautifulSoup keeps them
out of `get_text()`) and still are.

## Screenshots

N/A — server-side extraction.

## Checklist

- [x] This change requires a documentation update, included: **no**
- [x] I understand that this PR may be closed in case there was no previous
      discussion or issues. (This doesn't apply to typos!) — #42348 describes the
      bug with a reproduction.
- [x] I've updated the documentation accordingly. — not applicable
- [x] I ran `dev/reformat` (backend) and `cd web && npx lint-staged` (frontend)
      to appease the lint gods — `ruff check` and `ruff format --check` are clean
      on both changed files; no frontend change.

## Tests

Three cases added to `api/tests/unit_tests/core/rag/extractor/test_html_extractor.py`:
blocks are separated, an inline sentence is untouched, and `<br>` becomes a line
break.

```
$ pytest api/tests/unit_tests/core/rag/extractor/test_html_extractor.py -q
5 passed

# same tests against the current _load_as_text
FAILED test_block_elements_do_not_run_together
FAILED test_a_line_break_element_becomes_a_line_break
2 failed, 3 passed
```

`test_inline_elements_keep_the_sentence_intact` passes on both sides — it is the
guard against the separator version. The two tests that were already there pass
unchanged; the existing assertion compares
`"".join(page_content.split())`, which holds either way.


L4XB:
Pushed 64a16424a4: a reviewer of the same helper on LearningCircuit/local-deep-research#6520 measured that replacing every `<br>` in place is quadratic — `replace_with` finds the node through `parent.index(node)`, a scan of its siblings. Here 50,000 `<br>` in one element did not finish within 10 s.

`_load_as_text` now walks the tree once and leaves it unchanged: the text of each string `get_text()` would use (script and style text stays out as before), a newline at every `<br>`, and a newline after each block-level element. The output is the same string as before; only the time changes.

Two tests added: `test_many_line_breaks_are_read_in_linear_time` (50,000 `<br>` under `pytest.mark.timeout(10)`; it times out on 94def474) and `test_script_and_style_text_stays_out` (passes on both, pinning the string filter). `tests/unit_tests/core/rag/extractor`: 243 passed; `ruff check`, `ruff format --check` clean; `pyrefly` 0 diagnostics on the changed file, as before.

PR #42354, opened 2026-09-16
fix(rag): read the separator a CSV was written with

### LINKED ISSUE #42353: A semicolon-separated .csv is indexed as one column
### Self Checks

- [X] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [X] I confirm that I am using English to submit this report, otherwise it will be closed.
- [X] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）

### Dify version

main (`38f9d85d5a`)

### Cloud or Self Hosted

Self Hosted (Source)

### Steps to reproduce

Upload a `.csv` file that is not comma separated to a knowledge base. This is the ordinary output of `Save as → CSV` in Excel on a machine whose locale uses a semicolon as the list separator — German, French, Spanish, Italian, and most of Europe — and it is also what many "export as CSV" buttons produce when the data contains commas.

`data.csv`:

```
name;region;units
widget;EU;12
gadget;US;7
```

`api/core/rag/extractor/csv_extractor.py` reads it with

```python
df = pd.read_csv(csvfile, on_bad_lines="skip", **self.csv_args)
```

and `self.csv_args` carries no `sep`, so pandas uses its default, a comma.

Reproduced directly against the extractor:

```python
>>> [d.page_content for d in CSVExtractor("data.csv", encoding="utf-8").extract()]
['name;region;units: widget;EU;12', 'name;region;units: gadget;US;7']
```

### ✔️ Expected Behavior

The columns of the file are the columns of the document:

```
['name: widget;region: EU;units: 12', 'name: gadget;region: US;units: 7']
```

### ❌ Actual Behavior

The file is read as having a single column. The whole row lands in it, separators included, and the one column is named `name;region;units`.

Nothing fails and nothing is logged — `on_bad_lines="skip"` is not reached, because a one-column parse is perfectly well formed. The chunk that reaches the index is the raw line with a column name glued to the front, so retrieval on any field of the sheet cannot match, and an answer built from it quotes the separators back to the user.

A tab separated export saved as `.csv` behaves the same way.

I have a fix and tests ready and will open a PR referencing this issue.

L4XB:
## Summary

Fixes #42353.

`CSVExtractor` reads every uploaded `.csv` with

```python
df = pd.read_csv(csvfile, on_bad_lines="skip", **self.csv_args)
```

and `self.csv_args` carries no `sep`, so pandas uses its default: a comma. A file named `.csv` is often not comma separated — Excel writes the **list separator of the machine's locale**, a semicolon across most of Europe, and a tab separated export is routinely saved as `.csv`.

Reading such a file with a comma does not fail, which is why nothing is logged: a one-column parse is well formed, so `on_bad_lines="skip"` never fires. The whole row lands in that one column, and the column is named after the whole header line:

```python
# before, for  name;region;units / widget;EU;12 / gadget;US;7
['name;region;units: widget;EU;12', 'name;region;units: gadget;US;7']

# after
['name: widget;region: EU;units: 12', 'name: gadget;region: US;units: 7']
```

### What changed

`detect_csv_delimiter(sample)` in `core/rag/extractor/helpers.py`, next to `detect_file_encodings`: among `,`, `;`, `\t` and `|`, take the one that gives **the same column count on every row** of a sample (the first 20 non-blank rows, capped at 64 KB). A separator the file was not written with either does not occur at all, or occurs by accident and then the rows do not line up — which is what keeps a comma inside a sentence, or a semicolon inside a quoted field, from being read as one. When no candidate is consistent the comma is kept, so an ambiguous, single-column or ragged file is read exactly as before. On a tie the comma wins, because it is tried first and a later candidate has to be strictly better.

`_read_from_file` reads the sample, seeks back, and passes the detected `sep` — only when the caller has set neither `sep` nor `delimiter`, so `csv_args` still wins.

Pandas' own sniffing (`sep=None, engine="python"`) is deliberately not used: it searches the whole candidate space and on a single-column file splits the header `Note` into `No` and `e`.

## Screenshots

| Before | After |
| ------ | ----- |
| `'name;region;units: widget;EU;12'` | `'name: widget;region: EU;units: 12'` |

## Checklist

- [x] This change requires a documentation update, included: [Dify Document](https://github.com/langgenius/dify-docs) — not needed, no user-facing option changes

### Tests

`api/tests/unit_tests/core/rag/extractor/test_csv_extractor.py` — a new `TestCSVSeparatorDetection` class: the same table written with each of the four separators must extract to the same documents, plus three guards — a single-column file whose values contain semicolons, a semicolon inside a quoted comma-separated field, and an explicit `sep` in `csv_args` that must not be overridden.

| | `main` | this branch |
|---|---|---|
| `test_csv_extractor.py` | 3 failed, 19 passed | **22 passed** |

```
# on main
FAILED TestCSVSeparatorDetection::test_reads_the_separator_the_file_was_written_with[;]
FAILED TestCSVSeparatorDetection::test_reads_the_separator_the_file_was_written_with[\t]
FAILED TestCSVSeparatorDetection::test_reads_the_separator_the_file_was_written_with[|]
```

The 19 that pass on both sides include the existing `test_extract_honors_explicit_csv_args`, which pins that a caller-supplied `sep` reaches pandas untouched.

`ruff check`, `ruff format --check` and `mypy --check-untyped-defs --disable-error-code=import-untyped` are clean on the touched files.


api/core/rag/extractor/helpers.py
```diff
@@ -1,10 +1,68 @@
...
+    holding the whole line.
+
+    Pandas' own sniffing (`sep=None, engine="python"`) is not used: it searches
+    the whole candidate space and on a single-column file splits the header
+    `Note` into `No` and `e`.
+
+    Args:
+        sample: The first rows of the file, as text.
+    """
+    sample = sample[:CSV_DELIMITER_SAMPLE_CHARS]
```
jamalkamaladdin, line 70: When fewer than 20 rows fit in the 64 KB sample, the last row is cut in the middle, its column count does not match the rows before it, `_consistent_column_count` returns 0 for the real separator and detection falls back to a comma. Measured on a semicolon file of 480 KB, 3 columns, 4 KB per cell: detection returns `,`. With the sample trimmed to the last newline it returns `;`. Rows above 3.2 KB are common in files that carry long text.

```suggestion
    sample = sample[:CSV_DELIMITER_SAMPLE_CHARS]
    if len(sample) == CSV_DELIMITER_SAMPLE_CHARS and "\n" in sample:
        # the read cut the last row in half, and a half row never lines up
        sample = sample[: sample.rfind("\n") + 1]
```

jamalkamaladdin (commented):
One case the sample size leaves open.

api/core/rag/extractor/helpers.py
L4XB: Reproduced it: 60 rows of three columns with 4 KB cells, and the sample stops in the ninth row. The PR head returned `,`, and your trim returns `;`.

The trim still returns `,` in one case, though: when the last line break in the sample sits inside a quoted multi-line cell. Then the trimmed sample ends mid-row too. For example, three columns with a 60-line quoted middle cell: the 64 KB cut lands inside the quotes, and the row up to the last `\n` has two columns instead of three.

So 4107bf8 fixes it where the count is taken. A sample of `CSV_DELIMITER_SAMPLE_CHARS` is treated as cut from a longer file, and `_consistent_column_count` leaves its last row out, whether the read stopped between two fields or inside a quoted one; `csv.reader` hands back a partial row in both cases. A shorter sample is the whole file, so its last row still counts.

Three tests in `TestCSVSeparatorDetection`, each asserting its own precondition:
- `test_long_rows_cut_by_the_detection_sample_keep_their_separator`: your case. It checks the sample really ends mid-row, then that all 60 rows come back split on `;`.
- `test_a_sample_that_ends_inside_a_quoted_line_break_keeps_its_separator`: checks the cut falls inside a quoted field (an odd number of `"` in the sample).
- `test_the_last_row_of_a_complete_file_still_counts`: guard; a short file's last row still rules a separator out.

The first two fail on the previous head, and all 25 in the file pass now. Dropping the `truncated` condition fails the guard; never dropping the row, or never setting `truncated`, fails the other two. ruff, mypy and pyrefly are clean on `helpers.py`.

PR #41645, opened 2026-09-02
fix(web): render a meaningful message in the OAuth callback when no opener (#39752)

Taranum01:
Fixes #39752

## What problem does this PR solve?

The MCP OAuth callback page (`/oauth-callback`) was an empty `<div />` that only did anything when the page was opened as a popup with `window.opener` available. When the user opened the OAuth URL in a new tab (no opener), the hook did nothing and the page sat at an empty white screen — the exact "blank page" the issue reports.

## What is changed and how it works?

- `web/hooks/use-oauth.ts`: `useOAuthCallback` now returns a state object with `hasOpener`, `finished`, `error`, and `errorDescription`. The state is computed from the URL params and the presence of `window.opener` at first render (no `useState`-in-`useEffect` pattern, so the state is available on the first paint). The popup path is unchanged: it still posts the OAuth result to the opener and calls `window.close()`. As a small bonus, the sibling `openOAuthPopup` callback type is now `OAuthCallbackMessage` instead of `any`, which lets us drop the `no-explicit-any` lint suppression.
- `web/app/oauth-callback/page.tsx`: renders the empty `<div />` in popup mode and a meaningful success / error message in no-opener mode. Uses the four new i18n keys via the bracket-notation selector pattern (e.g. `t(($) => $['callback.success'], { ns: 'oauth' })`) because the keys are flat.
- `web/i18n/*/oauth.json`: adds the four new keys to every supported locale. English values are used as fallbacks for the non-English locales; a localization pass can replace them later.

## How it was tested?

```
$ pnpm test hooks/__tests__/use-oauth.spec.ts app/oauth-callback/__tests__/page.spec.tsx
unit  hooks/__tests__/use-oauth.spec.ts  (6 tests)
unit  app/oauth-callback/__tests__/page.spec.tsx  (4 tests)
10 passed
```

- `hooks/__tests__/use-oauth.spec.ts` (6 tests): popup success / error / fallback targetOrigin, the no-opener case (new tab), provider-error propagation, and the initial-state-mirrors-opener contract.
- `app/oauth-callback/__tests__/page.spec.tsx` (4 tests): empty `<div />` in popup mode, success message in new-tab mode, error message with provider description, error message without provider description.
- `tsslint` and `vp staged` pre-commit check clean on the touched files.

PR #42366, opened 2026-09-16
fix(web): redirect to console home when OAuth callback has no opener

### LINKED ISSUE #39752: MCP OAuth callback displays a blank page
### Self Checks

- [x] This is only for a bug report. Questions should be posted in [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have filled in all required fields without changing the report structure.

### Dify version

1.15.0

### Cloud or Self Hosted

Cloud

### Steps to reproduce

1. In the console, add an MCP server `https://mcp.notion.com/mcp` that uses OAuth authorization.
2. Start the authorization flow and approve access at the OAuth provider.
3. Let the provider redirect the browser to `/console/api/mcp/oauth/callback` with valid `code` and `state` query parameters.
4. Observe the page shown after the callback completes.

The API callback stores the OAuth data and redirects to `${CONSOLE_WEB_URL}/oauth-callback`. The frontend route renders an empty `<div />` and only closes itself when `window.opener` is available.

### ✔️ Expected Behavior

After the MCP OAuth callback completes, the popup should close and notify the console that authorization finished.

### ❌ Actual Behavior

The MCP OAuth callback displays a blank page.

Harsh23Kashyap:
## Summary

The OAuth callback hook only handled the popup flow (when `window.opener` is set). When the OAuth provider redirects to `/oauth-callback` as a top-level navigation — the pattern used by the MCP OAuth flow at `/console/api/mcp/oauth/callback` — the browser ends up on the callback page with no opener, the hook does nothing, and the page renders the empty `<div />` from `web/app/oauth-callback/page.tsx`.

This PR adds a single fallback branch: when there is no opener but the URL carries `subscription_id` or `error`, redirect to the console home origin. Server-side state is already correct (the OAuth credentials have been stored by the callback endpoint before the browser is redirected to the console), so the user only needs to be sent back into the console.

Fixes #39752

## Background

Issue #39752 reproduces the blank-page behaviour by walking through:

1. Add an MCP server that uses OAuth (`https://mcp.notion.com/mcp`).
2. Approve access at the OAuth provider.
3. The provider redirects to `/console/api/mcp/oauth/callback`.
4. The server stores the OAuth credentials and redirects the browser to `${CONSOLE_WEB_URL}/oauth-callback`.
5. The console's `/oauth-callback` page calls `useOAuthCallback`, which checks `window.opener`. There is no opener because this was a top-level navigation, not a popup, so the hook returns without doing anything.
6. The user sees a blank `<div />` and the authorization flow appears stuck.

The existing popup branches are unchanged. The new branch fires only when the page was reached without a parent window **and** has one of the two query parameters that indicate an OAuth result.

## Change

`web/hooks/use-oauth.ts` — add an `else if` after the existing popup branch:

```ts
} else if (subscriptionId || error) {
  // Top-level navigation flow (e.g., MCP OAuth server-side redirect).
  // The OAuth credentials are already stored on the server side; there
  // is no parent window to notify. Redirect back to the console home so
  // the user can continue normally instead of staring at a blank page.
  window.location.replace(new URL('/', window.location.origin).toString())
}
```

Six added lines, no behavioural change to the existing popup branches.

## Testing

```ts
// web/hooks/__tests__/use-oauth.spec.ts (new)
```

Seven scenarios cover both flows:

- popup + subscription_id: postMessage success + close (unchanged)
- popup + error: postMessage error + close (unchanged)
- popup + no params: postMessage default + close (unchanged)
- top-level + subscription_id: replace(console home)
- top-level + error: replace(console home)
- top-level + no params: no action (no false-positive redirect)
- top-level + unknown params: no action (no false-positive redirect)

The expected behaviour was verified by running both the pre-fix and post-fix effect bodies through a Node.js reimplementation against all seven scenarios. The two no-opener-with-relevant-params scenarios correctly demonstrate the bug under pre-fix (no action, blank page) and the fix under post-fix (replace to console home). All seven scenarios match the expected post-fix behaviour.

`web/node_modules` is not installed in this workspace, so the existing `vp test --project unit` runner cannot execute the new spec directly. The Node.js reimplementation exercises the same effect-body logic.

## Compatibility and risk

- Popup flow unchanged: postMessage targets and payloads are byte-identical to the existing implementation; `window.close()` still fires when the opener path executes.
- The new branch only fires when there is no opener AND the URL carries a meaningful result parameter, so it cannot interfere with deliberately-opened popup windows (which always have an opener) or with manual navigation to the callback URL without parameters.
- The redirect target is the console origin, not an external URL, so there is no open-redirect surface.
- No backend or schema changes.


crazywoola:

### Why this is being closed

For issue #39752, no competing PR was opened by the issue author, so we are keeping the earliest submission, #41645, under the FIFO rule. This PR overlaps with that retained implementation, so we are closing it as a duplicate to keep review focused.

### Next steps

Please add any unique test cases, edge cases, or implementation notes to #41645. If part of this change is independent, please open a focused PR for that scope and link its own tracking issue in the description.

PR #42367, opened 2026-09-16
fix(api): keep retrieval file grants visible to later workflow nodes

### LINKED ISSUE #42277: [Bug] #context_files# is silently empty for published-app end users on 1.16.1 (works in Studio preview) — retrieved knowledge attachments never reach the LLM
### Self Checks

- [x] This is only for bug report, if you have a question, please head to Discussions.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭。

### Dify version

`1.16.1` (Self Hosted / Docker). Image built from the 1.16.1 release plus a local Weaviate UUID patch (unrelated area).

**PR #36175 (`fix(api): allow LLM nodes to access retrieved knowledge files`, merged 2026-05-14, shipped in v1.14.2) is present in this build.** I verified both of its artifacts are in the running container:

- `core/app/file_access/scope.py` — the `granted_upload_file_ids` "execution-local" scope fields
- `core/workflow/node_factory.py` — `_build_retriever_segment_access_checker`

### Summary

For a **published** Advanced Chat app invoked through the **service API** as an **end user**, the LLM node never receives retrieved knowledge-base image attachments — `#context_files#` is not merely empty, the key is **absent** from `workflow_node_executions.inputs`. The same workflow, same dataset, same model and same vision configuration, invoked through **Studio preview**, does receive them.

There is **no error and no log line** — `#context#` is populated normally, the run succeeds, and the answer is produced from the caption text. This is a *silent* failure.

### This is likely NOT issue #36174

#36174 (fixed by #36175) has the same trigger — published app + end user — but fails **loudly**:

```
Node llm failed with ABORT strategy: Invalid upload file
RuntimeError: Invalid upload file
```

We do **not** get that error. In our build the loader path passes the ownership check without raising, and returns an empty list instead. So the #36175 fix appears to be in place, but something upstream of it still refuses the attachment for end users.

### Steps to reproduce

1. Knowledge base with per-segment image attachments (`segment_attachment_bindings` + `upload_files` are fully populated; attachment preview URLs return `200 image/jpeg`).
2. Advanced Chat app: `Knowledge Retrieval` node → `LLM` node. LLM node: **Context** enabled and bound to the retrieval node's `result`; **Vision** enabled with the file variable set; model has the `vision` feature (`deepseek-flash` / `deepseek-v4-flash-vision-exp`).
3. Run it in **Studio preview** as the workspace account → LLM node input contains `#context_files#` with **5** files.
4. **Publish**, then invoke the same app through `POST /v1/chat-messages` with the same question → LLM node input has **no `#context_files#` key**, 0 files.

### Evidence

Same app, **draft graph and published graph are field-by-field identical** (52 nodes; same context selector; same vision config; same model; same dataset on both retrieval nodes), so the only variable is the invoking identity:

| run | invocation | `user_from` | `#context#` length | `#context_files#` |
|---|---|---|---|---|
| A | service API `/v1/chat-messages` | `END_USER` | 6243 chars | **key absent** |
| B | service API `/v1/chat-messages` | `END_USER` | 10323 chars | **key absent** |
| C | service API `/v1/chat-messages` | `END_USER` | 4516 chars | **key absent** |
| D | **Studio preview** | `ACCOUNT` | 4505 chars | **5 files** ✅ |

Query used (authoritative — the SSE `node_started` event carries a *pre-context-collection* snapshot and must not be used for this):

```sql
SELECT created_at,
       (inputs::jsonb ? '#context_files#')                          AS has_context_files,
       jsonb_array_length(inputs::jsonb -> '#context_files#')        AS n_files,
       length(inputs::jsonb ->> '#context#')                         AS context_len
FROM workflow_node_executions
WHERE workflow_run_id = '<run_id>' AND title = '<llm node title>';
```

The retrieved segment is not the problem — it has attachments and they resolve:

```
segment_attachment_bindings for that segment_id : 3
of which JOIN upload_files                     : 3
```

### Environment / configuration that I believe is correct

- KB: high_quality, `is_multimodal = false`, text embedding `bge-m3` (attachments are produced by our own ingestion pipeline and written into segment content as markdown image references, which Dify converts into `upload_files` + `segment_attachment_bindings`).
- LLM node: `context.enabled = true`, `context.variable_selector` = the retrieval node's `result`; `vision.enabled = true`, `vision.configs.variable_selector = ["sys","files"]`, `detail = high`; model advertises `vision` in its plugin model schema.
- I also tested with a **direct** Knowledge Retrieval node added next to the LLM node (not inside an iteration) and rebound the context to it — same result (0 files over API).

### What I have already ruled out (to save you time)

| Hypothesis | Result |
|---|---|
| Iteration `is_parallel = true` (thread isolation) | ❌ Turning parallel mode **off** changes nothing |
| Retrieval inside an iteration sub-graph | ❌ A direct, non-iteration retrieval node reproduces it |
| Flat list → wrong Segment class | ❌ `build_segment([{...}])` and `build_segment_with_type(ARRAY_OBJECT, [...])` both yield `ArrayObjectSegment`, so the `isinstance(..., ArrayObjectSegment)` check in `_build_retriever_segment_access_checker` can pass |
| Iteration output loses `metadata` | ❌ DB shows the flattened output items carry `metadata._source = "knowledge"` and `metadata.segment_id` |
| A downstream Code node | ❌ The context is bound to the retrieval/iteration output, not to the Code node |
| `copy_context` / extra thread pool in graphon | ❌ `copy_context` does not appear anywhere in the `graphon` package, and `ThreadPoolExecutor` appears only in the iteration node — so retrieval and the LLM node appear to execute in the same context |

### Reading of the code, and where I get stuck

`DifyRetrieverAttachmentLoader.load()` (`core/workflow/node_runtime.py`) has two guards:

```python
if not is_retriever_segment_access_granted(segment_id):
    return []
if self._segment_access_checker is not None and not self._segment_access_checker(segment_id):
    return []
```

`is_retriever_segment_access_granted` (`core/app/file_access/scope.py`) returns `True` when `scope is None` or `not scope.requires_user_ownership` — i.e. it only actually checks the set for `END_USER`.

Since run **D** (ACCOUNT) successfully loads 5 files through the **second** guard, the second guard is satisfied by this graph. That implies the **first** guard is what returns `False` for `END_USER`, i.e. `granted_retriever_segment_ids` is empty at read time, even though the retrieval path calls `grant_retriever_segment_access([str(segment.id)])` (`core/rag/retrieval/dataset_retrieval.py`, in `knowledge_retrieval()`, in the same block that sets `retriever_from="workflow"` — and the stored retrieval output does show `retriever_from: "workflow"`, so that block definitely executed).

I could not reconcile "grant is written" with "grant is not visible", given there is no thread/context boundary that I can find in `graphon`. If you can tell me what else could separate the grant write from the read (or what to log to confirm it), I am happy to run that instrumentation on my instance and report back.

### Expected Behavior

For a published app invoked by an end user, retrieved knowledge attachments should resolve for the LLM node — as they do for the workspace account in Studio preview — while tenant isolation and the "segment must be present in the selected retrieval context" constraint are preserved.

### Actual Behavior

`#context_files#` is absent for end-user invocations, with no error and no log output. Vision-capable models therefore never see the retrieved images; only the text (including our caption text) reaches them.

msdianprince-7:
Fixes #42277

## Summary

When a published app is called by an end user (service API / web app), the LLM node never gets `#context_files#`, even though the same workflow works in Studio preview. There's no error, the attachments are just dropped.

The issue author's analysis was right: the first guard in `DifyRetrieverAttachmentLoader.load()` is the one returning `[]`, because `granted_retriever_segment_ids` is empty by the time the LLM node reads it. The reason the grant disappears is in how the graph engine runs nodes, not in the retrieval code.

`grant_retriever_segment_access()` records the grant by calling `ContextVar.set()` with a new copy of the `FileAccessScope`. But graphon's worker enters the run's `execution_context` around every node, and `ExecutionContext` / `FlaskExecutionContext.enter()` restores the contextvars snapshot that was taken when the run started (`for var, val in context_vars.items(): var.set(val)`). So once the knowledge retrieval node finishes, the next node starts from the original scope with no grants. It happens on the same worker thread too, so it isn't a thread issue.

Accounts don't hit this because `is_retriever_segment_access_granted()` only checks the set when `user_from == END_USER`, which is why Studio preview works.

## Changes

- `FileAccessScope` now keeps its granted upload file IDs and retriever segment IDs in a small lock-protected container that the scope object references. `grant_*` add to that container instead of replacing the scope in the context var, so every copy of the context for the run sees the same grants.
- `granted_upload_file_ids` and `granted_retriever_segment_ids` are still there as read-only `frozenset` properties, so callers like `DatabaseFileAccessController` don't change.
- Grants are still per scope. Each request binds a new `FileAccessScope`, so nothing is shared between runs or users, and the loader's second check (the segment has to be in the LLM node's selected context) is unchanged.

The upload-file grants had the same problem. It didn't cause a visible bug because the attachment loader grants the upload files again inside the LLM node, but they now carry over between nodes as well.

## Tests

- `test_file_access_grants_survive_execution_context_restore_between_nodes` captures the context the same way the workflow entry does, grants inside one "node", then checks the grant from a later "node" on the same worker thread and on a different one. It fails on `main` and passes with this change.
- `test_file_access_grants_do_not_leak_between_scopes` checks that a grant made under one scope isn't visible under a new scope for the same user.
- Existing file access, retrieval, node runtime and app generator tests still pass.


crazywoola:
Hi @msdianprince-7, thanks for contributing to Dify.

### Why this is being closed

For issue #42277, no competing PR was opened by the issue author, so we are keeping the earliest submission, #42317, under the FIFO rule. This PR overlaps with that retained implementation, so we are closing it as a duplicate to keep review focused.

### Next steps

Please add any unique test cases, edge cases, or implementation notes to #42317. If part of this change is independent, please open a focused PR for that scope and link its own tracking issue in the description.

PR #42331, opened 2026-09-16
fix(api): resolve 500 on MCP provider endpoints when provider_id is a server identifier

mcjjin:
## Summary

Fixes a 500 error on the MCP provider console endpoints (`GET /workspaces/current/tool-provider/mcp/tools/<id>`, `GET .../mcp/update/<id>`, `DELETE .../mcp`) when the path parameter is a server identifier instead of a UUID.

## Root cause

- `ToolTransformService.mcp_provider_to_user_provider()` returns the **server identifier** as the provider `id` for the list API (`api/services/tools/tools_transform_service.py`: `response["id"] = db_provider.server_identifier if not for_list else db_provider.id`).
- The frontend forwards that `id` in the `provider_id` path parameter of the detail/update/delete endpoints.
- `MCPToolManageService.get_provider()` only compared `provider_id` against the **UUID primary key**, so PostgreSQL raised `invalid input syntax for type uuid` → HTTP 500.

Server log:

```
psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type uuid: "monday_mcp"
SQL: SELECT ... FROM tool_mcp_providers
WHERE tool_mcp_providers.tenant_id = %(tenant_id_1)s::UUID AND tool_mcp_providers.id = %(id_1)s::UUID
GET /console/api/workspaces/current/tool-provider/mcp/tools/monday_mcp 500
```

## Fix

In `get_provider()`, when `provider_id` is not a valid UUID string, fall back to a `server_identifier` lookup. UUID inputs keep resolving via the primary key, and explicit `server_identifier=` calls are unchanged.

## Tests

Added a regression test (`test_get_mcp_provider_by_identifier_in_provider_id_slot`) covering:
- a non-UUID `provider_id` resolves via `server_identifier`
- a UUID `provider_id` still resolves via the primary key

Fixes #41512
Fixes #41649

Issue #41512, opened 2026-08-31
[Bug] MCP provider detail endpoint returns 500 "invalid input syntax for type uuid" — server_identifier passed where a provider UUID is expected


xiongyidxx:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.17.0

### Cloud or Self Hosted

Self Hosted (Kubernetes)

### Steps to reproduce

1. Register an MCP tool provider (Tools → MCP → add a streamable-http server). In `tool_mcp_providers` the row gets a UUID `id` plus a human-readable `server_identifier` (e.g. `my-mcp-tools`).
2. Open the provider's detail page in the console (the view that lists its tools). The front-end calls:
   ```
   GET /console/api/workspaces/current/tool-provider/mcp/tools/my-mcp-tools
   ```
   i.e. the **server_identifier** is passed in the path (`web/service/use-tools.ts` → `useMCPTools(providerID)`).
3. The API returns HTTP 500.

Root cause chain (verified against the `1.17.0` tag and today's `main`):

- `api/controllers/console/workspace/tool_providers.py:1589` (`ToolMCPDetailApi.get`) calls `service.get_provider(provider_id=provider_id, tenant_id=...)` — the path segment is treated as the UUID `id`; `server_identifier` is never passed.
- `api/services/tools/mcp_tools_manage_service.py::get_provider` therefore builds `WHERE tool_mcp_providers.id = 'my-mcp-tools'::UUID`, and PostgreSQL rejects it.
- The identifier confusion originates in `api/services/tools/tools_transform_service.py:277`:
  ```python
  response["id"] = db_provider.server_identifier if not for_list else db_provider.id
  ```
  so non-list responses (create/update) expose the **server_identifier** as the entity `id`, which the front-end then round-trips into subsequent detail calls.
[...]

PR #42395, opened 2026-09-17
refactor: reduce ambiguity in mcp queries

wylswz:
> [!IMPORTANT]
>

## Summary

Related #41512 #41360 #42331

### Inconsistent IDs

[Here](https://github.com/langgenius/dify/blob/e18c171f74c1eb07d15ac4746b435a91145dffed/api/services/tools/tools_transform_service.py#L269), the value of `id` depends on the caller-supplied `for_list` parameter. There is no guarantee that callers choose the value that matches their intended use. When these IDs are subsequently consumed by DSLs, this ambiguity can lead to inconsistent references.

### Ambiguous provider lookup

`MCPToolManageService.get_provider` combines two lookup behaviors, selected by the `by_server_id` parameter, rather than exposing each through a dedicated method. This makes the caller's intent less explicit: omitting the parameter or passing the wrong value can silently select the wrong lookup behavior and return an unexpected result.

## Screenshots

| Before | After |
| ------ | ----- |
| ...    | ...   |

## Checklist

- [ ] I've verified the change and added or updated tests where meaningful regression risk justifies coverage.


web/features/agent-v2/agent-detail/configure/components/orchestrate/prompt-editor/slash.tsx
hjlarry: maybe we should also use `getProviderReference` here instead of `provider.id` to make sure the prompt correct?

hjlarry: same as above

wylswz: fixed

hjlarry (approved):
LGTM

PR #42402, opened 2026-09-17
fix(api): resolve Agent App MCP tool inputs from the Agent Soul

### LINKED ISSUE #42387: Agent App MCP tool schema ignores Agent Soul app variables (required ones fail every tools/call)
Found while working on #42380 / #42383; this one is a different code path and is not fixed by that PR.

### What happens

An Agent App can declare app variables on its Agent Soul (`AgentSoulConfig.app_variables`). When such an app is registered as an MCP server, `tools/list` never advertises them: the tool schema only has `query`. That part is observable today, independently of #42380.

The runtime does enforce those variables — `AgentAppGenerator` validates `inputs` against the Soul projection — so a `required` variable makes the call fail, and a schema-driven client has no way to discover the missing parameter:

```json
{"jsonrpc": "2.0", "id": 1, "error": {"code": -32602, "message": "branch_id is required in input form"}}
```

On current `main` that error is masked by the call-path bug in #42380: `tools/call` fails earlier with "Agent App only supports streaming mode" (fixed by #42383). The schema gap is not masked — `tools/list` answers today, just without the declared variables.

### Steps to reproduce

1. Create an Agent App whose Soul declares `branch_id` (`text-input`, required). Today this is done through the console API or DSL import; the composer has no app-variables module yet.
2. Register the app as an MCP server and call `tools/list` — the schema lists only `query`.
3. Once the call path is fixed (#42383), call the tool without `branch_id` — the client gets the `INVALID_PARAMS` error above.

A client that already knows the variable name can still pass it, because `prepare_tool_arguments` copies unknown arguments into `inputs`. The failure is that the schema never reveals it.

### Where it comes from

`controllers/mcp/mcp.py::_get_user_input_form` resolves the form for every non-workflow mode from the legacy row:

```python
features_dict = app.app_model_config.to_dict()
raw_user_input_form = features_dict.get("user_input_form", [])
```

An Agent App's legacy row is created empty (`services/app_service.py`, the `AppMode.AGENT` branch): its model, prompt and variables live in the bound Agent Soul. The runtime projects the Soul instead — `AgentAppConfigManager` fills `user_input_form` from `agent_app_variables_to_user_input_form(agent_soul.app_variables)` — so the variable really is required by the time the tool call runs.

The rest of the platform already resolves this through the app-definition layer:

| face | resolver | Soul variables visible |
| ---- | -------- | ---------------------- |
| service API `GET /parameters` | `app_definitions.get_public_parameters` -> `_get_public_agent_parameter_config` (published Soul) | yes |
| webapp parameters (`controllers/web/app.py`) | same | yes |
| OpenAPI `get app` / `input_schema` | `controllers/openapi/_input_schema.py::resolve_app_config` (legacy row) | not reachable: `SupportedAppType` excludes the roster-owned `agent` type |
| MCP `tools/list` | `controllers/mcp/mcp.py::_get_user_input_form` (legacy row) | **no** |

### Executed evidence

Probing the real helpers (`MCPAppApi._get_user_input_form`, `agent_app_variables_to_user_input_form`, `build_parameter_schema`) and calling `AgentAppGenerator._prepare_user_inputs` directly, since a live `tools/call` on current `main` fails earlier for the reason above. Two DB-backed lookups are stubbed (the feature-row lookup, and the annotation-reply lookup inside `AppModelConfig.to_dict`):

```
1) what the Agent Soul declares, projected by the platform owner
    [{"text-input": {"label": "branch_id", "variable": "branch_id", "required": true}}]
2) what the MCP controller reads for the same Agent App
   raw form: []
3) the tool schema that reaches the MCP client
    {"type": "object", "properties": {"query": {"type": "string", "description": "User Input/Question content"}}, "required": ["query"]}
4) the schema the client would get with the projected form
    {"type": "object", "properties": {"query": {...}, "branch_id": {"description": "", "type": "string"}}, "required": ["query", "branch_id"]}
5) what the app requires when the tool call omits the variable
   -> ValueError: branch_id is required in input form
```

### Expected behavior

`tools/list` advertises the Soul variables with their types and required flags, matching what the runtime enforces.

### Fix direction

For `AppMode.AGENT`, `_get_user_input_form` could resolve the form through the app-definition owner (`AppDefinitionQueryService.get_public_parameters`, which returns `user_input_form` from the published Soul) instead of reading the legacy row. Two things to keep in mind:

- that reader raises `AppDefinitionNotPublishedError` / `AppDefinitionUnavailableError` when the Agent has no active snapshot; MCP should surface that rather than answering with an empty form (note the MCP route renders `MCPRequestError` as an opaque HTTP 500 today — pre-existing for every validation error in that controller);
- `_convert_user_input_form` already handles the form keys the projection emits (`text-input`, `paragraph`, `number`, `checkbox`).

### Open question

For Agent Apps today, app variables are validated and passed to input moderation, but they do not reach the agent: `AgentAppRuntimeRequestBuilder` sends `agent_soul_prompt` / `user_prompt` and no variable layer, and the Agent App runner does not render a prompt template with them. So fixing the schema makes the advertised contract match the enforced one, but does not by itself let callers influence the run. That part belongs to #41779 (`tool_input_bindings`); worth deciding whether the schema fix should wait for it or land first as a bug fix.

### Related

#41779 asks for application-controlled values bound to MCP tool parameters — a larger feature (per-tool bindings plus a composer UI for declaring variables). This issue is narrower: the MCP tool schema is built from the wrong source, so the declared variables never reach the client.

gold-beyond:
> [!IMPORTANT]
>

## Summary

Fixes #42387.

An Agent App keeps model, prompt and app variables on the Agent Soul; the legacy `app_model_config` row is created empty for that mode. `controllers/mcp/mcp.py::_get_user_input_form` read only the legacy row for every non-workflow mode, so `tools/list` advertised just `query` while the runtime validated `inputs` against the Soul projection — a required variable therefore failed the call with an error the client could not act on.

What the change does:

- for `AppMode.AGENT`, resolves the raw input form through the app-definition service (`get_public_parameters`), the same reader the service API and the webapp use, so the tool schema matches what `AgentAppGenerator` enforces (names, types, required flags);
- maps the reader's `AppDefinitionNotPublishedError` / `AppDefinitionUnavailableError` to the controller's existing "App is not published" / "App is unavailable" errors instead of answering with an empty form;
- leaves every other mode on its existing source;
- imports the app-definition container lazily, because it also loads the human-input and file services and their native file-type detection, which this transport does not use.

Relation to #42383: that PR fixes the `tools/call` path for Agent Apps (streaming flag + answer extraction) and is a separate, open change with no file overlap. The gap here is observable on its own through `tools/list`; the `branch_id is required in input form` failure becomes the client-visible one once the call path works.

## Evidence

Probes on `main` (`1203aa72`) and on this branch:

| probe | main | branch |
| ----- | ---- | ------ |
| MCP form for an Agent App declaring `branch_id` | `[]` (legacy row) | `[{"text-input": {"label": "branch_id", "variable": "branch_id", "required": true}}]` |
| `tools/list` `inputSchema` | `{"properties": {"query": …}, "required": ["query"]}` | `branch_id` present and required, `legacy_var` from a non-empty legacy row not merged in |
| Agent with no published version | schema answered as before | `App is not published` |

Tests: `api/tests/unit_tests/controllers/mcp/test_mcp.py` — 38 passed. Reverting the source changes makes 5 of them fail (both parametrized cases of `test_tools_list_advertises_soul_variables`, the not-published and unavailable cases, and the non-agent guard), so the new cases pin the behavior they guard. `ruff format --check`, `ruff check`, `pyrefly` (with `dev/pyrefly-check-local` flags), `api/dev/lint_response_contracts.py` (528 valid, 0 mismatch) and `mypy --check-untyped-defs` on the changed module are clean.

## Screenshots

Backend-only change, no UI involved. The before/after table above is the behavioral evidence.

## Known behavior change

For an Agent App whose Agent has no published version, the input form now resolves to an error instead of an empty form, and `_get_user_input_form` runs before method dispatch — so `initialize` and `tools/list` on such a server now fail rather than answering. Such a server can never complete a `tools/call` anyway. Note the route renders `MCPRequestError` as an opaque HTTP 500 because no handler is registered for it — pre-existing for every validation error in this controller (`Server Not Found`, `App is unavailable`, `Invalid user_input_form`, …), not introduced here.

## Checklist


gold-beyond:
Superseded by #42393, which reached the same fix first. Closing this to avoid duplicate work.

If the two extra regression tests here are useful — a non-empty legacy `app_model_config` row must not be merged into an Agent App's schema, and the published/unavailable cases surfaced separately — they are on the branch `gold-beyond:fix/agent-app-mcp-schema`.

PR #42419, opened 2026-09-17
fix: download preview images instead of opening a new tab

CodingOnStar:
The image preview Download button opened a new tab instead of saving an image. It passed the original URL with target=_blank, and images without a title received no download attribute. Cross-origin URLs could also ignore download semantics.

Fetch the image bytes and use the existing Blob download helper with a nonempty filename. Keep the original image bytes and format; derive a default filename from its media type when no title is available. Fetch failures show the existing localized download error and leave the preview open without navigating.

Validation: 25 focused preview tests pass, including real download-helper anchor behavior, untitled same-origin images, remote and data URLs, and network/HTTP failures. Targeted Vite+ formatting, lint, and type checks passed; commit hooks ran normally. Full build was not repeated.


lin-snow (dismissed):
LGTM

lin-snow (approved):
LGTM

PR #42476, opened 2026-09-18
fix(web): preserve i18next instance across I18nClientProvider rerenders

### LINKED ISSUE #42443: [Refactor/Chore] Preserve the i18next instance across provider rerenders
### Self Checks

### Description

Preserve the i18next instance for the lifetime of a mounted `I18nClientProvider`.

`web/app/components/provider/i18n.tsx` calls `createI18nextInstance(locale, resource)` directly in its render body. The factory in `web/i18n-config/client.ts` creates and initializes a new instance, including `initReactI18next`, each time.

A component-level diagnostic using the actual source and fully populated en-US/zh-Hans resource objects:

1. Invoke the provider with `locale="en-US"` and capture the instance on its returned `I18nextProvider` element.
2. Await `instance.changeLanguage('zh-Hans')`.
3. Invoke the provider again with identical props.
4. Compare the returned instances and translations.

Observed result: `sameInstance=false`, old instance language `zh-Hans`, new instance language `en-US`; the translated marker reverts to English.

This diagnostic directly invokes the component function; it is **not a mounted React rerender test or an end-to-end reproduction of a particular navigation path**. It demonstrates that a render constructs a new instance and loses the previous instance's language state. A mounted regression test and a real triggering path should be established during implementation. Share/embedded chat consumers change language on the existing instance, so its lifetime matters.

### Motivation

Unrelated provider rerenders should not recreate the translation context, reinitialize react-i18next, or discard client language state.

Acceptance criteria:

- A mounted same-props rerender retains the instance and a runtime language change.
- Intentional locale/resource prop changes have defined, tested behavior.
- Consumer subscriptions continue working without repeated initialization.
- Keep SSR request isolation; do not replace per-provider ownership with a process-global singleton.
- Verify a representative share/embedded-chat language override through a parent rerender.

### Additional Context

Observed on a local Dify 1.17.1 development branch based on `06e473015a9a1abb71da88ac0a0e92d96a992008`, with local DevTools integration patches; Vinext 1.0.0-beta.10, Vite+ core 0.3.2, Node 24.21.0. Recheck against current main before implementation. Build measurements are diagnostic baselines, not controlled production network benchmarks.

### Consolidated follow-up from #42462

Track provider lifetime and initialization readiness here rather than in two overlapping issues. Continue after the runtime-boundary work in #42471.

Additional verification scope:

- Establish first-render readiness for a non-English locale with `partialBundledLanguages: true` and English fallback. The factory currently returns without awaiting `init()`; dependency-level testing demonstrated asynchronous fallback loading, but a visible hydration or flicker regression has not been established.
- Use a real i18next instance to verify initial translations, missing-key English fallback, language switching, and locale/resource updates. Preserve fallback behavior rather than disabling it to make initialization appear synchronous.
- Verify SSR/hydration and concurrent server-language isolation in the supported runtime; mocked React cache tests do not prove request isolation.
- Preserve share/embedded-app language overrides and the browser HMR instance contract.

Keep directory migration and route-level namespace loading outside this issue. Establish runtime evidence before deciding which initialization changes are necessary.

Souravrajvi0:
Fixes #42443

## Summary

`I18nClientProvider` created a new i18next instance on every render, which dropped runtime `changeLanguage` state (notably in share/embedded chat flows).

This change keeps one instance per mounted provider via `useRef`, and only reacts when `locale` or `resource` props intentionally change.

## Test plan

- [x] vitest run app/components/provider/__tests__/i18n.spec.tsx


hyoban:
It seems there were some issues with the original issue, so this PR will be temporarily closed.

PR #42512, opened 2026-09-18
chore: improve accessibility on the datasets list page

iamjoel:
## Summary

- Make dataset entries, actions, and statistics accessible by keyboard while preserving permission-based navigation.
- Announce settled empty search results and give the tag management dialog an accessible title and responsive width.
- Keep dataset actions visible without obscuring unavailable labels.

Fixes SNP-636

Linear: https://linear.app/dify/issue/SNP-636

## Validation

- 150 unit tests and 2 Chromium regression tests passed.
- Full repository static checks passed with existing warnings; scoped checks and commit hooks passed.
- Full authenticated-page and screen-reader testing remain pending.

## Checklist

From Codex


lyzno1 (dismissed):
Reviewed all 13 changed files and the relevant card, permission, overlay, and testing contracts at 824284c9cf6d2423d739a5959892b4ecdd64d652. No blocking findings.

The browser regression test is worth retaining: it exercises real production OperationsDropdown and CornerLabels components with CSS and native Tab navigation. I verified that restoring the previous hover-only implementation makes both cases fail, and independently reverting the unavailable-state offset makes the label-overlap assertion fail. This is additional evidence beyond happy-dom. The fixture proves these component-level contracts, not complete dataset-page integration or focus transfer into the real rename dialog.

Validation: 159 focused unit tests and 2 Chromium tests passed locally; the browser tests passed again after restoring the reviewed head. The PR diff passes whitespace checks, and the current head CI is green. Full authenticated-page and screen-reader behavior was not independently verified in this review.

lyzno1 (approved):

PR #41391, opened 2026-08-27
chore: remove redundant server_default from Skill model definitions

QianJinGuo:
## Summary

Remove 6 redundant `server_default` entries from `api/models/skill.py` that duplicate their corresponding application-level `default=` values:

- `Skill.icon` - `default="📄"` already set, removed `server_default="📄"`
- `Skill.description` - `default=""` already set, removed `server_default=""`
- `Skill.name_manually_edited` - `default=False` already set, removed `server_default=sa.false()`
- `Skill.visibility` - `default="workspace"` already set, removed `server_default="workspace"`
- `SkillVersion.version_name` - `default=""` already set, removed `server_default=""`
- `SkillVersion.publish_note` - `default=""` already set, removed `server_default=""`

The `server_default` parameter only affects DDL generation (the `DEFAULT` clause in `CREATE TABLE`), not runtime insert behavior. When an equivalent `default=` is already present, keeping `server_default` is redundant and causes confusion for new model authors who reference existing definitions as examples.

This is a scoped slice of #29314, following the same pattern as #39886.

Fixes #29314

## Screenshots

N/A — behavior-preserving model definition cleanup, no UI changes.

## Checklist

- [x] This change requires a documentation update, included: N/A
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. N/A — behavior-preserving cleanup, existing tests cover model schema validity
- [x] I've updated the documentation accordingly. N/A
- [x] I ran `make lint && make type-check` (backend) and `vp staged` (frontend) to appease the lint gods

## Verification

- `uv run ruff check models/skill.py` — All checks passed
- `uv run ruff format --check models/skill.py` — 1 file already formatted
- `uv run pytest tests/unit_tests/test_sqlite_fixtures.py -v` — 7 passed
- `uv run pytest tests/unit_tests/migrations/ -v` — 22 passed

Issue #29314, opened 2025-12-09
[Chore/Refactor] Remove `server_default` in model definitions


QuantumGhost:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactoring, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

## Summary

PR #28188 introduced MySQL support and adjusted migration accordingly. However, the model definition is not adjusted, and most models still contains the `server_default=sa.text('uuid_generate_v4()')` option. While the `server_default` argument is only used to generate migration and not applied while running application, it still causes confusion, especially for new model definitions, as implementors of new models (both real humans and AIs) generally references existing model definitions as examples.

## Solution

Remove the `server_default` parameter from UUID column definitions in model files, allowing the application layer to handle database-specific UUID generation by setting the `default=uuid.uuid4` option for those models.

AkhmetovOlzhass:
Hi! I'd like to take this issue

tysoncung:
I'm interested in working on this issue. Could you provide more context about what you're looking for? Any additional details about requirements or constraints would be helpful.

AkhmetovOlzhass:
I searched through all SQLAlchemy ORM models and confirmed that
all UUID primary keys are generated at the application level
(uuid4 / uuidv7).

The only occurrences of uuid_generate_v4() are in historical Alembic
migration files, not in runtime models.

No changes are required. This issue appears to be already resolved

majiayu000:
I ran a no-op verification on current `upstream/main`:\n\n- `rg -n "uuid_generate_v4\(\)" api/models -g '*.py'` => no matches\n- `rg -n "uuid_generate_v4\(\)" api -g '*.py'` => matches are only under `api/migrations/**` (historical migrations)\n\nSo there is no remaining runtime model definition using `server_default=sa.text('uuid_generate_v4()')` to clean up. This issue appears already resolved in code state.

[...]

PR #42521, opened 2026-09-18
chore(models): remove redundant server_default from provider, tools and workflow models

CodeTrainerMan:
## Summary

Part of #29314

Third slice of the redundant `server_default` cleanup, following #42519 (`trigger.py`) and #42520 (`oauth.py`, `source.py`, `onboarding.py`, `comment.py`). As discussed in the issue, `server_default` only affects migration generation and is never applied while the application runs, so keeping both it and an equivalent application-level `default=` on the same column is confusing, especially since new models are usually written by copying existing ones.

| Model | Column | Removed `server_default` | Kept application-level `default` |
| --- | --- | --- | --- |
| `Provider` | `provider_type` | `text("'custom'")` | `ProviderType.CUSTOM` |
| `Provider` | `is_valid` | `text("false")` | `False` |
| `ProviderModel` | `is_valid` | `text("false")` | `False` |
| `ProviderModelSetting` | `enabled` | `text("true")` | `True` |
| `ProviderModelSetting` | `load_balancing_enabled` | `text("false")` | `False` |
| `LoadBalancingModelConfig` | `enabled` | `text("true")` | `True` |
| `ProviderCredential` | `visibility` | `sa.text("'all_team_members'")` | `PermissionEnum.ALL_TEAM` |
| `BuiltinToolProvider` | `is_default` | `sa.false()` | `False` |
| `BuiltinToolProvider` | `credential_type` | `sa.text("'api-key'")` | `CredentialType.API_KEY` |
| `BuiltinToolProvider` | `expires_at` | `sa.text("-1")` | `-1` |
| `BuiltinToolProvider` | `visibility` | `sa.text("'all_team_members'")` | `PermissionEnum.ALL_TEAM` |
| `MCPToolProvider` | `timeout` | `sa.text("30")` | `30.0` |
| `MCPToolProvider` | `sse_read_timeout` | `sa.text("300")` | `300.0` |
| `MCPToolProvider` | `identity_mode` | `sa.text("'off'")` | `"off"` |
| `Workflow` | `kind` | `sa.text("'standard'")` | `WorkflowKind.STANDARD` |
| `Workflow` | `marked_name` | `""` | `""` |
| `Workflow` | `marked_comment` | `""` | `""` |

Two columns were deliberately left alone because they are *not* an equivalent duplicate: `Provider.quota_type` has `server_default=text("''")` but `default=None`, and `WorkflowToolProvider.privacy_policy` has `server_default=""` but `default=None`. Removing those would change what a fresh `CREATE TABLE` produces (empty string vs. NULL), which is a semantic change rather than a cleanup. Happy to follow up on them separately if maintainers prefer otherwise.

The `created_at` / `updated_at` `server_default=func.current_timestamp()` entries are untouched, per the decision in the issue: they have no application-level default, so removing them would change insert behavior.

Scope: these three files only. Other model files are untouched to avoid overlapping with the in-flight PRs #39886 (`agent.py`, `credential_permission.py`) and #41391 (`skill.py`).

## Verification

1. **AST diff (no other semantic change).** Parsed all `mapped_column()` calls in the three files before and after: 17 columns changed, and every one of them changed by *nothing but* the removal of `server_default=`. Column counts are unchanged (89 / 115 / 152), and the per-file `server_default` counts drop exactly as expected: `provider.py` 28 -> 21, `tools.py` 31 -> 24, `workflow.py` 24 -> 21.
2. **Rendered DDL diff.** Rendered `CREATE TABLE` with the PostgreSQL dialect for all 8 touched tables, before vs. after, and diffed line by line. The only differences are the disappearance of the `DEFAULT` clause on exactly those 17 columns; every column type and `NOT NULL` marker is byte-identical. Example:
   ```diff
   -  is_valid BOOLEAN DEFAULT false NOT NULL,
   +  is_valid BOOLEAN NOT NULL,
   ```
3. **Runtime default values unchanged.** Constructed an instance of each affected model (the models are `MappedAsDataclass`, so the application-level `default` is applied in `__init__`) and read the touched columns back, before vs. after. All values are identical, e.g. `provider_type=custom`, `is_valid=False`, `enabled=True`, `load_balancing_enabled=False`, `visibility=all_team_members`, `credential_type=api-key`, `expires_at=-1`, `timeout=30.0`, `sse_read_timeout=300.0`, `identity_mode=off`. This is the direct evidence that the application-level `default` - not `server_default` - is what populates these columns at runtime. (`Workflow` is a plain declarative model rather than a dataclass mapping, so its column defaults are applied by SQLAlchemy at flush time, equally independent of `server_default`.)
4. **No raw SQL depends on the removed DDL defaults.** Searched the whole `api/` tree for `INSERT INTO` against the 8 affected tables. The only non-migration hits are the `legacy_model_type_migration` tests, and they all list `is_valid` / `enabled` / `load_balancing_enabled` explicitly. One historical migration (`de95f5c77138`, the serpapi one) inserts into `tool_builtin_providers` without those columns, but a migration runs against a schema built by the migration chain itself, not by the model definitions, so its behavior is unaffected.
5. **Lint.** `ruff check` and `ruff format --diff` clean on all three files (ruff 0.16.6, the version pinned in `api/pyproject.toml`); `python -m py_compile` passes. `make type-check` (pyrefly) was not run locally because the optional vdb dependencies cannot be built in this Windows environment; CI covers it.

## Screenshots

| Before | After |
| ------ | ----- |
| N/A - backend only, no UI change | N/A - backend only, no UI change |

## Checklist

- [x] I ran `make lint; make type-check` (backend) and `vp staged` (frontend) to appease the lint gods


CodeTrainerMan:
Regarding the failing `Run Web Full-Stack E2E / Chromium Full (1/3)` check: I looked into the job log and I believe it is unrelated to this change.

What actually failed: 18 of 19 scenarios passed. The single failure is `Disable and restore a published workflow Web App` (`e2e/features/apps/web-app-service.feature:4`), at this step:

```
Then the published workflow Web App should be unavailable   # e2e/features/step-definitions/apps/web-app-service.steps.ts:86
  Error: expect(locator).toBeVisible() failed
  Locator: getByRole('heading', { name: '404' })
  Expected: visible
  Timeout: 15000ms
  Error: element(s) not found
```

Why it cannot be caused by this PR:

1. `server_default` is only used when SQLAlchemy emits DDL, i.e. when a migration is generated. This PR touches no migration, and the E2E environment builds its schema through the alembic migration chain (docker compose with `docker-db_postgres`, `alembic==1.17.2`), so the tables in that environment are unchanged.
2. The columns keep their application-level `default=`, which is what actually populates them at runtime. I verified this directly by constructing each affected model and reading the columns back before vs. after - values are identical (`provider_type=custom`, `is_valid=False`, `enabled=True`, `load_balancing_enabled=False`, `visibility=all_team_members`, `credential_type=api-key`, `expires_at=-1`, `timeout=30.0`, `sse_read_timeout=300.0`, `identity_mode=off`).
3. All backend checks pass: API Unit Tests (all shards), API Integration Tests (all shards), DB Migration Test, Pyrefly type coverage unchanged, codecov clean.

Local verification I ran for this change:

- AST diff of every `mapped_column()` call: the 17 touched columns changed by nothing but the removal of `server_default=`; column counts unchanged.
- Rendered PostgreSQL `CREATE TABLE` before vs. after for all 8 touched tables: the only differences are the lost `DEFAULT` clause; types and `NOT NULL` are byte-identical.
- ORM-constructed default values before vs. after: identical (see above).
- Ran `tests/unit_tests/commands/test_legacy_model_type_migration.py` (the tests that exercise `provider_models`, `provider_model_settings` and `load_balancing_model_configs`) against this branch and against `main`: identical results (`2 failed, 32 passed, 6 errors` in both, same test names). Those failures are pre-existing in my ad-hoc local environment (a hand-assembled dependency set, not the project venv), not caused by this change.

I do not have permission to re-run the job myself (`POST .../rerun-failed-jobs` returns 403 - admin rights required), so I pushed an empty commit to trigger a fresh run. Happy to drop that commit once the check is green, and if it fails again on the same scenario I will dig into that specific E2E test.

wylswz (approved):

PR #42525, opened 2026-09-18
chore(models): remove redundant server_default from dataset models

CodeTrainerMan:
> **Note:** this is a follow-up slice of #29314, following the same pattern as #39886, #42519, #42520, #42521 and #42524.

## What this PR does

Removes 12 redundant `server_default` entries from `api/models/dataset.py`. Every one of them duplicates an equivalent application-level `default=` on the same column, so `server_default` is dead weight: it only affects DDL emitted when a migration is generated, while the value that actually gets written comes from the application-level `default`.

| Model | Columns |
|---|---|
| `Dataset` | `permission`, `is_multimodal` |
| `DocumentSegment` | `enabled`, `status` |
| `ChildChunk` | `type` |
| `DatasetKeywordTable` | `data_source_type` |
| `DatasetPermission` | `has_permission` |
| `DatasetAutoDisableLog` | `notified` |
| `Pipeline` | `is_public`, `is_published` |
| `DocumentSegmentSummary` | `status`, `enabled` |

`server_default` count in `api/models/dataset.py`: **69 -> 57**.

## What this PR does not do

The `created_at` / `updated_at` columns that rely **solely** on `server_default=func.current_timestamp()` are left untouched, since they have no application-level default and removing them would change insert behaviour. Same as the earlier slices (#42519, #42520, #42521, #42524).

## Verification

- **Rendered DDL**: rendered PostgreSQL `CREATE TABLE` for all 8 touched tables before vs. after. The only difference is the lost `DEFAULT` clause; column types and `NOT NULL` are identical. For example:
  ```diff
  -	enabled BOOLEAN DEFAULT true NOT NULL,
  +	enabled BOOLEAN NOT NULL,
  -	status VARCHAR(255) DEFAULT 'waiting' NOT NULL,
  +	status VARCHAR(255) NOT NULL,
  ```
- **Column metadata**: compared `default` / `server_default` / `type` / `nullable` for each touched column before vs. after. Every application-level `default=` is unchanged, `server_default` is gone, and the removed DDL defaults match the app-level defaults one-for-one: `'only_me'` / `DatasetPermissionEnum.ONLY_ME`, `false` / `False`, `true` / `True`, `'waiting'` / `SegmentStatus.WAITING`, `'automatic'` / `SegmentType.AUTOMATIC`, `'database'`, `true` / `True`, `false` / `False`, `false` / `False`, `false` / `False`, `'generating'` / `SummaryStatus.GENERATING`, `true` / `True`.
- **Unit tests**: ran the tests that exercise these tables, before vs. after on the same machine:
  ```
  tests/unit_tests/services/test_dataset_service_segment.py
  tests/unit_tests/services/test_dataset_service_lock_not_owned.py
  tests/unit_tests/services/rag_pipeline/pipeline_template/test_customized_retrieval.py
  tests/unit_tests/test_sqlite_fixtures.py
  ```
  `67 passed` on this branch, `67 passed` on `main` - identical.
- **AST check**: the touched `mapped_column()` calls changed by nothing but the removal of `server_default=`; no column was added, removed or retyped.

## Remaining work in #29314

After this PR the only remaining file with redundant `server_default` entries is `models/model.py` (21). I will pick that up in a follow-up PR unless someone else wants it.


CodeTrainerMan:
The `Run Web Full-Stack E2E / Chromium Full (1/3)` check failed, but not because of any test failure - the job died while pulling the Docker images, before a single scenario ran:

```
ssrf_proxy Error Head "https://registry-1.docker.io/v2/ubuntu/squid/manifests/latest":
  Get "https://auth.docker.io/token?...": context deadline exceeded
Error response from daemon: Head "https://registry-1.docker.io/v2/ubuntu/squid/manifests/latest": ... context deadline exceeded
Command failed (1): docker compose -f .../docker-compose.middleware.yaml --profile postgresql --profile weaviate up -d db_postgres redis weaviate sandbox ssrf_proxy plugin_daemon
```

That is a Docker Hub / network timeout on the runner, i.e. infrastructure flakiness. All backend checks pass (API Unit Tests and Integration Tests on every shard, DB Migration Test, Pyrefly coverage unchanged, codecov clean).

I don't have permission to re-run the job (`POST /actions/runs/{id}/rerun-failed-jobs` returns 403 - admin rights required), so I pushed an empty commit to trigger a fresh run. Happy to drop that commit once the check is green.

For context: this PR only removes `server_default` entries that duplicate an equivalent application-level `default=`, and the E2E environment builds its schema through the alembic migration chain anyway, so the model definitions cannot affect that environment either way. Details of the verification are in the PR description.

CodeTrainerMan:
Update: this PR has now hit two E2E failures, and they are **different scenarios with different causes** - which is the signature of an unstable suite rather than a code problem.

**Run 1** - `Chromium Full (1/3)` died before a single scenario ran:
```
ssrf_proxy Error Head "https://registry-1.docker.io/v2/ubuntu/squid/manifests/latest":
  Get "https://auth.docker.io/token?...": context deadline exceeded
Command failed (1): docker compose ... up -d db_postgres redis weaviate sandbox ssrf_proxy plugin_daemon
```
Docker Hub pull timeout on the runner - pure infrastructure.

**Run 2** - `Chromium Full (2/3)`, 18 of 19 scenarios passed. The single failure:
```
1) Run a minimal workflow app # features/apps/workflow-run.feature:4
     Then the workflow run should succeed # features/step-definitions/apps/workflow-run.steps.ts:21
       Error: expect(locator).toBeVisible() failed
       Locator: getByRole('status').getByText('SUCCESS', { exact: true })
       Timeout: 55000ms
```
A workflow run did not reach `SUCCESS` within 55s.

Neither one can be caused by this PR:

1. The E2E environment builds its schema through the alembic migration chain, so `server_default` entries in model definitions never reach it - removing them cannot change that environment.
2. This PR touches only `api/models/dataset.py`: `Dataset`, `DocumentSegment`, `ChildChunk`, `DatasetKeywordTable`, `DatasetPermission`, `DatasetAutoDisableLog`, `Pipeline`, `DocumentSegmentSummary`. The failing scenario runs a workflow app and waits for a run status; nothing on that path reads a DDL default from those columns - and every one of them keeps an equivalent application-level `default=`, which is what populates them.
3. The same E2E suite passed on my other slices from today (#42524 and #42526 both finished fully green, including all three Chromium shards), on essentially the same code base - that is what suite-level flakiness looks like.
4. For extra context, the `Web Full-Stack E2E` workflow only has 3 historical runs on record (all on `feat/agent-v2-e2e-test`, July) and all 3 failed, so the suite is new and still settling.

All backend checks have been green on every run of this PR (API Unit Tests and Integration Tests on all shards, DB Migration Test, Pyrefly coverage unchanged, codecov clean).

I do not have permission to re-run jobs myself (`POST /actions/runs/{id}/rerun-failed-jobs` returns 403 - admin rights required), so I pushed another empty commit to trigger a third run. If it fails again on yet another scenario I will stop retriggering and leave it to a maintainer to judge, rather than keep adding commits. Happy to drop the trigger commits once the check is green.

CodeTrainerMan:
Third run, and this time a backend job fell over - but again before any test executed. `Run API Tests / API Integration Tests (3.12, shard 2/4)`:

```
INTERNALERROR>   File ".../api/conftest.py", line 109, in pytest_sessionstart
INTERNALERROR>     stack.up()
INTERNALERROR>   File ".../api/tests/pytest_dify.py", line 130, in up
INTERNALERROR>     raise subprocess.CalledProcessError(
INTERNALERROR> subprocess.CalledProcessError: Command '['docker', 'compose', ..., 'up', '-d', '--wait',
INTERNALERROR>   '--wait-timeout', '180', 'db_postgres', 'redis', 'sandbox', 'ssrf_proxy']' returned non-zero exit status 1.
```

pytest died during session startup because the middleware containers did not come up within 180s, so zero tests ran. That is the same infrastructure problem as run 1 (the E2E job hit `context deadline exceeded` pulling `ubuntu/squid` from Docker Hub, in the same time window) - the runner's Docker/registry access is unreliable right now, not the code.

Shards 1/3/4 of the same integration suite passed on this run, and API Unit Tests have been green on every single run of this PR.

Nothing in this PR can affect a `docker compose up` of postgres/redis/sandbox/ssrf_proxy. I will wait a while for the registry pressure to ease and then trigger one final run; if it fails again for the same infrastructure reason I will stop retriggering and leave it for a maintainer to re-run, since I have no rerun permission (403 - admin rights required).

CodeTrainerMan:
Fourth run is fully green (18 success / 12 skipped / 0 failure), including all three Chromium shards and WebKit Browser Smoke of the E2E suite, and all four API Integration Test shards.

That confirms the three earlier failures were environmental:
1. E2E `Chromium Full (1/3)` - `docker compose` could not pull `ubuntu/squid` (`context deadline exceeded` against Docker Hub); zero scenarios ran.
2. E2E `Chromium Full (2/3)` - 18/19 scenarios passed, single flaky failure in `features/apps/workflow-run.feature:4` waiting for a run status.
3. API Integration `shard 2/4` - `docker compose up --wait` for the middleware containers did not finish within 180s, pytest died at session start with zero tests executed.

The change itself is identical across all four runs - no code was touched between them, only empty commits to trigger the workflows.

I'll leave the three trigger commits in place since the branch is already green; they can be squashed away at merge time, or I can drop them with a force-push if you'd prefer a single-commit branch - just say the word.

wylswz (approved):

PR #42541, opened 2026-09-19
fix(account): make case-fallback email lookup work in both directions

GareArc:
## Summary

`AccountService.get_account_by_email_with_case_fallback` only retried a lookup with a lowercased email when the *input itself* contained uppercase letters. Callers that already lowercase the email before calling it — workspace invites, login, forgot-password, email registration, webapp auth — could never match an existing account whose email was stored with different casing (for example, an SSO-provisioned account that keeps the identity provider's original casing). This caused a second, duplicate account to be created instead of reusing the existing one, and the account actually used for login could end up without the workspace membership the invite was supposed to grant.

The fix makes the fallback genuinely case-insensitive by comparing against `Account.normalized_email` (already indexed, already populated on every account-creation and email-change path) instead of only retrying in one direction.

This is tracked internally as CUS-1658; there's no public issue number to link.

## Test plan

- Added a regression test reproducing the failure mode: an account stored with mixed-case email, looked up with an already-lowercased email — fails against the old code, passes with the fix.
- Updated two existing test fixtures that constructed `Account` rows without `normalized_email`, which doesn't reflect how real accounts are created.
- `uv run --project api pytest api/tests/unit_tests/services/test_account_service.py api/tests/unit_tests/controllers/console/test_workspace_account.py` plus the other callers of this function (members, login, forgot-password, email registration, oauth, webapp auth, account commands) — all pass.
- `mypy` on the changed file and `ruff check`/`ruff format` on all changed files — clean (also ran via the pre-commit hook).

## Checklist


api/services/account_service.py
```diff
@@ -838,16 +838,21 @@ def get_owner_transfer_data(cls, token: str) -> dict[str, Any] | None:
...
         """
         account = session.execute(select(Account).where(Account.email == email)).scalar_one_or_none()
-        if account or email == email.lower():
+        if account:
             return account
 
-        return session.execute(select(Account).where(Account.email == email.lower())).scalar_one_or_none()
+        return session.execute(
+            select(Account).where(Account.normalized_email == normalize_email(email))
+        ).scalar_one_or_none()
```
jamalkamaladdin, line 858: `normalized_email` carries `account_normalized_email_idx`, a plain index, and migration 9b7c6d5e4f3a states that existing installations may already hold equivalent accounts. Two rows with the same normalized value make `scalar_one_or_none()` raise `MultipleResultsFound`. `has_account_with_normalized_email` guards the same query with `.limit(1)`.

```suggestion
        return session.execute(
            select(Account).where(Account.normalized_email == normalize_email(email)).limit(1)
        ).scalar_one_or_none()
```

jamalkamaladdin (commented):
One row count case on the fallback query.

wylswz (approved):

PR #42554, opened 2026-09-20
feat(agent): record E2B sandbox runtime usage 

zyssyz123:
## Summary

Agent message latency does not capture all E2B compute usage: sandbox initialization, file access, snapshots and automatic pauses can occur independently of LLM calls. This records provider execution duration and resource snapshots in a separate ledger so sandbox, vCPU and GiB hours can be reconciled without adding accounting work to the business path.

- Use existing Celery Beat and `ops_trace` workers to trigger a bounded `POST /internal/e2b/usage/collect` request. The Agent backend performs one scan using request-scoped clients and its existing E2B key. There is no Agent-process polling loop, leader election, outbox, observation worker or new queue.
- Keep binding create/get/resolve, generic leases, runner, file/snapshot/destroy services and E2B SDK operations free of accounting callbacks, HTTP reports, usage UUIDs, purpose labels and usage contexts. Their original retry, timing and cancellation behavior is retained.
- Persist provider lifecycle events and project one row per physical execution atomically. Unique event/execution keys make overlapping scans safe; only complete scans advance checkpoints. Steady-state state reads no longer acquire the project's ingestion lock.
- Resolve optional ownership during background ingestion by validating existing E2B metadata against the Binding → Workspace → App/Agent tenant chain and physical sandbox reference. Missing/deleted records remain unresolved while complete provider usage still contributes to project totals; previously verified ownership snapshots survive later deletion. A resource deleted before collection may remain unattributed.
- Validate optional metering configuration at scheduling/collection time. Invalid accounting settings fail or skip accounting only. Keep immutable activation T0, no environment column, no historical backfill, and unchanged table retention/schema for this review revision.
- The unrelated stdout-only sandbox file-error fix is separated into [#42624](https://github.com/langgenius/dify/pull/42624). It remains deployed and independently tested on dev, but is absent from this PR's diff.

Collection is bounded by the backend's 240-second deadline and caller's 260-second HTTP timeout. Celery configures 270/300-second soft/hard limits, whose availability depends on the worker pool; the operator guide explicitly covers gevent limitations. Scheduled messages expire after one interval, HTTP retry is disabled, and later Beat scans can safely redeliver provider IDs.

Implementation details: [E2B runtime audit document](https://langgenius.feishu.cn/wiki/ZEqMwAVVBiQgUSkUTSXcnLYqncb). This implements the reviewed internal design; no matching GitHub tracking issue was supplied.

## Validation

- Dev: 946 complete Agent local tests passed. Main PR: 944 complete Agent local tests passed (the two separated file-error cases are tested in #42624); 152 related API/schema tests passed. Agent Pyrefly, full unit-test strict API Pyrefly, relevant production Pyrefly/mypy, Ruff and formatting checks passed.
- Deployed implementation `46eb28401625db243642f29e5082b90d7feb692b` from the existing `deploy/agent` branch: [build](https://github.com/langgenius/dify/actions/runs/35555976299), [deployment](https://github.com/langgenius/dify/actions/runs/35556373003). All six related containers used that revision with zero restarts. Verified existing worker consumers, 60-second interval, unchanged project/T0, internal network policy and absence of old outbox/leader keys.
- Observed natural Beat → `ops_trace` worker → collection → advancing checkpoint, including a real business binding whose owner was resolved before deletion. Isolated bad metering settings and failing metering-service stubs did not affect real binding create/get: foreground metering calls were zero.
- Regressed real model success/failure/cancellation, file list/read/download and path errors, snapshots/clones, automatic pause and cleanup. Four sandbox resources produced 14 executions totaling **25,640 ms**, matching provider durations/resources exactly; the eight-second automatic-pause case contributed 8,172 ms.
- The real business fixture's execution was resolved (540 ms); 13 direct-backend fixture executions intentionally lacked business rows (25,100 ms), remained metered/unresolved and stayed in project totals. Previously verified owner snapshots survived business deletion. No application operation events were emitted.
- Verified missing-Bearer 401 and wrong-project 403. Two authorized complete scans preserved 14 executions / 25,640 ms while checkpoints advanced and T0 stayed fixed. A test-report Decimal serialization issue after the first scan's passing assertions was fixed in the harness; only the remaining second scan was run. All sandbox/snapshot and generated ToolFile fixtures were cleaned up.
- Restored the original shared-client shutdown sequence after removing the old accounting cleanup wrapper; 53 affected app/scheduler/route tests passed on dev and 38 app/route/collector tests passed on the PR branch.
- Final dev revision `509d5508bb2fdc4b0600a928eb617affebdc4be8`: [build](https://github.com/langgenius/dify/actions/runs/35557955547), [deployment](https://github.com/langgenius/dify/actions/runs/35558244275). A fresh isolated sandbox passed real model success, active cancellation and cleanup. This supplemental fixture is separate from the 14-execution reconciliation above. The normal deployment forced the old container to stop after ten seconds, so graceful shutdown completion is not claimed.
- A main-only test-isolation fix restores Celery current/default app and finalizer state after schedule tests. The failing schedule → existing-plugin-task order was reproduced before the fix; all 52 related ordered tests then passed. Production code and dev configuration are unchanged.
- Main-only migration parent and typing fixes are retained. This branch does not overwrite `deploy/agent`.

- Final head `2fb0a348f683c501132172f777c82dfca44056dc`: [Main CI](https://github.com/langgenius/dify/actions/runs/35559091754) succeeded; 46 checks succeeded and 13 were skipped, with no failed or pending checks. Codecov patch is 94.19% and project is 87.58%; both status checks passed. PRs remain open for review and are not merged.

## Screenshots

Backend-only change. The linked document contains seven native UML views of the final component boundaries, business attribution, single collection, state/quality, schema, deployment and failure isolation, plus verified regression evidence.

## Checklist

- [x] The operator guide and guide entry page describe scheduled, isolated provider collection.
- [x] I've verified the change and added or updated meaningful regression tests.
- [x] The linked implementation document update is complete.
- [ ] I ran `make lint && make type-check` (backend) and `vp staged` (frontend). The actual checks run are listed above; frontend code is unchanged.

From Codex


BeautyyuYanli (dismissed):
Requesting changes on two architectural points:

1. E2B usage polling should be triggered by the existing **Celery Beat** infrastructure rather than a separate polling loop in each Agent server process. Ingestion already guarantees idempotency, so a separate distributed leader-election mechanism is unnecessary for correctness.
2. E2B observability introduces too much coupling into shared runtime and business logic. Please refactor this integration around clear separation of concerns: provider-specific accounting should remain isolated, shared abstractions should retain provider-independent responsibilities, and optional observability should minimize changes to business behavior, execution timing, and failure semantics.

<details>
<summary>Concrete examples of coupling introduced by this PR</summary>

- **Binding lookups now perform accounting transactions.** In `api/services/agent/workspace_service.py`, `get_active_binding()` and `resolve_active_binding_for_scope()` previously returned the queried binding. They now call `_register_usage_allocation()` before returning. With metering enabled, even an already-registered binding enters a separate transaction and attribution checks. Metering configuration errors, ownership conflicts, or database failures can therefore fail an otherwise valid business lookup. `create_binding()` also requires accounting registration to succeed before allocating the resource.

- **Foreground lookups contend with background ingestion.** `SandboxUsageService._write()` locks the project's `metering_activated` row. Binding registration, event ingestion, and `get_state()` all use this method. Consequently, a background ingestion transaction can delay a foreground binding lookup through the shared lock.

- **The generic lease helper now carries metering responsibilities.** `runtime_backend/leases.py::open_runtime_lease()` adds a `purpose` argument, generates a UUID, and establishes a `runtime_usage_context` around acquisition, execution, and release. This happens for every backend using the helper, including non-E2B backends and deployments with metering disabled.

- **Usage-specific context spreads across shared execution and business services.** `AgentRunRunner.run()` now wraps the entire run in a usage context containing `run_id` and the request idempotency key. `DifyRuntimeLayer`, file list/read/download methods, snapshot creation, and binding destruction now supply accounting purpose labels such as `agent_run`, `file_read`, `snapshot`, and `cleanup`. These shared components must now participate in E2B observation plumbing.

- **Diagnostic delivery changes resource-operation timing and cancellation handling.** In `runtime_backend/e2b.py`, create/connect/pause/kill now await requested/observed HTTP reports, potentially adding roughly two seconds per SDK attempt. `_run_e2b_idempotent_operation()` now also tracks reporting cancellation and defers its propagation during pause/kill cleanup. Separately, `_best_effort_close_data_plane()` and `_best_effort_pause()` now propagate `CancelledError` that was previously caught; those changes also apply when metering is disabled.

- **An unrelated file-error behavior change is bundled in.** The sandbox scripts in `server/binding_files.py` now redirect stderr to stdout, changing how missing-file/directory errors are classified on stdout-only transports. This applies independently of metering and should be separated from the accounting integration.

</details>

zhaohao1004 (approved):

zhaohao1004 (approved):

MRZHUH (approved):

PR #42608, opened 2026-09-21
feat(cli): difyctl v2 over the server catalog

GareArc:
## Summary

This PR does three things: it replaces the v1 command tree with three commands driven by the catalog the server publishes, it introduces the plugin system those commands run on, and it turns the single embedded agent skill into a collection with list and install commands.

### 1. Operation commands over the server catalog

```
difyctl ops                       # list every operation the server exposes
difyctl ops describe <op-id>      # one operation: input schema, usage line, pinned values
difyctl call <op-id> --input ...  # run it; --stream/--only for sse ops, --output for file ops
```

- The server publishes its operations at `/openapi/v1/_catalog`. The CLI caches that document per server, sends its sha256 as `X-Dify-Catalog` on every request, and refetches once when the server answers `412 catalog_stale`. A `404` refetches once too, and resends only if the catalog moved, since a dropped route answers before the server's catalog guard runs. An unknown op id triggers one refetch before exit 6.
- `protocol/` holds the frozen catalog shape: response kinds, parameter binding, pins, and the per-event fold table. `call/` holds the flag table by response kind, input loading (`--input '<json>'`, `@file`, `@-`), JSON Schema validation, request building and one renderer per kind.
- Streaming ops fold to `{status, text, outputs?, message_id?, conversation_id?, error?, hints}` by default; `--stream` prints each event as one JSON line. Any response may carry `hints`, the next calls the server filled in.
- Exit codes are frozen at 0/1/2/4/6/7. Ctrl-C ends a stream as `incomplete` with exit 0; there is no 130.
- Static commands kept: `version login logout workspace list|use config get|set|unset cache refresh|clear skills list|install`. The agent skill teaches the discovery flow instead of enumerating operations.
- The v1 command tree, contract client and version-compat layer are removed.

### 2. Plugin system

The CLI is a set of plugins built on demand by a small kernel. The design follows cordis, the plugin kernel behind DeepSeek Harness, with the parts a short-lived CLI does not need left out: no event system, no hot reload, no effect rollback. A plugin here is `{ name, needs, build }`, which maps one to one onto cordis's `{ name, inject, apply }`, so a long-lived mode later could move onto cordis without touching the commands.

- `definePlugin({ name, needs, build })`. `needs` lists the plugin objects `build` may read through `ctx.get`, and TypeScript checks `build`'s context against that list. It is a compile-time allow-list, never a load list.
- `Context.get(plugin)` builds lazily and caches per invocation; a plugin never builds if nothing asks for it. `ctx.defer(fn)` queues cleanup that runs after the command, in reverse order, on success, failure or Ctrl-C.
- Eleven plugins, one folder each: `argv global-flags env config session token catalog http ops io commands`. Code outside a plugin folder imports only the plugin's index and reaches its behaviour through `ctx.get`; a helper only commands use lives outside `src/plugins/`.
- `commands` is the one pipeline: resolve the command from the generated tree, parse argv against its Zod input, validate, `run(input, ctx)`, print. Commands are plain classes and never branch on which command they are.
- Global flags (`--verbose`) are one Zod object in `global-flags`, parsed off argv before the command is resolved, and never reach a command's input. Every write goes through the `io` service (`line`, `document`, `notice`, `raw`).
- `main(argv, overrides)` is the embedding seam: tests and any future host seed `argv`, `io` or a mock `http` as overrides and run the same pipeline in process.

### 3. Skill collection

```
difyctl skills list                                  # every skill: name and description
difyctl skills install <dir>                         # write the whole collection under <dir>/<name>/
difyctl skills install <dir> --skill <name>          # one skill (a scenario skill brings the basic one)
```

- Skills are plain markdown under `skills/` at the repo root, one folder per skill. Only the basic skill `difyctl` ships in this PR; it points at the others through an "Other skills" section. `npx skills add langgenius/dify --skill difyctl -g` installs the same files.
- `cli/scripts/generate-skill-collection.ts` embeds the folder into `cli/src/skills/collection.generated.ts`, committed and drift-checked by `pnpm skills:check` in `pnpm ci`, the same way the command tree is.
- `<dir>` is the agent's skills root (`~/.claude/skills`, `~/.codex/skills`, ...). The caller names it; difyctl does not detect agents, stamp copies, or remember roots. Existing copies are overwritten, so re-running the command after an upgrade is the refresh.
- The old TS skill template, its `--stdout`/`--yes` flags, and the v1 agent detection are removed.

Things a reviewer should know:

- A shipped v1 binary gets `412` from a catalog-gated server; there is no compat shim.
- `cli-e2e.yml` is manual-only and currently runs zero suites (helpers retained). `cli-smoke.yml`'s runner has no `bun` and no `DIFY_TOKEN` provisioning; both pre-existing.
- `http.timeout` is a deadline for response headers, not the whole exchange, so an SSE run can stream past it.
- Server side: `feat/openapi-catalog` (this PR's base).

Verified with `pnpm type-check`, `vp check cli`, `pnpm build`, `pnpm tree:check`, `pnpm skills:check`, the unit suite against the in-process mock (448 tests), the compiled binary against the mock for the operation commands, and the skills commands run from source into a scratch directory.

## Screenshots

| Before | After |
| ------ | ----- |
| n/a (CLI) | n/a (CLI) |

## Checklist


laipz8200 (commented):
The part of `.github/` is LGTM

wylswz:
**Human-facing UX of the v2 surface — is the agent-first trade-off intended to be this total?**

No objection to the architecture. Catalog-driven `call` is the right answer to v1's real problem: `get app --mode` hardcoded a whitelist in the client, and the v2 commit rightly names *version compat* as one of the things being deleted. Server adds an op, CLI ships nothing — that's a hard win and I don't want to trade it back.

My concern is narrower: for a human at a terminal, v2 currently has **no** affordances, and I think a couple of them are cheap enough that they don't cost the architecture anything.

Three concrete symptoms, all from a first session on this branch:

1. `call` is a fixed tax on every business operation. The static command tree is only login/logout/config/cache/workspace/skills/ops/version — everything a user actually came for is `call <op-id>`. That's fine as the machine contract, but it means the human never gets a verb-first entry point.

2. Op ids are hard to type, and the two id syntaxes are silently different.

```
$ difyctl console_app list
{"error":{"code":"usage_invalid_flag","message":"unknown command: console_app list","hint":"run difyctl help"}}

$ difyctl ops console_app.list
{"error":{"code":"usage_invalid_flag","message":"unexpected argument \"console_app.list\""}}
```

3. Help is JSON-only, which makes the first five minutes hard.

### What kubectl actually does, and where we already match it

kubectl is usually cited as "extensible *and* usable", here are a few designs they made, some of which are similar to what we already have today.

| kubectl mechanism | What it buys | difyctl today |
| --- | --- | --- |
| **Discovery**. `/api/v1` and `/apis/<group>/<version>` return an `APIResourceList`; kubectl caches it under `~/.kube/cache/discovery/` and newer servers aggregate it into a single endpoint. | Resource types are learned at runtime, so a new CRD needs no client release. | **`/openapi/v1/_catalog`**, plus the fingerprint header and the 412 `catalog_stale` refetch. Same idea, arguably cleaner. |
| **Bounded verb set**. The API verbs are 8: `get list watch create update patch delete deletecollection`. kubectl's UX verbs are a larger but still *closed, client-owned* list. | The verb is guessable without a lookup, and it means the same thing on every resource. | The verb is the trailing segment of a free-form op id: `set_role`, `revoke_current`, `import_confirm`, `check`, `switch`. Open set, not quite guessable. |
| **Long composite nouns with server-declared abbreviations**. `persistentvolumeclaim`/`pvc`, `horizontalpodautoscaler`/`hpa`. A CRD declares `spec.names`: `plural` and `kind` required, `singular` defaulting to the lowercased kind, plus optional `shortNames` and `categories`. | Composite nouns are tolerable *because* the server ships the abbreviation; the client never invents names. | Op ids are `.` connected segments|
| **Subresources**. `deployments/scale`, `pods/log`, `pods/exec`, `serviceaccounts/token` are separate discovery entries with their own verbs, surfaced through dedicated commands (`kubectl scale`, `kubectl logs`). | Non-CRUD actions become **noun structure**, so the verb set does not have to grow for every new action. `scale` is a `patch` on a subresource, not a ninth verb. | Our AIP custom methods are already subresource-shaped — `/apps/{id}/workflow:run`, `/apps/{id}/tasks/{id}:stop`, `/apps/{id}/human-input-forms/{token}:submit`, `/apps/{id}/dependencies:check` — but the catalog flattens them into the op id, so they read as new verbs instead of subresources. |

Two consequences worth pulling out:

1. **Guessability comes from the bounded verb set, not from the `verb noun` shape.**
   Renaming `call workspace.members.set_role` to `set_role workspacemember` buys
   nothing on its own: the user still has to look up that the verb is `set_role`
   and not `set-role`, `grant`, or `update`. The ergonomic win only arrives if the
   verb set is closed and short.

2. **Subresources are how kubectl keeps that set closed**, and it is the mechanism
   we are missing for exactly the ops that break a naive `verb noun` mapping.

GareArc:
> **Human-facing UX of the v2 surface — is the agent-first trade-off intended to be this total?**
> 
> No objection to the architecture. Catalog-driven `call` is the right answer to v1's real problem: `get app --mode` hardcoded a whitelist in the client, and the v2 commit rightly names _version compat_ as one of the things being deleted. Server adds an op, CLI ships nothing — that's a hard win and I don't want to trade it back.
> 
> My concern is narrower: for a human at a terminal, v2 currently has **no** affordances, and I think a couple of them are cheap enough that they don't cost the architecture anything.
> 
> Three concrete symptoms, all from a first session on this branch:
> 
> 1. `call` is a fixed tax on every business operation. The static command tree is only login/logout/config/cache/workspace/skills/ops/version — everything a user actually came for is `call <op-id>`. That's fine as the machine contract, but it means the human never gets a verb-first entry point.
> 2. Op ids are hard to type, and the two id syntaxes are silently different.
> 
> ```
> $ difyctl console_app list
> {"error":{"code":"usage_invalid_flag","message":"unknown command: console_app list","hint":"run difyctl help"}}
> 
> $ difyctl ops console_app.list
> {"error":{"code":"usage_invalid_flag","message":"unexpected argument \"console_app.list\""}}
> ```
> 
> 3. Help is JSON-only, which makes the first five minutes hard.
> 
> ### What kubectl actually does, and where we already match it
> kubectl is usually cited as "extensible _and_ usable", here are a few designs they made, some of which are similar to what we already have today.
> 
> kubectl mechanism	What it buys	difyctl today
> **Discovery**. `/api/v1` and `/apis/<group>/<version>` return an `APIResourceList`; kubectl caches it under `~/.kube/cache/discovery/` and newer servers aggregate it into a single endpoint.	Resource types are learned at runtime, so a new CRD needs no client release.	**`/openapi/v1/_catalog`**, plus the fingerprint header and the 412 `catalog_stale` refetch. Same idea, arguably cleaner.
> **Bounded verb set**. The API verbs are 8: `get list watch create update patch delete deletecollection`. kubectl's UX verbs are a larger but still _closed, client-owned_ list.	The verb is guessable without a lookup, and it means the same thing on every resource.	The verb is the trailing segment of a free-form op id: `set_role`, `revoke_current`, `import_confirm`, `check`, `switch`. Open set, not quite guessable.
> **Long composite nouns with server-declared abbreviations**. `persistentvolumeclaim`/`pvc`, `horizontalpodautoscaler`/`hpa`. A CRD declares `spec.names`: `plural` and `kind` required, `singular` defaulting to the lowercased kind, plus optional `shortNames` and `categories`.	Composite nouns are tolerable _because_ the server ships the abbreviation; the client never invents names.	Op ids are `.` connected segments
> **Subresources**. `deployments/scale`, `pods/log`, `pods/exec`, `serviceaccounts/token` are separate discovery entries with their own verbs, surfaced through dedicated commands (`kubectl scale`, `kubectl logs`).	Non-CRUD actions become **noun structure**, so the verb set does not have to grow for every new action. `scale` is a `patch` on a subresource, not a ninth verb.	Our AIP custom methods are already subresource-shaped — `/apps/{id}/workflow:run`, `/apps/{id}/tasks/{id}:stop`, `/apps/{id}/human-input-forms/{token}:submit`, `/apps/{id}/dependencies:check` — but the catalog flattens them into the op id, so they read as new verbs instead of subresources.
> Two consequences worth pulling out:
> 
> 1. **Guessability comes from the bounded verb set, not from the `verb noun` shape.**
>    Renaming `call workspace.members.set_role` to `set_role workspacemember` buys
>    nothing on its own: the user still has to look up that the verb is `set_role`
>    and not `set-role`, `grant`, or `update`. The ergonomic win only arrives if the
>    verb set is closed and short.
> 2. **Subresources are how kubectl keeps that set closed**, and it is the mechanism
>    we are missing for exactly the ops that break a naive `verb noun` mapping.

@wylswz 

### 1. Every business operation goes through `call`

In v1 the CLI was tied to a Dify version, and every command was written by hand against one endpoint. To give users a verb-first command like `list apps`, the CLI would need either its own table mapping verbs to op ids, or a command tree generated from the catalog. Either way the CLI is back to knowing business names, which is what v2 removes.

What the user pays for this is one extra word, and it is always the same word. `kubectl get` and `gh pr` have the same first word on every command too. You are right that `set_role workspacemember` would not be any easier than `call workspace.members.set_role`. What makes ids easy to guess is good naming, covered below.

### 2. Two id syntaxes

Agreed that two shapes is a problem. It should be fixed in this PR, but toward the dotted form, not the spaced one. Static commands become `workspace.use <id>`, `cache.clear`, `config.get <key>`, `skills.install`. Single-word commands stay.

Why dotted rather than spaced:

- **It is domain-first naming.** `workspace.members.invite` reads resource, then sub-resource, then action. That is the order a user or agent thinks in when operating on Dify: confirm what you are touching, then what you do to it. The spaced form `workspace members invite` reads as a command tree, which is verb-noun grammar and sets up the verb-first expectation I do not want. Confirming a verb first does not buy any entropy reduction.
- **It marks the id as data, not syntax.** The dot says "this is a name, look it up", as opposed to `call`, `help`, `login`, which are the CLI's own words. Precedent is any task runner with namespaced ids: `rake db:migrate`, `gradle app:build`, `pnpm run web:build`.

- **`call` already owns the verb slot.** `call workspace.members.invite` is one verb and one subject. With spaced ids it would be `call workspace members invite`: two grammars on one line, and the parser has to guess where the id ends and the arguments start. That is confusing for a human and for an agent.
- **One token travels unchanged.** The same string appears in `help` output, in the usage line, in a server hint's `op` field, in the skill, and on the command line. An agent copies it from one place to the other without transformation. A spaced id would need joining and splitting at every boundary.

With one grammar, `difyctl console_app list` cannot happen. Two static commands go away: `workspace list` collides with the op `workspace.list`, and `ops describe` prints the same document as `help <id>`.

No short names for now. Ids are dotted paths and `help <namespace>` shows a group. If aliases are ever needed they will be catalog data, not client data.

### 3. Human-facing output

A human-friendly surface here is four things: help, errors, results and streams. v1 had hand-written formats for all of them, and they cost code and rotted when the server moved. A text help on its own would be a lonely island: the next thing the user sees is a result, a stream or an error, so the real question is whether those can be made human-friendly without dedicated handling.

Help and errors could. I own both shapes, so a text renderer is cheap. Results and streams cannot. A response has many fields and the client has no basis for picking which ones deserve a column. Most responses do not fit a table at all: nested objects, lists inside objects, free text. There is no elegant way to render arbitrary JSON as a good visual, and every attempt is a hand-written column list per op, which is the v1 code this branch deletes. kubectl has the same limit: `get` shows a few hand-picked columns, and anything past that is `-o yaml` or `-o json`, the raw object. Streams are the same. A readable stream view needs per-endpoint choices about which events to show, beyond the one fixed, event-keyed fold table I allow.

So results and streams stay JSON. Since they are what the user sees most of the time, help and errors follow the same shape. One shape to learn.

One small change that can improve UX: bare `difyctl` will print a pointer to `help <words>` and the skill. Still JSON.

### On the kubectl comparison

**Bounded verb set.** kubectl gets by with eight verbs because every resource is the same thing: a declarative object with `spec` and `status`, reconciled the same way. Non-CRUD actions hide behind subresources plus `patch`, and that works because a controller does the real work: a `scale` only changes the declaration, not a direct action, unlike Dify.

Dify is imperative and cross-domain. Run an app, install a plugin, submit a human-input form, confirm an import, check dependencies, hit-test a knowledge base, publish a workflow, re-index a document. These actions have their own semantics and the resources do not share one lifecycle. Forcing them behind `patch` or any general modification verb hides the intent in a body field the user has to look up anyway. When a user or agent operates on Dify, the resource is the first thing to confirm, then the action. So the verb set stays open and domain-named.

The model I follow is the one you mentioned: AIP custom methods (`workflow:run`, `tasks/{id}:stop`, `forms/{token}:submit`, `dependencies:check`). The fitting comparison is gcloud, built on the same rules: `instances start`, `instances stop`, `instances attach-disk`. Discoverability comes from nesting plus `help`, which is what `help <namespace>` and `help <words>` do here. Try `difyctl help account.sessions`.

The op id should mirror the path, including nesting: a standard verb for a standard action, otherwise the `:action` in the path. Next step is to rename the op ids that break that rule:

| Today | Path | After |
| --- | --- | --- |
| `account.sessions.revoke_current` | `DELETE sessions/self` | `account.sessions.self.revoke` |
| `workspace.members.set_role` | `PATCH members/{id}` | `workspace.members.role.set` |
| `console_app.dsl.import_confirm` | `POST imports/{id}:confirm` | `console_app.dsl.imports.confirm` |
| `console_app.file.upload` | `POST apps/{id}/files` | `console_app.files.create` |

PR #42615, opened 2026-09-21
ci: add deterministic pull request size labels

### LINKED ISSUE #42529: fix size label
dosu 's is buggy

https://github.com/facebook/pyrefly/blob/main/.github/workflows/size_labeler.yml

meta has a better one

@dosu what do you think

surbadest:
## Summary

Adds a deterministic PR size-labeling workflow for #42529. It counts additions and deletions, excluding common JavaScript and Python lockfiles, and uses the thresholds documented by the existing size:XS through size:XXL labels. It replaces only stale size labels and leaves other labels intact.

The workflow reads GitHub API metadata only: it never checks out or runs PR code. Incomplete file listings are skipped rather than assigned an incorrect size. Includes 16 mocked API tests for boundaries, exclusions, idempotency, and label replacement.

## Validation

- node --test .github/scripts/size-labeler.test.cjs: 16 passed
- actionlint 1.7.12: passed (existing depot-ubuntu-24.04 runner label allowed)
- git diff --cached --check: passed
- Live label writes have not been tested; this pull_request_target workflow takes effect after merge.

## Rollout

Maintainers should disable Dosu's size labeling when enabling this workflow, so the two labelers do not overwrite each other. The existing repository thresholds are preserved instead of copying Pyrefly's different thresholds. Only the explicitly listed lockfiles are excluded; this is not general generated-file detection.

fix #42529

From Codex


crazywoola:
please fix the ci lint errors

crazywoola (approved):

PR #42678, opened 2026-09-22
fix(web): remove a Code node output variable once the removal is confirmed

### LINKED ISSUE #42677: [Bug] Code node output variable is not removed after the removal is confirmed when other nodes reference it
### Self Checks

- [x] I have read the [contributing guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md).
- [x] I have searched for existing issues — no duplicate found.
- [x] I can reproduce the issue on the latest `main` (`151680ee`).

### Dify version

1.16.1 (self-hosted). The code path is unchanged since the new workflow engine (#3160), so it affects every version since 1.0, including `main` (`151680ee`).

### Affected area

Web / Workflow editor → **Code node** → Output Variables.

### Steps to reproduce

1. Add a **Code** node with two output variables, e.g. `ret` and `body`.
2. Reference one of them (`ret`) from a downstream node (e.g. an Answer node).
3. In the Code node panel, click the trash icon on the `ret` row under **Output Variables**.
4. Dify asks *"The variable is used in other nodes. Do you still want to remove it?"* → click **Confirm**.

### Actual behaviour

- The reference inside the downstream node **is** cleared.
- The `ret` row is **still shown** in the Code node's output variable list, and the variable is still present in the node data (`outputs` + `outputKeyOrders`), so it is still saved into the DSL and still exported.

This leaves the editor in a half-applied state: the variable no longer exists anywhere except in the node that declares it.

### Expected behaviour

After confirming, the row should disappear from the Code node in exactly the same way it does when the variable is not referenced by any other node — i.e. the variable is gone from both the list and the node data.

### Root cause

`web/app/components/workflow/nodes/_base/hooks/use-output-var-list.ts`

| # | Location | Behaviour |
|---|---|---|
| 1 | `output-var-list.tsx:116` | the row's trash button calls `onRemove(index)` |
| 2 | `use-output-var-list.ts:131-139` (`handleRemoveVariable`) | if `isVarUsedInNodes([id, key])` it only opens the confirm dialog, stores `removedVar` and **returns** — the `index` argument is dropped |
| 3 | `use-output-var-list.ts:121-130` (`removeVarInNode`, wired to `onRemoveVarConfirm`) | on confirm it deletes the inspect var, calls `removeUsedVarInNodes(removedVar)` and closes the dialog. It never touches `inputs.outputs` / `outputKeyOrders` |
| 4 | `use-output-var-list.ts:141-162` | the non-referenced path *does* perform the removal, so the two paths disagree |

Step 3 is why the UI ends up half applied: the downstream references are cleared, but the variable itself is never removed.

The same interaction is implemented correctly in the two other places that show this dialog:

- RAG pipeline input field list — `rag-pipeline/components/panel/input-field/field-list/hooks.ts:75-86` removes the input field on confirm
- Variable Assigner node — `variable-assigner/use-config.ts:136-147` removes the group on confirm

Code node output variables are the only one of the three missing the removal.

The backend is not involved: nothing is ever sent, because the removal never happens on the client in the first place.

### Proposed fix

Extract the removal into a single `removeOutputVariable(index)`, remember the row index when the dialog is opened, and call it from both the direct path and the confirm handler. A PR will follow.

Neo1Noir:
Fixes #42677.

## Summary

The confirmation dialog shown when removing a **Code node output variable that other nodes still reference** only cleared those downstream references. `removeVarInNode` (wired to `onRemoveVarConfirm`) never removed the row from the node itself, so the variable stayed in the output list and in the saved DSL — the editor ended up half applied. The row `index` was dropped when the dialog was opened (`handleRemoveVariable` early-returns right after `setRemovedVar`), which left the confirmation with no way to know which row to remove.

| # | Location (before this PR) | Behaviour |
|---|---|---|
| 1 | `output-var-list.tsx:116` | the row's trash button calls `onRemove(index)` |
| 2 | `use-output-var-list.ts:131-139` | `isVarUsedInNodes([id, key])` → open dialog, store `removedVar`, **return** (the `index` is dropped) |
| 3 | `use-output-var-list.ts:121-130` | on confirm: `deleteInspectVar` + `removeUsedVarInNodes` + close dialog — never touches `outputs` / `outputKeyOrders` |
| 4 | `use-output-var-list.ts:141-162` | the non-referenced path *does* remove the row, so the two paths disagree |

The backend is untouched: nothing is ever sent, because the removal never happens on the client.

## Fix

The removal is extracted into `removeOutputVariable(index)` and called from both the direct path and the confirm handler, with the row index kept in state while the dialog is open:

```diff
   const removeVarInNode = useCallback(() => {
-    const varId = nodesWithInspectVars
-      .find((node) => node.nodeId === id)
-      ?.vars.find((varItem) => varItem.name === removedVar[1])?.id
-    if (varId) deleteInspectVar(id, varId)
+    // The confirmation only covers variables that other nodes still reference,
+    // the row itself has to be removed from this node as well.
+    removeOutputVariable(removedIndex)
     removeUsedVarInNodes(removedVar)
     setIsShowRemoveVarConfirm(false)
-  }, [deleteInspectVar, id, nodesWithInspectVars, removeUsedVarInNodes, removedVar])
+  }, [removeOutputVariable, removedIndex, removeUsedVarInNodes, removedVar])
```

The confirmed path now also drops the inspect var and recomputes `default_value` (Code node with `error_strategy: default-value`) exactly like the direct path — before, only the direct path did either.

## Tests

New spec: `web/app/components/workflow/nodes/_base/hooks/__tests__/use-output-var-list.spec.ts` — 6 cases.

| Case | Asserts |
|---|---|
| variable not referenced | row, `outputs` entry and inspect var removed right away |
| referenced → confirm | downstream references cleared **and** the row removed from `outputs` / `outputKeyOrders` |
| referenced → confirm (second row) | removes by its own index, not by the shared name |
| referenced → cancel | nothing changes |
| two rows sharing a name | the `outputs` entry is kept while another row still uses the name |
| `error_strategy: default-value` | `default_value` recomputed on the confirmed path |

**Ablation verification**: with `removeOutputVariable(removedIndex)` removed from the confirm handler, exactly the 4 confirmation-related cases fail and the other 2 stay green. Restored → 6/6 green.

## Verification

- `npx vp test --project unit --run app/components/workflow/nodes/_base/hooks/__tests__/use-output-var-list.spec.ts app/components/workflow/nodes/code/` → **34 passed**
- `npx vp test --project unit --run app/components/workflow/ app/components/rag-pipeline/` → **4259 passed, 2 failed**. Both failures are in `panel/chat-record/__tests__/index.spec.tsx` and are **pre-existing**: they fail identically on a clean `151680ee` checkout with this change stashed.
- `npx vp check --no-lint <2 files>` → all files correctly formatted; `npx eslint <2 files> --no-warn-ignored` → clean.
- Note on `vp staged` / oxlint: my local `node_modules` is older than this base and is missing `@shadcn/lint`, so oxlint could not start locally. Formatting was applied and re-verified with `vp check --fix --no-lint`, and the pre-commit hook was skipped for that reason only.

## Screenshots

| Before | After |
| ------ | ----- |
| Confirming the dialog leaves the row (and the variable) in the Code node | The row is gone, the downstream references are cleared |

## Checklist

- [ ] I ran `make lint && make type-check` (backend) and `vp staged` (frontend) to appease the lint gods — no backend change was made here, and `vp staged` could not run locally (stale `node_modules` missing `@shadcn/lint`), so formatting and eslint were run directly on the two touched files as described above.

From WorkBuddy


web/app/components/workflow/nodes/_base/hooks/use-output-var-list.ts
```diff
@@ -174,6 +159,30 @@ function useOutputVarList<T>({
     ],
   )
 
+  const removeVarInNode = useCallback(() => {
+    // The confirmation only covers variables that other nodes still reference,
+    // the row itself has to be removed from this node as well.
+    removeOutputVariable(removedIndex)
```
WTW0313, line 182: Thanks for fixing the confirmation flow. I noticed one related case: when a Code node uses “Default value” error handling, `removeOutputVariable` rebuilds the defaults for every remaining output. For example, if `primary` has a fallback of "sentinel" and `secondary` is referenced downstream, confirming the removal of `secondary` resets `primary` to "". Could we preserve the configured values for outputs that remain, and add a test with a non-empty fallback?

WTW0313 (commented):
While checking the surrounding code, I found two related behaviors that predate this PR. Adding an output through `handleAddVariable` resets fallback values already configured for other outputs. Renaming or changing an output’s type through `handleVarsChange` does the same; directly removing an unreferenced output also follows that path. They appear to share the same default-value rebuilding logic. These could be addressed together with this change or tracked as a follow-up.

web/app/components/workflow/nodes/_base/hooks/use-output-var-list.ts
Neo1Noir: Good catch, thanks — fixed in 425863e5.

`removeOutputVariable` now goes through a new `mergeDefaultValue` helper: the list is rebuilt from the node's current outputs, but a key that is still there **with the same type** keeps its configured value, and a key that disappeared is dropped. So in your example `primary` keeps `"sentinel"` when `secondary` is removed.

The test you asked for is included (`keeps the configured fallbacks of the outputs that stay behind a confirmed removal`, non-empty fallback `"sentinel"`), plus the same assertion for the direct-removal path.

Neo1Noir:
Addressed the review in 425863e5 (single push), and updated the description with the new tests and verification numbers.

Besides the confirmed-removal case from the inline comment, the two behaviours mentioned in the review body are covered too, since they now share the same call site: `handleAddVariable` keeps the fallbacks of the outputs that stay, and `handleVarsChange` carries a renamed output's fallback over to its new name. A type change still resets only that output, so a value that no longer matches its type is never carried over.

Neo1Noir:
Heads-up: I amended the second commit and force-pushed (`cd815942`) to fix the `Style Check / TS Common` failure on the previous head.

That failure was mine: the new `mergeDefaultValue` call sites added three `typescript/no-explicit-any` findings beyond the per-file baseline recorded in `oxlint-suppressions.json`, and `removeOutputVariable`'s dependency array had lost `id`. My own code now adds no new findings — the file sits at 5 recorded `any` usages against its baseline of 8 — and the dependency array is back to its previous shape. Behaviour is unchanged, so the review feedback above is unaffected.

Verification after the amend (`npx vp test --project unit --run ...`): hook spec + `nodes/code/` + `error-handle/` → **50 passed**, same as before.

One note for transparency: oxlint itself panics on my Windows checkout (`apps/oxlint/src/output_formatter/default.rs`, the default formatter), so the lint side is reasoned against the documented count baseline rather than executed locally — the CI is the judge.

WTW0313 (approved):

Issue #38833, opened 2026-07-13
[Refactor/Chore] Migrate frontend linting to Vite+ with an ESLint fallback


hyoban:
### Self Checks

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md) and [Language Policy](https://github.com/langgenius/dify/issues/1542).
- [x] This is only for refactors or chores; if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify/issues), including closed ones.
- [x] I confirm that I am using English to submit this report, otherwise it will be closed.
- [x] 【中文用户 & Non English User】请使用英语提交，否则会被关闭 ：）
- [x] Please do not modify this template :) and fill in all the required fields.

### Description

Migrate the frontend lint workflow from ESLint-first checks to Vite+ linting backed by Oxlint, while retaining ESLint only for non-code languages that Oxlint cannot parse.

The migration should:

- make `vp lint` the JavaScript and TypeScript linter;
- translate only the rules that are actually enabled in the repository's current ESLint configurations;
- prefer native Oxlint rules, then compatible Oxlint JavaScript plugins;
- limit executable ESLint configuration to JSON, JSONC, JSON5, YAML, TOML, and Markdown;
- record unsupported code-only behavior as accepted migration tradeoffs instead of keeping an ESLint code path;
- avoid depending on `@antfu/eslint-config`;
- preserve existing diagnostics through separate Oxlint and ESLint suppression baselines;
- integrate the two paths with package scripts, staged fixes, CI, editor settings, and contributor documentation;
- keep `packages/contracts/**` lint-ignored and format-only.

The accepted code-only gaps include six core fallback rules, JavaScript `dot-notation`, the former 223-rule declaration snapshot, the CLI declaration import restriction, and three Dify UI Tailwind rules.

### Motivation

The existing ESLint-first workflow is slower and carries a large preset-driven configuration surface. Vite+ provides a faster integrated lint and formatting workflow, while ESLint remains useful for structured data and Markdown languages that Oxlint does not parse.

Keeping ESLint away from code files makes the ownership boundary explicit: Oxlint handles code, ESLint handles non-code languages, and unsupported code checks are documented rather than silently retained as a second code linter.

### Additional Context

Related prior discussions:

- #22583 proposed trying Oxlint for ESLint performance.
[...]

PR #42714, opened 2026-09-22
refactor(web): replace legacy icon barrel imports

lyzno1:
## Summary

Replace legacy icon barrel imports with direct imports of the same leaf components, removing 41 indexes and their 200 `no-barrel-files` suppressions. Update the generator so it does not recreate the indexes, and update affected test mocks to target the leaf default exports.

Teach `prefer-tailwind-icons` to recognize direct default imports by their source filename. Use the actual custom Iconify collections to produce valid class suggestions, preserve their casing, and skip unavailable or ambiguous names. This also fixes invalid suggestions exposed by the import migration. The icon implementations, SVG data, consumer JSX, props, event handlers, and styles remain unchanged.

Related to #38833.

## Validation

- Independently verified all 337 migrated import bindings resolve to the same original leaf modules; consumer non-import syntax and test mock behavior are unchanged.
- Compared old and new generators in isolated output directories: all 1,032 non-index artifacts are byte-identical.
- Compared 200 icons across three prop cases and light/dark themes in Chromium: all 1,200 SVG instances have identical DOM, geometry, and computed colors; both 1440×2296 screenshots have zero changed pixels. Real FileIcon, CornerLabel, and ContentSwitch compositions and navigation/disabled behavior also pass.
- Affected mock suites: 17 files, 177 tests passed.
- Icon lint diagnostics were audited against the actual collections: preserve 447 valid diagnostics across 190 migrated TSX consumers, correct 19 class spellings, and remove 2 suggestions for an unavailable icon. The 501 configured custom classes exactly match the real Tailwind plugin registrations; all 933 final suggestions across 485 relevant files reference available classes.
- Rule regression tests: 13 passed, covering named/default imports, aliases, JSX/runtime references, unavailable classes, case normalization, ambiguous names, and asset/type-only exclusions.
- `vp run -w check`, `vp run dify-web#lint:tss` (6,215 checks), `vp run knip`, and `git diff --check` passed.

## Checklist

- [x] Added or updated tests where meaningful regression risk justifies coverage.
- [x] No user documentation changes are required for this refactor.
- [x] Completed `vp run -w check` and frontend staged checks.


lyzno1:
Superseded by #42716, which migrates the consumers to Iconify CSS and removes the complete legacy React icon system. The replacement also handles embedded bitmap assets as feature-owned images and verifies sizing and behavior.

PR #42726, opened 2026-09-22
fix: improve accessibility in legacy knowledge creation

iamjoel:
## Summary

- Fix keyboard interaction, accessible names, headings, and selection semantics across the legacy knowledge creation flow and its shared settings controls.
- Announce upload, preview, and indexing status; handle preview failures and restore focus after closing previews. Keep the mobile preview drawer with explicit open/close state and a close button.
- Adapt navigation and creation panels to narrow screens, and stack retrieval controls based on container width. Keep the original first-step padding and leave color contrast unchanged.

Fixes [SNP-640](https://linear.app/dify/issue/SNP-640/旧的-knowledge-创建流程)

## Validation

- Relevant component and regression tests passed, including 200 tests covering the final mobile preview drawer changes.
- Full repository `vp run -w check` passed (0 errors; warnings remain), including formatting, lint/type diagnostics, and ESLint checks.
- The final commit-hook fixes passed 80 file-preview and first-step regression tests.
- A separate earlier `tsc --noEmit` run reported stale `.next/dev/types/validator.ts` generated-route errors.
- Manually checked LICENSE.txt upload, file preview status, and focus restoration in the local application.
- Full browser verification of all three steps and every responsive layout has not been completed.

## Checklist

- [x] Full repository `vp run -w check` and normal pre-commit hooks passed.

From Codex


web/app/components/datasets/settings/option-card.tsx
```diff
@@ -47,22 +48,32 @@ const OptionCard = <T,>({
...
-        onClick?.(id)
-      }}
     >
-      <div className={cn('relative flex rounded-t-xl p-2', className)}>
+      <Toggle
+        pressed={!!isActive}
+        disabled={disabled || !onClick}
+        onPressedChange={(pressed) => {
+          if (pressed) onClick?.(id)
+        }}
```
lyzno1: [P2] Keep Economy parameter editing consistent with the selected mode

For an existing Economy knowledge base, select High Quality and then edit the still-visible Economy keyword count. Moving the selection handler from the card to this header means the keyword count changes without selecting Economy; saving still submits `indexing_technique: high_quality`. `IndexMethod` renders the keyword controls with unconditional `showChildren`, so this is reachable through the real settings UI. A focused interaction test passes with the pre-change card and fails on this revision (`onKeywordNumberChange` fires, but `onChange(economy)` does not). Please hide/disable the inactive mode's parameters, or explicitly coordinate parameter editing and mode selection in the owning component.

web/app/components/datasets/create/step-two/components/option-card.tsx
```diff
@@ -98,39 +116,51 @@ export const OptionCard: FC<OptionCardProps> = ({ ref, ...props }) => {
...
-        isActive={isActive && !noHighlight}
-        activeClassName={activeHeaderClassName}
-        effectImg={effectImg}
+      <Toggle
+        pressed={!!isActive}
         disabled={disabled}
-      />
+        onPressedChange={(pressed) => {
+          if (pressed) onSwitched?.()
+        }}
```
lyzno1: [P2] Use radio-group semantics for mutually exclusive choices

Chunking mode and indexing method each represent one persistent selection, but these standalone Toggles expose independent `aria-pressed` buttons. Activating the selected button cannot toggle it off because the `false` transition is discarded, and there is no radio-group relationship or arrow-key selection behavior. The same mismatch exists in `datasets/settings/option-card.tsx` for indexing and retrieval methods. Please let the owning section provide a labelled `RadioGroup`, render each selectable header as a `RadioItem`, and keep parameter controls outside the radio itself. This follows the repository's `packages/dify-ui/docs/selection.md` contract and the [WAI-ARIA radio-group pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/).

web/app/components/datasets/create/file-preview/index.tsx
```diff
@@ -16,35 +18,36 @@ type IProps = {
...
-  }
+  const headingId = useId()
+  const fileID = file?.id || ''
+  const { data, isPending, isError } = useQuery(
+    queryOptions({
+      queryKey: commonQueryKeys.filePreview(fileID),
+      queryFn: () => fetchFilePreview({ fileID }),
+      enabled: !!fileID,
+      retry: false,
+    }),
```
lyzno1, line 26: [P2] Use the generated query options for the preview migration

This migrates the preview request from an Effect into Query but introduces another handwritten `queryKey`/`queryFn` around the legacy REST helper. The matching JSON operation is already available as `consoleQuery.files.byFileId.preview.get.queryOptions()`, with `input.params.file_id` and the generated `TextContentResponse`. `web/AGENTS.md` requires generated APIs for migrated surfaces, and the data contract requires `skipToken` when required input is missing. Please use those generated options directly instead of the legacy key/helper and empty-string ID. This is a project-contract violation, not a demonstrated runtime failure.

lyzno1 (commented):
Rechecked the current revision. The three issues below remain, including a reproduced regression when editing Economy parameters. The 19 changed test files (518 tests) pass on this revision, but do not cover that interaction. Browser layout and screen-reader behavior have not been fully verified.

web/app/components/datasets/create/step-two/index.tsx
```diff
@@ -267,61 +275,70 @@ const StepTwo: FC<StepTwoProps> = ({
...
-            onSummaryIndexSettingChange={segmentation.handleSummaryIndexSettingChange}
-          />
-        )}
+        </h1>
+        <RadioGroup<ChunkingMode>
+          className="block"
+          aria-label={t(($) => $['stepTwo.segmentation'], { ns: 'datasetCreation' })}
+          value={currentDocForm === ChunkingMode.qa ? ChunkingMode.text : currentDocForm}
+          onValueChange={handleDocFormChange}
+        >
```
lyzno1, line 293: [P1] Isolate parameter-input arrow keys from radio-group navigation

Wrapping the complete option cards in this RadioGroup also puts their parameter controls inside Base UI's composite keyboard handler. With General selected, focus the delimiter input, place the caret at the end, and press ArrowRight: the event bubbles to the group, selects Parent-child, unmounts the input, and calls `handleDocFormChange`, which also resets the preview. I reproduced this through the real StepTwo component. The same regression occurs in `settings/index-method/index.tsx`: ArrowRight at the end of the Economy keyword input selects High Quality and hides the input. Keeping the fields outside RadioItem alone does not isolate them from the containing RadioGroup. Please exclude parameter-panel keyboard events from the group's navigation handling while preserving arrow-key selection on the radio headers, and cover both input-editing scenarios.

lyzno1 (commented):
The previous three findings are addressed in this revision: inactive Economy parameters are hidden, the selection controls use radio groups, and file preview uses generated query options. One new keyboard regression remains below. All 19 changed test files (465 tests) pass, but two additional focused interaction tests reproduce unintended mode changes from text-input arrow keys in StepTwo and IndexMethod. These are component-test results; full browser layout and screen-reader verification is still outstanding.

lyzno1 (approved):

PR #42747, opened 2026-09-22
fix(web): make human input button style selector accessible

balancetheworld:
## Summary

  Fixes #<issue-number>

  Improve the accessibility of the Human Input button style selector.

  ### Changes

  - Replace non-semantic clickable elements with accessible buttons.
  - Implement the style options using a radio group.
  - Add keyboard navigation and selection support.
  - Add accessible labels for the selector and each button style option.
  - Add localized labels for all supported locales.
  - Update tests to cover keyboard interaction, radio selection, and the disabled state.
  - Remove the related oxlint accessibility suppressions.

  ## Screenshots

  Not applicable.

  ## Checklist
  - [ ] I ran `make lint && make type-check` and `vp staged`.

  From Codex.


web/app/components/workflow/nodes/human-input/components/button-style-dropdown.tsx
```diff
@@ -74,56 +75,63 @@ const ButtonStyleDropdown: FC<Props> = ({ text = 'Button Text', data, onChange,
...
-          <div className="mt-2 flex w-81 flex-wrap gap-1">
-            <div
-              className={cn(
-                'box-border flex h-20 w-40 cursor-pointer items-center justify-center rounded-lg border-[1.5px] border-transparent bg-background-section hover:bg-background-section-burn',
-                data === UserActionButtonType.Primary &&
-                  'border-components-option-card-option-selected-border',
-              )}
-              onClick={() => onChange(UserActionButtonType.Primary)}
+          <RadioGroup
+            aria-label={t(($) => $[`${i18nPrefix}.userActions.chooseStyle`], { ns: 'workflow' })}
```
WTW0313: [P2] Give the popover dialog an accessible name. `PopoverContent` renders a `dialog`, but the visible heading above this group is a plain `div`. Base UI only sets the dialog's `aria-labelledby` when `PopoverTitle` is present, so opening the picker exposes an unnamed dialog; this group's `aria-label` does not name its parent. Render the heading as `PopoverTitle` and reference it from the radio group with `aria-labelledby`.

web/app/components/workflow/nodes/human-input/components/button-style-dropdown.tsx
```diff
@@ -74,56 +75,63 @@ const ButtonStyleDropdown: FC<Props> = ({ text = 'Button Text', data, onChange,
...
+            value={data}
+            onValueChange={(value) => onChange(value)}
+            disabled={readonly}
+            className="mt-2 flex w-81 flex-wrap gap-1"
+          >
+            <RadioItem
+              value={UserActionButtonType.Primary}
+              aria-label={`${text}, ${t(($) => $[`${i18nPrefix}.userActions.buttonStyle.primary`], { ns: 'workflow' })}`}
```
WTW0313, line 90: [P2] Include the action in the trigger's accessible name as well. The new radio names include the action text, but every row's popover trigger is still named only "Choose a button style". The panel renders one selector per action, so someone navigating by buttons cannot distinguish the Approve selector from the Reject selector until opening it. Include the action text in a localized trigger name. This is an existing gap that the accessibility change leaves in place.

web/app/components/workflow/nodes/human-input/components/__tests__/button-style-dropdown.spec.tsx
```diff
@@ -25,11 +31,22 @@ describe('ButtonStyleDropdown', () => {
...
+    await user.keyboard('{Enter}')
+    const radioGroup = screen.getByRole('radiogroup', {
+      name: 'nodes.humanInput.userActions.chooseStyle',
+    })
+    expect(radioGroup).toBeInTheDocument()
+
+    const ghostRadio = screen.getByRole('radio', { name: 'Approve, Ghost' })
+    expect(ghostRadio).toHaveAttribute('aria-checked', 'true')
+    const primaryRadio = screen.getByRole('radio', { name: 'Approve, Primary' })
+    primaryRadio.focus()
```
WTW0313: [P2] Exercise keyboard navigation instead of focusing the unchecked radio imperatively. `primaryRadio.focus()` bypasses the radio group's roving tab stop, so this test still passes if Arrow-key navigation breaks. Move from the selected Ghost radio with an Arrow key, then assert focus, the checked state, and `onChange`.

WTW0313 (commented):
Accessibility review findings are noted inline.

web/app/components/workflow/nodes/human-input/components/button-style-dropdown.tsx
balancetheworld: 收到

WTW0313 (approved):
The three review findings have been addressed. The focused component tests and Web CI pass.

PR #42750, opened 2026-09-22
fix(workflow): coerce non-uuid variable ids and warn on skipped branches

### LINKED ISSUE #34358: [Refactor/Chore] Add preflight workflow validation and actionable diagnostics for missing variable references / unreachable paths

IlIIIIIIlI:
## Summary

A conversation variable whose id is not a UUID still imported, then the first Preview died in `INSERT INTO workflow_conversation_variables` with `invalid input syntax for type uuid`. Import of a YAML that is not an app export also showed two toasts, and the specific one did not say which keys were present. Publishing a graph where a node can read a producer on a skipped branch succeeded with no signal.

This keeps the publish. It rewrites a non-UUID conversation variable id to a name-stable uuid5 before that insert, names the top-level keys when `app` is missing, shows that backend error once, and returns an advisory `warning` when a reachable node can run without the node it reads. The warning does not block publish. The old stream hang is already gone on main and is left alone.

The checked-in fixture `api/tests/fixtures/workflow/minimal_non_uuid_conversation_variable.yml` is a fake-node repro: one hand-written id `opt-comp-prompt-var`, one if-else, one producer only on the true branch, and an answer that can run without that producer. It replaces attaching the full enterprise DSL.

Fixes #34358
close https://github.com/langgenius/dify/pull/37196

## Why

#37196 and #37131 were the two open pieces of this report. #37337 is already covered on main. Combining the crash fix and the advisory warning here follows the request on #34358 to bring those two changes together, without folding in the stream handler again.

## Screenshots

Not applicable in this commit. The import toast and the publish warning are asserted in the web specs. A local Docker check of the minimal fixture follows separately.

## Validation

- `uv run --project api --no-sync pytest --noconftest api/tests/unit_tests/factories/test_conversation_variable_id.py api/tests/unit_tests/services/test_workflow_variable_reference_validator.py -q`: 22 passed.
- `uv run --project api --no-sync ruff check` on the touched Python files: passed.
- Not run: `make type-check`, full `make test`, and `vp staged`. The clean worktree sync cannot build `chroma-hnswlib` on this Windows machine (MSVC is missing), and local Node is 20 while web expects Node 24, so the web specs were not executed here.

## Checklist

- [x] I understand that this PR may be closed if there was no previous discussion.
- [x] Tests cover the behavior that can regress.
- [x] Documentation is unchanged because this is internal.

From Grok


IlIIIIIIlI:
@crazywoola

Rebased onto current main (`0f6a622`). The checks that failed last time are fixed.

This push has not run CI. The workflows are waiting for approval. Could you approve them?

Locally I ran ruff, pyrefly, and the tests for this change (21 passed). The full API and web suites do not run on my machine. Is there a good way to replicate the same CI locally?

IlIIIIIIlI:
Follow-up on `3bcf15d`: publish still returns only `result` and `created_at` when there is no advisory finding. `warning` is included only when a skipped-branch reference exists, so existing publish callers are unchanged.

crazywoola (dismissed):

IlIIIIIIlI:
@crazywoola it seems 2 of the 58 ci check failed again. I have fixed those two check. Can you check again please? 
If this failed again, i would manually check my updates when I off-work later today. Just follow up my last question,, if there's a way to simulate all ci checks locally? I have followed the contribute.md and learnt from previous PRs for this one, but it still didn't 100% guarantee the ci pass...

IlIIIIIIlI:
@crazywoola, fixed the reproduced variable-ID failure in b684886. All 4,911 Linux controller tests and both backend unit shards pass. Local Web coverage still has two timeouts reproduced on the unchanged base; Next build is blocked by local OOM. The PR body now records these limits and the advisory's scope. Ready for another CI run and review.

IlIIIIIIlI:
@crazywoola Good morning, The prior Web failure was one Skills Builder test timing out at 5,059 ms against the 5,000 ms limit. This PR does not modify that test; the other three Web shards passed. The new CI run is awaiting another approval. Codecov will update after it runs.

crazywoola (dismissed):

IlIIIIIIlI:
Thanks for the review, @crazywoola! The PR checks are now all green. 

The merge queue failure appears to be an existing Redis mock isolation issue in TestUseProxyContext.test_returns_data_and_deletes_key, the same failure also occurred in https://github.com/langgenius/dify/pull/42056
Could you please re-add this PR to the merge queue for another run or manually merge it? Thanks!

IlIIIIIIlI:
@crazywoola, updated from main and resolved the conflict. Could you approve CI and re-add this to the queue? If my PR checks pass but the queue fails on an unrelated issue again, may I merge it directly?

